PR #1, opened 2022-07-07
Copy prototype to new repo

quiiver:
Copy prototype to the new repo. This can provide a starting point for the new version of merino


Dockerfile
```diff
@@ -0,0 +1,21 @@
+# Use the official lightweight Python image.
```
ncloudioj, line 1: To make the deployment setup easier, can you adjust this Dockerfile by following the [Dockerflow](https://github.com/mozilla-services/Dockerflow)? In particular, the `USER` and the `WORKDIR` bits.

Here is an [example](https://github.com/mozilla-services/telescope/blob/main/Dockerfile) for reference.

merino/providers/adm.py
```diff
@@ -0,0 +1,47 @@
...
+                id = len(self.results)
+                for kw in suggestion.get("keywords"):
+                    self.suggestions[kw] = id
+                self.results.append({k:suggestion[k] for k in suggestion if k != 'keywords'})
+        icon_items = [i for i in suggest_settings if i['type'] == 'icon']
+        for icon in icon_items:
+            id = int(icon['id'].replace('icon-', ''))
+            self.icons[id] = icon["attachment"]["location"]
+
+    async def query(self, q: str):
```
ncloudioj, line 28: Looks like `async` is not required for this method.

Separately, shall define `query` in the `BaseProvider/DefaultProvider`?

merino/remotesettings.py
```diff
@@ -0,0 +1,18 @@
...
+    def __init__(self) -> None:
+        self.client = kinto_http.Client(server_url=self.host)
+        self.attachment_host = self.client.server_info()["capabilities"]["attachments"]["base_url"]
+
+    def get(self, bucket, collection):
+        return self.client.get_records(collection=collection, bucket=bucket)
+
+    def fetch_attachment(self, attachement_uri):
+        uri = urljoin(self.attachment_host, attachement_uri)
+        return requests.get(uri)
```
ncloudioj, line 18: This should be fine for now. Given the current suggestion pool on RS, I think we will change this to async fetching in production to avoid blocking the eventloop.

merino/providers/adm.py
ncloudioj: Makes sense. Shall we make `BaseProvider` an `ABC` and then add `query` as an async abstract method so that subclasses can implement it accordingly?

Dockerfile
quiiver: 👍 updated

merino/remotesettings.py
quiiver: makes sense.

PR #4, opened 2022-07-25
feat: Add models for suggest endpoint

ncloudioj:
This fixes https://mozilla-hub.atlassian.net/browse/CONSVC-1892.

### DIFF
--- a/merino/web/models_v1.py
+++ b/merino/web/models_v1.py
@@ -1,0 +1,61 @@
+from typing import Optional
+from uuid import UUID, uuid4
+
+from pydantic import BaseModel, Field, HttpUrl
+
+
+class ProviderResponse(BaseModel):
+    """
+    Model for the `providers` API response.
+    """
+
+    id: str
+    availability: str
+
+
+class BaseSuggestion(BaseModel):
+    """
+    Base model for suggestions, can be extended for sponsored and nonsponsored
+    suggestions.
+    """
+
+    block_id: int
+    full_keyword: str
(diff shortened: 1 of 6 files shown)


tests/web/test_api_v1.py
```diff
@@ -35,6 +36,17 @@ def test_no_suggestion():
     assert len(response.json()["suggestions"]) == 0
 
 
+@pytest.mark.parametrize("query", ["sponsored", "nonsponsored"])
+def test_suggest_from_missing_providers(query):
+    """
+    Despite the keyword is available for other providers, it should not return any suggestions if
```
pjenvey, line 42:
```suggestion
    Despite the keyword being available for other providers, it should not return any suggestions if
```

merino/web/models_v1.py
```diff
@@ -0,0 +1,61 @@
...
+    click_url: Optional[HttpUrl] = None
+
+
+class SuggestResponse(BaseModel):
+    """
+    Model for the `suggest` API response.
+    """
+
+    suggestions: list[SponsoredSuggestion | NonsponsoredSuggestion]
+    client_variants: list[str] = []
```
ncloudioj, line 59: My past Python experience reminds me to be extra careful about using a mutable object as the default value, but Pydantic has some magic making this okay here. So we don't need to write `Field(default_factory=list)`.

pjenvey (approved):

quiiver (approved):

PR #7, opened 2022-07-29
feat: Add dockerflow and MozLog

ncloudioj:
This closes [CONSVC-1849](https://mozilla-hub.atlassian.net/browse/CONSVC-1849).

### DIFF
--- a/version.json
+++ b/version.json
@@ -1,0 +1,6 @@
+{
+  "source": "https://github.com/mozilla-services/merino-py",
+  "version": "dev",
+  "commit": "TBD",
+  "build": "TBD"
+}
--- a/Makefile
+++ b/Makefile
@@ -22,7 +22,7 @@ format: $(INSTALL_STAMP)
 	$(POETRY) run black $(APP_TEST_DIR)
 
 test: $(INSTALL_STAMP)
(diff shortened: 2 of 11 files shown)


version.json
```diff
@@ -0,0 +1,6 @@
+{
```
quiiver, line 1: Not a blocker but can we use the version information in the `tool.poetry` section of the `pyproject.toml` file? Ideally we wouldn't have to duplicate version information.

quiiver (approved):

version.json
ncloudioj: Are you referring to this particular test or to the Dockerflow endpoint in general?

This version file only serves as a placeholder for testing as some of those fields would only be available at the build/deploy time. As a result, we usually just replace this one with another file generated automatically during CI. For example, [this](https://github.com/mozilla-services/merino/blob/main/.circleci/config.yml#L305-L317) shows how it gets generated for the old Merino repo.

Another thing worth noting is that the `version` field gets trickier to populate/maintain as now we've switched over to CD for Merino, we no longer bump/tag version numbers when we merge PRs to the main branch. I believe @hackebrot and the team are working on a solution to this.

hackebrot: That's correct. We're working on a recommendation for `version` in the Rapid Release Model working group.

PR #8, opened 2022-08-05
feat: Add a configuration manager

ncloudioj:
Added a configuration manager (using Dynaconf). It also replaced various hardcoded constants with configurations.

Documented inline about the basic usage, but a more detailed instruction doc should be provided in the future.  

This closes [CONSVC-1903](https://mozilla-hub.atlassian.net/browse/CONSVC-1903).

### DIFF
--- a/merino/config.py
+++ b/merino/config.py
@@ -1,35 +1,15 @@
-from logging.config import dictConfig
-from os import environ
+from dynaconf import Dynaconf
 
-log_level: str = environ.get("LOG_LEVEL", "INFO").upper()
-# TODO: Add another formatter for development
-log_format: str = "json"
+# `root_path` = The root path for Dynaconf, DO NOT CHANGE.
+# `envvar_prefix` = Export envvars with `export MERINO_FOO=bar`.
+# `settings_files` = Load these files in the order.
+# `environments` = Enable layered environments such as `development`, `production`, `testing` etc.
+# `env_switcher` = Switch environments by `export MERINO_ENV=production`. Default: `development`.
 
-
-def configure_logging() -> None:  # pragma: no cover
-    """
-    Configures logging with MozLog.
-    """
-
-    dictConfig(
-        {
-            "version": 1,
-            "formatters": {
-                "json": {
-                    "()": "dockerflow.logging.JsonLogFormatter",
-                    "logger_name": "merino",
-                },
-            },
-            "handlers": {
-                "console": {
-                    "level": log_level,
-                    "class": "logging.StreamHandler",
-                    "formatter": log_format,
-                }
-            },
-            "loggers": {
-                "merino": {"handlers": ["console"], "level": log_level},
-                "request.summary": {"handlers": ["console"], "level": log_level},
-            },
-        }
-    )
+settings = Dynaconf(
+    root_path="merino",
+    envvar_prefix="MERINO",
+    settings_files=["configs/default_settings.toml", "configs/settings.toml"],
+    environments=True,
+    env_switcher="MERINO_ENV",
+)
--- a/merino/config.py
+++ b/merino/config.py
@@ -1,10 +1,18 @@
-from dynaconf import Dynaconf
+from dynaconf import Dynaconf, Validator
+
+# Validators for Merino settings.
+_validators = [
+    Validator("logging.level", is_in=["DEBUG", "INFO", "WARNING", "ERROR", "CRITICAL"]),
+    Validator("logging.format", is_in=["json"]),
+    Validator("remote_settings.record_type", is_in=["data", "offline-expansion-data"]),
+]
 
 # `root_path` = The root path for Dynaconf, DO NOT CHANGE.
 # `envvar_prefix` = Export envvars with `export MERINO_FOO=bar`.
 # `settings_files` = Load these files in the order.
 # `environments` = Enable layered environments such as `development`, `production`, `testing` etc.
 # `env_switcher` = Switch environments by `export MERINO_ENV=production`. Default: `development`.
+# `validators` = Define validators for Merino settings.
 
 settings = Dynaconf(
     root_path="merino",
@@ -12,4 +20,5 @@ settings = Dynaconf(
     settings_files=["configs/default_settings.toml", "configs/settings.toml"],
     environments=True,
     env_switcher="MERINO_ENV",
+    validators=_validators,
 )
--- a/merino/config.py
+++ b/merino/config.py
@@ -17,7 +17,11 @@ _validators = [
 settings = Dynaconf(
     root_path="merino",
     envvar_prefix="MERINO",
-    settings_files=["configs/default_settings.toml", "configs/settings.toml"],
+    settings_files=[
+        "configs/default.toml",
+        "configs/development.toml",
+        "configs/production.toml",
+    ],
     environments=True,
     env_switcher="MERINO_ENV",
     validators=_validators,
--- a/merino/providers/adm.py
+++ b/merino/providers/adm.py
@@ -37,13 +37,13 @@ class Provider(BaseProvider):
         suggest_settings = rs.get(
             settings.remote_settings.bucket, settings.remote_settings.collection
         )
-        data_items = [
-            i
-            for i in suggest_settings
-            if i["type"] == settings.remote_settings.record_type
+        records = [
+            record
+            for record in suggest_settings
+            if record["type"] == settings.remote_settings.record_type
         ]
-        for item in data_items:
-            res = rs.fetch_attachment(item["attachment"]["location"])
+        for record in records:
+            res = rs.fetch_attachment(record["attachment"]["location"])
             for suggestion in res.json():
                 id = len(self.results)
                 for kw in suggestion.get("keywords"):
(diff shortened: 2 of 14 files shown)


hackebrot (commented):
Hi @ncloudioj! 👋🏻 

Last week at the engineering meeting, we went over a few options for configuration management for Python web applications:

* python-decouple
* dynaconf
* pydantic

Looks like you've decided to integrate dynaconf. I'd love to know why you decided it was the right tool for merino-py!

ncloudioj:
Hi @hackebrot, sure, I've added the decision matrix of those three candidates in [here](https://docs.google.com/document/d/1Ikt-pB5qRSUBkjEKA_Ojvd9_Bzo6PL7ENEF5_IzlJdo/edit?usp=sharing) for the record. To me, Dynaconf is the no-brainer for its rich feature set and flexibility, I think we can use it to implement the same configuration manager as its Rust predecessor.

merino/config.py
```diff
@@ -1,35 +1,15 @@
...
+# `settings_files` = Load these files in the order.
+# `environments` = Enable layered environments such as `development`, `production`, `testing` etc.
+# `env_switcher` = Switch environments by `export MERINO_ENV=production`. Default: `development`.
+
+settings = Dynaconf(
+    root_path="merino",
+    envvar_prefix="MERINO",
+    settings_files=["configs/default_settings.toml", "configs/settings.toml"],
+    environments=True,
+    env_switcher="MERINO_ENV",
```
taddes, line 14: To follow up on validation comment above, if we want to [add validators,](https://www.dynaconf.com/validation/#validating-in-python-programmatically) which I've read evaluate lazily, we either define them as a list of validators, or the registration of validators after the Dynaconf instance is created with `settings.validators.register()`. What's really neat is that we can verify these at the environment level with the `env` parameter, which may end up being really useful, given that we may want different behaviors in the future based on providers.  

Ex:
```
settings = Dynaconf(
    ...
    validators=[
        # Ensure some parameters exists (are required)
        Validator('ENVVAR_PREFIX', 'ENV_SWITCHER', 'ROOT_PATH', must_exist=True),

        # Ensure some password cannot exist
        Validator('PASSWORD', must_exist=False),

        # Ensure some parameter mets a condition
        # conditions: (eq, ne, lt, gt, lte, gte, identity, is_type_of, is_in, is_not_in)
        Validator('SETTINGS_FILES', is_type_of="list"),

        # under the env [production] its value must be >=8000 and <=9000
        Validator("DB.PORT",gte=8000, lte=9000, env="production"),

        # validate a value is eq in specific env
        Validator('PROJECT', eq='hello_world', env='production'),

        # Ensure some parameter (string) meets a condition
        # conditions: (len_eq, len_ne, len_min, len_max, cont)
        # Determines the minimum and maximum length for the value
        Validator("NAME", len_min=3, len_max=125),

        # Signifies the presence of the value in a set, text or word
        Validator("DEV_SERVERS", cont='localhost'),

        # Checks whether the length is the same as defined.
        Validator("PORT", len_eq=4),
    ]
)
```

Then we can call `settings.validators.validate()`. I did see a pydantic schema can be used for this, so I guess it remains open as to whether we use this Dynaconf feature.

merino/providers/adm.py
```diff
@@ -33,8 +34,14 @@ class Provider(BaseProvider):
 
     def __init__(self) -> None:
         rs = remotesettings.Client()
-        suggest_settings = rs.get("main", "quicksuggest")
-        data_items = [i for i in suggest_settings if i["type"] == "data"]
+        suggest_settings = rs.get(
+            settings.remote_settings.bucket, settings.remote_settings.collection
+        )
+        data_items = [
+            i
```
taddes, line 41: Might I suggest the use of a more declarative name in the expression, like `item`? I know we're using it below to iterate through `data_items`, but I think it doesn't hurt given it's within the body of the comprehension.  It just makes understanding the individual collection item easier, in my opinion.

taddes (approved):
Looks great, Nan! I think this will end up being the best choice for configuration management. Getting to review the docs shows a lot of flexibility and promise for all that we intend to do, especially managing possible future providers and settings. 

I added several comments, none of which are action items but discussion points on  architecture and validation enhancements. We can discuss further and see if the features are valuable for addition.

merino/config.py
```diff
@@ -1,35 +1,15 @@
...
-                    "formatter": log_format,
-                }
-            },
-            "loggers": {
-                "merino": {"handlers": ["console"], "level": log_level},
-                "request.summary": {"handlers": ["console"], "level": log_level},
-            },
-        }
-    )
+from dynaconf import Dynaconf
```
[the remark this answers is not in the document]
ncloudioj: Good catch, I will add some validators here.

merino/config.py
```diff
@@ -1,35 +1,15 @@
...
+# `root_path` = The root path for Dynaconf, DO NOT CHANGE.
+# `envvar_prefix` = Export envvars with `export MERINO_FOO=bar`.
+# `settings_files` = Load these files in the order.
+# `environments` = Enable layered environments such as `development`, `production`, `testing` etc.
+# `env_switcher` = Switch environments by `export MERINO_ENV=production`. Default: `development`.
+
+settings = Dynaconf(
+    root_path="merino",
+    envvar_prefix="MERINO",
+    settings_files=["configs/default_settings.toml", "configs/settings.toml"],
```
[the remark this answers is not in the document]
ncloudioj: While I don't think we will use this feature in production, we might be able to use it for development in the future, that's why I added those to ignore files to avoid inadvertent inclusion later. I didn't include them here as they're currently not in use, doing so would force everyone to create a dummy file otherwise it will complain.

merino/config.py
ncloudioj: Yeah, we can use this feature for sure at least for settings with obvious validation specs.

merino/config.py
```diff
@@ -1,35 +1,15 @@
...
+    environments=True,
+    env_switcher="MERINO_ENV",
+)
```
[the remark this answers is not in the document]
ncloudioj: I think we will use them both, specifically SRE prefers env vars for deployment, TOML files are better in other cases. Also, env vars will be provided separately by the container orchestration manager rather than via dot env files. That's why I didn't include `configs/.env` here.

I expect that we could further clarify those desicions in a dedicated configuration doc.

merino/providers/adm.py
ncloudioj: Sure, we can do that. Thanks for pointing this out!

merino/config.py
```diff
@@ -1,35 +1,24 @@
...
-            "loggers": {
-                "merino": {"handlers": ["console"], "level": log_level},
-                "request.summary": {"handlers": ["console"], "level": log_level},
-            },
-        }
-    )
+settings = Dynaconf(
+    root_path="merino",
+    envvar_prefix="MERINO",
+    settings_files=["configs/default_settings.toml", "configs/settings.toml"],
```
quiiver, line 20: what do you think about changing the names to be more explicit in their purpose? perhaps something like `default.toml` and `dev.toml`. that would give us a path to `prod.toml` if we needed it. we could also drop the `settings` suffix since i assume everything in config would just be settings `toml` files

quiiver (approved):
looks good, i had some minor feedback around the naming of the config files. otherwise 👍

merino/config.py
ncloudioj: I think that's a great idea, let's do that!

hackebrot:
> Hi @hackebrot, sure, I've added the decision matrix of those three candidates in [here](https://docs.google.com/document/d/1Ikt-pB5qRSUBkjEKA_Ojvd9_Bzo6PL7ENEF5_IzlJdo/edit?usp=sharing) for the record. To me, Dynaconf is the no-brainer for its rich feature set and flexibility, I think we can use it to implement the same configuration manager as its Rust predecessor.
> 
> Would love to hear your thoughts as well.

Thank you for the link to the decision matrix. That is helpful! 📋 

Since we're aiming to make it easier for project members to contribute to Merino, I want to make sure that we think it's necessary to pull in an additional dependency (especially if it's a central piece of the service, such as configuration management, and the dependency has a rich feature set that may be new to people). Hence my suggestion to use the settings management capabilities [FastAPI](https://fastapi.tiangolo.com/advanced/settings/) and [Pydantic](https://pydantic-docs.helpmanual.io/usage/settings/) already provide.

You've mentioned that dynaconf makes it easier for us to implement the same configuration manager as its Rust predecessor. Do we need all of the features of the Merino Rust config? In other words could this be a case of Yagni ("You Aren't Gonna Need It")? If not, let's go ahead and use dynaconf. 🚀

ncloudioj:
Thanks, @hackebrot! Also appreciated your feedback.

> Do we need all of the features of the Merino Rust config?

Yes, I think so. In fact, pretty much all of the configuration features of Merino are now in use in production except for the remote configuration feature. And I am certain that we will use that once we ramp up the sponsorship campaigns in Merino.

Re: the complexity with this dependency, I have an opposite view on this as I believe that this config manager would largely simplify how we manage configurations for different environments. Admittedly, there is indeed a learning cost with the underlying Engine (Dynaconf) for advanced uses, but I assume most of us would work with the configuration files (default/development/production.toml) in practice. Also, a guidance document around configuration will be provided to further ease the onboarding.

PR #9, opened 2022-08-05
[CONSVC-1904] feat: add bandit

tiftran:


pyproject.toml
```diff
@@ -38,6 +42,7 @@ kinto-http = "^10.10.1"
 requests = "^2.28.1"
 dockerflow = "^2022.7.0"
 asgi-correlation-id = "^3.0.1"
+bandit = "^1.7.4"
```
ncloudioj, line 45: Let's move this to the dev dependencies to keep the docker image lean. N.B: dev deps won't be packaged into the docker image.

ncloudioj (approved):
Looks great, thanks for banding it up!

PR #14, opened 2022-08-09
CONSVC-1901 Add async apis to remote settings client

quiiver:
* Moved provider initialization out of the class __init__ method and
  into an abstract class method called `initialize`
* Added `httpx` as an async `requests` alternative

### DIFF
--- a/merino/providers/__init__.py
+++ b/merino/providers/__init__.py
@@ -1,8 +1,14 @@
+import asyncio
+import logging
+from timeit import default_timer as timer
+
 from merino.providers.adm import Provider as AdmProvider
 from merino.providers.base import BaseProvider
 
 providers: dict[str, BaseProvider] = {}
 default_providers: list[BaseProvider] = []
+
+logger = logging.getLogger("merino.providers")
 
 
 async def init_providers() -> None:
@@ -11,11 +17,17 @@ async def init_providers() -> None:
 
     This should only be called once at the startup of application.
     """
+    start = timer()
     providers["adm"] = AdmProvider()
+    await asyncio.gather(*[p.initialize() for p in providers.values()])
     default_providers.extend([p for p in providers.values() if p.enabled_by_default()])
+    logger.info(
+        "Provider initialization complete",
+        extra={"providers": [*providers.keys()], "elapsed": timer() - start},
+    )
 
 
-def get_providers() -> tuple[dict[str, BaseProvider], list[BaseProvider]]:
+async def get_providers() -> tuple[dict[str, BaseProvider], list[BaseProvider]]:
     """
     Return a tuple of all the providers and default providers.
     """
--- a/merino/providers/__init__.py
+++ b/merino/providers/__init__.py
@@ -8,7 +8,7 @@ from merino.providers.base import BaseProvider
 providers: dict[str, BaseProvider] = {}
 default_providers: list[BaseProvider] = []
 
-logger = logging.getLogger("merino.providers")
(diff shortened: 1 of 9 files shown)


merino/providers/__init__.py
```diff
@@ -1,21 +1,33 @@
...
+import logging
+from timeit import default_timer as timer
+
 from merino.providers.adm import Provider as AdmProvider
 from merino.providers.base import BaseProvider
 
 providers: dict[str, BaseProvider] = {}
 default_providers: list[BaseProvider] = []
 
+logger = logging.getLogger("merino.providers")
```
ncloudioj, line 11: nit: you could do `logging.getLogger(__name__)` as well.

merino/providers/__init__.py
```diff
@@ -1,21 +1,33 @@
...
+    await asyncio.gather(*[p.initialize() for p in providers.values()])
     default_providers.extend([p for p in providers.values() if p.enabled_by_default()])
+    logger.info(
+        "Provider initialization complete",
+        extra={"providers": [*providers.keys()], "elapsed": timer() - start},
+    )
 
 
-def get_providers() -> tuple[dict[str, BaseProvider], list[BaseProvider]]:
+async def get_providers() -> tuple[dict[str, BaseProvider], list[BaseProvider]]:
```
ncloudioj, line 30: Hmm, why is this an async func?

merino/providers/adm.py
```diff
@@ -31,23 +32,36 @@ class Provider(BaseProvider):
...
-        suggest_settings = rs.get("main", "quicksuggest")
+        suggest_settings = await rs.get("main", "quicksuggest")
         data_items = [i for i in suggest_settings if i["type"] == "data"]
-        for item in data_items:
-            res = rs.fetch_attachment(item["attachment"]["location"])
+
+        fetch_tasks = [
+            rs.fetch_attachment(item["attachment"]["location"]) for item in data_items
+        ]
+        for routine in as_completed(fetch_tasks):
```
ncloudioj, line 47: nit: can you rename `routine` to `done_task` please?

merino/providers/__init__.py
quiiver: sure

merino/providers/base.py
```diff
@@ -7,6 +7,15 @@ class BaseProvider(ABC):
     Abstract class for suggestion providers.
     """
 
+    @abstractmethod
+    async def initialize(self) -> bool:
```
ncloudioj, line 11: Do we need a return value for this method? Looks like the provider in `adm.py` doesn't return a boolean. When getting unrecoverable initialization errors, normally the application would log them and exit. Unless we want to handle those init errors, I don't think the return value would be necessary here.

quiiver: oh yeah i meant to drop that. thanks

ncloudioj (changes requested):
This looks great. Having a few nits.

I've tested it locally, the async concurrent loading cuts the loading time from 3.5s to 0.7s on my laptop!!!

merino/providers/__init__.py
ncloudioj: Ah, i see. Let's drop it. The test/util function can be a reg function as well, we don't need to call `initialize()` for those.

ncloudioj (approved):
Good stuff! 🚢

quiiver:
this fixes https://mozilla-hub.atlassian.net/browse/CONSVC-1901

PR #15, opened 2022-08-10
fix: Provide an init value for attachment_host

ncloudioj:
Noticed this oops while working on other things. It will hang the initilize method of the adm provider.


quiiver (approved):
weird i assumed the conditional in `fetch_attachment` would cover that

PR #16, opened 2022-08-11
feat(provider): Periodically resync data from Remote Setting

ncloudioj:
This adds a background job to resync data from Remote Settings for the adm provider. The implementation is identical to the Rust version. Also, we should add some test coverage to this module as it gets more stuff in there. 

This fixes [CONSVC-1902](https://mozilla-hub.atlassian.net/browse/CONSVC-1902).

### DIFF
--- a/merino/configs/default.toml
+++ b/merino/configs/default.toml
@@ -34,3 +34,9 @@ bucket = "main"
 collection = "quicksuggest"
 # Can be "data" or "offline-expansion-data"
 record_type = "data"
+
+[default.providers.adm]
+# The cron job should tick more frequently than `resync_interval_sec` so that
+# the resync failure can be retried soon.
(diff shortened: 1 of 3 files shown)


merino/configs/default.toml
```diff
@@ -34,3 +34,9 @@ bucket = "main"
 collection = "quicksuggest"
 # Can be "data" or "offline-expansion-data"
 record_type = "data"
+
+[default.providers.adm]
+# The cron job should tick more frequently than `resync_interval_sec` so that
+# the resync failure can be retried soon.
+cron_interval_sec = 60
```
ncloudioj, line 41: These values reflect what we currently use in production. We could tune them now if needed.

quiiver (approved):
looks good!

PR #17, opened 2022-08-12
feat: enable cors CONSVC-1906

taddes:
Added simple CORS configurations to mernio-py as defined in the [FastAPI CORS docs](https://fastapi.tiangolo.com/tutorial/cors).

Looking at our current Merino-rs config, see [here in actix docs](https://docs.rs/actix-cors/latest/actix_cors/struct.Cors.html#method.permissive) and [here in source](https://docs.rs/actix-cors/latest/src/actix_cors/builder.rs.html#81) it was using the `permissive` setting, which for production is of course not desirable. While in the local testing phases, a permissible CORS configuration is desirable, but we will want to adjust it as time goes on. However, I did implement a few restrictions:

- Only permit GET, POST, OPTIONS, and HEAD requests (simple requests, excl OPTIONS)
- Did not permit credentials and cookies to be supported in cross-origin requests. Since the wildcard `"*"` is set for origins, this is not allowed. Not sure if we may want auth cooks enabled that match the API's domain down the road.

In discussing whether we should make any adjustments to the current config, I want to pose the following questions to reviewers:
- In an ideal world, we could consider the specific origins from which we allow requests. Do we have a list or potential collection we could partially validate against? Given my relative newness to quicksuggest and the general architecture, I can't confidently know what this should or shouldn't be. Can we make some assumptions about the origins of some of our requests? This will differ significantly from development to production. Is there a world in which we store and dynamically add origins to our list? If so, I have seen some possible implementations using Starlette's classes.
- We want to consider the returned `Access-Control-Request-Method`  and `Access-Control-Request-Headers`  headers and if we want to set up a whitelist. For request headers, by default  `Accept`, `Accept-Language`, `Content-Language` and `Content-Type` are allowed.
- Consider what response headers we may want to expose to the browser. Default: `expose_headers=[]`. Can I get some input on what responses we may want to expose?
- Is the 3600 age setting good or do we want to increase the TTL for the browser to use to cache the response? Default is 800.

I also added a CONTRIBUTING.md file here. I know it's not part of the task, so I can take it out, but I thought it useful at this initial phase. Based on some of our other Mozilla-services projects with merino-py specific revisions.

### DIFF
--- a/merino/main.py
+++ b/merino/main.py
@@ -46,17 +46,9 @@ app.add_middleware(logging.LoggingMiddleware)
 app.include_router(dockerflow.router)
 app.include_router(api_v1.router, prefix="/api/v1")
 
-cors_origins = [
-    "http://localhost:8000",
-    "http://127.0.0.1:8000",
-    "http://localhost:80",
-]
-
 app.add_middleware(
     CORSMiddleware,
-    allow_origins=cors_origins,
-    allow_origin_regex=r"http://localhost:^([1-9][0-9]{0,3}|[1-5][0-9]{4}"
-    / r"|6[0-4][0-9]{3}|65[0-4][0-9]{2}|655[0-2][0-9]|6553[0-5])$",
+    allow_origins=["*"],
     allow_credentials=False,
     allow_methods=["GET", "POST", "OPTIONS", "HEAD"],
     allow_headers=["*"],
--- a/merino/main.py
+++ b/merino/main.py
@@ -50,7 +50,6 @@ app.add_middleware(
     CORSMiddleware,
     allow_origins=["*"],
     allow_credentials=False,
-    allow_methods=["GET", "POST", "OPTIONS", "HEAD"],
+    allow_methods=["GET", "OPTIONS", "HEAD"],
     allow_headers=["*"],
-    max_age=3600,
 )
--- a/merino/main.py
+++ b/merino/main.py
@@ -51,5 +51,4 @@ app.add_middleware(
     allow_origins=["*"],
     allow_credentials=False,
     allow_methods=["GET", "OPTIONS", "HEAD"],
-    allow_headers=["*"],
 )
--- a/merino/main.py
+++ b/merino/main.py
@@ -49,6 +49,7 @@ app.include_router(api_v1.router, prefix="/api/v1")
 cors_origins = [
     "http://localhost:8000",
     "http://127.0.0.1:8000",
+    "http://localhost:80",
 ]
 
 app.add_middleware(
--- a/merino/main.py
+++ b/merino/main.py
@@ -2,6 +2,7 @@ from asgi_correlation_id import CorrelationIdMiddleware
 from fastapi import FastAPI, status
 from fastapi.encoders import jsonable_encoder
 from fastapi.exceptions import RequestValidationError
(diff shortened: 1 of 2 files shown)


merino/main.py
```diff
@@ -44,3 +45,12 @@ async def validation_exception_handler(_, exc) -> JSONResponse:
 
 app.include_router(dockerflow.router)
 app.include_router(api_v1.router, prefix="/api/v1")
+
+app.add_middleware(
+    CORSMiddleware,
+    allow_origins=["*"],
```
ncloudioj, line 51: I think it's fine to use `*` here. The requests of Merino are made directly from the browser rather than from any particular sites (origins).

merino/main.py
```diff
@@ -44,3 +45,12 @@ async def validation_exception_handler(_, exc) -> JSONResponse:
...
+    allow_credentials=False,
```
ncloudioj, line 52: Good choice!

merino/main.py
```diff
@@ -44,3 +45,12 @@ async def validation_exception_handler(_, exc) -> JSONResponse:
...
+    allow_methods=["GET", "POST", "OPTIONS", "HEAD"],
```
ncloudioj, line 53: Merino currently doesn't have any POST endpoints. So let's drop "POST" from the allow list.

merino/main.py
```diff
@@ -44,3 +45,12 @@ async def validation_exception_handler(_, exc) -> JSONResponse:
...
+    allow_headers=["*"],
```
ncloudioj, line 54: Let's use the default allow headers, those are should be sufficient for now.

merino/main.py
```diff
@@ -44,3 +45,12 @@ async def validation_exception_handler(_, exc) -> JSONResponse:
...
+    max_age=3600,
```
ncloudioj, line 55: I believe this max_age only affects how long the preflight OPTION results will be cached. As Merino only exposes GET methods, which belong to the simple requests and will not trigger the OPTION preflight requests, it really doesn't matter here. So using the default max_age should be fine.

ncloudioj:
@taddes Thanks for the patch and the thorough consideration! Also the CONTRIBUTING.md 👍

It looks good, having a few minor comments. Also, highly recommended to read this doc about [CORS](https://developer.mozilla.org/en-US/docs/Web/HTTP/CORS), it explains how things work quite well.

merino/main.py
taddes: Sounds good! I didn't think of it that way... duh...

merino/main.py
taddes: Ok, I will remove it then and allow the default to kick in.

merino/main.py
taddes: Sounds good. So, for testing allowing all headers ok for now?  The Accept, Accept-Language, Content-Language and Content-Type headers are always allowed for CORS. Enabling the wildcard allows all, so is that ok for now or shall I remove it and only allow those 4?

ncloudioj: Accept, Accept-Language, Content-Language and Content-Type are good enough, so let's remove the wildcard for now, we will add more in the future if needed.

ncloudioj (approved):
Looks good, thanks!

taddes:
Ok, changes made @ncloudioj !

PR #19, opened 2022-08-16
[CONSVC-1933] feat: add sid and seq_no to logging for suggest api

tiftran:


jrconlin (approved):

pyproject.toml
```diff
@@ -44,6 +44,7 @@ dockerflow = "^2022.7.0"
 asgi-correlation-id = "^3.0.1"
 dynaconf = "^3.1.9"
 httpx = "^0.23.0"
+mock = "^4.0.3"
```
ncloudioj, line 47: I believe that mock is now part of the standard library `unittest.mock`, so we don't need to install it separately. Also, `pytest-mock` is already installed and exposed as a fixture to facilitate its use.

tests/web/test_api_v1.py
```diff
@@ -61,3 +62,16 @@ def test_providers():
     assert set([provider["id"] for provider in providers]) == set(
         ["sponsored-provider", "nonsponsored-provider"]
     )
+
+
+def test_request_logs_contain_required_info():
```
ncloudioj, line 67:
```suggestion
def test_request_logs_contain_required_info(mocker):
    import logging

    spy = mocker.spy(logging.Logger, "info")
    query = "nope"
    sid = "deadbeef-0000-1111-2222-333344445555"
    seq = "22"
    root_path = "/api/v1/suggest"
    client.get(f"{root_path}?q={query}&sid={sid}&seq={seq}")
    extra_dict = spy.call_args[1].get("extra")
    assert extra_dict["path"] == root_path
    assert extra_dict["session_id"] == sid
    assert extra_dict["sequence_no"] == seq
```

merino/middleware/logging.py
```diff
@@ -26,6 +26,8 @@ async def dispatch(self, request: Request, call_next):
             "time": datetime.fromtimestamp(time.time()).isoformat(),
             # Provided by the asgi-correlation-id middleware.
             "rid": response.headers["X-Request-ID"],
+            "session_id": request.query_params.get("sid"),
+            "sequence_no": request.query_params.get("seq"),
```
ncloudioj, line 30: @tiftran By specs, `sequence_no` should be zero-based integers, let's do the conversion here.
```suggestion
            "sequence_no": int(seq) if (seq := request.query_params.get("seq")) else None
```

tiftran: oh snap you're right

ncloudioj: I think everything in `query_params` is a string (i could be wrong please verify), and "0" is a truthy value in python. so we should be good.

tiftran: Oooh 🤦‍♀️  you're right it's "0" not 0

ncloudioj (approved):
Looks good, thanks!

PR #22, opened 2022-08-19
Feature Flags using Dynaconf

quiiver:
As a little bit of a side quest, I wrote a basic implementation of featureflags using dynaconf as the configuration system. It is a simple implementation that supports two bucketing schemes `random` and `session`. Random does what it says on the tin. It generates a random bucketing id for every flag check. Session bucketing uses the session id of the request as the bucketing key so that feature checks within a given search session would be consistent. Each flag has a very simple configuration:

```
[default.flags.<flag_name>]
scheme = 'session'
enabled = 0.5 
```
`scheme` - This is the bucketing scheme for the flag. Allowed values are 'random' and 'session'
`enabled` - This represents the % enabled for the flag and must be a float between 0 and 1


merino/featureflags.py
```diff
@@ -0,0 +1,77 @@
...
+        self.flags = (flags if flags is not None else _flags).get("flags", {})
+
+    def is_enabled(self, flag: str, bucket_for: str | bytes | None = None) -> bool:
+        config = self.flags.get(flag)
+        if config is None:
+            return False
+
+        enabled = config.get("enabled", 0.0)
+        # enabled should be gt 0 and lt 1
+        if enabled <= 0.0 or enabled > 1.0:
```
ncloudioj, line 34: I guess that the passed-in `flags` argument in `__init__` and this check here is mainly for testing, correct? For other scenarios, feature flags will be defined by `flags.toml` or env vars. Either way, this can also be checked and enforced by Dynaconf's validator like what we've done [here](https://github.com/mozilla-services/merino-py/blob/main/merino/config.py#L4-L10). Can you add a validator for this?

merino/featureflags.py
```diff
@@ -0,0 +1,77 @@
...
+        "configs/flags.toml",
+    ],
+    environments=True,
+    env_switcher="MERINO_ENV",
+)
+
+session_id_context: ContextVar[str | None] = ContextVar("session_id", default=None)
+
+
+class FeatureFlags:
```
ncloudioj, line 21: Do you mind adding some documents here, like the PR summary?

merino/featureflags.py
ncloudioj: The `scheme` field can be validated by an enum of `["random", "session"]` as well.

quiiver: I added validators but i'm not entirely sure the appropriate way to test them.

quiiver: Also, I switched to using a `testing` environment instead of passing in the flags

merino/middleware/featureflags.py
```diff
@@ -0,0 +1,16 @@
+from starlette.middleware.base import BaseHTTPMiddleware
+from starlette.requests import Request
+
+from merino.featureflags import session_id_context
+
+
+class FeatureFlagsMiddleware(BaseHTTPMiddleware):
+    """
+    Sets the a ContextVar for session_id so that it can be used
```
ncloudioj, line 9:
```suggestion
    Sets a ContextVar for session_id so that it can be used
```

merino/featureflags.py
```diff
@@ -0,0 +1,107 @@
...
+    `scheme` - This is the bucketing scheme for the flag. Allowed values are 'random' and 'session'
+    `enabled` - This represents the % enabled for the flag and must be a float between 0 and 1
+    """
+
+    flags: dict
+
+    def __init__(self) -> None:
+        self.flags = _flags.get("flags", {})
+
+    def is_enabled(self, flag: str, bucket_for: str | bytes | None = None) -> bool:
```
ncloudioj, line 53: Since this is the main method of `FeatureFlags`, can you add a few descriptions for the arguments of this function?

merino/featureflags.py
```diff
@@ -0,0 +1,107 @@
...
+        """
+        Checks if a given flag is enabled via a feature flag configuration block.
+        As a principal it fails closed as quickly as possible.
```
ncloudioj, line 56:
```suggestion
        As a principal, it fails as quickly as possible.
```

merino/configs/flags/default.toml
```diff
@@ -0,0 +1,5 @@
+# Example feature flag configuration block.
+# 
+# [default.flags.example]
+# scheme="random"
```
ncloudioj, line 4: Nit: let's add some brief comments here for each configuration, e.g. what possible values (range) it could take. This allows folks to make quick changes with enough context without further looking into the more detailed doc in `FeatureFlags.py`.

merino/featureflags.py
ncloudioj: Using the testing environment is a good idea. Alternatively, I think we can also mock out `_config.flags` by using the `mocker` fixture.

```py
import merino.featureflags import _flags

def test_foo(mocker):
    mocker.patch.object(_config, 'flags', {"test-enabled": { "enabled": 1, "scheme": "random" }})

    flags = FeatureFlags()
    assert flags.is_enabled("test-missing") is False    
```

merino/featureflags.py
```diff
@@ -0,0 +1,107 @@
...
+
+        bucket_scheme = config.get("scheme")
+        if (bucketing_id := self._get_bucketing_id(bucket_scheme, bucket_for)) is None:
+            return False
+
+        bucket_value = self._bytes_to_interval(bucketing_id)
+        return bucket_value <= enabled
+
+    def _get_bucketing_id(
+        self, scheme: str, bucket_for: str | bytes | None
```
ncloudioj, line 75: Nit: let's change `_get_bucketing_id()`, `_get_random()`, `_get_digest()`, and `_bytes_to_interval()` to static methods as none of them need `self`.

merino/featureflags.py
```diff
@@ -0,0 +1,107 @@
...
+    ) -> bytes | None:
+        # override bucketing id if specified in args
+        if bucket_for is not None:
+            bucket_type = type(bucket_for)
+            if bucket_type is str:
+                return self._get_digest(cast(str, bucket_for))
+            elif bucket_type is bytes:
+                return cast(bytes, bucket_for)
```
ncloudioj, line 83: Shall we throw over here when seeing an invalid `bucket_for` type rather than return `None` implicitly? If we expect it to be either str or bytes if present, having other types sounds like a bug to me, and I think we should fail loud.

merino/featureflags.py
```diff
@@ -0,0 +1,107 @@
...
+    def _get_random(self) -> bytes:
+        return randbytes(32)
+
+    def _get_digest(self, id: str) -> bytes:
+        hash = hashlib.sha256(bytes(id, "utf-8"))
+        return hash.digest()
+
+    def _bytes_to_interval(self, bucketing_id: bytes) -> float:
+        out = 0
+        max = 1 << len(bucketing_id)  # this should always be 32 bytes long
```
ncloudioj, line 103: If `bucketing_id` is guaranteed (enforced by `_get_bucketing_id()`) to be 32 bytes long, why not use a constant here?

merino/featureflags.py
```diff
@@ -0,0 +1,107 @@
...
+        for i in bucketing_id:
```
ncloudioj, line 104: Out of curiosity, is this a known algorithm? It's kinda fascinating.

I thought that perhaps some bit shifting could be a little faster than an if statement, though the benchmark didn't suggest so. Actually, it was always 5-10% slower 😂

```py
def branch_free(bs):
    out = 0
    for i, b in enumerate(bs):
        out |= (b >> 7) << i   # (31 - i) would yield the same result as the original 
    return out / (1 << 32)
```

tests/test_featureflags.py
```diff
@@ -0,0 +1,44 @@
+from merino.featureflags import FeatureFlags, session_id_context
+
+
+def test_missing():
+    flags = FeatureFlags()
+    assert flags.is_enabled("test-missing") is False
```
ncloudioj, line 6: Usually, we'd just do `assert [!]predicate` if the `predicate` is of type boolean. Any particular reason for using `is True/False` here?

tests/test_featureflags.py
```diff
@@ -0,0 +1,44 @@
...
+
+
+def test_enabled_perc():
+    flags = FeatureFlags()
+    assertions = {
+        b"\x00\x00\x00\x00": True,
+        b"\xff\xff\xff\xff": False,
+    }
+
+    for bid, test in assertions.items():
```
ncloudioj, line 31: Nit: you can also do [`@pytest.mark.parameterize`](https://docs.pytest.org/en/6.2.x/parametrize.html) for this. It pretty much does the same thing with less boilerplate and it will display all test combinations in the test results.

ncloudioj (changes requested):
Sorry for the delay, got distracted earlier.

It looks good, just had a few comments and nits for your consideration.

hackebrot (changes requested):
Hi @quiiver! 👋🏻 

Thank you for tagging me on this. I added my comments inline.

merino/featureflags.py
quiiver: We could throw here but I wouldn't want those exceptions to bubble up past `is_enabled`. I don't want `is_enabled` checks to ever throw. They should fail to False to prevent extra branching around catching exceptions for every flag check. I would be down with adding logging or metrics in situations like this. that would allow us to have visibility into issues

tests/test_featureflags.py
quiiver: I'm just not terribly familiar with the correct patterns so i usually default to the most explicit :D

tests/test_featureflags.py
quiiver: parameterize sounds great, thanks for the heads up

Makefile
```diff
@@ -23,7 +23,7 @@ format: $(INSTALL_STAMP)
 	$(POETRY) run black $(APP_TEST_DIR)
 
 test: $(INSTALL_STAMP)
-	$(POETRY) run pytest -v $(TEST_DIR) --cov $(APP_DIR)
+	MERINO_ENV=testing $(POETRY) run pytest -v $(TEST_DIR) --cov $(APP_DIR)
```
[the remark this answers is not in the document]
quiiver: Yeah I waffled a bit on that and eventually landed on setting it via the environment variable. since we will be defining production environments with env variables it felt like a good opportunity to 'test' that out. The other reason was that directly calling functions on the `settings` module generated a deprecation warning about not modifying the settings module directly

merino/featureflags.py
quiiver: I should change that comment but the reason for the length check is because with adhoc bucketing ids we don't enforce a length.

merino/featureflags.py
```diff
@@ -0,0 +1,107 @@
...
+        if config is None:
+            return False
+
+        enabled = config.get("enabled", 0.0)
+        # enabled should be gt 0 and lt 1
+        if enabled <= 0.0 or enabled > 1.0:
+            return False
+
+        bucket_scheme = config.get("scheme")
+        if (bucketing_id := self._get_bucketing_id(bucket_scheme, bucket_for)) is None:
```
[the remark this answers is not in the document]
quiiver: scheme is not optional and defaults to 'session'

merino/featureflags.py
```diff
@@ -0,0 +1,107 @@
...
+from dynaconf import Dynaconf, Validator
+
+_flags = Dynaconf(
+    root_path="merino",
+    envvar_prefix="MERINO",
+    settings_files=[
+        "configs/flags/default.toml",
+        "configs/flags/testing.toml",
+    ],
+    validators=[
```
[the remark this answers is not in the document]
quiiver: I want to continue using dynaconf and not add another configuration system but I agree with you on the validation. I can try to spend some more time looking at how to use pydantic with dynaconf for additionally structure / schema

merino/featureflags.py
```diff
@@ -0,0 +1,154 @@
...
+        """
+        # Override bucketing id if specified in args
+        if bucket_for is not None:
+            match bucket_for:
+                case str():
+                    return self._get_digest(bucket_for)
+                case bytes():
+                    return bucket_for
+                case _:
+                    raise RuntimeError(
```
ncloudioj, line 114: Maybe this is a `TypeError`?

merino/featureflags.py
```diff
@@ -0,0 +1,154 @@
...
+            # If bucket_for is None use the scheme specified in the config
+            match scheme:
+                case "random":
+                    return self._get_random()
+                case "session" if (session_id := session_id_context.get()) is not None:
+                    # If session_id is None, try the next case block.
+                    # Otherwise return the digest of the session_id.
+                    return self._get_digest(session_id)
+                case _:
+                    raise RuntimeError(
```
ncloudioj, line 127: It sounds like a `ValueError` to me. Actually, maybe we can define an Enum for `scheme`.

merino/middleware/featureflags.py
```diff
@@ -0,0 +1,16 @@
...
+    to consistently bucket flags within a search session
+    """
+
+    async def dispatch(self, request: Request, call_next):
+        session_id = request.query_params.get("sid")
+        session_id_context.set(str(session_id))
```
ncloudioj, line 15:
```suggestion
        session_id_context.set(session_id)
```

`request_params.get("sid")` is already a str if present.  If `sid` is not set in `request.query_params`, `str(session_id)` will produce a string `"None"` which is probably not what we want to set for `session_id_context`.

ncloudioj (approved):
LGTM. r+ with a few more comments.

merino/featureflags.py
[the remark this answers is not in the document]
quiiver: Updated to use BaseModel for both `FeatureFlag` and `FeatureFlagConfig`.

merino/featureflags.py
```diff
@@ -0,0 +1,193 @@
...
+        Args:
+            scheme: The bucketing scheme. Can be `random` or `session`
+            bucket_for:
+                An optional bucketing id that will be used in place of the specified scheme.
+                Can be either a string or bytes. If given a string we will create a digest
+                using sha256.
+        Returns:
+            bytes
+        Raises:
+            RuntimeError
```
quiiver, line 146:
```suggestion
            TypeError, ValueError
```

merino/featureflags.py
```diff
@@ -0,0 +1,195 @@
...
+        config = self.flags[flag_name]
+        if config is None:
+            return False
+
+        # Short circuits for enabled values of 0 or 1
+        match config.enabled:
+            case 0.0:
+                return False
+            case 1.0:
+                return True
```
hackebrot, line 119: Interesting! The GitHub syntax highlighting for this seems to be broken, but I don't see any syntactical errors here.

merino/featureflags.py
```diff
@@ -0,0 +1,195 @@
...
+        """
+        Hashes a string into a 32 byte bytearray
+        """
+        hash = hashlib.sha256(bytes(id, "utf-8"))
+        return hash.digest()
+
+    @staticmethod
+    def _bytes_to_interval(bucketing_id: bytes) -> float:
+        """
+        Takes an arbitrarily long bytearray and maps it to a float between [0,1)
```
hackebrot, line 190:
```suggestion
        Takes an arbitrarily long bytearray and maps it to a float between [0, 1].
```

merino/featureflags.py
```diff
@@ -0,0 +1,195 @@
...
+    scheme = 'session'
+    enabled = 0.5
+    ```
+
+    `scheme` - This is the bucketing scheme for the flag. Allowed values are 'random' and 'session'
+    `enabled` - This represents the % enabled for the flag and must be a float between 0 and 1
+    """
+
+    flags: FeatureFlagConfigs
+    default_scheme: BucketingScheme = BucketingScheme.session
```
hackebrot, line 89: See comment above.

hackebrot (changes requested):
Hi @quiiver! 👋🏻 

This is great! I have added one suggestion for the `scheme` default value, but then is good to go from my side.

merino/featureflags.py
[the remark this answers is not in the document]
hackebrot: Looks good! Thanks! 👍🏻

merino/featureflags.py
quiiver: that's not a typo its how you noteate an interval range that includes the upper boundary https://en.wikipedia.org/wiki/Interval_(mathematics)#Notations_for_intervals

merino/featureflags.py
```diff
@@ -0,0 +1,195 @@
...
+
+
+class BucketingScheme(str, Enum):
+    random = "random"
+    session = "session"
+
+
+class FeatureFlag(BaseModel):
+    enabled: Enabled
+    scheme: Optional[BucketingScheme]
```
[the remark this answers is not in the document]
quiiver: I'll check but i believe mypy doesn't understand that it will never have a None value so it complains when you use `Flag.scheme` directly

merino/featureflags.py
hackebrot: TIL! 💡

merino/featureflags.py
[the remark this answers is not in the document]
hackebrot: Right, we can drop `Optional` and make it:

```python
scheme: BucketingScheme = BucketingScheme.session
```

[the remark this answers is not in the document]
quiiver: done

hackebrot (approved):
LGTM! 🚀

PR #18, opened 2022-08-16
WIP feat: initial configuration of sentry middleware

taddes:
Initial configuration

PR #23, opened 2022-08-23
Sentry integration consvc 1851

ncloudioj:
This replaces #18.

It closes [CONSVC-1851](https://mozilla-hub.atlassian.net/browse/CONSVC-1851).


pjenvey (commented):
Want to add a `/__error__` endpoint to the dockerflow module while you're here?

ncloudioj:
> Want to add a /__error__ endpoint to the dockerflow module while you're here?

Good point, will do!

pjenvey:
I'll also note that willkg built a sentry fake for testing here: https://github.com/willkg/kent/

(just as an FYI, I haven't used it nor am I suggesting we add it) cc @Trinaa

quiiver:
I don't fully understand the difference between integration and unit testing here. in this case it looks like we are still just unit testing via log capture methods. am i missing something? it looks like the main difference is the configuration change. the feature flags pr introduced a `testing.toml` config. can we use that instead of having two testing configurations?

ncloudioj:
> I don't fully understand the difference between integration and unit testing here. in this case it looks like we are still just unit testing via log capture methods. am i missing something?

The main difference between the two is that in unit tests, a couple of things are disabled (providers, logging, sentry, etc), and all the startup handlers will be ignored. Whereas we enable all of them in integration tests in order to make sure they work as a whole.

> it looks like the main difference is the configuration change. the feature flags pr introduced a testing.toml config. can we use that instead of having two testing configurations?

In addition to that, the other change is how we initialize the test client in integration tests. While it looks subtle, it will trigger the startup and shutdown handlers of FastAPI (see more details [here](https://fastapi.tiangolo.com/advanced/testing-events/)). I was aware of the addition of `testing.toml`, but decided to create another one for integration tests as certain configurations cannot be easily applied to both. For instance, we don't need to hook up with Sentry for unit tests.

Admittedly, the boundary between those two tests is not clear, plus we also have contract tests. So I can otherwise be convinced if we didn't want to do this for now.

ncloudioj:
After a second thought, I've decided to drop the integration tests bit as they're not quite as what I originally wanted them to be. Will revisit them in the future.

quiiver (approved):
Non blocking comment, otherwise lgtm

PR #26, opened 2022-08-25
feat: Metrics - initial implementation

quiiver:
Adding metrics for merino. Currently the following are instrumented:
* Total provider initialization timing
* Individual provider initialization timing
* Total request timing (including middleware)
* Count of request status codes (per url path)
* Total provider query timing 
* Individual provider timing

Novel thing here is monkeypatching a local logger for dev that does not send messages across the upd socket and insteads logs the metrics. I like the visibility in development so if there's a better way please let me know.

### DIFF
--- a/merino/config.py
+++ b/merino/config.py
@@ -27,7 +27,6 @@ settings = Dynaconf(
     envvar_prefix="MERINO",
     settings_files=[
         "configs/default.toml",
-        "configs/testing.toml",
         "configs/development.toml",
         "configs/production.toml",
         "configs/ci.toml",
--- a/merino/config.py
+++ b/merino/config.py
@@ -5,6 +5,7 @@ from dynaconf import Dynaconf, Validator
 _validators = [
     Validator("logging.level", is_in=["DEBUG", "INFO", "WARNING", "ERROR", "CRITICAL"]),
     Validator("logging.format", is_in=["mozlog", "pretty"]),
+    Validator("metrics.port", gte=0),
     Validator("providers.adm.cron_interval_sec", gt=0),
     Validator("providers.adm.resync_interval_sec", gt=0),
     Validator("providers.adm.score", gte=0, lte=1),
--- a/merino/config.py
+++ b/merino/config.py
@@ -27,6 +27,7 @@ settings = Dynaconf(
     envvar_prefix="MERINO",
     settings_files=[
         "configs/default.toml",
+        "configs/testing.toml",
         "configs/development.toml",
         "configs/production.toml",
         "configs/ci.toml",
--- a/merino/config.py
+++ b/merino/config.py
@@ -5,7 +5,9 @@ from dynaconf import Dynaconf, Validator
 _validators = [
     Validator("logging.level", is_in=["DEBUG", "INFO", "WARNING", "ERROR", "CRITICAL"]),
     Validator("logging.format", is_in=["mozlog", "pretty"]),
-    Validator("metrics.port", gte=0),
+    Validator("metrics.host", is_type_of=str),
+    Validator("metrics.port", gte=0, is_type_of=int),
+    Validator("metrics.dev_logger", is_type_of=bool),
     Validator("providers.adm.cron_interval_sec", gt=0),
     Validator("providers.adm.resync_interval_sec", gt=0),
     Validator("providers.adm.score", gte=0, lte=1),
--- a/merino/configs/development.toml
+++ b/merino/configs/development.toml
@@ -3,19 +3,16 @@
 
 # This would be the default environment for development.
 
-
 [development]
 debug = true
+# For `list` or `table` settings, `dynaconf_merge` allows you to merge settings with the default
+# settings. This enables merge for the entire `development` environment.
+dynaconf_merge = true
 
 [development.metrics]
 dev_logger = true
-dynaconf_merge = true
 
 [development.logging]
 # Any of "DEBUG", "INFO", "WARNING", "ERROR", "CRITICAL"
 level = "DEBUG"
 format = "pretty"
-# For `list` or `table` settings, `dynaconf_merge` allows you to merge settings with the default
-# settings. For instance, `logging.level` in this case will override its counterpart defined in
-# `default_settings`.
-dynaconf_merge = true
--- a/merino/configs/development.toml
+++ b/merino/configs/development.toml
@@ -7,7 +7,15 @@
 [development]
 debug = true
 
+[development.metrics]
+dev_logger = true
+dynaconf_merge = true
+
 [development.logging]
 # Any of "DEBUG", "INFO", "WARNING", "ERROR", "CRITICAL"
 level = "DEBUG"
 format = "pretty"
+# For `list` or `table` settings, `dynaconf_merge` allows you to merge settings with the default
+# settings. For instance, `logging.level` in this case will override its counterpart defined in
+# `default_settings`.
+dynaconf_merge = true
--- a/merino/main.py
+++ b/merino/main.py
@@ -18,9 +18,7 @@ app = FastAPI()
 
 @app.on_event("startup")
 async def startup_configuration() -> None:
-    """
-    Set up various configurations such as logging.
-    """
+    """Set up various configurations such as logging."""
     configure_logging()
     configure_sentry()
     await configure_metrics()
@@ -34,9 +32,7 @@ async def startup_providers() -> None:
 
 @app.on_event("shutdown")
 async def shutdown() -> None:
-    """
-    Clean up for the application shutdown.
-    """
+    """Clean up for the application shutdown."""
     await get_metrics_client().close()
 
 
--- a/merino/main.py
+++ b/merino/main.py
@@ -7,25 +7,37 @@ from fastapi.middleware.cors import CORSMiddleware
 from fastapi.responses import JSONResponse
 
 from merino import providers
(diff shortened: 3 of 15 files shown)


quiiver:
Example logs:

```
{"Timestamp": 1661446878757428992, "Type": "merino.metrics", "Logger": "merino", "Hostname": "wstuckey-q19r7w", "EnvVersion": "2.0", "Severity": 7, "Pid": 16944, "Fields": {"data": "merino.providers.initialize.adm:387|ms|#application:merino", "msg": "sending metrics"}}
{"Timestamp": 1661446878757558016, "Type": "merino.metrics", "Logger": "merino", "Hostname": "wstuckey-q19r7w", "EnvVersion": "2.0", "Severity": 7, "Pid": 16944, "Fields": {"data": "merino.providers.initialize:387|ms|#application:merino", "msg": "sending metrics"}}
{"Timestamp": 1661446880365125120, "Type": "merino.metrics", "Logger": "merino", "Hostname": "wstuckey-q19r7w", "EnvVersion": "2.0", "Severity": 7, "Pid": 16944, "Fields": {"data": "merino.providers.adm.query:0|ms|#application:merino", "msg": "sending metrics"}}
{"Timestamp": 1661446880365938944, "Type": "request.summary", "Logger": "merino", "Hostname": "wstuckey-q19r7w", "EnvVersion": "2.0", "Severity": 6, "Pid": 16944, "Fields": {"agent": "Mozilla/5.0 (Macintosh; Intel Mac OS X 10.15; rv:105.0) Gecko/20100101 Firefox/105.0", "path": "/api/v1/suggest", "method": "GET", "lang": "en-US,en;q=0.5", "querystring": {"q": "ama"}, "errno": 0, "code": 200, "time": "2022-08-25T12:01:20.365876", "rid": "5f0106c5df1c4060a7da6560ff0a4173", "session_id": null, "sequence_no": null, "country": null, "region": null, "city": null, "dma": null}}
{"Timestamp": 1661446880366594048, "Type": "merino.metrics", "Logger": "merino", "Hostname": "wstuckey-q19r7w", "EnvVersion": "2.0", "Severity": 7, "Pid": 16944, "Fields": {"data": "api.v1.suggest.timing:11|ms|#application:merino", "msg": "sending metrics"}}
{"Timestamp": 1661446880366808064, "Type": "merino.metrics", "Logger": "merino", "Hostname": "wstuckey-q19r7w", "EnvVersion": "2.0", "Severity": 7, "Pid": 16944, "Fields": {"data": "api.v1.suggest.status_codes.200:1|c|#application:merino", "msg": "sending metrics"}}
```

merino/middleware/metrics.py
```diff
@@ -0,0 +1,30 @@
...
+
+logger = logging.getLogger(__name__)
+
+
+class MetricsMiddleware(BaseHTTPMiddleware):
+    async def dispatch(
+        self, request: Request, call_next: RequestResponseEndpoint
+    ) -> Response:
+        try:
+            async with get_client() as client:
```
ncloudioj, line 18: Looks like `get_client()` is just a regular function, how does it get wrapped into an async context manager?

merino/middleware/metrics.py
```diff
@@ -0,0 +1,30 @@
...
+                with client.timeit(f"{metric_name}.timing"):
+                    response = await call_next(request)
+                client.increment(f"{metric_name}.status_codes.{response.status_code}")
+                return response
+        except Exception as e:
+            logger.exception(e)
+            raise HTTPException(status_code=500, detail=e)
+
+
+def build_metric_name(req: Request) -> str:
```
ncloudioj, line 29: I think we can build a lookup table for this instead of processing this repeatedly.

```py
class MertricsMiddleware(BaseHTTPMiddleware):
    def __init__(self, app):
          super().__init__(app)
          self._metric_names = {
              route.path: route.path.lstrip("/").replace("/", ".") for route in app.routes
          }
   
    async def dispatch(...):
        ...
        metric_name = self._metric_names.get(request.url.path)

```

merino/middleware/metrics.py
quiiver: yeah it returns a client which has the context manager impl. however i no longer need that and will be taking it out

merino/middleware/metrics.py
quiiver: i like that

merino/configs/development.toml
```diff
@@ -3,9 +3,14 @@
 
 # This would be the default environment for development.
 
-
 [development]
 debug = true
+# For `list` or `table` settings, `dynaconf_merge` allows you to merge settings with the default
+# settings. This enables merge for the entire `development` environment.
+dynaconf_merge = true
```
ncloudioj, line 10: Nice find👍

merino/main.py
```diff
@@ -29,6 +31,14 @@ async def startup_providers() -> None:
     await providers.init_providers()
 
 
+@app.on_event("shutdown")
+async def shutdown() -> None:
+    """
+    Set up various configurations such as logging.
```
ncloudioj, line 37:
```suggestion
    Clean up for the application shutdown.
```

merino/main.py
```diff
@@ -29,6 +31,14 @@ async def startup_providers() -> None:
...
+    """
+    await get_client().close()
```
ncloudioj, line 39: I'd recommend renaming this to `get_metrics_client()` or using `metrics.get_client()` to provide a little more context here.

merino/middleware/metrics.py
```diff
@@ -0,0 +1,37 @@
...
+            return response
+        except HTTPException as e:
+            status_code = e.status_code
+            raise e
+        finally:
+            client.increment(f"{metric_name}.status_codes.{status_code}")
+
+    @cache
+    def _build_metric_name(self, method: str, path: str) -> str:
+        return "{}.{}".format(
```
ncloudioj, line 35: We might end up caching a lot of 404 paths in memory as FastAPI's middleware always gets executed even for those 404 requests. To avoid that potential attacking surface, I'd recommend building this lookup table based only on the registered routes and methods of the app.

merino/metrics.py
```diff
@@ -0,0 +1,38 @@
...
+
+
+async def configure_metrics() -> Client:
+    client = get_client()
+    try:
+        if settings.metrics.dev_logger:
+            client._protocol = LocalDatagramLogger()
+        await client.connect()
+    except Exception as e:
+        logger.exception(e)
```
ncloudioj, line 28: If `connect()` fails, shall we do anything else (retry or just bail out) apart from logging? Running the app with a malfunctioning metrics client might trigger more chain effects such as flying blind or generating more errors on subsequent calls on this client.

merino/metrics.py
```diff
@@ -0,0 +1,38 @@
...
+
+logger = logging.getLogger(__name__)
+
+
+@cache
+def get_client() -> Client:
+    return Client(
+        host=settings.metrics.host,
+        port=settings.metrics.port,
+        constant_tags={"application": "merino"},
```
ncloudioj, line 17: Using this prefix might collide with the existing metrics of Merino (need confirm). If so, we might want to use a different prefix for the time being.

merino/middleware/metrics.py
```diff
@@ -0,0 +1,37 @@
...
+from merino.metrics import get_client
+
+logger = logging.getLogger(__name__)
+
+
+class MetricsMiddleware(BaseHTTPMiddleware):
+    async def dispatch(
+        self, request: Request, call_next: RequestResponseEndpoint
+    ) -> Response:
+        client = get_client()
```
ncloudioj, line 18: I think renaming it to `get_metrics_client()` would make this more readable.

merino/config.py
```diff
@@ -5,6 +5,7 @@
 _validators = [
     Validator("logging.level", is_in=["DEBUG", "INFO", "WARNING", "ERROR", "CRITICAL"]),
     Validator("logging.format", is_in=["mozlog", "pretty"]),
+    Validator("metrics.port", gte=0),
```
ncloudioj, line 8: Let's also add this:
```
Validator("metrics.dev_logger", is_type_of=bool)
```

tests/web/util.py
```diff
@@ -79,28 +79,49 @@ async def query(self, q: str) -> list[dict[str, Any]]:
             return []
 
 
-async def get_providers() -> tuple[dict[str, BaseProvider], list[BaseProvider]]:
+class CorruptProvider(BaseProvider):
     """
-    Return a tuple of all the providers and default providers.
+    A test nonsponsored provider that only responds to query "nonsponsored".
```
ncloudioj, line 84:
```suggestion
    A test corrupt provider that always raises a `RuntimeError` for any query it receives.
```

ncloudioj (approved):
Looks great. r+ with review comments.

merino/middleware/metrics.py
```diff
@@ -0,0 +1,59 @@
...
+        try:
+            response = await call_next(request)
+            status_code = response.status_code
+        except HTTPException as e:
+            status_code = e.status_code
+            raise e
+        finally:
+            duration = (loop.time() - started_at) * 1000
+            # don't track NOT_FOUND statuses by path.
+            # Instead we will track those within a general `status_codes` bucket.
```
quiiver, line 45:
```suggestion
            # Instead we will track those within a general `response.status_codes` metric.
```

PR #28, opened 2022-08-26
Fix contract tests

ncloudioj:
Fix contract tests in #27


merino/web/models_v1.py
```diff
@@ -25,8 +25,8 @@ class BaseSuggestion(BaseModel):
     advertiser: str
     provider: str
     is_sponsored: bool
-    icon: HttpUrl
     score: float
+    icon: str | None = None
```
ncloudioj, line 29: @Trinaa This change was made because, in the contract tests, the hostname of the icon url (e.g. `http://remote-settings-attachemnt/foo.icon`) was not a valid TLDN, and Pydantic wasn't happy about it. I couldn't find a better way to walk around, so ended up making this change.

merino/providers/adm.py
```diff
@@ -97,11 +97,11 @@ async def _fetch(self) -> None:
         suggest_settings = await rs.get(
             settings.remote_settings.bucket, settings.remote_settings.collection
         )
-        records = [
-            record
-            for record in suggest_settings
-            if record["type"] == settings.remote_settings.record_type
-        ]
+
+        # Falls back to "data" records if "offline-expansion-data" records do not exist
```
ncloudioj, line 101: For context, this is actually what Merino-rs does now. So we've aligned merino-py with that now.

test-engineering/contract-tests/volumes/client/scenarios.yml
```diff
@@ -131,7 +131,7 @@ scenarios:
                 url: 'https://example.com/target/coffee'
                 impression_url: 'https://example.com/impression/coffee'
                 click_url: 'https://example.com/click/coffee'
-                provider: 'test_remote_settings'
+                provider: 'adm'
```
ncloudioj, line 134: Unlike merino-rs, as part of the simplification, all the provider names are hardcoded in merino-py, updating fixtures to reflect that.

test-engineering/contract-tests/volumes/client/scenarios.yml
```diff
@@ -226,8 +226,8 @@ scenarios:
                 url: 'https://example.org/target/banana'
                 impression_url: 'https://example.org/impression/banana'
                 click_url: 'https://example.org/click/banana'
-                provider: 'test_remote_settings'
+                provider: 'adm'
                 advertiser: 'Example.org'
-                is_sponsored: true
+                is_sponsored: false
```
ncloudioj, line 231: By specs, only the IAB category "22 - Shopping" are deemed to be "sponsored" tiles, other categories are all non-sponsored.

test-engineering/contract-tests/volumes/kinto/data-01.json
```diff
@@ -19,7 +19,7 @@
         "url": "https://example.org/target/banana",
         "click_url": "https://example.org/click/banana",
         "impression_url": "https://example.org/impression/banana",
-        "iab_category": "1 - Education",
+        "iab_category": "5 - Education",
```
ncloudioj, line 22: Updated various IAB categories to reflect the actual values.

merino/providers/wiki_fruit.py
```diff
@@ -0,0 +1,35 @@
...
+                "provider": "test_wiki_fruit",
+                "advertiser": "test_advertiser",
+                "is_sponsored": False,
+                "icon": "https://en.wikipedia.org/favicon.ico",
+                "score": 0,
+            }
+        ]
+
+    def enabled_by_default(self) -> bool:
+        return True
```
quiiver, line 35: if this is `True` it will show up in the `api/v1/providers` endpoint

test-engineering/contract-tests/volumes/client/scenarios.yml
```diff
@@ -226,8 +185,49 @@ scenarios:
...
                 impression_url: 'https://example.org/impression/banana'
                 click_url: 'https://example.org/click/banana'
-                provider: 'test_remote_settings'
+                provider: 'adm'
                 advertiser: 'Example.org'
+                is_sponsored: false
+                icon: null
+                score: 0.3
+
+  - name: remote_settings__coffee
```
ncloudioj, line 194: I think we have some timing issues somewhere. Moving this scenario down here will make it pass. Otherwise, it will fail even with a longer delay.

ncloudioj:
@Trinaa All the tests passed locally. Looks like there is a permission issue on CI, investigating.

merino/providers/wiki_fruit.py
quiiver: Actually, maybe we enforce a `providers.<provider_name>.enabled` config flag and get rid of this method. then use a little meta programming to configured the providers on startup

ncloudioj: Oops, was gonna add a comment here but forgot 🤦‍♂️.

To facilitate testing, I ended up adding a provider level toggle [here](https://github.com/mozilla-services/merino-py/pull/28/files#diff-18ddb9d44bf21da9baf8584b894a8c5eb716a04ed3584f83582cace5db753024R48) and it only gets [created](https://github.com/mozilla-services/merino-py/pull/28/files#diff-7fe7d6e70f4aca4c64c129832043c299b8ff27ab517c08b94cc0f8defa9921a9R24) if that toggle is enabled. I find that using this method alone is sort of awkward for certain things. We might want to refactor this later.

ncloudioj: FWIW, I added a similar [toggle](https://github.com/mozilla-services/merino-py/pull/23/files#diff-18ddb9d44bf21da9baf8584b894a8c5eb716a04ed3584f83582cace5db753024R52) for the adm provider in another PR.

test-engineering/contract-tests/volumes/client/scenarios.yml
Trinaa: That's indicative of an interaction imo. I will look more at that...

test-engineering/contract-tests/volumes/client/scenarios.yml
Trinaa: Since this is a copy of merino scenarios, I'm wondering if there is a problem there or if this is changed in merino-py?

test-engineering/contract-tests/volumes/kinto/data-01.json
Trinaa: Should we make these changes in merino as well?

test-engineering/contract-tests/docker-compose.yml
```diff
@@ -37,7 +37,7 @@ services:
       /wait-for-it.sh merino:8000 --strict -- pytest -vv
 
   kinto:
-    image: mozilla/kinto-dist:25.0.0
+    image: mozilla/kinto-dist:27.0.2
```
ncloudioj, line 40: Piggyback bumping up kinto to the latest version.

test-engineering/contract-tests/volumes/kinto/data-01.json
ncloudioj: Yes, I think so.

test-engineering/contract-tests/volumes/client/scenarios.yml
ncloudioj: It just occurred to me that we're doing this logic slightly differently in merino-py and merino-rs:

```
# merino-rs
is_sponsored = iab_cateogry != "5 - Education"

# merino-py
is_sponsored = idb_category == "22 - Shopping"
``` 

I think merino-py is the correct one.

test-engineering/contract-tests/volumes/client/scenarios.yml
Trinaa: @ncloudioj 
- The tests `remote_settings__coffee` and `multiple_providers__banana` are giving a false positive when executing after `remote_settings__offline_expansion_orange`. 
- `remote_settings__offline_expansion_orange` has an additional set of attachments defined in `offline-expansion-data-01.json` that is interacting, this doesn't matter in merino, but is a factor in merino-py.
- To isolate for now, i recommend changing the order of tests to:
    1. `remote_settings__coffee` 
    2. `multiple_providers__banana`
    3. `remote_settings__offline_expansion_orange`
- Note there is a task for improved isolation: [CONSVC-1840](https://mozilla-hub.atlassian.net/browse/CONSVC-1840)

test-engineering/contract-tests/volumes/client/scenarios.yml
Trinaa: Should we create a task and modify merino?

test-engineering/contract-tests/volumes/kinto/data-01.json
Trinaa: Note. Parity branch [merino-py-parity-CONSVC-1855](https://github.com/mozilla-services/merino/tree/test/merino-py-parity-CONSVC-1855)

test-engineering/contract-tests/docker-compose.yml
Trinaa: I created a parity branch in [merino](https://github.com/mozilla-services/merino/tree/test/merino-py-parity-CONSVC-1855). Copying over the scenario, kinto.ini and docker-compose changes, but the CI is failing atm.

dev/kinto.ini
```diff
@@ -20,6 +23,24 @@ kinto.attachment.base_url =
 kinto.attachment.extra.base_url = http://localhost:8889/
 kinto.attachment.folder = {bucket_id}/{collection_id}
 
+kinto.signer.resources =
```
hackebrot, line 26: Hi @ncloudioj! 👋🏻 

What are these changes for? I'm surprised we have to modify this Kinto configuration in order to get the existing contract tests to pass for merino-py. Is this because of the Kinto upgrade (see https://github.com/mozilla-services/merino-py/pull/28#discussion_r956207898)?

test-engineering/contract-tests/volumes/client/scenarios.yml
ncloudioj: For context, we currently only have two suggestion categories (`5 - Education` and `22 - Shopping` and this is enforced by the suggestion ingestion task) in the source, and we don't foresee having other cateogries in the near future. So the above handling in merino-py and merino-rs are logically identical. Hence, I'd recommend updating our test fixture of the contract tests to only use those exact two categories to reflect what we have in production.

@Trinaa what do you think?

dev/kinto.ini
ncloudioj: Correct, those are entirely for the Kinto upgrade. We're using a quite old version of Kinto, that's why I piggybacked it along with this PR. If we preferred to do that in a separate PR, that's also fine with me.

test-engineering/contract-tests/volumes/client/scenarios.yml
Trinaa: Thank you for explaining. I copied these changes to merino for parity and the outcomes are the same.

test-engineering/contract-tests/docker-compose.yml
Trinaa: Since the kinto-dist changes are causing CI failure in merino-rs. I suggest we do this upgrade at a later point, to preserve parity for now between merino-py and merino-rs.

dev/kinto.ini
Trinaa: see https://github.com/mozilla-services/merino-py/pull/28#discussion_r957537763

ncloudioj: Have dropped this change to minimize the scope. We can upgrade kinto in a follow-up.

Trinaa (approved):

quiiver (approved):
This looks good. We should create a followup ticket for improving the api for enabling providers and configuring the defaults. The initial class instance based method was created before we had a configuration system and could use a refactoring.

PR #32, opened 2022-08-29
[CONSVC-1941] feat: separate suggest log

tiftran:
Tried to mimick merino logs as best as I can for merino-py. What's still currently missing from the logs compared to the rust version of merino:

- os_family
- accepts_english
- form_factor
- browser

### DIFF
--- a/merino/middleware/logging.py
+++ b/merino/middleware/logging.py
@@ -13,7 +13,7 @@ suggest_request_logger = logging.getLogger("web.suggest.request")
 logger = logging.getLogger("request.summary")
 
 # The path pattern for the suggest API
-PATTERN: Pattern = re.compile(r"/api/v[1-9][0-9]*/suggest$")
+PATTERN: Pattern = re.compile(r"/api/v[1-9]\d*/suggest$")
 
 
 class LoggingMiddleware(BaseHTTPMiddleware):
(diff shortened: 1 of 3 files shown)


merino/middleware/logging.py
```diff
@@ -6,6 +6,7 @@
 from starlette.requests import Request
 
 logger = logging.getLogger("request.summary")
```
ncloudioj, line 8: Nit: let's add a brief comment here to describe what those two loggers are used for.

merino/middleware/logging.py
```diff
@@ -15,26 +16,44 @@ class LoggingMiddleware(BaseHTTPMiddleware):
...
-            "sequence_no": int(seq)
-            if (seq := request.query_params.get("seq"))
-            else None,
-            "country": request.state.location.country,
-            "region": request.state.location.region,
-            "city": request.state.location.city,
-            "dma": request.state.location.dma,
-        }
-        logger.info("", extra=data)
+        if request.url.path.endswith("/api/v1/suggest"):
```
ncloudioj, line 19: To free us from coming back to here again when bumping up the API version in the future, let's do a regex matching here instead:

 ```suggestion
    import re
    from typing imoprt Pattern
    
    # The path pattern for the suggest API 
    PATTERN: Pattern = re.compile(r"/api/v[1-9][0-9]*/suggest$")
 
    if PATTERN.match(request.url.path):
         ...
```

ncloudioj (changes requested):
Looks good, had two review comments.

ncloudioj (approved):
Looks great 🚢

PR #33, opened 2022-08-30
refactor: improve testability of business logic

hackebrot:
This is work in progress. 🚧

The goal of this pull request is to demonstrate how composition and protocols can help us increase testability. I've added a unit test for the new cron job to show how we can test async functions.

This change increases test coverage for `merino/providers/adm.py` from 33% to 38%. The new `merino/providers/cron.py` file has 100% test coverage. 📈

### DIFF
--- a/merino/providers/adm.py
+++ b/merino/providers/adm.py
@@ -8,6 +8,7 @@ from typing import Any
 from merino import remotesettings
 from merino.config import settings
 from merino.providers.base import BaseProvider
+from merino.providers.cron import CronJob
 
 logger = logging.getLogger(__name__)
 
@@ -42,37 +43,14 @@ class Provider(BaseProvider):
 
     async def initialize(self) -> None:
         await self._fetch()
-        self.cron_task = asyncio.create_task(self._cron())
-
-    async def _cron(self):
-        """
-        A cron job that resyncs data from Remote Settings in the background.
-
-        Note that the cron job interval and the resync interval are configured separately to
-        facilitate the retry logic upon sync failures.
-        """
-
-        last_tick: float = time.time()
-        while True:
-            if self._should_fetch():
-                begin = time.perf_counter()
-                try:
-                    await self._fetch()
-                except Exception as e:
-                    logger.warning(
-                        "Cron: failed to fetch data from Remote Setting",
-                        extra={"error message": f"{e}"},
-                    )
-                else:
-                    logger.info(
-                        "Cron: resync data from Remote Settings",
-                        extra={"duration": time.perf_counter() - begin},
-                    )
-            sleep_duration = max(
-                0, settings.providers.adm.cron_interval_sec + last_tick - time.time()
-            )
-            await asyncio.sleep(sleep_duration)
-            last_tick = time.time()
+        # Run a cron job that resyncs data from Remote Settings in the background.
+        cron_job = CronJob(
+            name="resync_rs_data",
+            interval=settings.providers.adm.resync_interval_sec,
+            condition=self._should_fetch,
+            task=self._fetch,
+        )
+        asyncio.create_task(cron_job())
 
     def _should_fetch(self) -> bool:
         """
--- a/merino/providers/adm.py
+++ b/merino/providers/adm.py
@@ -5,10 +5,9 @@ from asyncio import as_completed
 from enum import Enum, unique
 from typing import Any
 
-from merino import remotesettings
+from merino import cron, remotesettings
 from merino.config import settings
 from merino.providers.base import BaseProvider
-from merino.providers.cron import CronJob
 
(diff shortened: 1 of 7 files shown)


merino/providers/cron.py
```diff
@@ -0,0 +1,66 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
```
quiiver, line 1: Should we move this to the top level module since it's no longer specific to `providers` ?

hackebrot: Yes, good idea! 👍🏻

pyproject.toml
```diff
@@ -57,6 +57,7 @@ pytest-cov = "^3.0.0"
 pre-commit = "^2.20.0"
 pytest-mock = "^3.8.2"
 bandit = "^1.7.4"
+pytest-asyncio = "^0.19.0"
```
quiiver, line 60: `@pytest.mark.asyncio` works for me without having to include this dependency

quiiver (approved):
Nice, I like the testability and that this keeps the provider code specific to the purpose of providing suggestions.

tests/providers/test_cron.py
```diff
@@ -0,0 +1,84 @@
...
+async def test_cron(caplog: Any, cron_job: CronJob, numbers: list[int]) -> None:
+    """Test for the CronJob implementation."""
+
+    caplog.set_level(logging.INFO)
+
+    # Schedule the task like adm.Provider does it
+    cron_task = asyncio.create_task(cron_job())
+
+    # Cancel the task after 3 seconds
+    await asyncio.sleep(3)
```
Trinaa, line 65: Is it possible to have a shorter task run, in order to keep the suite running quickly?

Trinaa (approved):
Very Nice 🚀. I suggestion about time as discussed earlier.

merino/providers/adm.py
```diff
@@ -42,37 +43,14 @@ class Provider(BaseProvider):
...
-            await asyncio.sleep(sleep_duration)
-            last_tick = time.time()
+        # Run a cron job that resyncs data from Remote Settings in the background.
+        cron_job = CronJob(
+            name="resync_rs_data",
+            interval=settings.providers.adm.resync_interval_sec,
+            condition=self._should_fetch,
+            task=self._fetch,
+        )
+        asyncio.create_task(cron_job())
```
ncloudioj, line 53: This task needs to be stored here otherwise it will get garbage collected because asyncio's runtime only holds a weak reference to it.

merino/providers/cron.py
```diff
@@ -0,0 +1,66 @@
...
+
+
+class Condition(Protocol):
+    """Check whether the cron task should run."""
+
+    def __call__(self) -> bool:  # pragma: no cover
+        ...
+
+
+class Task(Protocol):
```
ncloudioj, line 20: Nit: this would collide with `asyncio.Task`, shall we consider renaming this to something like `CronTask`?

ncloudioj (changes requested):
I like this refactor, thanks! @hackebrot 

Had a few review comments, otherwise, it looks great.

pyproject.toml
hackebrot: Does pytest run the test for you without the marker? I can run the suite without failures without the plugin, but pytest skips the marked test and logs a message about the missing plugin.

merino/providers/cron.py
hackebrot: That's a good question! @Trinaa and I recently had a similar conversation about `requests` in the scope of contract tests. My recommendation would be to not prefix our objects to accommodate for dependencies but refer to objects with conflicting names using their respective module name.

```python
import asyncio
from merino import cron

asyncio_task: asyncio.Task
cron_task: cron.Task
```

In this particular case `cron.Task` is a `Protocol` and is unlikely to be referenced outside of unit tests or the `cron.py` module itself. Objects that implement that protocol will be considered a `cron.Task` implicitly. I'll update `adm.py` to use `cron.Job`.

tests/providers/test_cron.py
hackebrot: Yes! I changed `interval` to be a `float` and managed to get the tests working with `0.1` intervals and a shorter sleep.

merino/providers/cron.py
ncloudioj: Sounds good to me. Tanks for sharing the context!

ncloudioj (approved):
:shipit:

tests/providers/test_cron.py
Trinaa: Ran them locally. It's much faster now ty!

hackebrot:
Thanks for the reviews, team! 🥳

PR #34, opened 2022-08-30
[CONSVC-1964] feat(logging): Add an exception log hander and pretty formatter

ncloudioj:
This adds:
* An error log handler to capture/log exceptions
* A pretty log formatter (based on [rich](https://github.com/Textualize/rich)) for development. Note that this is meant to be used only for development. In production, we will stick to the vanilla MozLog. Hence the package `rich` was included as a dev dependency.

This fixes [CONSVC-1964](https://mozilla-hub.atlassian.net/browse/CONSVC-1964).


quiiver (approved):
🚢

PR #36, opened 2022-09-01
WIP: pydocs

tiftran:
WIP: enabling pydocstyle as part of `make lint` as well as editing/adding docstrings


tiftran (commented):
So there were a few things I wasn't sure how to describe, especially the feature flag stuff, was hoping I could get some guidance?

Turns out you can pick a subset of rules and ignore a few at the same time after all 😄. My rationale to ignore the rules that I chose were either, they were conflicting with another rule, or trying to comply to the rule was too much trouble ie the spacing is not right or the first line has to end with a "."

merino/cron.py
```diff
@@ -1,3 +1,4 @@
+"""Cron Module"""
```
tiftran, line 1: this probs need some work too

ncloudioj: We can do: "A utility class that can be used to implement cron jobs based on asyncio tasks".

merino/cron.py
```diff
@@ -14,13 +15,15 @@ class Condition(Protocol):
     """Check whether the cron task should run."""
 
     def __call__(self) -> bool:  # pragma: no cover
+        """TODO"""
```
ncloudioj, line 18: Is there a way to skip checking for this line? Like what we did for coverage `# pragma: no cover`.

merino/cron.py
```diff
@@ -35,14 +38,14 @@ class Job:
     def __init__(
         self, *, name: str, interval: float, condition: Condition, task: Task
     ) -> None:
+        """Init Job"""
```
ncloudioj, line 41: This is ugly. I don't think we need to document those magic methods in python. __init__, __call__, __eq__, etc. Let's manually ignore checking for those methods if there isn't a rule to do so.

merino/featureflags.py
```diff
@@ -1,3 +1,4 @@
+"""TODO"""
```
ncloudioj, line 1: The implementation of feature flags (feature toggles) for Merino.

merino/featureflags.py
```diff
@@ -64,14 +74,15 @@ def __getitem__(self, flag_name) -> FeatureFlag | None:
...
 class FeatureFlags:
-    """
-    A very basic implementation of featureflags using dynaconf as the configuration system.
-    It supports two bucketing schemes `random` and `session`. Random does what it says on the tin.
-    It generates a random bucketing id for every flag check. Session bucketing uses the session id
-    of the request as the bucketing key so that feature checks within a given search session would
-    be consistent. Additionally you can pass a custom bucketing_id via the `bucket_for` parameter.
-    This is useful when you have an ad-hoc bucketing identifier that is not supported via one of
-    the standard schemes.
+    """A very basic implementation of featureflags using dynaconf as the
```
ncloudioj, line 77: Let's break this down to:
```suggestion
    """A very basic implementation of feature flags using dynaconf as the
     configuration system.
     
    It supports two bucketing schemes `random` and
    `session`. Random does what it says on the tin. It generates a random
    bucketing id for every flag check. Session bucketing uses the session id of
    the request as the bucketing key so that feature checks within a given
    search session would be consistent. Additionally you can pass a custom
    bucketing_id via the `bucket_for` parameter. This is useful when you have
    an ad-hoc bucketing identifier that is not supported via one of the
    standard schemes.
```

merino/middleware/featureflags.py
```diff
@@ -1,3 +1,4 @@
+"""TODO:Feature Flag Module."""
```
ncloudioj, line 1:
```suggestion
"""The middleware that configures features flags for Merino"""
```

merino/middleware/featureflags.py
```diff
@@ -6,11 +7,12 @@
 
 class FeatureFlagsMiddleware(BaseHTTPMiddleware):
     """
-    Sets the a ContextVar for session_id so that it can be used
-    to consistently bucket flags within a search session
+    Sets a ContextVar for session_id so that it can be used
```
ncloudioj, line 10:
```suggestion
    """Sets a ContextVar for session_id so that it can be used
```

merino/middleware/geolocation.py
```diff
@@ -1,3 +1,4 @@
+"""TODO: Geolocation."""
```
ncloudioj, line 1:
```suggestion
"""The middleware that parses geolocation from the client IP address."""
```

merino/middleware/logging.py
```diff
@@ -1,3 +1,4 @@
+"""Logging Value Configuration"""
```
ncloudioj, line 1:
```suggestion
"""The middleware that records various access logs for Merino."""
```

merino/providers/adm.py
```diff
@@ -111,12 +104,15 @@ async def _fetch(self) -> None:
         self.last_fetch_at = time.time()
 
     def enabled_by_default(self) -> bool:
+        """TODO: help please"""
```
ncloudioj, line 107:
```suggestion
        """Enabled by default."""
```

merino/providers/adm.py
```diff
@@ -111,12 +104,15 @@ async def _fetch(self) -> None:
...
         return True
 
     def hidden(self) -> bool:
+        """TODO: help please"""
```
ncloudioj, line 111: ditto

merino/providers/base.py
```diff
@@ -18,16 +17,20 @@ async def initialize(self) -> None:
 
     @abstractmethod
     async def query(self, query: str) -> list[dict[str, Any]]:
+        """TODO:"""
```
ncloudioj, line 20:
```suggestion
        """Query against this provider.
        
        Args:
          - `query`: the query string.
        """
```

merino/providers/base.py
```diff
@@ -18,16 +17,20 @@ async def initialize(self) -> None:
...
         ...
 
     @abstractmethod
     def enabled_by_default(self) -> bool:
+        """TODO"""
```
ncloudioj, line 25:
```suggestion
        """Is this provider enabled by default?"""
```

merino/providers/base.py
```diff
@@ -18,16 +17,20 @@ async def initialize(self) -> None:
...
         ...
 
     def hidden(self) -> bool:
+        """TODO"""
```
ncloudioj, line 29:
```suggestion
        """The hidden flag of the provider."""
```

merino/providers/base.py
```diff
@@ -18,16 +17,20 @@ async def initialize(self) -> None:
...
         return False
 
     def availability(self) -> str:
+        """TODO"""
```
ncloudioj, line 33:
```suggestion
        """Return the status of this provider."""
```

merino/remotesettings.py
```diff
@@ -1,3 +1,4 @@
+"""RemoteSettings Client"""
```
ncloudioj, line 1:
```suggestion
"""A utility class for Remote Settings client"""
```

merino/remotesettings.py
```diff
@@ -7,25 +8,32 @@
 
 
 class Client:
+    """Remote Settings Client Class."""
```
ncloudioj, line 11:
```suggestion
    """A utility class for Remote Settings client."""
```

merino/remotesettings.py
```diff
@@ -7,25 +8,32 @@
...
+
     client: kinto_http.AsyncClient
     attachment_host: str = ""
 
     def __init__(self) -> None:
+        """Init Remote Settings Client"""
         self.client = kinto_http.AsyncClient(server_url=settings.remote_settings.server)
 
     async def fetch_attachment_host(self) -> str:
+        """TODO: Get attachment host"""
```
ncloudioj, line 21:
```suggestion
        """Fetch the attachment host from the Remote Settings server."""
```

merino/remotesettings.py
```diff
@@ -7,25 +8,32 @@
...
         server_info = await self.client.server_info()
         return server_info["capabilities"]["attachments"]["base_url"]
 
     async def get(self, bucket, collection) -> list[dict]:
+        """TODO: Get records from specified collection and bucket"""
```
ncloudioj, line 26:
```suggestion
        """Get records from Remote Settings server.
        
        Args:
          - `collection`: the collection name
          -  `bucket`: the bucket name
        """
```

merino/remotesettings.py
```diff
@@ -7,25 +8,32 @@
...
     async def fetch_attachment(self, attachement_uri) -> httpx.Response:
+        """TODO: Get attachment"""
         if not self.attachment_host:
             self.attachment_host = await self.fetch_attachment_host()
         uri = urljoin(self.attachment_host, attachement_uri)
         async with httpx.AsyncClient() as client:
             return await client.get(uri)
 
     def get_icon_url(self, icon_uri: str) -> str:
+        """TODO: Get Icon Url"""
```
ncloudioj, line 38:
```suggestion
        """Get the URL for an icon.
        
        Args:
          - `icon_uri`: a URI path for an icon stored on Remote Settings
        """
```

ncloudioj (approved):
@tiftran Thanks for taking the lead in this effort! This is super important for us to enforce/maintain the source-code level doc quality for this project.

Dropped some comments to fill out those TODO placeholders. Also let's see if we can find a way to skip checking those builtin methods, if not, feel free to use the placeholders for now.

Feel free to merge once all those review comments are resolved.

PR #37, opened 2022-09-01
refactor: Simplify suggestion index building for adM provider

ncloudioj:
As we're getting more keywords and suggestions from adM, this should save us some allocations and dict lookups while building indices for adM suggestions.

### DIFF
--- a/merino/providers/adm.py
+++ b/merino/providers/adm.py
@@ -95,8 +95,12 @@ class Provider(BaseProvider):
 
     async def _fetch(self) -> None:
         """Fetch suggestions, keywords, and icons from Remote Settings."""
+        # A dictionary keyed on suggestion keywords, each value stores an index
+        # (pointer) to one entry of the suggestion result list.
         suggestions: dict[str, int] = {}
+        # A list of suggestion results.
         results: list[dict[str, Any]] = []
+        # A dictionary of icon IDs to icon URLs.
         icons: dict[int, str] = {}
 
         suggest_settings = await self.backend.get(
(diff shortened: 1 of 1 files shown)


merino/providers/adm.py
```diff
@@ -118,11 +122,11 @@ async def _fetch(self) -> None:
             res = await done_task
             for suggestion in res.json():
                 id = len(results)
-                for kw in suggestion.get("keywords"):
+                for kw in suggestion.pop("keywords", []):
```
quiiver, line 125: TIL about `dict.pop` cool

quiiver (approved):
looks good

merino/providers/adm.py
ncloudioj: Just realized that Merino-py currently uses the original keyword (or the partial) as the `full_keyword` (which is used by Firefox to highlight the current query string and the matched keyword) in the API response, whereas both Merino-rs and Firefox (Offline) would use the minimal-covering-keywords as the `full_keyword`. For example, `wikipedia` would be the `full_keyword` for all the partials/keywords below.

```
wiki
wikip
wikipe
wikiped
wikipedi
wikipedia
```

So, if we pop out the `keywords` list here, we will need to pre-calculate the `full_keyword` field for every single keyword in the list.

ncloudioj: I will file a follow-up ticket for that.

quiiver: What is the current logic for calculating the `full_keyword`

ncloudioj: See [here](https://searchfox.org/mozilla-central/source/browser/components/urlbar/UrlbarQuickSuggest.sys.mjs#199-220), Merino has the identical implementation.

I filed [CONSVC-1966](https://mozilla-hub.atlassian.net/browse/CONSVC-1966) to track this.

PR #38, opened 2022-09-01
[CONSVC-1932] Refactor provider to depend on RS backend protocol

hackebrot:
## Description

Following the theme of #33, this pull request refactors the adm provider to improve testability.

It updates the provider to explicitly define the interfaces/protocols it depends on. This allows us to set up a fake remote settings client that implements that interface and use that in automated tests. It returns fake data that we can use to verify the provider implementation.

Outside of tests, we pass the live remote settings client to the provider (see `merino/providers/__init__.py`).

This pull request increases the test coverage for `merino/providers/adm.py` from 37% to 91%. I recommend adding unit tests for the `query()` method in a separate pull request. 🤖 

## Issue(s)

[CONSVC-1932](https://mozilla-hub.atlassian.net/browse/CONSVC-1932)

### DIFF
--- a/merino/providers/adm.py
+++ b/merino/providers/adm.py
@@ -15,17 +15,17 @@ from merino.providers.base import BaseProvider
 logger = logging.getLogger(__name__)
 
 
-class RemoteSettingsClient(Protocol):
-    """Protocol for a Remote Settings client that this provider depends on.
+class RemoteSettingsBackend(Protocol):
+    """Protocol for a Remote Settings backend that this provider depends on.
 
-    Note: This only defines the methods used by the provider. The actual
-    provider might define additional methods and attributes which this provider
-    doesn't directly depend on.
+    Note: This only defines the methods used by the provider. The actual backend
+    might define additional methods and attributes which this provider doesn't
+    directly depend on.
     """
 
     async def get(
-        self, bucket: Any, collection: Any
-    ) -> list[dict[Any, Any]]:  # pragma: no cover
+        self, bucket: str, collection: str
+    ) -> list[dict[str, Any]]:  # pragma: no cover
         """Get records from Remote Settings."""
         ...
 
@@ -64,11 +64,11 @@ class Provider(BaseProvider):
     icons: dict[int, str] = {}
     last_fetch_at: float
     cron_task: asyncio.Task
-    rs_client: RemoteSettingsClient
+    backend: RemoteSettingsBackend
 
-    def __init__(self, rs_client: RemoteSettingsClient, **kwargs: Any) -> None:
-        """Store the given Remote Settings client on the provider."""
-        self.rs_client = rs_client
+    def __init__(self, backend: RemoteSettingsBackend, **kwargs: Any) -> None:
+        """Store the given Remote Settings backend on the provider."""
+        self.backend = backend
         super().__init__(**kwargs)
 
     async def initialize(self) -> None:
@@ -99,7 +99,7 @@ class Provider(BaseProvider):
         results: list[dict[str, Any]] = []
         icons: dict[int, str] = {}
 
-        suggest_settings = await self.rs_client.get(
+        suggest_settings = await self.backend.get(
             settings.remote_settings.bucket, settings.remote_settings.collection
         )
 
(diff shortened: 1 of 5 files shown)


merino/providers/adm.py
```diff
@@ -3,15 +3,42 @@
...
+class RemoteSettingsClient(Protocol):
+    """Protocol for a Remote Settings client that this provider depends on.
+
+    Note: This only defines the methods used by the provider. The actual
+    provider might define additional methods and attributes which this provider
+    doesn't directly depend on.
+    """
+
+    async def get(
+        self, bucket: Any, collection: Any
```
ncloudioj, line 26: Both of them are of type `str`.

merino/providers/adm.py
```diff
@@ -3,15 +3,42 @@
...
+    ) -> list[dict[Any, Any]]:  # pragma: no cover
```
ncloudioj, line 27:
```suggestion
    ) -> list[dict[str, Any]]:  # pragma: no cover
```

merino/providers/adm.py
```diff
@@ -3,15 +3,42 @@
...
+import httpx
+
+from merino import cron
 from merino.config import settings
 from merino.providers.base import BaseProvider
 
 logger = logging.getLogger(__name__)
 
 
+class RemoteSettingsClient(Protocol):
```
ncloudioj, line 17: Does it make more sense to rename this to `SuggestionBackend` or `ProviderBackend`? There is already a `Client` defined in `remotesettings.py` which is backed by Remote Settings. In the testing environment, we're using a local in-memory backend.

@hackebrot What do you think?

tests/providers/test_adm.py
```diff
@@ -0,0 +1,135 @@
...
+import pytest
+
+from merino.providers.adm import Provider
+
+
+class FakeRSClient:
+    """Fake Remote Settings client that returns suggest data for tests."""
+
+    async def get(self, bucket: Any, collection: Any) -> list[dict[Any, Any]]:
+        """Return fake records.."""
```
ncloudioj, line 18:
```suggestion
        """Return fake records."""
```

ncloudioj (approved):
Nice DI usage here 👍

Looks good, just a few nits for your consideration.

tests/providers/test_adm.py
```diff
@@ -0,0 +1,135 @@
...
+
+def test_hidden(adm: Provider) -> None:
+    """Test for the hidden method."""
+
+    assert adm.hidden() is False
+
+
+@pytest.mark.asyncio
+async def test_initialize(adm: Provider) -> None:
+    """Test for the the initialize() method of the adM provider."""
```
Trinaa, line 119:
```suggestion
    """Test for the initialize() method of the adM provider."""
```

merino/providers/adm.py
```diff
@@ -3,15 +3,42 @@
...
+    """
+
+    async def get(
+        self, bucket: Any, collection: Any
+    ) -> list[dict[Any, Any]]:  # pragma: no cover
+        """Get records from Remote Settings."""
+        ...
+
+    async def fetch_attachment(
+        self, attachement_uri: Any
```
Trinaa, line 32:
```suggestion
        self, attachment_uri: Any
```

Trinaa (approved):
🎉 Woohoo! Increased Coverage!

merino/providers/adm.py
hackebrot: That spelling error is intentional. I should have added a comment about that.

It matches the signature of `merino/remotesettings.py`. I'll push a commit to correct it in both places.

merino/providers/adm.py
hackebrot: Indeed. I set up the protocol to match the signature of the existing client implementation in `merino/remotesettings.py` which currently doesn't specify the types for these arguments.

I'll get that updates in both places. Thanks!

merino/providers/adm.py
hackebrot: I think that's a great idea! 👍🏻

hackebrot: The API defined by this protocol is very specific to Remote Settings currently. My recommendation is to rename this to `RemoteSettingsBackend` as a first step and then update the API to be backend agnostic at a later point if needed.

For instance `fetch_attachment()` is not something I expect to see in a general backend for suggestions. Instead I would expect methods such as `download_suggestions()`.

ncloudioj: Makes sense. Let's use `RemoteSettingsBackend`.

ncloudioj (approved):
🚢

Trinaa (approved):

hackebrot:
Thanks for the reviews @Trinaa and @ncloudioj! 🥳

PR #43, opened 2022-09-06
[CONSVC-1840] test: clear records between scenario executions in merino contract tests

Trinaa:
## Description
* add teardown fixture in merino contract tests
* add delete records function to kinto module
* increase delay in contract test scenarios to assure remote settings resync in merino
* update README

## Issue(s)
[CONSVC-1840](https://mozilla-hub.atlassian.net/browse/CONSVC-1840)

### DIFF
--- a/test-engineering/contract-tests/client/tests/test_merino.py
+++ b/test-engineering/contract-tests/client/tests/test_merino.py
@@ -7,7 +7,13 @@ from typing import Callable, Dict, List, Set, Tuple
 
 import pytest
 import requests
-from kinto import KintoEnvironment, KintoRequestRecord, upload_attachment, upload_icons
+from kinto import (
+    KintoEnvironment,
+    KintoRequestRecord,
+    delete_records,
+    upload_attachment,
+    upload_icons,
+)
 from models import (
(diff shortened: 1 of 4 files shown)


test-engineering/contract-tests/client/tests/test_merino.py
```diff
@@ -186,6 +192,15 @@ def assert_200_response(
             continue
 
 
+@pytest.fixture(scope="function", autouse=True)
```
hackebrot, line 195: We define pytest fixtures in both `conftest.py` and in `test_merino.py`. I think it makes sense to have this fixture here. We probably need guidelines for where to define fixtures going forward.

It may be useful to define fixtures in a specific file or collection of files and load them as local plugins in `conftest.py`. See https://docs.pytest.org/en/7.1.x/how-to/writing_plugins.html#requiring-loading-plugins-in-a-test-module-or-conftest-file

I can help with that if that's what you want to do, @Trinaa. 🤖

hackebrot (approved):
LGTM! 🚀

ncloudioj (approved):
LGTM.

PR #45, opened 2022-09-07
[CONSVC-1935] feat: Parse & log user agent

ncloudioj:
This adds the user agent parsing & logging feature to merino-py.

Have tried a couple of UA parsing libs: [python-user-agents](https://github.com/selwin/python-user-agents), [woothee-python](https://github.com/woothee/woothee-python), and [uap-python](https://github.com/ua-parser/uap-python). Picking uap-python as it's the fastest one (10x faster than the other two) and it's actively developed & maintained compared to others.
 
This fixes [CONSVC-1935](https://mozilla-hub.atlassian.net/browse/CONSVC-1935).

### DIFF
--- a/merino/middleware/user_agent.py
+++ b/merino/middleware/user_agent.py
@@ -1,0 +1,115 @@
+"""The middleware that parses the "User-Agent" from the HTTP request header.
+
+Note that Merino is a service made for Firefox users, this middleware only
+focuses on Firefox related user agents.
+"""
+from typing import Any
+
+from pydantic import BaseModel
+from starlette.middleware.base import BaseHTTPMiddleware
+from starlette.requests import Request
+from ua_parser import user_agent_parser
+
+
+class UserAgent(BaseModel):
+    """Data model for user agent information.
+
+    `browser`: The browser and possibly the version if detected. E.g. 'Firefox(104.0.1)'.
+               'Other' is used if the browser cannot be parsed.
+    `os_family`: The OS family, One of "windows", "macos", "linux", "ios", "android",
+                 "chromeos", or "other".
+    `form_factor`: One of "desktop", "phone", "tablet", or "other".
+    """
+
+    browser: str
+    os_family: str
+    form_factor: str
+
+
+class UserAgentMiddleware(BaseHTTPMiddleware):
+    """A middleware to populate user agent information from the HTTP request header.
+
+    The parsed result `UserAgent` is stored in `Request.state.user_agent`.
+    """
+
+    async def dispatch(self, request: Request, call_next):
+        """Provide user agent information before handling request"""
+        browser, os_family, form_factor = UserAgentMiddleware.parse(
+            request.headers["User-Agent"]
+        )
+        request.state.user_agent = UserAgent(
+            browser=browser, os_family=os_family, form_factor=form_factor
+        )
+
+        return await call_next(request)
+
+    @staticmethod
+    def parse(ua_str: str) -> tuple[str, str, str]:
+        """Parse the "User-Agent" string for browser, os family, and form factor."""
+        ua = user_agent_parser.Parse(ua_str)
+        browser = UserAgentMiddleware._parse_browser(ua["user_agent"])
+        os_family = UserAgentMiddleware._parse_os_family(ua["os"])
+        form_factor = UserAgentMiddleware._parse_form_factor(ua["device"], os_family)
+        return browser, os_family, form_factor
+
+    @staticmethod
+    def _parse_browser(user_agent: dict[str, Any]) -> str:
+        """Parse the browser family and version from the user agent dictionary."""
+        match user_agent:
+            case {
+                "family": "Firefox"
+                | "Firefox iOS"
+                | "Firefox Mobile"
+                | "Firefox Alpha",
+                "major": major,
+                "minor": minor,
+                "patch": patch,
+            }:
+                version = f"{major or ''}.{minor or ''}.{patch or ''}"
+                return "Firefox" if major is None else f"Firefox({version.rstrip('.')})"
+            case {"family": browser}:
+                # Do not bother parsing the browser version for non-Firefox user agents.
+                return browser
+            case _:
+                return "Other"
+
+    @staticmethod
+    def _parse_os_family(operating_system: dict[str, Any]) -> str:
+        """Parse the OS family from the os dictionary."""
+        match operating_system:
+            case {"family": "Windows"}:
+                return "windows"
+            case {"family": "iOS"}:
+                return "ios"
+            case {"family": "Mac OS X"}:
+                return "macos"
+            case {"family": "Android"}:
+                return "android"
+            case {"family": "Chrome OS"}:
+                return "chromeos"
+            case {"family": "Ubuntu" | "Fedora" | "Debian" | "Arch Linux" | "Linux"}:
+                # This should cover the most of main-stream Linux distros
+                return "linux"
+            case _:
+                return "other"
+
+    @staticmethod
+    def _parse_form_factor(device: dict[str, Any], os_family: str) -> str:
+        """Parse the form factor from the device dictionary.
+
+        It akes an extra argument `os_family` to facilitate parsing for Windows
+        and Linux form factors, as the underlying parser doesn't support well
+        for those two platforms.
+        """
+        match device:
+            case {"family": "iPhone" | "Generic Smartphone"}:
+                return "phone"
+            case {"family": "iPad" | "Generic Tablet"}:
+                return "tablet"
+            case {"family": "Mac"}:
+                return "desktop"
+            case {"family": "Other"} if os_family in ["linux", "windows"]:
+                # Hardcode to "desktop" if the os_family is Linux or Windows
+                return "desktop"
+            case _:
+                return "other"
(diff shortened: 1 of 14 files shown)


ncloudioj:
@tiftran With this change, is  `accept-english` the only missing field for the suggest API log?

quiiver (approved):
looks good, just some small comments but non-blocking

merino/middleware/user_agent.py
```diff
@@ -0,0 +1,115 @@
...
+            request.headers["User-Agent"]
+        )
+        request.state.user_agent = UserAgent(
+            browser=browser, os_family=os_family, form_factor=form_factor
+        )
+
+        return await call_next(request)
+
+    @staticmethod
+    def parse(ua_str: str) -> tuple[str, str, str]:
```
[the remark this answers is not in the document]
quiiver: @ncloudioj and i chatted about this and it's probably not worth adding

merino/middleware/user_agent.py
```diff
@@ -0,0 +1,115 @@
...
+    """Data model for user agent information.
+
+    `browser`: The browser and possibly the version if detected. E.g. 'Firefox(104.0.1)'.
+               'Other' is used if the browser cannot be parsed.
+    `os_family`: The OS family, One of "windows", "macos", "linux", "ios", "android",
+                 "chromeos", or "other".
+    `form_factor`: One of "desktop", "phone", "tablet", or "other".
+    """
+
+    browser: str
```
hackebrot, line 24: I wonder if we should rename this to `client` given that there may be consumers of the API which are not browsers. 🤔

tests/middleware/test_user_agent.py
```diff
@@ -0,0 +1,86 @@
+import pytest
+
+from merino.middleware.user_agent import UserAgentMiddleware
+
+FIXTURE = [
```
hackebrot, line 5:
```suggestion
SCENARIO = tuple[str, str, str, str]

SCENARIOS: list[SCENARIO] = [
```

tests/web/test_user_agent.py
```diff
@@ -0,0 +1,33 @@
...
+
+    headers = {
+        "User-Agent": (
+            "Mozilla/5.0 (Macintosh; Intel Mac OS X 11.2;"
+            " rv:85.0) Gecko/20100101 Firefox/103.0"
+        )
+    }
+    client.get("/api/v1/suggest?q=nope", headers=headers)
+
+    assert len(caplog.records) == 1
```
hackebrot, line 27: This might break if we add additional logging unrelated to the user-agent middleware.

Should we filter by logger here?

hackebrot (changes requested):
Hi @ncloudioj! 👋🏻 

The `UserAgentMiddleware` is really light-weight. I like that!

I don't think `parse`, `_parse_browser`, `_parse_os_family`, and `_parse_form_factor` need to be static methods on the starlette middleware class since they are not specific to an HTTP middleware or request. Instead I would define them as functions on the module and probably drop the leading underscore.

I see you've added a number of test scenarios for `parse()` and the code coverage for `merino/middleware/user_agent.py` is at 100%. That's great! 👍🏻 

For the sake of making it easier to debug user-agent parsing related issues, please also add unit tests for the individual functions. Thanks!

merino/middleware/user_agent.py
```diff
@@ -0,0 +1,115 @@
...
+                return "ios"
+            case {"family": "Mac OS X"}:
+                return "macos"
+            case {"family": "Android"}:
+                return "android"
+            case {"family": "Chrome OS"}:
+                return "chromeos"
+            case {"family": "Ubuntu" | "Fedora" | "Debian" | "Arch Linux" | "Linux"}:
+                # This should cover the most of main-stream Linux distros
+                return "linux"
```
Trinaa, line 92:
```suggestion
            case {"family": "Ubuntu" | "Fedora" | "Debian" | "Arch Linux" | "Linux"}:
                # This should cover most of main-stream Linux distros
                return "linux"
```

merino/middleware/user_agent.py
```diff
@@ -0,0 +1,115 @@
...
+            case _:
+                return "other"
+
+    @staticmethod
+    def _parse_form_factor(device: dict[str, Any], os_family: str) -> str:
+        """Parse the form factor from the device dictionary.
+
+        It akes an extra argument `os_family` to facilitate parsing for Windows
+        and Linux form factors, as the underlying parser doesn't support well
+        for those two platforms.
```
Trinaa, line 102:
```suggestion
        It takes an extra argument `os_family` to facilitate parsing for Windows
        and Linux form factors, as the underlying parser doesn't adequately identify
        those two platforms.
```

ncloudioj:
@hackebrot Thanks for your feedback!

> I don't think parse, _parse_browser, _parse_os_family, and _parse_form_factor need to be static methods on the starlette middleware class since they are not specific to an HTTP middleware or request. Instead I would define them as functions on the module and probably drop the leading underscore.

That's a fair point. I think I will define a utility module for ua parsing instead of putting them here.

> For the sake of making it easier to debug user-agent parsing related issues, please also add unit tests for the individual functions. Thanks!

All unit tests are defined in `tests/middleware/test_user_agent.py`, and you're right that only the `parse()` functions were called there, but they should also cover those helper functions as `parse()` is just a wrapper around those underscore functions.

merino/middleware/user_agent.py
```diff
@@ -0,0 +1,115 @@
...
+        request.state.user_agent = UserAgent(
+            browser=browser, os_family=os_family, form_factor=form_factor
+        )
+
+        return await call_next(request)
+
+    @staticmethod
+    def parse(ua_str: str) -> tuple[str, str, str]:
+        """Parse the "User-Agent" string for browser, os family, and form factor."""
+        ua = user_agent_parser.Parse(ua_str)
```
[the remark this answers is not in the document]
ncloudioj: It's not type annotated, but I believe it's of type `dict[str, Any]`

merino/middleware/user_agent.py
```diff
@@ -0,0 +1,115 @@
...
+                | "Firefox Alpha",
+                "major": major,
+                "minor": minor,
+                "patch": patch,
+            }:
+                version = f"{major or ''}.{minor or ''}.{patch or ''}"
+                return "Firefox" if major is None else f"Firefox({version.rstrip('.')})"
+            case {"family": browser}:
+                # Do not bother parsing the browser version for non-Firefox user agents.
+                return browser
```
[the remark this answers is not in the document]
ncloudioj: Don't think so. The value of the `family` key is guaranteed to be a string with "Other" as the fallback. This is enforced by a unit test.

merino/middleware/user_agent.py
ncloudioj: I think `browser` should be fine to use here in this context.

tests/web/test_user_agent.py
ncloudioj: Good point. I will make the change.

hackebrot:
Hi @ncloudioj! 👋🏻 

> All unit tests are defined in `tests/middleware/test_user_agent.py`, and you're right that only the `parse()` functions were called there, but they should also cover those helper functions as `parse()` is just a wrapper around those underscore functions.

That's correct. We have test coverage for the helper functions via `test_ua_parsing` in `tests/middleware/test_user_agent.py`. However should there be a bug in any of these functions that test can only detect that `UserAgentMiddleware.parse(ua)` produces an unexpected result. But it cannot say which function specifically is causing that failure.

A fine grained unit test design pays off longer term by identifying specific causes of failure, enabling existing or new project members to diagnose issues more confidently and develop and deploy fixes more quickly.

merino/middleware/user_agent.py
```diff
@@ -0,0 +1,115 @@
...
+class UserAgentMiddleware(BaseHTTPMiddleware):
+    """A middleware to populate user agent information from the HTTP request header.
+
+    The parsed result `UserAgent` is stored in `Request.state.user_agent`.
+    """
+
+    async def dispatch(self, request: Request, call_next):
+        """Provide user agent information before handling request"""
+        browser, os_family, form_factor = UserAgentMiddleware.parse(
+            request.headers["User-Agent"]
```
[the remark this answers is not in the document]
ncloudioj: I believe that Starlette.request.Header ensure that those following headers are always present:

```
    "accept",
    "accept-encoding",
    "accept-language",
    "cache-control",
    "user-agent"
```

But yeah, let's not take any changes. I will do `.get("User-Agent", "")` instead.

ncloudioj:
> That's correct. We have test coverage for the helper functions via test_ua_parsing in tests/middleware/test_user_agent.py. However should there be a bug in any of these functions that test can only detect that UserAgentMiddleware.parse(ua) produces an unexpected result. But it cannot say which function specifically is causing that failure.

All those static methods are now relocated to their own module. Worth pointing out that only the `parse()` function is exposed but not the case for the rest of the helper functions as they don't mean to be used by other modules.  

> A fine grained unit test design pays off longer term by identifying specific causes of failure, enabling existing or new project members to diagnose issues more confidently and develop and deploy fixes more quickly.

Totally agreed. Though I don't think that'll be a major concern in this case. Should any of those helper functions break, it'll be straightforward to tell what the breakage(s) is as `browser`, `os_family`, and `form_factor` don't overlap with each other. If we expose those helper functions in order to make them easier to test, that doesn't look like a good idea to me either.

Trinaa (approved):

PR #48, opened 2022-09-09
Test/add contract tests for setting refresh consvc 1866 [do not deploy]

taddes:
## Description
* Suggestions from Remote Settings are made available in testing though a call to Kinto to upload new suggestions. 
* Added new data json files to test and verify different impression and click url for same search term.
* Made additions to scenarios file to check same search term and verify result differs between both results.
* Matching search term is verified by a call to merino.
* Local suggestions removed if their counterpart is removed or updated in Remote Settings.

## Issue(s)
[CONSVC-1866](https://mozilla-hub.atlassian.net/browse/CONSVC-1866)

### DIFF
--- a/test-engineering/contract-tests/volumes/kinto/data-04.json
+++ b/test-engineering/contract-tests/volumes/kinto/data-04.json
@@ -1,17 +1,32 @@
 [
     {
-        "id": 7,
+        "id": 6,
         "url": "https://example.org/target/tree2",
         "click_url": "https://example.org/click/tree2",
         "impression_url": "https://example.org/impression/tree2",
         "iab_category": "22 - Shopping",
-        "icon": "2",
+        "icon": "1",
         "advertiser": "Example.org",
         "title": "Tree 2",
         "keywords": [
             "tree",
             "birch tree",
             "maple tree"
+        ]
+    },
+    {
+        "id": 7,
+        "url": "https://example.org/target/flower",
+        "click_url": "https://example.org/click/flower",
+        "impression_url": "https://example.org/impression/flower",
+        "iab_category": "22 - Shopping",
+        "icon": "2",
+        "advertiser": "Example.org",
+        "title": "Flower",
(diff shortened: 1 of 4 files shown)


taddes:
We're noting here that locally, the assertions pass but when ran on CircleCI, we're not getting a match between the refreshed suggestion. Investigating. Didn't happen before so not sure if this is a CircleCI thing and if the order of tests is running as expected.

test-engineering/contract-tests/volumes/kinto/data-04.json
```diff
@@ -0,0 +1,17 @@
...
+        "click_url": "https://example.org/click/tree2",
+        "impression_url": "https://example.org/impression/tree2",
+        "iab_category": "22 - Shopping",
+        "icon": "2",
+        "advertiser": "Example.org",
+        "title": "Tree 2",
+        "keywords": [
+            "tree",
+            "birch tree",
+            "maple tree"
```
Trinaa, line 14: @ncloudioj 

I was wondering about the accuracy of this updated data. For instance should the `id` value also be 6?

ncloudioj: In production, our partner might either add new suggestions with new IDs or make in-place updates to the existing suggestions.

So I'd recommend:
- Adding another suggestion to `data-04.json` with a different set of keywords, and then adding another request/response step in the scenario to ensure Merino can find it. This can be used to verify that the new suggestions are loaded by Merino
- Modifying the existing suggestion for `"id": 6` just like the current record in the pr (i.e. just changed id from `7` to `6`). This can be used to verify that the updated suggestions are loaded by Merino as well

Trinaa: thank you @ncloudioj. I've adjusted the scenario as suggested.

ncloudioj (approved):
LGTM, thanks!

PR #51, opened 2022-09-13
refactor: Switch to geoip2 library from maxminddb

quiiver:
Whe i was looking at some mypy errors i noticed that the maxminddb library we were using hasn't been updated in a year and doesn't have a nice of type definitions for the results of a geoip database lookup. This switches them and makes the test mypy compliant

### DIFF
--- a/merino/middleware/geolocation.py
+++ b/merino/middleware/geolocation.py
@@ -5,8 +5,9 @@ from typing import Optional
 import geoip2.database
 from geoip2.errors import AddressNotFoundError
 from pydantic import BaseModel
-from starlette.middleware.base import BaseHTTPMiddleware
+from starlette.middleware.base import BaseHTTPMiddleware, RequestResponseEndpoint
 from starlette.requests import Request
+from starlette.responses import Response
 
 from merino.config import settings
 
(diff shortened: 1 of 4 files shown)


merino/middleware/geolocation.py
```diff
@@ -35,16 +36,18 @@ async def dispatch(self, request: Request, call_next):
         # `request.client.host` should be the first remote client address of `X-Forwarded-For`.
         record = None
         try:
-            record = reader.get(request.client.host)
+            record = reader.city(request.client.host if request.client else "")
```
ncloudioj, line 39:
```suggestion
            record = reader.city(request.client.host or "")
```


Per Starlette's doc: `request.client` is always exposed as a namedtuple. But `request.client.host` could be None.

> Client Address
>
> The client's remote address is exposed as a named two-tuple `request.client`. Either item in the tuple may be None.
>
> The hostname or IP address: request.client.host
> The port number from which the client is connecting: request.client.port

quiiver: I tried that at first because i prefer that syntax. however, per the starlette code `request.client` is of type `Optional[Address]` and the definition of `Address` is:

```py
class Address(typing.NamedTuple):
    host: str
    port: int
```

merino/middleware/geolocation.py
```diff
@@ -35,16 +36,18 @@ async def dispatch(self, request: Request, call_next):
...
         except ValueError:
```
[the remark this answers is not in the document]
quiiver: It still throws for invalid addresses

merino/middleware/geolocation.py
ncloudioj: Ah, then we have to do `request.client or request.client.host or ""`. Because of that note:

> Either item in the tuple may be None.

ncloudioj: oh nope, that will return `request.client` in most cases. 😅

ncloudioj (approved):
r+ with a nit.

PR #52, opened 2022-09-14
[CONSVC-1947] test: align contract test tech stack with application

Trinaa:
## Description
* update contract-tests client and kinto-setup from python 3.8 to 3.10
* merge 'tests' and 'test-engineering' directories
* align mypy checks
* migrate contract-tests client and kinto-setup from pip-tools to poetry

## Issue(s)
[CONSVC-1947](https://mozilla-hub.atlassian.net/browse/CONSVC-1947)


Trinaa:
@taddes There may be some overlap with your mypy work in the following files:
- .pre-commit-config.yaml
- Makefile
- pyproject.toml

tests/contract/client/Dockerfile
```diff
@@ -8,7 +8,13 @@ ENV PATH="${PYTHON_VENV}/bin:${PATH}"
 
 RUN python -m pip install --upgrade pip
 
-COPY requirements.txt /tmp/requirements.txt
+# Setup poetry and extract requirements
+RUN python -m pip install --no-cache-dir --quiet poetry
+WORKDIR /tmp
+COPY ./pyproject.toml ./poetry.lock /tmp/
+RUN poetry export --no-interaction --dev --output requirements.txt --without-hashes
```
hackebrot, line 15: TIL! 💡

tests/contract/client/pyproject.toml
```diff
@@ -0,0 +1,19 @@
...
+pydantic = "1.10.2"
+pytest = "7.1.3"
+PyYAML = "6.0"
+requests = "2.28.1"
+
+[tool.poetry.dev-dependencies]
+
+[build-system]
+requires = ["poetry-core>=1.0.0"]
+build-backend = "poetry.core.masonry.api"
```
hackebrot, line 19: How did you choose this backend?

tests/contract/client/tests/conftest.py
```diff
@@ -76,13 +76,13 @@ def fixture_kinto_records(request: Any) -> Dict[str, KintoRequestRecord]:
...
-    kinto_records: Dict[str, KintoRequestRecord],
-) -> Callable:
+    kinto_records: dict[str, KintoRequestRecord],
+) -> Callable[[str], str]:
     """Return a function that will query for an icon URL from a suggestion title"""
 
     attachments_url: str = request.config.option.kinto_attachments_url
 
-    def fetch_icon_url(*, suggestion_title: str) -> str:
+    def fetch_icon_url(suggestion_title: str) -> str:
```
hackebrot, line 85: The `*` is so that the method only accepts keyword arguments.

Do you think we should change that?

tests/contract/client/tests/test_merino.py
```diff
@@ -131,8 +133,8 @@ def merino_step(step: Step) -> None:
 
 @pytest.fixture(scope="session", name="step_functions")
 def fixture_step_functions(
-    kinto_step: Callable, merino_step: Callable
-) -> Dict[Service, Callable]:
+    kinto_step: Callable[[Step], None], merino_step: Callable[[Step], None]
```
hackebrot, line 136: We could use a `typing.Protocol` or type alias for this:

```python
StepFunction = Callable[[Step], None]

@pytest.fixture(scope="session", name="step_functions")
def fixture_step_functions(
    kinto_step: StepFunction, merino_step: StepFunction
) -> dict[Service, StepFunction]:
    pass
```

tests/contract/kinto-attachments/.gitignore
```diff
@@ -1,2 +1,2 @@
-*
+main
```
hackebrot, line 1: I believe we have this here so that we can have an empty directory in the repo. What is this change for?

tests/load/tests/locustfile.py
```diff
@@ -9,12 +9,11 @@
 
 import faker
 import kinto_http
+from client_info import DESKTOP_FIREFOX, LOCALES
```
hackebrot, line 12: What this order determined by isort?

tests/load/tests/locustfile.py
```diff
@@ -96,7 +95,7 @@ def request_suggestions(client: HttpSession, query: str) -> None:
     if PROVIDERS:
         params = {**params, "providers": PROVIDERS}
 
-    headers: Dict[str, str] = {
+    headers: Dict[str, str] = {  # nosec
```
hackebrot, line 98: What does `# nosec` do?

hackebrot (changes requested):
Hi @Trinaa! 👋🏻 

I like the new structure! It makes a lot of sense to me. 👍🏻 

I've added a few comments inline.

pyproject.toml
```diff
@@ -62,13 +68,15 @@ black = "^22.6.0"
...
 pydocstyle = "^6.1.1"
-mypy = "^0.961"
+mypy = "^0.971"
 pytest-cov = "^3.0.0"
 pre-commit = "^2.20.0"
 pytest-mock = "^3.8.2"
 bandit = "^1.7.4"
 rich = "^12.5.1"
 pytest-asyncio = "^0.19.0"
+types-PyYAML = "^6.0.11"
```
ncloudioj, line 78: Are these two deps for mypy?

tests/contract/client/pyproject.toml
Trinaa: This must be the default. This file was generated with `poetry init`. Is there a value that would be better?

tests/contract/client/tests/conftest.py
Trinaa: It's that mypy raises the following error. I find the suggestion verbose and opted for less specificity. Perhaps there is another way?
image

tests/contract/client/tests/test_merino.py
Trinaa: This is much better.

tests/contract/kinto-attachments/.gitignore
Trinaa: That's a good catch! ty! I didn't change this explicitly, an IDE side-effect possibly.

tests/load/tests/locustfile.py
Trinaa: Yes.
image

tests/load/tests/locustfile.py
Trinaa: It quiets bandit.
image

pyproject.toml
Trinaa: Yes that's right. Without them we get 'Library stubs not installed' errors.

tests/contract/client/pyproject.toml
hackebrot: I don't have a lot of experience with poetry myself. Looks like we're using the same backend for the merino package. So that seems reasonable to me! 👍🏻

tests/contract/client/tests/conftest.py
hackebrot: Ok, got it! Let's go with your suggest change then!

tests/load/tests/locustfile.py
hackebrot: Okay! 🤖

hackebrot (approved):
It's possible I've missed something in this 70 files diff, but the updated revision looks good to me. Let's merge it! 🚀

ncloudioj (approved):
:shipit:

PR #53, opened 2022-09-15
chore: Do not export dev dependencies to docker iamge

ncloudioj:
Came across this while reviewing another PR. I must've added the `--dev` flag by accident 😅.

We don't need those dev dependencies to run Merino, also they'd bloat the docker image over time.


taddes (approved):
Makes sense to me, @ncloudioj , good catch!

PR #59, opened 2022-09-21
[CONSVC-1965] fix: add exception handling to adm provider for the init fetch

ncloudioj:
This adds an exception handler to the `adm` provider for the initial fetch to Remote Settings. Also added some unit tests to make `adm.py` 100% covered. 

This fixes [CONSVC-1965](https://mozilla-hub.atlassian.net/browse/CONSVC-1965).

### DIFF
--- a/merino/providers/adm.py
+++ b/merino/providers/adm.py
@@ -73,7 +73,17 @@ class Provider(BaseProvider):
 
     async def initialize(self) -> None:
         """Initialize cron job."""
-        await self._fetch()
+        try:
+            await self._fetch()
+        except Exception as e:
+            logger.warning(
+                "Failed to fetch data from Remote Settings, will retry it soon",
+                extra={"error message": f"{e}"},
+            )
+            # Set the last fetch timestamp to 0 so that the cron job will retry
+            # the fetch upon the next tick.
+            self.last_fetch_at = 0
+
         # Run a cron job that resyncs data from Remote Settings in the background.
         cron_job = cron.Job(
             name="resync_rs_data",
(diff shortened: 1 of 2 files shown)


merino/providers/adm.py
```diff
@@ -149,20 +159,20 @@ def hidden(self) -> bool:  # noqa: D102
     async def query(self, q: str) -> list[dict[str, Any]]:
         """Provide suggestion for a given query."""
         if (id := self.suggestions.get(q)) is not None:
-            if (res := self.results[id]) is not None:
```
ncloudioj, line 152: Turns out this check is redundant since it's part of the invariant for the search index of the `adm` provider. I.e, the suggestions in the `self.results` list should never be `None`. Also, `id` should always be a valid index of `self.results` as both `self.suggestions` and `self.results` are immutable once created. All the above invariants are enforced/tested by tests.

tests/unit/providers/test_adm.py
```diff
@@ -133,3 +137,63 @@ async def test_initialize(adm: Provider) -> None:
...
+    records = filter_caplog(caplog.records, "merino.providers.adm")
+
+    assert adm.last_fetch_at == 0
+
+    assert len(records) == 1
+    assert records[0].__dict__["error message"] == "The remote server was unreachable"
+
+    # Clean up the cron task. Unlike other test cases, this action is necessary here
+    # since the cron job has kicked in as the initial fetch fails.
+    adm.cron_task.cancel()
```
Trinaa, line 166: In the event of an assertion failure or other exception, I believe this line won't execute.

Perhaps encase this line in a finally block or use a pytest fixture to assure its execution.

ncloudioj: Good idea. Let's not leave any pending task behind in all cases :)

ncloudioj: Ended up using `try..finally..` here, the following fixture didn't seem to work for me 🤔:
```py
@pytest.fixture(name="adm_with_teardown"):
def fixture_adm_with_teardown() -> Generator[Provider, None, None]:
    adm = Provider(backend=FakeBackend())
    yield adm
    adm.cron_task.cancel()
```

Trinaa: @ncloudioj 

I am unresolving to say, I liked where you are going with the dedicated setup/teardown fixture and I have it working with a few tweaks:
1. The return type of the fixture should be 'Provider'
2. Use 'adm_with_teardown' instead of 'adm' in test signature

Example:
```
@pytest.fixture(name="adm_with_teardown")
def fixture_adm_with_teardown() -> Provider:
    adm = Provider(backend=FakeBackend())
    yield adm
    adm.cron_task.cancel()

@pytest.mark.asyncio
async def test_initialize_failure(
    adm_with_teardown: Provider,
    mocker: MockerFixture,
    caplog: Any,
) -> None:
    """Test exception handling for the initialize() method."""
```


That said, what do you think of modifying the existing `fixture_adm` as follows? Then we don't need the extra fixture or finally clause.

```
@pytest.fixture(name="adm")
def fixture_adm() -> Provider:
    """Supply an adM provider that uses a fake remote settings client."""

    adm: Provider = Provider(backend=FakeBackend())

    yield adm

    # Clean up the cron task if initialized.
    # This action is necessary for tests that may kick-off the cron job, for instance
    # when the initial fetch fails in `test_initialize_failure`
    if hasattr(adm, 'cron_task'):
        adm.cron_task.cancel()
```

ncloudioj: > The return type of the fixture should be 'Provider'

Needed to wrap it inside of `Generator` to appease mypy.

> Use 'adm_with_teardown' instead of 'adm' in test signature

All the tests passed for me, but I was also seeing this at the end of test report:

```
Task was destroyed but it is pending!
task: <Task pending name='Task-7' coro=<Job.__call__() running at /Users/path/to/mozilla/merino-py/merino/cron.py:67> wait_for=<Future cancelled>>
```

I like your above proposal, let me try it to see if that warning would go away.

ncloudioj: Hmm, weird, I was constantly getting that warning when calling `adm.cron_task.cancel()` inside of the `adm` fixture. Didn't happen when calling the cancellation inside of the test function. 🤔

ncloudioj: The only difference being this particular unit test is an async function, whereas the `adm` fixture is not. We could make `adm` an async fixture, but it's also used by non-async test functions. 🤷‍♂️

Trinaa: > Needed to wrap it inside of `Generator` to appease mypy.
> 
> > Use 'adm_with_teardown' instead of 'adm' in test signature
😿 that makes sense ...

I'm coming back around to using the finally block. It works and seems far less complicated.

PR #62, opened 2022-09-26
chore: specify streams for log handlers

ncloudioj:
The default stream of log handlers is `stderr` in Python, which confuses Cloud Logging as it treats all log records from stderr as `ERROR`.

Note that `console-pretty` is used in dev environments only, and it has its own stream implicitly set, so we don't need to configure it separately.


quiiver (approved):
lgtm

PR #63, opened 2022-09-27
[CONSVC-1955][Part 1] fix: enable mypy checks on merino directory

ncloudioj:
Took a first stab at appeasing `mypy` against everything in the `merino` directory. Partially enabled mypy in linting (`make lint`) and pre-commit checking.

@taddes Apologize in advance if this conflicts with anything in your working branch. `mypy` is indeed not 100% accurate, had to manually ignore several cases as I believed those are definitely false alarms.

### DIFF
--- a/merino/providers/__init__.py
+++ b/merino/providers/__init__.py
@@ -41,7 +41,7 @@ async def init_providers() -> None:
             case ProviderType.ADM:
                 providers["adm"] = AdmProvider(
                     backend=(
-                        LiveBackend()
+                        LiveBackend()  # type: ignore
                         if setting.backend == "remote-settings"
                         else TestBackend()
                     ),
--- a/merino/providers/adm.py
+++ b/merino/providers/adm.py
@@ -30,7 +30,7 @@ class RemoteSettingsBackend(Protocol):
         ...
 
     async def fetch_attachment(
-        self, attachment_uri: Any
+        self, attachment_uri: str
     ) -> httpx.Response:  # pragma: no cover
         """Fetch the attachment for the given URI."""
         ...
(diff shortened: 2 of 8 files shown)


merino/providers/__init__.py
```diff
@@ -41,7 +41,7 @@ async def init_providers() -> None:
             case ProviderType.ADM:
                 providers["adm"] = AdmProvider(
                     backend=(
-                        LiveBackend()
+                        LiveBackend()  # type: ignore
```
ncloudioj, line 44: Looks like `mypy` doesn't recognice `LiveBackend` and `TestBackend` as the underlying `RemoteSettingsBackend` protocol, already checked that all methods and argument types matched perfectly. Not sure why 🤷‍♂️

merino/providers/adm.py
```diff
@@ -120,7 +120,7 @@ async def initialize(self) -> None:
 
     def _should_fetch(self) -> bool:
         """Check if it should fetch data from Remote Settings."""
-        return (
+        return (  # type: ignore
```
ncloudioj, line 123: Somehow, `mypy` believed that the return value was of type `Any`.

merino/util/user_agent_parsing.py
```diff
@@ -29,7 +29,7 @@ def _parse_browser(user_agent: dict[str, Any]) -> str:
             return "Firefox" if major is None else f"Firefox({version.rstrip('.')})"
         case {"family": browser}:
```
ncloudioj, line 30: Alternatively, we could match against a `str` value via `case {"family": str(browser)}`. But that would come with a runtime type check, whereas `cast(str, browser)` won't as we know `browser` here is indeed a "str".

merino/util/user_agent_parsing.py
```diff
@@ -54,7 +54,7 @@ def _parse_os_family(operating_system: dict[str, Any]) -> str:
             return "other"
 
 
-def _parse_form_factor(device: dict[str, Any], os_family: str) -> str:
+def _parse_form_factor(device: dict[str, Any], os_family: str) -> str:  # type: ignore
```
ncloudioj, line 57: Looks like another false alarm. The same pattern match above worked just fine.

merino/providers/__init__.py
taddes: I had the same issue. Annoying. I think when I did a little research, `mypy` can struggle with custom types that are imported.

merino/util/user_agent_parsing.py
taddes: :thumbsup:

ncloudioj:
Close this to favor #69

PR #65, opened 2022-09-28
refactor: cache score from settings to avoid loading overhead

ncloudioj:
Notice this while profiling merino-py. Turns out that the Dynaconf setting lookups can be expensive if the target setting is deeply nested in the hierarchy. The profiling result suggests that repeatedly referencing those nested settings such as in middlewares or route handlers would incur a significant overhead. To mitigate that, we can cache those settings whenever possible in those scenarios.

I've checked that this is the only case we have had so far. Will add a note to the relevant document for future reference.

### DIFF
--- a/docs/ops.md
+++ b/docs/ops.md
@@ -60,6 +60,15 @@ such as `config_sentry.py` or `config_logging.py`.
 - `format` (`MERINO_LOGGING__FORMAT`) - Controls the format of outputted logs in
   either `pretty` or `mozlog` format. See [../config_logging.py][log].
 
+### Caveat
+
+Be extra careful whenever you need to reference those deeply nested settings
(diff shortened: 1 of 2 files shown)


quiiver (approved):
lgtm. i'm a little surprised that these values aren't cached by the library.

ncloudioj:
@quiiver Thanks for the review!

Yep, that's indeed surprising. Another interesting finding in `adm.py` is that `import httpx` is 2nd most expensive call in this module 🤣, and it also comes with a 3.6MB memory footprint. I am not entirely sure how accurate that was but looks like we only need it for type annotation. I thought the import overheads for typing should be generally cheap, looks like that varies from case to case, the worst case scenario is that we'd have to use `typing.TYPE_CHECKING` to avoid those expensive mod imports for type checking.

PR #66, opened 2022-09-29
Add an AccuWeather provider

0c0w3:


Trinaa (commented):
Hi @0c0w3  👋🏻 

I hope you don't mind some drive-by comments! 

I also recommend looking into running mypy against the new files. [Looks like that will be enabled soon](https://github.com/mozilla-services/merino-py/pull/63) and become blocking here.

0c0w3:
@Trinaa Thanks for your comments! I will definitely address them. I haven't done it yet because my git/GitHub workflow is rusty and I wanted to make sure I didn't mess anything up before Nan got a chance to look at this.

merino/configs/default.toml
```diff
@@ -54,6 +54,15 @@ env = "dev"
...
 traces_sample_rate = 0.1
 
+[default.providers.accuweather]
+enabled_by_default = false
+score = 0.3
+api_base_url = "https://apidev.accuweather.com"
+api_param_key = "apikey"
+api_forecasts_path = "/forecasts/v1/daily/1day/{location_key}.json"
+api_postalcodes_path = "/locations/v1/postalcodes/{country_code}/search.json"
+api_postalcodes_param_query = "q"
```
ncloudioj, line 64: I'd recommend defining `MERINO_CLIENT_IP_OVERRIDE` here to be consistent with other configurations. Such as

```
# This can be set to facilitate manual testing during development.
client_ip_override = ""
```

```
# Define this at the module level. See here: https://mozilla-services.github.io/merino-py/ops.html#caveat
CLIENT_IP_OVERRIDE: str = settings.providers.accuweather.client_ip_override


ip_address =  CLIENT_IP_OVERRIDE or (
    request.client.host or "" if request.client else ""
)
```

Then this configuration can be overridden via this environment variable `MERINO_PROVIDERS__ACCUWEATHER__CLIENT_IP_OVERRIDE`.

merino/providers/__init__.py
```diff
@@ -38,6 +40,11 @@ async def init_providers() -> None:
     # register providers
     for provider_type, setting in settings.providers.items():
         match provider_type:
+            case ProviderType.ACCUWEATHER:
+                providers["accuweather"] = AccuWeatherProvider(
+                    enabled_by_default=setting.enabled_by_default,
+                )
+                pass
```
ncloudioj, line 47:
```suggestion

```

merino/configs/default.toml
```diff
@@ -54,6 +54,15 @@ env = "dev"
 # A setting for the tracing sample rate. Should be a float in range [0, 1.0].
 traces_sample_rate = 0.1
 
+[default.providers.accuweather]
+enabled_by_default = false
+score = 0.3
+api_base_url = "https://apidev.accuweather.com"
+api_param_key = "apikey"
```
ncloudioj, line 61:
```suggestion
api_param_key = ""
```
It'll be easier to validate this configuration with an empty string.

merino/providers/accuweather.py
```diff
@@ -0,0 +1,154 @@
...
+    async def initialize(self) -> None:
+        """Initialize the provider."""
+        ...
+
+    def hidden(self) -> bool:  # noqa: D102
+        return False
+
+    async def handle_request(self, request: Request) -> list[BaseSuggestion]:
+        """Provide suggestions for a given request."""
+        api_key = os.environ.get("MERINO_ACCUWEATHER_API_KEY")
```
ncloudioj, line 59:
```suggestion
        api_key = settings.providers.accuweather.api_param_key
        if api_key == "":
```

merino/providers/accuweather.py
```diff
@@ -0,0 +1,154 @@
...
+        if api_key is None:
+            logger.warning("AccuWeather API key not specified")
+            return []
+
+        country = None
+        postal_code = None
+        try:
+            country = request.state.location.country
+            postal_code = request.state.location.postal_code
```
ncloudioj, line 68: As part of the middleware rewrite, we no longer store the location data inside of `request.state`. Instead we use a `ContextVar` to store that for each request. Here is an [example](https://github.com/mozilla-services/merino-py/blob/main/merino/middleware/logging.py#L41).

```suggestion
            # need to import the location ContextVar
            from merino.middleware.geolocation import ctxvar_geolocation
            
            # `try except` is not needed here
            location = ctxvar_geolocation.get()
            country = location.country
            postal_code = location.postal_code
            if country is None or postal_code is None:
                 logger.warning("Country and/or postal codes unknown")
                 return []
```

merino/providers/accuweather.py
```diff
@@ -0,0 +1,154 @@
...
+        except AttributeError:
+            logger.warning("Country and/or postal codes unknown")
+            return []
```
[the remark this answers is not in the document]
ncloudioj: Close this to favor https://github.com/mozilla-services/merino-py/pull/66/files#r990559812

merino/configs/default.toml
ncloudioj: Looking at the code below, looks like `api_param_key` is not used to store the API key to Accuweather, correct? We can add a new configuration `api_key` and set its default to `""`.  We will override it with an environment variable (`MERINO_PROVIDERS__ACCUWEATHER__API_KEY`) in production.

merino/providers/accuweather.py
```diff
@@ -0,0 +1,154 @@
...
+                city_name=location.get("LocalizedName"),
+                temperature_unit=get_value(
+                    forecast, ["Temperature", "Maximum", "Unit"]
+                ),
+                high=get_value(forecast, ["Temperature", "Maximum", "Value"]),
+                low=get_value(forecast, ["Temperature", "Minimum", "Value"]),
+                day_summary=get_value(forecast, ["Day", "IconPhrase"]),
+                day_precipitation=get_value(forecast, ["Day", "HasPrecipitation"]),
+                night_summary=get_value(forecast, ["Night", "IconPhrase"]),
+                night_precipitation=get_value(forecast, ["Night", "HasPrecipitation"]),
```
ncloudioj, line 152: I think we can extract all those data points by using a pattern matching statement against `forecast` as follows:

```
match forecast:
    case { 
        "Tempreture": {
            "Maximum": { "Value": high, "Unit": tempreture_unit },
            "Minimum": { "Value": low },
        },
        "Day": { "IconPhrase": day_summary, "HasPrecipitation": day_precipitation },
        "Night": { "IconPhrase": night_summary, "HasPrecipitation": night_precipitation },
     }:
        Suggestion(title=..., temperature_unit=temperature_unit, high=high, low=low, ...)
    case _:
         # Does not match, perhaps do some error handling here
```

ncloudioj (changes requested):
@0c0w3 Apologies again for the delayed review.

Took the first pass and overall it looks good. Let's resolve those review comments first, I will take another closer look next week. Thanks!

merino/configs/default.toml
0c0w3: Thanks, I see. I think it makes more sense for this to live under the `location` config than `providers.accuweather`, so I'll put it there if that's OK.

merino/configs/default.toml
0c0w3: Right, it's the name of the query param that should be set to the key. (It looks like the provider name has to be specified in lowercase as part of the environment variable, like `MERINO_PROVIDERS__accuweather__API_KEY`.)

0c0w3: I'll come up with better names for these.

merino/providers/accuweather.py
0c0w3: I see. That means my commit where I replaced `query()` with `handle_request()` isn't necessary at all, nice.

merino/providers/accuweather.py
0c0w3: Nice, thanks!

merino/providers/accuweather.py
```diff
@@ -0,0 +1,154 @@
...
+logger = logging.getLogger(__name__)
+
+
+def get_value(d: dict, keys: list, default_value: Any = None) -> Any:
+    """Get a nested value in a sequence of dictionaries."""
+    for k in keys:
+        if d is None or not isinstance(d, dict):
+            return default_value
+        d = d.get(k)
+    return d
```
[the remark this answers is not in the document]
0c0w3: Thanks, this function turned out not to be necessary with Nan's `match` suggestion.

merino/middleware/geolocation.py
```diff
@@ -36,10 +38,15 @@ async def dispatch(
         self, request: Request, call_next: RequestResponseEndpoint
     ) -> Response:
         """Provide Geolocation before handling request"""
+        # The `MERINO_CLIENT_IP_OVERRIDE` environment variable can be set to
+        # facilitate manual testing during development.
         # `request.client.host` should be the first remote client address of `X-Forwarded-For`.
         record = None
+        ip_address = os.environ.get("MERINO_CLIENT_IP_OVERRIDE") or (
+            request.client.host or "" if request.client else ""
+        )
```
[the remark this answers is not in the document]
0c0w3: I think this looks nicer now with Nan's `CLIENT_IP_OVERRIDE` suggestion, so I went ahead and left this as-is.

merino/providers/accuweather.py
0c0w3: I ended up leaving my `handle_request()` change because I think it still makes sense when some providers depend on certain query params and others don't. I'm not sure if that's the best way to handle it, let me know if not.

0c0w3:
OK, I addressed comments and this is ready for review again. As I mentioned in another comment, I left my `handle_request()` change because it seems reasonable when some providers depend on certain query params (and potentially other parts of the request) and others don't. I'm open to other ways of handling that.

merino/providers/accuweather.py
```diff
@@ -0,0 +1,161 @@
...
+        return False
+
+    async def handle_request(self, request: Request) -> list[BaseSuggestion]:
+        """Provide suggestions for a given request."""
+        suggestions = await self.query()
+        return suggestions
+
+    async def query(self) -> list[BaseSuggestion]:
+        """Provide suggestions that don't depend on a client query string."""
+        api_key = settings.providers.accuweather.api_key
```
quiiver, line 55: We should move config lookups to either the module level or the class `__init__`

ncloudioj: See more details [here](https://mozilla-services.github.io/merino-py/ops.html#caveat). And here is an [example](https://github.com/mozilla-services/merino-py/blob/main/merino/providers/adm.py#L81-L83).

merino/providers/accuweather.py
```diff
@@ -0,0 +1,169 @@
...
+            return []
+
+        location = ctxvar_geolocation.get()
+        country = location.country
+        postal_code = location.postal_code
+        if country is None or postal_code is None:
+            logger.warning("Country and/or postal code unknown")
+            return []
+
+        async with httpx.AsyncClient(app=self._app, base_url=URL_BASE) as client:
```
ncloudioj, line 79: Any specific reason for providing the `app` argument here? I believe the doc says it's optional, if it's specified, the request will be sent to this ASGI `app` rather than sending the actual network request.

Also, since `query()` will be called by each request handler of this provider, creating a new client for each one of them might have high overhead. We might want to use a connection pool to alleviate that. Could be done in a followup.

0c0w3: The app argument lets the test create a provider instance with a mock server and mock responses essentially, see test_accuweather.py. I don't know if that's the proper way to write a test for this but it's an approach I found that works well. Is there a better way?

> Also, since `query()` will be called by each request handler of this provider, creating a new client for each one of them might have high overhead.

That makes sense and it's the kind of thing I might need help on. Ideally Merino would also cache responses from AccuWeather. `AsyncClient` doesn't seem to do that on its own? I found this: https://github.com/johtso/httpx-caching

So I guess I would like to leave this and other caching to a follow up, if that's OK.

0c0w3: The [docs](https://www.python-httpx.org/advanced/#more-efficient-usage-of-network-resources) say `Client` instances use connection pooling. So would pooling be as simple as creating the client in `__init__()` as an attribute?

merino/providers/accuweather.py
```diff
@@ -0,0 +1,169 @@
...
+                    {
+                        URL_PARAM_API_KEY: API_KEY,
+                        URL_POSTALCODES_PARAM_QUERY: postal_code,
+                    }
+                ),
+                "",
+            )
+        )
+        try:
+            location_resp = await client.get(location_url)
```
ncloudioj, line 107: Nit: `client.get()` takes an argument `params` (see API [doc](https://www.python-httpx.org/api/)), which automatically takes cares of url encoding.

 ```suggestion
             location_resp = await client.get(
                 URL_POSTALCODES_PATH.format(country_code=country),
                 params={
                      URL_PARAM_API_KEY: API_KEY,
                      URL_POSTALCODES_PARAM_QUERY: postal_code,
                 }
             )
```

File: merino/providers/accuweather.py [hunk 13]
```diff
@@ -0,0 +1,169 @@
...
+                urlencode(
+                    {
+                        URL_PARAM_API_KEY: API_KEY,
+                    }
+                ),
+                "",
+            )
+        )
+        try:
+            forecasts_resp = await client.get(forecasts_url)
```
ncloudioj, line 129: Ditto.

merino/providers/accuweather.py
```diff
@@ -0,0 +1,169 @@
...
+                        city_name=location.get("LocalizedName"),
+                        temperature_unit=temperature_unit,
+                        high=high,
+                        low=low,
+                        day_summary=day_summary,
+                        day_precipitation=day_precipitation,
+                        night_summary=night_summary,
+                        night_precipitation=night_precipitation,
+                    )
+                ]
```
ncloudioj, line 166: Nit: while it's logically correct, having a non-exhaustive pattern matching could trigger a lint warning (e.g. by mypy). We can add a catch all case here and moving the logging and return statement in it.
```suggestion
                ]
            case _:
                logger.warning("Unexpected AccuWeather response")
                return []
```

tests/unit/web/test_api_v1.py
```diff
@@ -56,7 +56,8 @@ def test_suggest_from_missing_providers(query):
 
 def test_no_query_string():
     response = client.get("/api/v1/suggest")
-    assert response.status_code == 400
+    assert response.status_code == 200
```
ncloudioj, line 59: The current suggest API spec expects the query parameter `q` to be always present in the request URL. Any particular reason for this change? For zero-prefix weather recommendations, I think we can leave `q` empty and specify a provider to make the request. E.g. `merino.com/api/api/suggest?q=&providers=accuweather`.

What do you think?

merino/providers/adm.py
```diff
@@ -72,6 +77,30 @@ class IABCategory(str, Enum):
 MISSING_ICON_ID = "-1"
 
 
+class SponsoredSuggestion(BaseSuggestion):
```
ncloudioj, line 80: Why moving those two models to this particular provider? It's possible that they can be reused by other providers.

merino/providers/adm.py
```diff
@@ -178,23 +207,36 @@ async def _fetch(self) -> None:
     def hidden(self) -> bool:  # noqa: D102
         return False
 
-    async def query(self, q: str) -> list[dict[str, Any]]:
+    async def handle_request(self, request: Request) -> list[BaseSuggestion]:
+        """Provide suggestion for a given request."""
```
ncloudioj, line 211: Q: I've noticed that this pretty much shares the same implementation as other providers that take the query parameter `q` for `query(q: str)`. In fact, they are also similar to the `accuweather` provider. The downside is that this generates some non-DRY boilerplate code across the providers.

Should we consider providing a default implementation for `BaseProvider.handle_request()` so that the subclasses can just reuse it if they don't require any custom handling? Otherwise, they can override it. We can also include `async query(q: str|None = None)` to `BaseProvider` to make this work.

ncloudioj (changes requested):
Added a few more comments. We're getting closer :)

merino/providers/accuweather.py
ncloudioj: > The app argument lets the test create a provider instance with a mock server and mock responses essentially, see test_accuweather.py. I don't know if that's the proper way to write a test for this but it's an approach I found that works well. Is there a better way?

Ah, I see. I am not aware of a better way offhand. It just felt a little counter-intuitive to see a provider depending on the top-level FastAPI app. Before we find a better solution to this, do you mind adding a note to `self._app` describing why it's needed here? 

> That makes sense and it's the kind of thing I might need help on. Ideally Merino would also cache responses from AccuWeather. AsyncClient doesn't seem to do that on its own? I found this: https://github.com/johtso/httpx-caching

Yep, to limit the complexity of this PR. Let's leave all those performance optimizations to follow ups.

> The [docs](https://www.python-httpx.org/advanced/#more-efficient-usage-of-network-resources) say Client instances use connection pooling. So would pooling be as simple as creating the client in __init__() as an attribute?

Yes, I believe so. The other benefit of creating the pool inside of `__init__()` is that we could use some dependence injection trick there to use a different connection pool (like mocks) in the tests. We use the same technique to ease the Remote Settings dependency for the `adm` provider.

0c0w3: > We use the same technique to ease the Remote Settings dependency for the adm provider.

I did notice that the adm provider takes a backend as an argument. I was going to do something similar with accuweather until I found the app approach. Like, a class with a `get()` or `fetch()` method that talks to the server that tests could replace with a mock object.

tests/unit/web/test_api_v1.py
0c0w3: Would you prefer that? I feel like I should leave it up to you all as the Merino developers, but to me it makes more sense not to force all providers take a `q` param when they might not need it. Maybe I'm overthinking it. If you would prefer to keep the `q`, then I can revert the part of my change that replaced `provider.query()` with `handle_request()`.

merino/providers/adm.py
0c0w3: To me it seems like the suggestion payload is going to tightly depend on its provider. The accuweather provider has a way different payload than the adm provider. If we were to add more adm-style sponsored partners in the future, would they really have the same payload with `block_id` and all that? It really seems adm-specific. There will be some commonality which is why I kept some attributes in the `BaseSuggestion` class (maybe there should be more?). So that's why I moved all the suggestion classes to their providers.

That said, I know this is a significant-ish architectural change and I'll definitely defer to you on it.

merino/providers/adm.py
0c0w3: > I've noticed that this pretty much shares the same implementation as other providers that take the query parameter `q` for `query(q: str)`.

Yeah, this is related to the earlier comment about `q`. If you'd like for me to revert that change and make the accuweather provider take an empty `q`, we can do that and remove this.

> In fact, they are also similar to the `accuweather` provider.

True, they all ended up delegating to a `query()` method. The only reason I did it that way is so that tests don't need to create a mock `Request` in order to call `handle_request()`. Instead they can call `query()`.

So in summary here, let me know if you'd prefer to go back to `q` for all providers, and then we can go from there.

0c0w3:
I updated the PR for new review comments marked as resolved.

tests/unit/web/test_api_v1.py
ncloudioj: Yes, I'd recommend keeping `q` as a required query param for this endpoint as what the [API spec](https://mozilla-services.github.io/merino-py/api.html#query-parameters) suggests. Otherwise, the spec of this endpoint would be coupled with specific providers, which can be confusing and hard to maintain in the long run.

That being said, if a provider's request/response spec diverges significantly from the current one, we can consider adding a new endpoint for it.

merino/web/models_v1.py
```diff
@@ -11,44 +11,10 @@ class ProviderResponse(BaseModel):
...
-
-    impression_url: Optional[HttpUrl] = None
-    click_url: Optional[HttpUrl] = None
-
-
 class SuggestResponse(BaseModel):
     """Model for the `suggest` API response."""
 
-    suggestions: list[SponsoredSuggestion | NonsponsoredSuggestion]
+    suggestions: list[BaseSuggestion]
```
ncloudioj, line 17: Would this change break the client? The only difference between `SponsoredSuggestion` and `NonsponsoredSuggestion` is that `impression_url` and `click_url` are optional for the latter. That should not affect Firefox as it also treats them as optional fields. Now that the suggestion model is totally different from the others, let's say if we assign a large score to AccuWeather suggestions, I think that will break Firefox as the payload is different than that of adM suggestions, correct?

merino/providers/adm.py
ncloudioj: I see. Yep, we will need to think about how to handle this as more providers get added over time. I've dropped a related comment [here](https://github.com/mozilla-services/merino-py/pull/66/files#r994722529).

Currently, all suggestions are stored in the same field called `suggestions`. Perhaps we could expand it to a JSON object as follows:

```js
"suggestions": {
  "provider-foo": list[suggestion-model-foo],
  "provider-bar": list[suggestion-model-bar],
  ...
}
```

merino/web/models_v1.py
0c0w3: Yeah, it would break it -- it might cause exceptions but if not the suggestion in the UI would probably be missing some things.

This is the first time we're adding a new type of suggestion to Merino that's different from adM suggestions, right? Do you have more thoughts on how to maintain compatibility with the client? I don't think we can rely on all types of suggestions being the same shape. Weather suggestions are just too different from adM suggestions. So it seems like the client will either need to know which Merino providers it supports, or it will need to inspect the shape of returned suggestions to see if they're close enough to what it supports. The second option is more complex but it would allow some forward compatibility for older clients that don't know about newer providers, as long as the new providers' suggestions look similar enough to the old ones.

We support different providers in Firefox's urlbar by having a [base result class](https://searchfox.org/mozilla-central/source/browser/components/urlbar/UrlbarResult.sys.mjs) that composes [specific payload objects per result type](https://searchfox.org/mozilla-central/source/browser/components/urlbar/UrlbarUtils.sys.mjs#1246-1250). The base class has properties that are common to all results like a title, and then the payload object has all the type-specific data. Different providers can return results of the same type, which means we can add new providers and as long as they return results of an existing type, we don't need to do anything extra to support them. Of course there are some cases where we also have provider-specific handling.

Would it make sense to introduce a similar type system to Merino suggestions?

merino/providers/adm.py
0c0w3: I like the idea of Merino returning a single list of heterogeneous suggestions sorted by score. That makes it simple for the client to pick the first one or to show more than one sorted by score. Of course it's not a big deal since the client could do the sorting itself. Also, suggestion objects contain the `provider` property, so we could go the other way too: The client could generate an object/map of suggestions keyed on provider, if necessary.

quiiver: I also prefer the heterogeneous list approach. Given that we want to add a Best Bets Suggestion type soon it might make sense to get together to chat about this

ncloudioj: > I like the idea of Merino returning a single list of heterogeneous suggestions sorted by score.

Sounds good. We can stick to the current schema then. I think the only thing we need to be careful of is maintaining backward compatibility. I.e. if we're adding new suggestion models that could otherwise break the old versions of Firefox. We'd need to bump the API version, e.g. `api/v2/suggest` so that newer versions of Firefox can hit the new endpoint and they know how to handle suggestions for different providers.

merino/web/models_v1.py
ncloudioj: > This is the first time we're adding a new type of suggestion to Merino that's different from adM suggestions, right?

Correct. For other upcoming new providers (navigational queries & wikipedia dynamic match), they more or less share the same schema as adM suggestions. (N.B: we currently use dummy values for fields that don't apply. e.g. `block_id = 0` for all Wikipedia suggestions, this is not ideal but 'block_id' is a required field in our reporting database for FS). But it's not the case for Accuweather suggestions, that's why I was thinking about adding a new endpoint for it, e.g. `api/v1/weather`.

> Do you have more thoughts on how to maintain compatibility with the client? I don't think we can rely on all types of suggestions being the same shape

Normally, we'd either bump up the API version to introduce breaking changes or create new endpoints for those specific providers. Either way, the client needs to know how to use those suggestions for different providers by using the `provider` field or pinging a different endpoint. The old versions of clients can continue using the old endpoint, at some point once that user base becomes small enough, then we can deprecate the old endpoint(s).

> Would it make sense to introduce a similar type system to Merino suggestions?

Absolutely, I am all ears to solutions like what we do for Firefox urlbar. I think we can spin off another follow-up just for that expansion.

ncloudioj: A note for myself, our telemetry for FS should evolve accordingly as well as more heterogeneous providers get added to FS.

merino/providers/__init__.py
```diff
@@ -38,6 +40,10 @@ async def init_providers() -> None:
     # register providers
     for provider_type, setting in settings.providers.items():
         match provider_type:
+            case ProviderType.ACCUWEATHER:
+                providers["accuweather"] = AccuWeatherProvider(
+                    enabled_by_default=setting.enabled_by_default,
```
ncloudioj, line 45: Nit: I know that `name` takes a default value "accuweather", but let's pass it explicitly just in case we change the configuration in the future.
```suggestion
                    name=provider_type,
                    enabled_by_default=setting.enabled_by_default,
```

merino/web/models_v1.py
0c0w3: > But it's not the case for Accuweather suggestions, that's why I was thinking about adding a new endpoint for it

That's actually how I started implementing this, but I think at some point (ideally) Merino would start returning weather suggestions (and other non-adM types) in response to non-empty query strings at the usual `/suggest` endpoint. e.g., if the user searches for "weather", then Merino might return Amazon suggestions for rain boots along with a weather forecast. Goes back to our discussion about heterogeneous suggestions.

So that's why I did it under `/suggest`. If we want to punt that to some later date and just simply add a `/weather` endpoint for now, that's OK with me, but I think we'll need to resolve that sooner or later given the other types of suggestions we want to support?

ncloudioj (approved):
LGTM. Thanks for taking the first crack at it!

As this is not enabled by default, I think we can land it as is. As mentioned before, we will add the following things as follow-ups:
* Documentation for the new provider
* HTTP client pooling (if necessary)
* Accuweather location key prefetching & caching
* Accuweather forecast response caching
* Metrics
* Replace the existing suggestion model with a more flexible one that can be expanded for other providers

PR #67, opened 2022-10-03
[CONSVC-1982] refactor: rewrite Merino middleware as pure ASGI middleware

ncloudioj:
Rewrite middlewares as pure ASGI middlewares to address the performance overheads with Starlette-based middlewares.

This is the first part for:
- geolocation
- user agent
- logging

Will convert the remainder (metrics and feature flags) in follow-ups. 

This closes [CONSVC-1982](https://mozilla-hub.atlassian.net/browse/CONSVC-1982).

### DIFF
--- a/merino/middleware/geolocation.py
+++ b/merino/middleware/geolocation.py
@@ -1,13 +1,13 @@
 """The middleware that parses geolocation from the client IP address."""
 import logging
+from contextvars import ContextVar
 from typing import Optional
 
 import geoip2.database
 from geoip2.errors import AddressNotFoundError
 from pydantic import BaseModel
-from starlette.middleware.base import BaseHTTPMiddleware, RequestResponseEndpoint
 from starlette.requests import Request
-from starlette.responses import Response
+from starlette.types import ASGIApp, Receive, Scope, Send
 
 from merino.config import settings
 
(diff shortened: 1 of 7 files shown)


merino/middleware/logging.py
```diff
@@ -17,53 +21,67 @@
...
+        if scope["type"] != "http":  # pragma: no cover
+            await self.app(scope, receive, send)
+            return
+
+        async def send_wrapper(message: Message) -> None:
+            if message["type"] == "http.response.start":
+                request = Request(scope=scope)
+                if PATTERN.match(request.url.path):
+                    location = ctxvar_geolocation.get()
+                    ua = ctxvar_user_agent.get()
```
quiiver, line 42: I'm pretty sure `ContextVar.get` returns an `Optional[T]` we should probably specify default values or have a guard to check against a None value

ncloudioj:
Load test comparision (before/after)

Test environment: running Merino-py (`MERINO_LOGGING__FORMAT=mozlog MERINO_LOGGING__LEVEL=INFO make run`) in the single-threaded mode on a 2021 MBP M1 Max.

Load tests from [merino-explorations](https://github.com/quiiver/merino-explorations). Using 100 and 200 VUs (virtual users).

## 100 VUs
Before: 775 rps & 160 ms (P95 latency)
After: 2500 rps & 52 ms

## 200 VUs
Before: 666 rps & 352 ms
After: 2537 rps & 70 ms

It's about 3-4x faster in both throughput and latency.

merino/middleware/geolocation.py
```diff
@@ -25,18 +25,31 @@ class Location(BaseModel):
...
+        """Initialize."""
+        self.app = app
+
+    async def __call__(self, scope: Scope, receive: Receive, send: Send) -> None:
+        """Parse geolocation through client's IP address and store the result
+        (if any) to the `ContextVar`.
+        """
+        if scope["type"] != "http":  # pragma: no cover
+            await self.app(scope, receive, send)
+            return
```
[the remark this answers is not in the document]
ncloudioj: This is used quite frequently in Starlette & FastAPI (e.g. [here](https://github.com/encode/starlette/blob/master/starlette/middleware/sessions.py#L40)). I guess it's because it's not super easy to issue an ASGI non-HTTP scope in the test, so it wound up having this check uncovered in the tests.

merino/middleware/logging.py
ncloudioj: Yep, initially, I did provide a default value here just in case the get call fails. But after the second thought, I decided to take it out as the default value has already been set in the middleware, if the get ever fails, that should be a code error and it should fail fast and loud (as each request/task should always have a ctxvar associated as enforced by the middleware). Providing a default value here would otherwise silence that error inadvertently, which is probably not ideal in this case.

Does that make sense?

quiiver: Yeah I was thinking for the purposes of static analysis not yelling at us. if we want some error reporting for a None result, then what about a guard for the results of the get and log an error if the are None?

ncloudioj: Yep, but I was thinking about letting Sentry handle this.

Per its [doc](https://docs.python.org/3/library/contextvars.html#contextvars.ContextVar.get), `ContextVar.get()` will raise a `LookupError` if the value is not set and the default is not provided. And that exception (assuming it's rare) would be captured automatically by Sentry. For this particular request, an HTTP 500 error will be responded and it won't affect other requests since raising an exception inside of a middleware won't crash the app.

Do you think that's sufficient?

quiiver (approved):
looks good thanks!

PR #68, opened 2022-10-07
fix: Metric names did not share a common prefix. Setting it to `merino`

quiiver:
This was more annoying than I expected because in the original implementation I used module name in a few places to get the provider name. To fix that I added another method to the base provider class `BaseProvider:name`

### DIFF
--- a/merino/providers/adm.py
+++ b/merino/providers/adm.py
@@ -93,6 +93,7 @@ class Provider(BaseProvider):
     ) -> None:
         """Store the given Remote Settings backend on the provider."""
         self.backend = backend
+        self._name = "adm"
         self._enabled_by_default = enabled_by_default
         super().__init__(**kwargs)
 
--- a/merino/providers/base.py
+++ b/merino/providers/base.py
@@ -6,6 +6,7 @@ from typing import Any
 class BaseProvider(ABC):
     """Abstract class for suggestion providers."""
 
+    _name: str
     _enabled_by_default: bool
 
     @abstractmethod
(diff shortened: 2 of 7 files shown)


merino/providers/base.py
```diff
@@ -43,3 +44,7 @@ def availability(self) -> str:
             return "enabled_by_default"
         else:
             return "disabled_by_default"
+
+    def name(self) -> str:
```
ncloudioj, line 48:
```suggestion
    @property
    def name(self) -> str:
```

ncloudioj: While you're here, do you mind replacinh this constant [here](https://github.com/mozilla-services/merino-py/blob/main/merino/providers/adm.py#L193) with this property?

merino/providers/adm.py
```diff
@@ -93,6 +93,7 @@ def __init__(
     ) -> None:
         """Store the given Remote Settings backend on the provider."""
         self.backend = backend
+        self._name = "adm"
```
ncloudioj, line 96: I think we have already named each provider implicitly in the configuration files. E.g. `default.providers.adm`, shall we add an argument `name: str` to `__init__()` so that we can pass in the name whenever it's available?

For example, we can do this when creating the [providers](https://github.com/mozilla-services/merino-py/blob/main/merino/providers/__init__.py#L42-L49):

```py
case ProviderType.ADM:
    providers[ProviderType.ADM] = AdmProvider(
        name=ProviderType.ADM,
        backend=...,
        enabled_by_default=...,
    )
```

ncloudioj (approved):
Looks good, r+ with review comments.

ncloudioj (approved):

PR #69, opened 2022-10-10
type annotations, ignore statements with context, mod to pyproject.to…

taddes:
…ml, cast calls and Nan revisions from other branch

### DIFF
--- a/merino/config_sentry.py
+++ b/merino/config_sentry.py
@@ -1,5 +1,5 @@
 """Sentry Configuration"""
-from typing import NoReturn
+
 import sentry_sdk
 from sentry_sdk.integrations.fastapi import FastApiIntegration
 from sentry_sdk.integrations.starlette import StarletteIntegration
--- a/merino/config_sentry.py
+++ b/merino/config_sentry.py
@@ -10,7 +10,7 @@ from merino.config import settings
 def configure_sentry() -> None:  # pragma: no cover
     """Configure and initialize Sentry integration."""
     if settings.sentry.mode == "disabled":
-        return None
+        return
 
     sentry_sdk.init(
         dsn=settings.sentry.dsn,
--- a/merino/config_sentry.py
+++ b/merino/config_sentry.py
@@ -1,4 +1,5 @@
 """Sentry Configuration"""
+from typing import NoReturn
 import sentry_sdk
 from sentry_sdk.integrations.fastapi import FastApiIntegration
 from sentry_sdk.integrations.starlette import StarletteIntegration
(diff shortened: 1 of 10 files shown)


merino/config_sentry.py
```diff
@@ -1,15 +1,16 @@
...
 
 from merino.config import settings
 
 
-def configure_sentry():  # pragma: no cover
+def configure_sentry() -> None:  # pragma: no cover
     """Configure and initialize Sentry integration."""
     if settings.sentry.mode == "disabled":
-        return
+        return None
```
ncloudioj, line 13: I think `None` is unnecessary here.

merino/remotesettings.py
```diff
@@ -21,7 +21,7 @@ def __init__(self) -> None:
     async def fetch_attachment_host(self) -> str:
         """Fetch the attachment host from the Remote Settings server."""
         server_info = await self.client.server_info()
-        return server_info["capabilities"]["attachments"]["base_url"]
+        return str(server_info["capabilities"]["attachments"]["base_url"])
```
ncloudioj, line 24: We can also do a `cast(str, ...)` here.

merino/util/user_agent_parsing.py
```diff
@@ -29,7 +29,7 @@ def _parse_browser(user_agent: dict[str, Any]) -> str:
             return "Firefox" if major is None else f"Firefox({version.rstrip('.')})"
         case {"family": browser}:
             # Do not bother parsing the browser version for non-Firefox user agents.
-            return browser
+            return str(browser)
```
ncloudioj, line 32: Ditto: `cast(str, browser)`

merino/config_sentry.py
taddes: Agreed, I was being nitpicky and explicit, heh

merino/remotesettings.py
taddes: Happy to modify to `cast()`, which I considered, but the `str()` wrapper did explicitly silence the error, and where `cast(str, ...)` only operates for mypy, what approach do you think is the best here? In some ways, `str()` is indeed redundant, but it actually does a conversion if it needs to be there. We know the value will be a string in all likelihood, regardless, it's just that MyPy isn't getting the hint.  This question stems from `cast()` being essentially just be a no-op that silences the type checker, as looking at the [src here](https://github.com/python/cpython/blob/f64de53ff01e734d48d1d42195443d7d1646f220/Lib/typing.py#L1628-L1636).  Just wanted to ask your opinion on this, in the meantime I can change to `cast(str, ...)`. Thanks!

taddes: Makes sense regarding the desire for the type checker to infer that. I suppose at runtime we won't know from the type checker since it's static, but overall that makes solid sense. Changes pushed!

PR #74, opened 2022-10-14
[CONSVC-2006] test: add geolocation unit tests

Trinaa:
## Description
* Contributing this refactor done during integration test investigation spike as it's proving difficult to keep it up to date in a separate branch.

* Refactored middleware unit tests (geolocation so far)
    * `web/test_geolocation.py`, assures proper propagation of happy and un-happy path behaviour through internal interfaces.
    * `middleware/test_geolocation.py`, checks decision logic directly against the middleware class for better coverage accuracy and to increase coverage.
    image

* fix mypy error for `types-geoip2`
image

## Issue(s)
[CONSVC-2006](https://mozilla-hub.atlassian.net/browse/CONSVC-2006)


merino/middleware/logging.py
```diff
@@ -58,6 +58,7 @@ async def send_wrapper(message: Message) -> None:
                         "region": location.region,
                         "city": location.city,
                         "dma": location.dma,
+                        "postal_code": location.postal_code,
```
Trinaa, line 61: @ncloudioj or @0c0w3 
I am not sure if I should have added this?

ncloudioj: Since we haven't done the data review (for data collection) for `postal_code` nor have we added `postal_code` to the query log schema, let's *not* add this field to the log for now.

tests/unit/middleware/test_geolocation.py
```diff
@@ -0,0 +1,156 @@
...
+    geolocation_middleware: GeolocationMiddleware,
+    expected_location: Location,
+    client_ip_and_port: list,
+) -> None:
+    """
+    Test the proper assignment of Location properties given a request IP address.
+    """
+    scope: Scope = {"type": "http", "client": client_ip_and_port}
+    receive_mock: Receive = Mock()
+    send_mock: Send = Mock()
```
ncloudioj, line 59: Looks like these two mocks are defined separately for each test case, shall we just create two dummy fixtures for them to avoid the boilerplate?

ncloudioj: Ditto to `scope`, perhaps we can define a base fixture as `{"type": "http"}`. Test cases can augment it with other keys if needed.

tests/unit/middleware/test_geolocation.py
```diff
@@ -0,0 +1,156 @@
...
+    Location,
+    ctxvar_geolocation,
+)
+
+
+@pytest.fixture(name="geolocation_middleware")
+def fixture_geolocation_middleware() -> GeolocationMiddleware:
+    """Creates a GeolocationMiddleware object for test"""
+
+    asgiapp_mock = AsyncMock(spec="starlette.types.ASGIApp")
```
ncloudioj, line 23: Ah, fancy 👖!

As an aside, does `AsyncMock` only take a string for `spec`, looks a little odd to me?

Trinaa: [4e8bfcd](https://github.com/mozilla-services/merino-py/pull/74/commits/4e8bfcd5f291e3ab62d7b409341ee7744443063a)

AsyncMock can take either a string or object as a `spec` definition. Now that you've asked this question, I will opt for the object, since I believe this works better with refactor tools.

tests/unit/middleware/test_geolocation.py
Trinaa: [d894db5](https://github.com/mozilla-services/merino-py/pull/74/commits/d894db562d17c335fa296b65413e962f1bbb753b)

Added fixtures, let me know if it's not what you had in mind.

merino/middleware/logging.py
Trinaa: [4e8bfcd](https://github.com/mozilla-services/merino-py/pull/74/commits/4e8bfcd5f291e3ab62d7b409341ee7744443063a)

I removed `postal_code`.

I have concern that when the logging scheme is updated to include the `postal_code`, that the tests currently in `web/test_geolocation.py`  will not be up to date, since they assert against a log subset.

Have we considered defining the log scheme in a data object? From a test perspective it's easier to assert against.

Trinaa:
@ncloudioj should be ready for a second look.

tests/unit/middleware/test_geolocation.py
```diff
@@ -0,0 +1,176 @@
...
+from starlette.types import ASGIApp, Receive, Scope, Send
+
+from merino.middleware.geolocation import (
+    GeolocationMiddleware,
+    Location,
+    ctxvar_geolocation,
+)
+
+
+@pytest.fixture(name="geolocation_middleware")
```
ncloudioj, line 19: Nit: how about we auto use those fixtures if they're present everywhere in this test module?

ncloudioj (approved):
Looks great, thanks!

merino/middleware/logging.py
ncloudioj: That's a good point. Do you mind filing a ticket for this? Thanks!

Trinaa: Task created [CONSVC-2044](https://mozilla-hub.atlassian.net/browse/CONSVC-2044)

ncloudioj: Thanks!

PR #80, opened 2022-10-18
feat: prefer scope for data produced by middlewares

### LINKED ISSUE #73: Share middleware data via its Scope
To improve test comprehensibility and general app maintainability, store values produced from middlewares such as the `ctxvar_geolocation` value, found in `middleware/geolocation.py`, to the `scope` dictionary, an injected dependency.

!image-20221013-220930.png|width=697,height=546!

**Related documentation:**

[https://github.com/encode/starlette/blob/master/docs/middleware.md#passing-information-to-endpoints](https://github.com/encode/starlette/blob/master/docs/middleware.md#passing-information-to-endpoints|smart-link)

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/CONSVC-2021)
┆Attachments: <a href="https://mozilla-hub.atlassian.net/rest/api/2/attachment/content/29696">image-20221013-220930.png</a>
┆epic: Merino: maintenance

pjenvey:
Closes #73

### DIFF
--- a/merino/middleware/geolocation.py
+++ b/merino/middleware/geolocation.py
@@ -1,6 +1,5 @@
 """The middleware that parses geolocation from the client IP address."""
 import logging
-from contextvars import ContextVar
 from typing import Optional
 
 import geoip2.database
@@ -28,16 +27,12 @@ class Location(BaseModel):
     postal_code: Optional[str] = None
 
 
-# A `ContextVar` to store the geolocation result.
-ctxvar_geolocation: ContextVar[Location] = ContextVar("merino_geolocation")
-
-
 class GeolocationMiddleware:
     """An ASGI middleware to parse and populate geolocation from client's IP
     address.
 
-    The geolocation result `Location` (if any) is stored in a `ContextVar` called
-    `merino_geolocation`.
+    The geolocation result `Location` (if any) is stored in
+    `scope["merino_geolocation"]`.
     """
 
     def __init__(self, app: ASGIApp) -> None:
@@ -46,7 +41,7 @@ class GeolocationMiddleware:
 
     async def __call__(self, scope: Scope, receive: Receive, send: Send) -> None:
         """Parse geolocation through client's IP address and store the result
-        (if any) to the `ContextVar`.
+        to `scope`.
         """
         if scope["type"] != "http":
             await self.app(scope, receive, send)
@@ -64,7 +59,7 @@ class GeolocationMiddleware:
         except AddressNotFoundError:
             pass
 
-        ctxvar_geolocation.set(
+        scope["merino_geolocation"] = (
             Location(
                 country=record.country.iso_code,
                 region=record.subdivisions[0].iso_code,
--- a/merino/middleware/geolocation.py
+++ b/merino/middleware/geolocation.py
@@ -9,6 +9,7 @@ from starlette.requests import Request
 from starlette.types import ASGIApp, Receive, Scope, Send
 
 from merino.config import settings
+from merino.middleware import ScopeKey
 
 CLIENT_IP_OVERRIDE: str = settings.location.client_ip_override
 
@@ -32,7 +33,7 @@ class GeolocationMiddleware:
     address.
 
     The geolocation result `Location` (if any) is stored in
-    `scope["merino_geolocation"]`.
+    `scope[ScopeKey.GEOLOCATION]`.
     """
 
     def __init__(self, app: ASGIApp) -> None:
@@ -59,7 +60,7 @@ class GeolocationMiddleware:
         except AddressNotFoundError:
             pass
 
-        scope["merino_geolocation"] = (
+        scope[ScopeKey.GEOLOCATION] = (
             Location(
                 country=record.country.iso_code,
                 region=record.subdivisions[0].iso_code,
--- a/merino/middleware/logging.py
+++ b/merino/middleware/logging.py
@@ -9,6 +9,8 @@ from starlette.datastructures import Headers
 from starlette.requests import Request
 from starlette.types import ASGIApp, Message, Receive, Scope, Send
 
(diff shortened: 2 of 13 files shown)


merino/middleware/logging.py
```diff
@@ -38,8 +35,8 @@ async def send_wrapper(message: Message) -> None:
             if message["type"] == "http.response.start":
                 request = Request(scope=scope)
                 if PATTERN.match(request.url.path):
-                    location = ctxvar_geolocation.get()
-                    ua = ctxvar_user_agent.get()
+                    location = scope["merino_geolocation"]
+                    ua = scope["merino_user_agent"]
```
Trinaa, line 39: What do you think of capturing the `merino_geolocation` and `merino_user_agent` string keys in a ScopeKey Enum or Constant, so as to prevent errors from misspelling or facilitate future re-use/refactor?

tests/unit/providers/test_accuweather.py
```diff
@@ -244,65 +252,59 @@ async def test_invalid_location_key(accuweather: Provider) -> None:
...
+        city=None,
+        dma=None,
+        postal_code=None,
     )
 
     set_response_bodies()
 
-    res = await accuweather.query("")
+    res = await accuweather.query(SuggestionRequest("", geolocation))
     assert res == []
```
Trinaa, line 310: This is a bit out of scope, so let me know if you prefer I make a separate PR.

I realized looking at this changeset, that these last 3 tests (`test_no_client_country`, `test_no_client_postal_code` & `test_no_client_country_postal_code`) could be parameterized and should have a log check. 

```
@pytest.mark.parametrize(
    "geolocation",
    [
        Location(postal_code=94105),
        Location(country="US", region="CA", city="Some City", dma=555),
        Location(),
    ],
)
@pytest.mark.asyncio
async def test_no_client_country_or_postal_code(
    caplog: LogCaptureFixture, accuweather: Provider, geolocation: Location
) -> None:
    """
    Test that if a client has an unknown country or postal code, that a warning is
    logged and no suggestions are returned.
    """
    set_response_bodies()

    res = await accuweather.query(SuggestionRequest("", geolocation))

    assert res == []
    assert len(caplog.messages) == 1
    assert caplog.messages[0] == "Country and/or postal code unknown"
```

Trinaa (commented):
Thank you for this work!
From a test perspective it's powerful to be using injectable objects like this!

merino/middleware/logging.py
pjenvey: Great idea, added a ScopeKey enum

merino/providers/base.py
```diff
@@ -33,11 +44,11 @@ async def initialize(self) -> None:
         ...
 
     @abstractmethod
-    async def query(self, query: str) -> list[BaseSuggestion]:
+    async def query(self, srequest: SuggestionRequest) -> list[BaseSuggestion]:
         """Query against this provider.
 
         Args:
-          - `query`: the query string.
+          - `SuggestionRequest`: the query requested.
```
ncloudioj, line 51:
```suggestion
          - `SuggestionRequest`: the suggestion request object.
```

Since `geolocation` is also part of `SuggestionRequest`.

ncloudioj (approved):
Looks great, thanks!

I've run some quick local benchmarks, this runs on par with the `ContextVar` version, but this definitely comes with a couple of benefits over `ContextVar`. So yeah, let's use `scope` to store states onwards 👍

merino/middleware/geolocation.py
```diff
@@ -64,7 +60,7 @@ async def __call__(self, scope: Scope, receive: Receive, send: Send) -> None:
         except AddressNotFoundError:
             pass
 
-        ctxvar_geolocation.set(
+        scope[ScopeKey.GEOLOCATION] = (
```
hackebrot, line 63: I like the idea of using an enum for this to avoid naming conflicts. 👍🏻 

This uses the enum member as the mapping key. I want to make sure that's what we want.

From what I can tell starlette's `Scope` expects `typing.MutableMapping[str, typing.Any]`. This is satisfied for `ScopeKey` since we use `str` as one of the base classes, but it may not be what we want. If want to use `merino_geolocation` as a key in scope, we need to retrieve the value first:

```suggestion
        scope[ScopeKey.GEOLOCATION.value] = (
```

Trinaa (approved):

merino/middleware/geolocation.py
pjenvey: With it being a str subclass I'll argue using the enum variant itself doesn't matter much here as it all works just the same (maybe except you get an enum repr of these keys). Even `scope["merino_geolocation"]` works against the `Enum`.

However if you'd like I can just make `ScopeKey` a regular class of plain str constants instead of an enum.

hackebrot: I think using an `Enum` is the right call here. I just wanted to make sure we're OK with using the enum member for the key rather than the string value.

I didn't know that `scope["merino_geolocation"]` works with that. Let's use an enum!

pjenvey: 👍, reverted back to the enum

tests/unit/web/util.py
```diff
@@ -146,3 +147,8 @@ async def get_providers() -> tuple[dict[str, BaseProvider], list[BaseProvider]]:
 def filter_caplog(records: list[LogRecord], logger_name: str) -> list[LogRecord]:
     """Filter pytest captured log records for a given logger name."""
     return [record for record in records if record.name == logger_name]
+
+
+def srequest(q: str) -> SuggestionRequest:
```
hackebrot, line 152: Long-term, I'd recommend replacing these utility functions with pytest fixtures.

hackebrot (approved):
LGTM! 🚀

PR #81, opened 2022-10-18
[CONSVC-2047] fix: do not collect location data in the query log

ncloudioj:
This allows us to use the full-sized geolocation database rather than the dev one, which is required for features like the weather provider.

This fixes [CONSVC-2047](https://mozilla-hub.atlassian.net/browse/CONSVC-2047).

### DIFF
--- a/tests/unit/test_config_logging.py
+++ b/tests/unit/test_config_logging.py
@@ -27,3 +27,16 @@ def test_mozlog_production():
         assert "Log format must be 'mozlog' in production" in str(excinfo)
 
         settings.logging.format = old_format
+
+
+def test_collect_data_in_production():
+    with settings.using_env("production"):
+        old_value = settings.logging.collect_location
+        settings.logging.collect_location = True
+
+        with pytest.raises(ValueError) as excinfo:
+            configure_logging()
+
+        assert "`collect_location` should be `false` in production" in str(excinfo)
+
+        settings.logging.collect_location = old_value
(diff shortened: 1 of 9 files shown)


ncloudioj:
r? @quiiver 
cc: @0c0w3

quiiver:
Why can we not collect city level location if we use the full maxmind db?

ncloudioj:
Because the current [privacy policy](https://www.mozilla.org/en-US/privacy/firefox/) of Firefox Suggest (Online) discloses that:

> If you choose to enable "Improve the Firefox Suggest Experience," Firefox shares the following information to improve your browsing experience and to improve the service:

>    Searches: Firefox sends Mozilla what you type into the search bar and Mozilla may share that data with its [partners](https://support.mozilla.org/kb/firefox-suggest#w_who-are-mozillas-partners).

But it doesn't disclose whether or not the location data is stored along the search queries. The previous version of Merino doesn't log location data either. IIRC, location data was not included in the data collection review for Firefox Suggest Online. We'd go through the data review process If we wanted to collect it now.

tests/unit/test_config_logging.py
```diff
@@ -27,3 +27,16 @@ def test_mozlog_production():
...
+    with settings.using_env("production"):
+        old_value = settings.logging.collect_location
+        settings.logging.collect_location = True
+
+        with pytest.raises(ValueError) as excinfo:
+            configure_logging()
+
+        assert "`collect_location` should be `false` in production" in str(excinfo)
+
+        settings.logging.collect_location = old_value
```
Trinaa, line 42: If the assertion should fail, I believe this value will not be reset.

ncloudioj:
Confirmed with the legal team, we can collect the derived location data in the query log, but the IP address is not collectible which Merino doesn't collect anyway.

Closing this as the location collection is now appoved.

PR #88, opened 2022-10-22
feat: add current conditions to accuweather response

0c0w3:
Closes CONSVC-2049

### DIFF
--- a/merino/providers/accuweather.py
+++ b/merino/providers/accuweather.py
@@ -4,6 +4,7 @@ from typing import Any, Optional
 
 import httpx
 from fastapi import FastAPI
+from pydantic import BaseModel, HttpUrl
 
 from merino.config import settings
 from merino.providers.base import BaseProvider, BaseSuggestion, SuggestionRequest
@@ -15,6 +16,9 @@ SCORE: float = settings.providers.accuweather.score
 # Endpoint URL components
 URL_BASE: str = settings.providers.accuweather.url_base
 URL_PARAM_API_KEY: str = settings.providers.accuweather.url_param_api_key
+URL_CURRENT_CONDITIONS_PATH: str = (
+    settings.providers.accuweather.url_current_conditions_path
+)
 URL_POSTALCODES_PATH: str = settings.providers.accuweather.url_postalcodes_path
 URL_POSTALCODES_PARAM_QUERY: str = (
     settings.providers.accuweather.url_postalcodes_param_query
@@ -24,17 +28,37 @@ URL_FORECASTS_PATH: str = settings.providers.accuweather.url_forecasts_path
 logger = logging.getLogger(__name__)
 
 
+class Temperature(BaseModel):
+    """Model for temperature with C and F values."""
+
+    c: Optional[float] = None
+    f: Optional[float] = None
+
+
+class CurrentConditions(BaseModel):
+    """Model for AccuWeather current conditions."""
+
+    url: HttpUrl
+    summary: str
+    icon_id: int
+    temperature: Temperature
+
+
+class Forecast(BaseModel):
+    """Model for AccuWeather one-day forecasts."""
+
+    url: HttpUrl
+    summary: str
+    high: Temperature
+    low: Temperature
+
+
 class Suggestion(BaseSuggestion):
     """Model for AccuWeather suggestions."""
 
-    city_name: Optional[str] = None
-    temperature_unit: Optional[str] = None
-    high: Optional[float] = None
-    low: Optional[float] = None
-    day_summary: Optional[str] = None
-    day_precipitation: Optional[bool] = None
-    night_summary: Optional[str] = None
-    night_precipitation: Optional[bool] = None
+    city_name: str
+    current_conditions: CurrentConditions
+    forecast: Forecast
 
 
 class Provider(BaseProvider):
@@ -49,7 +73,7 @@ class Provider(BaseProvider):
         app: Optional[FastAPI] = None,
         name: str = "accuweather",
         enabled_by_default: bool = False,
-        **kwargs: Any
+        **kwargs: Any,
     ) -> None:
         self._app = app
         self._name = name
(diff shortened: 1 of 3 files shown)


merino/providers/accuweather.py
```diff
@@ -110,44 +151,101 @@ async def _get_forecast(
...
+                    return None
+
+                return Forecast(url=url, summary=summary, high=high, low=low)
+            case _:
+                return None
+
+    def _parse_forecast_temperature(
+        self, value: float, unit: str
+    ) -> Optional[Temperature]:
+        match unit:
```
ncloudioj, line 245: Using an enum against which the `unit` matches may silence the mypy warning. E.g

```py
from enum import Enum, unique

@unique
class Unit(str, Enum):
    CELCIUS = "C"
    FAHRENHEIT = "F"

try:
    unit = Unit(unit.upper())
except ValueError:
    logger.warning("Unexpected temperature unit")
    return None

match unit:
    case Unit.CELCIUS:
        ...
    case Unit.FAHRENHEIT:
        ...
    # no need to use `case _:` any more

```

merino/providers/accuweather.py
```diff
@@ -25,17 +29,37 @@
 logger = logging.getLogger(__name__)
 
 
+class Temperature(BaseModel):
+    """Model for temperature with C and F values."""
+
+    c: Optional[float] = None
+    f: Optional[float] = None
```
ncloudioj, line 36: Q: not sure how useful it would be, shall we automatically do the conversion and populate the other field if only one of `c` or `f` is provided?

merino/providers/accuweather.py
```diff
@@ -110,44 +151,101 @@ async def _get_forecast(
...
+                return Forecast(url=url, summary=summary, high=high, low=low)
+            case _:
+                return None
+
+    def _parse_forecast_temperature(
+        self, value: float, unit: str
+    ) -> Optional[Temperature]:
+        match unit:
+            case "C" | "c":
+                return Temperature(c=value, f=None)
```
ncloudioj, line 247: Nit: since both `f` and `c` use "None" as the default value, `f=None` and `c=None` can be omitted here.

merino/providers/accuweather.py
ncloudioj: Also noticed down below that in the API response, we provide both `c` and `f` for current conditions, but only one of them for the high & low temperatures for forecasts, which feels a little bit inconsistent to me.

ncloudioj (changes requested):
Looks good. Just a few comments on temperature handling.

merino/providers/accuweather.py
0c0w3: Unfortunately it didn't -- the problem seems to be that `data` is `Any`. The list structure in the pattern might also be part of the problem because mypy doesn't have a problem with the other pattern in this file that does not include a list, and the pattern in the [linked mypy issue](https://github.com/python/mypy/issues/12770) also uses a list.

So for now I kept this `_parse_forecast_temperature()` implementation the way it is since it seems a little simpler.

merino/providers/accuweather.py
```diff
@@ -110,44 +151,101 @@ async def _get_forecast(
...
+                "WeatherText": summary,
+                "WeatherIcon": icon_id,
                 "Temperature": {
-                    "Maximum": {"Value": high, "Unit": temperature_unit},
-                    "Minimum": {"Value": low},
-                },
-                "Day": {
-                    "IconPhrase": day_summary,
-                    "HasPrecipitation": day_precipitation,
+                    "Metric": {
```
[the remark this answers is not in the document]
0c0w3: AFAICT yes, the [doc](http://apidev.accuweather.com/developers/currentConditionsAPIParameters#responseParameters) includes both and doesn't say one or the other might be absent, and in my testing both are returned. Do you think it makes sense to be defensive and check? IMO it doesn't because I'm not sure where we would draw the line -- if we're defensive here then we should be defensive everywhere and not trust what their docs say. (That's kind of what I was doing in my original accuweather patch, where I wasn't using pattern matching.)

merino/providers/accuweather.py
0c0w3: Yeah, AccuWeather provides [both for current conditions](http://apidev.accuweather.com/developers/currentConditionsAPIParameters#responseParameters) but [only one for forecasts](http://apidev.accuweather.com/developers/forecastsAPIParameters#dailyforecast). For forecasts, they include a `Unit` property. It's a weird inconsistency.

We may well want to do the conversion ourselves at some point. In my testing the forecasts temperature is in F, I'm guessing since my IP is in the U.S. Of course another option is to do it on the client. Since this feature will roll out in only the U.S. at first, I think we can leave that to later if necessary.

ncloudioj: Makes sense. Let's leave it as is for now. The conversion can be done at a later time somewhere if necessary. I'd guess that we will let Firefox users pick whatever unit they prefer to use.

merino/providers/accuweather.py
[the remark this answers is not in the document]
ncloudioj: Agreed, no need to be defensive here as long as their API says both will be provided in the response.

merino/providers/accuweather.py
0c0w3: We would probably choose the unit appropriate for the user's current Firefox locale. If it's en-US, then F, and otherwise use C even if the user's region is the U.S. I guess that's something we could do even for the initial rollout since we'll likely target all en-* locales in the U.S. region like we've done before. So I'm definitely open to doing the conversion ourselves right now -- I just don't think it's a high priority ATM.

merino/providers/accuweather.py
ncloudioj: Hmm, I am seeing `data` here, did I miss anything?

Anyhow, mypy could be confused in certain cases. I think your current implementation is fine, let's keep it as is then.

ncloudioj (approved):
Looks good. Thanks!

merino/providers/accuweather.py
0c0w3: The mypy warning happens [on this line](https://github.com/mozilla-services/merino-py/pull/88/files#diff-4d942cadca923e41129b9530a95b09919c151aa1b3632127d0104de197c2d281R227) in `_parse_forecast()`, which has the `data` param. It doesn't seem to be related to parsing the unit.

PR #89, opened 2022-10-25
[CONSVC-2008] feat: add nav query top pick provider 

taddes:


dev/top_picks.json
```diff
@@ -0,0 +1,20 @@
+{
+    "domains": {
```
ncloudioj, line 2: Per the schema defined in [DENG-406](https://mozilla-hub.atlassian.net/browse/DENG-406), it should look like this:

```json
{
  "domains": [
    {
      "rank": 1,
      "title": "Mozilla",
      "domain": "mozilla",
      "url": "https://www.mozilla.org/",
      "icon": "chrome://devtools/skin/images/globe.svg",
      "categories": ["web-browser"],
      "similars": [
        "mozzilla",
        "mozila"
      ]
    }
  ]
}
```

N.B.:
* I was using the TLD (Top-Level Domain) for `domain` as we will get any queries from Firefox that contains `.`
* I was using the internal globe SVG icon in Firefox as an example

merino/configs/default.toml
```diff
@@ -79,3 +79,10 @@ backend = "remote-settings"
 cron_interval_sec = 60
 resync_interval_sec = 10800
 score = 0.3
+
+[default.providers.top_pick]
+enabled_by_default = false
+score = 0.25
+is_top_pick = true
+query_char_limit = 4
+top_pick_file_path = "merino/providers/top_picks.json"
```
ncloudioj, line 88:
```suggestion
top_pick_file_path = "./dev/top_picks.json"
```

merino/configs/default.toml
```diff
@@ -79,3 +79,10 @@ backend = "remote-settings"
 cron_interval_sec = 60
 resync_interval_sec = 10800
 score = 0.3
+
+[default.providers.top_pick]
+enabled_by_default = false
+score = 0.25
+is_top_pick = true
```
ncloudioj, line 86: Why having this boolean here? Do we have any case in which suggestions of this provider will not be served as Top Picks?

merino/configs/default.toml
```diff
@@ -79,3 +79,10 @@ backend = "remote-settings"
 cron_interval_sec = 60
 resync_interval_sec = 10800
 score = 0.3
+
+[default.providers.top_pick]
```
ncloudioj, line 83: To be consistent, let's use "Top Picks" in Merino.
 
```suggestion
[default.providers.top_picks]
```

merino/providers/__init__.py
```diff
@@ -26,6 +27,7 @@ class ProviderType(str, Enum):
 
     ACCUWEATHER = "accuweather"
     ADM = "adm"
+    TOP_PICK = "top_pick"
```
ncloudioj, line 30:
```suggestion
    TOP_PICKS = "top_picks"
```

merino/providers/top_pick.py
```diff
@@ -0,0 +1,165 @@
...
+
+    PRIMARY: Final = "PRIMARY"
+    SECONDARY: Final = "SECONDARY"
+
+
+class Suggestion(BaseSuggestion):
+    """Model for Top Pick Query Suggestion"""
+
+    block_id: int
+    icon: Optional[HttpUrl] = None
```
ncloudioj, line 43: We might need to use an internal icon URI (e.g. "chrome://devtools/skin/images/globe.svg"), which won't be a valid `HttpUrl`. So let's do `icon: str` for now. Also this field should be required per the schema.

merino/providers/top_pick.py
```diff
@@ -0,0 +1,165 @@
...
+    full_keyword: str
+    title: str
+    is_sponsored_suggestion: bool
+    top_pick: bool
+    score: float
+    impression_url: HttpUrl
```
ncloudioj, line 49: Both `impression_url` and `click_url` are not needed for this provider, let's drop them.

merino/providers/top_pick.py
ncloudioj: In fact, `BaseSuggestion` already defines the following fields:
```py
    title: str
    url: HttpUrl
    provider: str
    is_sponsored: bool
    score: float
    icon: str | None = None
```
No need to redefine those here.

merino/providers/top_pick.py
```diff
@@ -0,0 +1,165 @@
...
+    PRIMARY: Final = "PRIMARY"
+    SECONDARY: Final = "SECONDARY"
+
+
+class Suggestion(BaseSuggestion):
+    """Model for Top Pick Query Suggestion"""
+
+    block_id: int
+    icon: Optional[HttpUrl] = None
+    full_keyword: str
```
ncloudioj, line 44: We don't need `full_keyword` for this provider either.

merino/providers/top_pick.py
```diff
@@ -0,0 +1,165 @@
...
+    # FastAPI instance to fetch mock responses from it. See `__init__()`.
+    _app: Optional[FastAPI]
+
+    domain_list: dict[str, Any] = {}
+    suggestions: dict[str, int] = {}
+    primary_index: defaultdict = defaultdict(list)
+    primary_results: list[dict[str, Any]] = []
+    secondary_index: defaultdict = defaultdict(list)
+    secondary_results: list[dict[str, Any]] = []
+    icons: dict[int, str] = {}
```
ncloudioj, line 66: Icons are already stored in the suggestions, no need to store them separately here.

merino/providers/top_pick.py
```diff
@@ -0,0 +1,165 @@
...
+
+
+class Provider(BaseProvider):
+    """Top Pick Query Suggestion Provider"""
+
+    # In normal usage this is None, but tests can create the provider with a
+    # FastAPI instance to fetch mock responses from it. See `__init__()`.
+    _app: Optional[FastAPI]
+
+    domain_list: dict[str, Any] = {}
```
ncloudioj, line 60: I don't think we need to store this list once the indices are built. Let's drop it.

merino/providers/top_pick.py
```diff
@@ -0,0 +1,165 @@
...
+    primary_index: defaultdict = defaultdict(list)
+    primary_results: list[dict[str, Any]] = []
+    secondary_index: defaultdict = defaultdict(list)
+    secondary_results: list[dict[str, Any]] = []
+    icons: dict[int, str] = {}
+    # Store the value to avoid fetching it from settings every time as that'd
+    # require a three-way dict lookup.
+    titles: list[dict[str, Any]] = []
+    score: float = settings.providers.adm.score
+    last_fetch_at: float
```
ncloudioj, line 71: I believe we only load this domain list once at application startup, until we move this file to remote, we don't need this field.

merino/providers/top_pick.py
```diff
@@ -0,0 +1,165 @@
...
+        if not os.path.exists(LOCAL_TOP_PICK_FILE):
+            logger.warning("Local file does not exist")
+            raise FileNotFoundError
+        try:
+            with open(file, "r") as readfile:
+                domain_list: dict[str, Any] = json.load(readfile)
+                return domain_list
+        except Exception as e:
+            logger.warning("Cannot Process File: {e}")
+            return e
```
ncloudioj, line 124: Perhaps you want to re-throw the exception here instead of returning it?

merino/providers/top_pick.py
```diff
@@ -0,0 +1,165 @@
...
+
+    @staticmethod
+    def build_index(domain_list: dict[str, Any]) -> Any:
+        """Construct indexes and results from Top Picks"""
+        domains: list[dict[str, Any]] = domain_list["domains"]["items"]
```
ncloudioj, line 129: As mentioned above, `domain_list["domains"]` should be a list, there is no such property `items` in it.

merino/providers/top_pick.py
```diff
@@ -0,0 +1,165 @@
...
+        index: defaultdict = defaultdict(list)
+        results: list[dict[str, Any]] = []
+
+        for domain in domains:
+            index_key = len(results)
+            for chars in range(QUERY_CHAR_LIMIT, len(domain) + 1):
```
ncloudioj, line 135: Be mindful of the case when `len(domain) < QUERY_CHAR_LIMIT`, this will insert no keys into the index while appending the `domain` onto `results`, which would leave this domain unused. I think in that case, we can just skip this domain as a whole.

merino/providers/top_pick.py
```diff
@@ -0,0 +1,165 @@
...
+                # See configs/default.toml for character limit for Top Picks
+                index[domain["domain"][:chars]].append(index_key)
+            results.append(domain)
```
ncloudioj, line 138: I'd recommend building both the primary index and the second index within a single loop here, e.g.

```py
    for record in domains_list["domains"]:
        index_key = len(results)
        suggestion = { ... }  # create the suggestion based on `record`
        
        # insert keys into the primary index
        domain = record["domain"]
        for chars in range(QUERY_CHAR_LIMIT, len(domain)):
             primary_index[domain[:chars]].append(index_key)

        # insert keys into the secondary index
        for variant in record.get("similars", []):
             for chars in range(QUERY_CHAR_LIMIT, len(variant)):
                  secondary_index[variant[:chars]].append(index_key)

        results.append(suggestion)

    return {
         "primary_index": primary_index,
         "second_index": secondary_index,
         "results": results
    }
```

merino/providers/top_pick.py
```diff
@@ -0,0 +1,165 @@
...
+        for domain in alt_domains:
+            index_key = len(secondary_results)
+            for chars in range(QUERY_CHAR_LIMIT, len(domain) + 1):
+                # See configs/default.toml for character limit for Top Picks
+                secondary_index[domain["domain"][:chars]].append(index_key)
+            secondary_results.append(domain)
+        return (index, results), (secondary_index, secondary_results)
+
+    @staticmethod
+    def build_indices() -> Any:
```
ncloudioj, line 161: To make this more cohesive and testable, we can move all those static methods into a separate utility module.

merino/providers/top_pick_schema.json
```diff
@@ -0,0 +1,56 @@
+{
+    "$schema": "http://json-schema.org/draft-04/schema#",
+    "type": "object",
+    "properties": {
+        "domains": {
+            "type": "array",
+            "items": [
```
ncloudioj, line 7: FWIW, `items` are not the field of this array but only a term used by the JSON schema to describe the "items" of this array.

merino/providers/top_pick.py
```diff
@@ -0,0 +1,183 @@
...
+LOCAL_TOP_PICK_FILE: str = settings.providers.top_pick.top_pick_file_path
+QUERY_CHAR_LIMIT: int = settings.providers.top_pick.query_char_limit
+# Used whenever the `icon` field is missing from the top pick payload.
+MISSING_ICON_ID: Final = "-1"
+
+
+logger = logging.getLogger(__name__)
+
+
+class TopPickCategory(str, Enum):
```
ncloudioj, line 27: Looks like this never gets used anywhere, let's drop it for now.

merino/providers/top_pick.py
```diff
@@ -0,0 +1,183 @@
...
+
+        Args:
+            - `q`: the query string.
+        """
+        if len(q) < QUERY_CHAR_LIMIT:
+            return []
+        if (id := self.primary_index.get(q)) is not None:
+            res = self.primary_results[id]
+            suggestion_dict = {
+                "block_id": res.get("rank"),
```
ncloudioj, line 108: Let's use `0` as the dummy value for `block_id`.

merino/providers/top_pick.py
taddes: I'm actually planning on using that as a payload response in the suggestion. I thought it wise for telemetry purposes if the suggestion was flagged as primary or secondary, hence the enum. I hadn't yet implemented it in the code as I wanted to recommend it in our discussion.

merino/providers/top_pick.py
taddes: Sounds good!

merino/providers/top_pick.py
```diff
@@ -0,0 +1,183 @@
...
+
+    async def query(self, q: str) -> Optional[list[BaseSuggestion]]:  # type: ignore
+        """Query Top Pick provider.
+
+        Args:
+            - `q`: the query string.
+        """
+        if len(q) < QUERY_CHAR_LIMIT:
+            return []
+        if (id := self.primary_index.get(q)) is not None:
```
ncloudioj, line 105: `self.primary_index.get(q)` will return a list, rather than a single index. So you'd do (assuming that we only return the first one (i.e. the highest ranked one)):
```py
    if (ids := self.primary_index.get(q, [])):
        res = self.primary_results[ids[0]]
```
The same applies to secondary_index.

merino/providers/top_pick.py
```diff
@@ -0,0 +1,183 @@
...
+                "url": res.get("url"),
+                "impression_url": res.get("url"),
+                "click_url": res.get("url"),
+                "provider": self.name,
+                "is_top_pick": True,
+                "is_sponsored": False,
+                "icon": res.get(int(res.get("icon", MISSING_ICON_ID))),
+                "score": SCORE,
+            }
+            Suggestion(**suggestion_dict)
```
ncloudioj, line 122:
```suggestion
            return [Suggestion(**suggestion_dict)]
```

merino/providers/top_pick.py
```diff
@@ -0,0 +1,183 @@
...
+        elif (id := self.secondary_index.get(q)) is not None:
+            res = self.secondary_results[id]
```
ncloudioj, line 124:
```suggestion
            res = self.secondary_results[id]
            return Suggestion(..)
```

merino/providers/top_pick.py
```diff
@@ -0,0 +1,183 @@
...
+        else:
+            res = None
```
ncloudioj, line 126:
```suggestion
            return []
```

dev/top_picks.json
taddes: Got it, updating that now

merino/providers/top_pick.py
taddes: Dropped!

merino/providers/top_pick_schema.json
taddes: 🤦‍♂️ yup, my bad. Never took that out when making test files from schema. Duh

merino/providers/top_pick.py
```diff
@@ -0,0 +1,183 @@
...
+    ) -> None:
+        self._app = app
+        self._name = name
+        self._enabled_by_default = enabled_by_default
+        super().__init__(**kwargs)
+
+    async def initialize(self) -> None:
+        """Initialize the provider."""
+        try:
+            primary, secondary = await asyncio.to_thread(Provider.build_indices())
```
ncloudioj, line 86:
```suggestion
            primary, secondary = await asyncio.to_thread(Provider.build_indices)
```

taddes: Thank you for that. Classic mistake instead of passing ref.

merino/providers/top_pick.py
taddes: Will do, great idea. I can make the changes, or we can merge this in after the PR is reviewed and I can then extract them out to a utility module.

merino/providers/top_pick.py
taddes: Just need to ask you about icon again tomorrow, was passing an empty string and saw in another provider a Final was set with the value of -1, so wondering if we should do that until we have something, though I see we set the icon values to be typed as str.

dev/top_picks.json
```diff
@@ -0,0 +1,50 @@
+{
+    "domains": [
+        {
+            "rank": 1,
+            "title": "Example",
+            "domain": "example",
+            "url": "https://example.com",
+            "icon": -1,
```
ncloudioj, line 8: Per schema, I believe that `icon` is of type `string`

merino/configs/default.toml
```diff
@@ -79,3 +79,9 @@ backend = "remote-settings"
 cron_interval_sec = 60
 resync_interval_sec = 10800
 score = 0.3
+
+[default.providers.top_picks]
+enabled_by_default = false
+score = 0.2
```
ncloudioj, line 85: Firefox also assigns 0.2 to Remote Settings suggestions (see [here](https://searchfox.org/mozilla-central/source/browser/components/urlbar/UrlbarQuickSuggest.sys.mjs#62)). Let's use a higher score for this provider. 0.25 should be fine.

merino/configs/testing.toml
```diff
@@ -31,3 +31,10 @@ backend= "test"
 [testing.providers.wiki_fruit]
 # Whether or not this provider is enabled by default.
 enabled_by_default = true
+
+[testing.providers.top_picks]
+# Whether or not this provider is enabled by default.
+enabled_by_default = true
+score = 0.2
```
ncloudioj, line 38: Ditto.

merino/providers/top_picks.py
```diff
@@ -0,0 +1,162 @@
...
+        super().__init__(**kwargs)
+
+    async def initialize(self) -> None:
+        """Initialize the provider."""
+        try:
+            index_results_dict: dict = await asyncio.to_thread(Provider.build_indices)
+            self.primary_index = index_results_dict["primary_index"]
+            self.secondary_index = index_results_dict["secondary_index"]
+            self.results = index_results_dict["results"]
+            self.query_char_count_dict = index_results_dict["char_count_dict"]
```
ncloudioj, line 64: Let's make `char_count_dict` a tuple `(min_char_count, max_char_count)` and rename it to `index_char_range`.

merino/providers/top_picks.py
```diff
@@ -0,0 +1,162 @@
...
+                domain_list = json.load(readfile)
+                return domain_list
+        except Exception as e:
+            logger.warning("Cannot Process File: {e}")
+            raise e
+
+    @staticmethod
+    def build_index(domain_list: Any) -> Any:
+        """Construct indexes and results from Top Picks"""
+        primary_index: defaultdict = defaultdict(list)
```
ncloudioj, line 108: Could you add some docs to briefly describe the structure of those instance variables. You can reference [here](https://github.com/mozilla-services/merino-py/blob/main/merino/providers/adm.py#L161-L165)

merino/providers/top_picks.py
```diff
@@ -0,0 +1,162 @@
...
+            raise e
+
+    def hidden(self) -> bool:  # noqa: D102
+        return False
+
+    async def query(self, q: str) -> Any:
+        """Query Top Pick provider."""
+        if q.startswith("http"):
+            return []
+        if len(q) < self.query_char_count_dict["min"]:
```
ncloudioj, line 77:
```suggestion
        if not (self.query_char_count_dict["min"] <= len(q) <= self.query_char_count_dict["max"]):
```

merino/providers/top_picks.py
```diff
@@ -0,0 +1,162 @@
...
+            return []
+        if len(q) > self.query_char_count_dict["max"]:
+            return []
+
+        if ids := self.primary_index.get(q, []):
+            res = self.results[ids[0]]
+            logger.warning(res)
```
ncloudioj, line 84:
```suggestion
```

merino/providers/top_picks.py
```diff
@@ -0,0 +1,162 @@
...
+            "max": QUERY_CHAR_LIMIT,
+        }
+
+        for record in domain_list["domains"]:
+            index_key = len(results)
+
+            if len(record["domain"]) < QUERY_CHAR_LIMIT:
+                continue
+
+            suggestion = {
```
ncloudioj, line 122: Instead of creating a dict here, we can directly create a `Suggestion` object here, such as
```py
suggestion = Suggestion(
    block_id=0,
    title= record["title"],
    ...
)
```

So in the `query()`, it doesn't need to create a Suggestion object over and over again.

merino/providers/top_picks.py
```diff
@@ -0,0 +1,162 @@
...
+
+logger = logging.getLogger(__name__)
+
+
+class Suggestion(BaseSuggestion):
+    """Model for Top Pick Query Suggestion"""
+
+    block_id: int
+    domain: str
+    rank: int
```
ncloudioj, line 27: Both `domain` and `rank` are irrelevant to the API consumers, no need to be included in the response.

merino/providers/top_picks.py
```diff
@@ -0,0 +1,162 @@
...
+    is_top_pick: bool
+    score: float
```
ncloudioj, line 29: Both `score` and `is_sponsored` are already defined in `BaseSuggestion`, no need to redefine them again here.

merino/providers/top_picks.py
```diff
@@ -0,0 +1,162 @@
...
+        results: list[dict[str, Any]] = []
+        min_max_chars: dict[str, int] = {
+            "min": QUERY_CHAR_LIMIT,
+            "max": QUERY_CHAR_LIMIT,
+        }
+
+        for record in domain_list["domains"]:
+            index_key = len(results)
+
+            if len(record["domain"]) < QUERY_CHAR_LIMIT:
```
ncloudioj, line 119: I don't think we need this check here, as it also needs to check all the entries in `similars`.

merino/providers/top_picks.py
```diff
@@ -0,0 +1,162 @@
...
+                "domain": record["domain"],
+                "url": record["url"],
+                "provider": "top_picks",
+                "is_top_pick": True,
+                "icon": "",
+                "score": settings.providers.top_picks.score,
+            }
+
+            # Insertion of keys into primary index.
+            for chars in range(QUERY_CHAR_LIMIT, len(record["domain"]) + 1):
```
ncloudioj, line 135: Nit: let's define a variable `domain = record["doamin"]` instead of repeating it multiple tiles in this block.

merino/providers/top_picks.py
```diff
@@ -0,0 +1,162 @@
...
+                # See configs/default.toml for character limit for Top Picks
+                if chars > min_max_chars["max"]:
+                    min_max_chars["max"] = chars
```
ncloudioj, line 138: You will also need to track the minimal index length. It's possible that it's greater than `QUERY_CHAR_LIMIT`.

merino/providers/top_picks.py
```diff
@@ -0,0 +1,162 @@
...
+                primary_index[record["domain"][:chars]].append(index_key)
+
+            # Insertion of keys into primary index.
```
ncloudioj, line 141:
```suggestion
            # Insertion of keys into seconary index.
```

merino/providers/top_picks.py
```diff
@@ -0,0 +1,162 @@
...
+        self,
+        app: Optional[FastAPI] = None,
+        name: str = "top_picks",
+        enabled_by_default: bool = False,
+        **kwargs: Any,
+    ) -> None:
+        self._app = app
+        self._name = name
+        self._enabled_by_default = enabled_by_default
+        super().__init__(**kwargs)
```
quiiver, line 55: I don't think we need this line since there's no `__init__` defined on BaseProvider

tests/unit/providers/test_top_picks.py
```diff
@@ -0,0 +1,175 @@
...
+
+# from merino.config import settings
+from merino.providers.top_picks import Provider, Suggestion
+
+app = FastAPI()
+router = APIRouter()
+
+top_picks_example = {
+    "domains": {
+        "items": [
```
ncloudioj, line 20: Drop `items`.

tests/unit/providers/test_top_picks.py
```diff
@@ -0,0 +1,175 @@
...
+
+
+def test_read_domain_list(top_picks: Provider) -> None:
+    """Test that the JSON file containing the domain list can be processed"""
+    with pytest.raises(FileNotFoundError):
+        top_picks.read_domain_list("./wrongfile.json")
+    domain_list = top_picks.read_domain_list(
+        settings.providers.top_picks.top_picks_file_path
+    )
+    assert type(domain_list) == dict
```
ncloudioj, line 65: The next assert already verifies it's a `dict`.
```suggestion
```

tests/unit/providers/test_top_picks.py
ncloudioj: Hmm, how come this didn't break the tests? 🤔

tests/unit/providers/test_top_picks.py
```diff
@@ -0,0 +1,175 @@
...
+
+
+def test_local_file_exists() -> None:
+    """Test that the Top Picks Nav Query file exists locally"""
+    assert os.path.exists(settings.providers.top_picks.top_picks_file_path) is True
+
+
+def test_read_domain_list(top_picks: Provider) -> None:
+    """Test that the JSON file containing the domain list can be processed"""
+    with pytest.raises(FileNotFoundError):
```
ncloudioj, line 60: Nice, looks like you already have this covered. Could you split this into two unit tests since the former tests the failure case and the latter checks the happy one?

quiiver (approved):
This looks good, I have 1 minor comment in the code. 

Generally I'm curious about the naming/structure of the provider. `TopPick` is a UX feature on the client which could have many sources of data, we're just starting with domains. How would we evolve this provider to account for more `TopPick` data sources? In my mind there's two options when we get to that:

1. We take a per-source provider approach and move this provider to a more specific name like `domains_top_pick` and the second source would follow the same convention of `<source>_top_pick`. This would be easy to split experiment traffic across providers by using the `providers` query param but is probably the less DRY of the solutions.
2. We keep a single provider for the `top_pick` feature and encapsulate the indexes into their own modules. This is probably the most DRY of the solutions but leaves the question about how you bring up new sources.

Anyway, some food for thought. I don't think any immediate changes are necessary in this regard just wanted to plant the seed.

tests/unit/providers/test_top_picks.py
```diff
@@ -0,0 +1,175 @@
...
+
+
+def test_build_index(top_picks: Provider) -> None:
+    """Test constructing the primary and secondary indexes and suggestions"""
+    domain_list = top_picks.read_domain_list(
+        settings.providers.top_picks.top_picks_file_path
+    )
+    source_dict = top_picks.build_index(domain_list)
+
+    assert "exa" not in source_dict["primary_index"]
```
ncloudioj, line 77: this would break if we update `QUERY_CHAR_LIMIT` in the future. You can do `example[:max(QUERY_CHAR_LIMIT - 1, 0)] not in source_dict...`

tests/unit/providers/test_top_picks.py
```diff
@@ -0,0 +1,175 @@
...
+
+    assert len(short_domain_dict["results"]) == 0
+    assert short_domain_dict["results"] == []
+    assert short_domain_dict["primary_index"] == {}
+
+
+def test_build_indeces(top_picks: Provider) -> None:
+    """Test to build indexes and result data structures"""
+    source_dict = top_picks.build_indices()
+    source_dict["primary_index"]
```
ncloudioj, line 109: Did you forget to assert?

tests/unit/providers/test_top_picks.py
```diff
@@ -0,0 +1,175 @@
...
+    source_dict["secondary_index"]
+    source_dict["results"]
+
+
+@pytest.mark.asyncio
+async def test_initialize(top_picks: Provider) -> None:
+    """Test initialization of top pick provider"""
+    await top_picks.initialize()
+    assert type(top_picks.primary_index) == defaultdict
```
ncloudioj, line 118: Those assertions won't verify anything as you expected as those are all instance variables, their types won't change even if you call `initialize()` here. Perhaps you need to check what's inside them.

tests/unit/providers/test_top_picks.py
```diff
@@ -0,0 +1,175 @@
...
+    assert type(top_picks.secondary_index) == defaultdict
+    assert type(top_picks.results) == list
+
+
+@pytest.mark.asyncio
+async def test_initialize_exception(top_picks: Provider) -> None:
+    """Test that proper exception is thrown if initialization is unsuccessful"""
+    with pytest.raises(Exception):
+        top_picks.initialize(0)
```
ncloudioj, line 127: Hmm, I don't think this is the best way to simulate an exception inside `initialize()`, as we won't call it with this argument and w/o `await` for it. I think as you've already had some test cases to check exceptions raised in those helper functions, we don't need to do that again here at the provider level.

merino/providers/top_picks.py
```diff
@@ -0,0 +1,162 @@
...
+        if len(q) > self.query_char_count_dict["max"]:
+            return []
+
+        if ids := self.primary_index.get(q, []):
+            res = self.results[ids[0]]
+            logger.warning(res)
+            return Suggestion(**res)
+        elif ids := self.secondary_index.get(q, []):
+            res = self.results[ids[0]]
+            return Suggestion(**res)
```
ncloudioj, line 88:
```suggestion
            return [Suggestion(**res)]
```

dev/top_picks.json
taddes: Of course @ncloudioj . This came from looking at how the other providers handled icons not being present, so I implemented it the same way. I know the schema dictated strings. Can you comment on how we handle icons not being present? For the time being, I know we don't have that part sorted out so I was trying to stay ahead of it. Thanks!

merino/configs/default.toml
taddes: You got it!

merino/configs/testing.toml
taddes: Got it!

merino/providers/top_picks.py
ncloudioj: I was surprised that this didn't cause any test failures. Can you add some tests to actually send some queries to this privider? Like what we have in [test_api_v1.py](https://github.com/mozilla-services/merino-py/blob/main/tests/unit/web/test_api_v1.py)

merino/providers/top_picks.py
```diff
@@ -0,0 +1,162 @@
...
+        return []
+
+    @staticmethod
+    def read_domain_list(file: str) -> Any:
```
ncloudioj, line 92: Could you replace all the `Any` types in this module with their actual types? I think the function input and ouputs are pretty well-defined, shouldn't be too hard to fill them out. Thanks!

ncloudioj (changes requested):
I think it's now in good shape, we're getting closer. Just a few more review comments.

merino/providers/top_picks.py
taddes: Debug! Removed.

merino/providers/top_picks.py
taddes: Good point!

tests/unit/providers/test_top_picks.py
taddes: I had altered the test that originally used this, I had a note of this as I have a few more tests to write.

merino/providers/top_picks.py
taddes: My only concern with this is that as we iterate through the collection of domains when building indexes, we update the values for these min and max values, so I can assign to separate int variables for min and max and then pass them into a tuple at the end, but it results in three declarations opposed to one, which is why I opted for a dictionary for the start, which requires fewer vars. It is very explicit, since they key describes what the value is, opposed to accessing the index of a tuple which in a check may be confusing. Happy to still make it a tuple, just wanted to provide my rationale. I implemented the previous indices as tuples in the last review and it was opted they change to a dict for clarity, so I was trying to follow that pattern.

merino/providers/top_picks.py
taddes: You bet, I mentioned I'd be doing that in the Jira ticket. It became onerous with all the updates to manage pertaining to some of the type hints, appease the linter and commit. All the type annotations will now be updated.

merino/providers/top_picks.py
taddes: I actually do have the `test_query` function and it was successful, as I'd check the first position of the index to test that the suggestion data was the same. I changed the return type from this to just `res`, so a constructed `Suggestion` as you recommended, so that a new obj doesn't have to be made every time query is called. Should be good!

merino/providers/top_picks.py
taddes: That's a very good point @quiiver  and indeed worth keeping in mind. I suppose we can definitely take either approach, though for where we are at option 2 seems most sensible for where we are at, so long as you feel confident it provides the adequate extensibility desired for future provider sources and data.

merino/providers/top_picks.py
ncloudioj: I think that's fine. We can have any number of declarations as long as they make sense. For better readability, you can also do `min_len, max_len = index_char_range` before using it, or you can just use two instance variables to store them. Using a dict in this case just feels a bit odd to me.

merino/providers/top_picks.py
```diff
@@ -0,0 +1,160 @@
...
+            self.index_char_range: dict = index_results_dict["index_char_range"]
+
+        except Exception as e:
+            logger.warning(f"Could not instantiate Top Pick Provider: {e}")
+            raise e
+
+    def hidden(self) -> bool:  # noqa: D102
+        return False
+
+    async def query(self, q: str) -> Any:
```
ncloudioj, line 73: Let's stick to the abstract method defined in providesr/base.py.
```suggestion
    async def query(self, srequest: SuggestionRequest) -> list[BaseSuggestion]:
:
```

merino/providers/top_picks.py
```diff
@@ -0,0 +1,160 @@
...
+        if ids := self.primary_index.get(q, []):
+            res = self.results[ids[0]]
+            return [res]
+        elif ids := self.secondary_index.get(q, []):
+            res = self.results[ids[0]]
+            return [res]
+        return []
+
+    @staticmethod
+    def read_domain_list(file: str) -> dict[str, Union[str, int, list[str]]]:
```
ncloudioj, line 86:
```suggestion
    def read_domain_list(file: str) -> dict[str,Any]:
```

merino/providers/top_picks.py
```diff
@@ -0,0 +1,160 @@
...
+        try:
+            with open(file, "r") as readfile:
+                domain_list: dict = json.load(readfile)
+                return domain_list
+        except Exception as e:
+            logger.warning("Cannot Process File: {e}")
+            raise e
+
+    @staticmethod
+    def build_index(domain_list: dict) -> dict[str, Union[list, dict]]:
```
ncloudioj, line 100:
```suggestion
    def build_index(domain_list: dict[str, Any]) -> dict[str, Any]:
```

merino/providers/top_picks.py
```diff
@@ -0,0 +1,160 @@
...
+
+        return {
+            "primary_index": primary_index,
+            "secondary_index": secondary_index,
+            "results": results,
+            "index_char_range": index_char_range,
+        }
+
+    @staticmethod
+    def build_indices() -> dict[str, Union[list, dict]]:
```
ncloudioj, line 156:
```suggestion
    def build_indices() -> dict[str, Any]:
```

tests/unit/providers/test_top_picks.py
taddes: Good call!

merino/providers/top_picks.py
taddes: Because of how the logic is structured with the following check:
`            if len(record["domain"]) < QUERY_CHAR_LIMIT:
                continue
`
a record under that char limit as set in the config will not be appended, hence the minimum value will match the minimum value set in config. This way we don't deal with the logic here making that determination based on what domain lists we get. Thoughts? That's why I kept it this way, so 2 checks aren't necessary

merino/providers/top_picks.py
```diff
@@ -0,0 +1,158 @@
...
+            self.index_char_range: dict = index_results_dict["index_char_range"]
+
+        except Exception as e:
+            logger.warning(f"Could not instantiate Top Pick Provider: {e}")
+            raise e
+
+    def hidden(self) -> bool:  # noqa: D102
+        return False
+
+    async def query(self, srequest: str) -> list[BaseSuggestion]:
```
ncloudioj, line 71: The abstract method is defined [here](https://github.com/mozilla-services/merino-py/blob/main/merino/providers/base.py#L45). It's implemented by all other providers.

merino/providers/top_picks.py
```diff
@@ -0,0 +1,155 @@
...
+
+logger = logging.getLogger(__name__)
+
+
+class Suggestion(BaseSuggestion):
+    """Model for Top Pick Query Suggestion"""
+
+    block_id: int
+    is_top_pick: bool
+    score: float
```
ncloudioj, line 27: Forgot dropping `score`? Also, no need to redefine `is_sponsored` here, you can set it to `False` later upon the object instantiation.

merino/providers/top_picks.py
```diff
@@ -0,0 +1,155 @@
...
+class Provider(BaseProvider):
+    """Top Pick Query Suggestion Provider"""
+
+    # In normal usage this is None, but tests can create the provider with a
+    # FastAPI instance to fetch mock responses from it. See `__init__()`.
+    _app: Optional[FastAPI]
+
+    primary_index: defaultdict = defaultdict(list)
+    secondary_index: defaultdict = defaultdict(list)
+    results: list[dict[str, Union[str, list, int]]]
```
ncloudioj, line 40:
```suggestion
    results: list[dict[str, Any]
```

merino/providers/top_picks.py
```diff
@@ -0,0 +1,155 @@
...
+            logger.warning(f"Could not instantiate Top Pick Provider: {e}")
+            raise e
+
+    def hidden(self) -> bool:  # noqa: D102
+        return False
+
+    async def query(self, srequest: str) -> list[BaseSuggestion]:
+        """Query Top Pick provider."""
+        if srequest.startswith("http"):
+            return []
```
ncloudioj, line 72: Don't forget to add the length range check here.

merino/providers/top_picks.py
```diff
@@ -0,0 +1,155 @@
...
+            domain = record["domain"]
+
+            if len(record["domain"]) < QUERY_CHAR_LIMIT:
+                continue
+
+            suggestion = Suggestion(
+                block_id=0,
+                title=record["title"],
+                url=record["url"],
+                provider="top_picks",
```
ncloudioj, line 121:
```suggestion
                provider=self.name,
```

merino/providers/top_picks.py
```diff
@@ -0,0 +1,155 @@
...
+                is_top_pick=True,
+                icon="",
```
ncloudioj, line 123:
```suggestion
                icon=record["icon"],
```

merino/providers/top_picks.py
```diff
@@ -0,0 +1,155 @@
...
+                score=settings.providers.top_picks.score,
```
ncloudioj, line 124:
```suggestion
                score=SCORE,
```

tests/unit/web/test_top_picks.py
```diff
@@ -0,0 +1,44 @@
...
+from tests.unit.web.util import get_provider_factory
+
+client = TestClient(app)
+
+
+@pytest.fixture(scope="module", autouse=True)
+def inject_providers():
+    app.dependency_overrides[get_providers] = get_provider_factory(
+        {
+            "wiki_fruit": TopPicksProvider(name="top_picks", enabled_by_default=True),
```
ncloudioj, line 16:
```suggestion
            "top_picks": TopPicksProvider(name="top_picks", enabled_by_default=True),
```

merino/providers/top_picks.py
taddes: Yep, took  original out when I was refactoring

merino/providers/top_picks.py
taddes: This won't work was the `build_index` method is `@staticmethod`, so it doesn't have access to the instance, same issue with using `Provider` as the class isn't instantiated. Can I leave it as a str or do we do something else? I can set it as a setting in the `toml` file.

merino/providers/top_picks.py
taddes: Thanks @ncloudioj . So in each individual suggestion that is appended to the list in the `build_index` method, I set `is_sponsored` to `False`. That work?

merino/providers/top_picks.py
taddes: Hm, so again with the min range, perhaps I'm just missing something, but if we have the query limit set to say 4, it will always iterate through a domain string when building the index, meaning it will always, at minimum, be 4 characters.

Originally, when I had the check whether to index the suggestion or not, my thought was that if for some reason, a domain (and also its similars) could in a rare case be lower, causing an index error when iterating through the string. So, that's why I didn't add it. Edge case, but do you think for that reason alone I should add it back? It would be a silly reason for the indexing operation to fail.

As far as querying is concerned, the max is easy, based on the longest domain string in the collection of indexes. However, for the min, if the spec of the provider is to, at minimum, trigger a lookup at 4 characters, is it really worthwhile updating the min? Doesn't that essentially bypass the trigger limit of 4.   Perhaps I'm just confused on this point. We certainly won't update min any lower, but in terms of updating it higher, doesn't the char limit also set that no matter what, a query will at least yield a result so long as it hits the character limit, and if we are indexing using that same limitation, won't those values be the same? Thanks for discussing this.

taddes:
@ncloudioj Ready for review, though I know we need to add the new `base.py` changes @pjenvey landed and then I need to update the code to implement `SuggestionRequest`, so once final reviews are done (I responded in a few comments), I can rebase onto master. Thoughts? Thanks again!

merino/providers/top_picks.py
ncloudioj: Exactly. Also, can you remove `score` as well, it's already defined in the base class, no need to repeat that again here.

merino/providers/top_picks.py
```diff
@@ -0,0 +1,155 @@
...
+    """Top Pick Query Suggestion Provider"""
+
+    # In normal usage this is None, but tests can create the provider with a
+    # FastAPI instance to fetch mock responses from it. See `__init__()`.
+    _app: Optional[FastAPI]
+
+    primary_index: defaultdict = defaultdict(list)
+    secondary_index: defaultdict = defaultdict(list)
+    results: list[dict[str, Union[str, list, int]]]
+    index_char_range = dict[str, int]
```
ncloudioj, line 41: Let's not waste two dict lookups for each query call for this. Please use two variables instead `index_char_min` and `index_char_max`.

merino/providers/top_picks.py
ncloudioj: You're right. I was wrong about the minimum, sorry about that. Let's just track maximum then.

ncloudioj (approved):
Looks good. r+ with comments.

Thanks!

merino/providers/top_picks.py
taddes: You got it, artifact from removing the original test and dict lookup. Thx

taddes: Union above was removed in another commit, fyi. Must have not hit when commented on.

merino/providers/top_picks.py
taddes: 👍

merino/providers/top_picks.py
taddes: Thx. I took score out in a subsequent commit, so by the time you commented I had removed. Thx for confirming.

merino/providers/top_picks.py
taddes: Ok, thx!

merino/providers/top_picks.py
```diff
@@ -0,0 +1,157 @@
...
+        except Exception as e:
+            logger.warning(f"Could not instantiate Top Pick Provider: {e}")
+            raise e
+
+    def hidden(self) -> bool:  # noqa: D102
+        return False
+
+    async def query(self, srequest: str) -> list[BaseSuggestion]:
+        """Query Top Pick provider and return suggestion"""
+        if srequest.startswith("http"):
```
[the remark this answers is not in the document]
taddes: Always a good  to add some context

merino/providers/top_picks.py
```diff
@@ -0,0 +1,157 @@
...
+QUERY_CHAR_LIMIT: int = settings.providers.top_picks.query_char_limit
+
+
+logger = logging.getLogger(__name__)
+
+
+class Suggestion(BaseSuggestion):
+    """Model for Top Pick Query Suggestion"""
+
+    block_id: int
```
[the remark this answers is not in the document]
taddes: Yeah, it'll have a purpose later which is why we're keeping it, but I am not 100% sure.

merino/providers/top_picks.py
```diff
@@ -0,0 +1,157 @@
...
+class Provider(BaseProvider):
+    """Top Pick Query Suggestion Provider"""
+
+    # In normal usage this is None, but tests can create the provider with a
+    # FastAPI instance to fetch mock responses from it. See `__init__()`.
+    _app: Optional[FastAPI]
+
+    primary_index: defaultdict = defaultdict(list)
+    secondary_index: defaultdict = defaultdict(list)
+    results: list[dict[str, Any]]
```
pjenvey, line 38: Noticed a few type mismatches as well, I'm surprised mypy didn't catch any of them?

```suggestion
    results: list[Suggestion]
```

merino/providers/top_picks.py
```diff
@@ -0,0 +1,157 @@
...
+        self._app = app
+        self._name = name
+        self._enabled_by_default = enabled_by_default
+
+    async def initialize(self) -> None:
+        """Initialize the provider."""
+        try:
+            index_results_dict: dict = await asyncio.to_thread(Provider.build_indices)
+            self.primary_index: dict = index_results_dict["primary_index"]
+            self.secondary_index: list = index_results_dict["secondary_index"]
```
pjenvey, line 57:
```suggestion
            self.primary_index: defaultdict = index_results_dict["primary_index"]
            self.secondary_index: defaultdict = index_results_dict["secondary_index"]
```

pjenvey (changes requested):

merino/providers/top_picks.py
```diff
@@ -0,0 +1,157 @@
...
+    async def query(self, srequest: str) -> list[BaseSuggestion]:
+        """Query Top Pick provider and return suggestion"""
+        if srequest.startswith("http"):
+            return []
+        if len(srequest) < self.query_min or len(srequest) > self.query_max:
+            return []
+        if ids := self.primary_index.get(srequest, []):
+            res = self.results[ids[0]]
+            return [res]
+        elif ids := self.secondary_index.get(srequest, []):
```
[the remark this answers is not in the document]
taddes: Good point, walrus operator would be falsy either way.

taddes:
Thanks for the comments @pjenvey ! If you're ok with approving this now or adding another review so I can merge, that'd be great, merging currently blocked due to requested changes.

pjenvey (approved):

PR #90, opened 2022-10-26
feat: add full keyword to merino response

tiftran:
closes [CONSVC-2070](https://mozilla-hub.atlassian.net/browse/CONSVC-2070)


merino/providers/adm.py
```diff
@@ -184,11 +185,18 @@ async def _fetch(self) -> None:
...
             res = await done_task
             for suggestion in res.json():
-                id = len(results)
-                for kw in suggestion.pop("keywords", []):
-                    # Note that for adM suggestions, each keyword can only be mapped to
-                    # a single suggestion.
-                    suggestions[kw] = id
+                kw_id = len(results)
+                keywords = suggestion.pop("keywords")
+                full_keywords = suggestion.pop("full_keywords")
```
ncloudioj, line 190: Just a note that since Remote Settings records currently do not have this `full_keywords` field, this PR cannot be merged until we upload the new records to RS.

merino/providers/adm.py
```diff
@@ -184,11 +185,18 @@ async def _fetch(self) -> None:
         for done_task in as_completed(fetch_tasks):
             res = await done_task
             for suggestion in res.json():
-                id = len(results)
-                for kw in suggestion.pop("keywords", []):
-                    # Note that for adM suggestions, each keyword can only be mapped to
-                    # a single suggestion.
-                    suggestions[kw] = id
+                kw_id = len(results)
```
ncloudioj, line 188: Does `kw_id` stands for "keyword id"? I think `result_id` or just `id` is more accurate.

merino/providers/adm.py
```diff
@@ -184,11 +185,18 @@ async def _fetch(self) -> None:
...
+                full_keywords = suggestion.pop("full_keywords")
+                begin = 0
+                for full_keyword, n in full_keywords:
+                    for query, fkw in zip(
+                        islice(keywords, begin, begin + n), repeat(full_keyword, n)
+                    ):
+                        # Note that for adM suggestions, each keyword can only be mapped to
+                        # a single suggestion.
+                        suggestions[query] = (kw_id, full_keyword)
+                        begin += n
```
ncloudioj, line 199: Incrementing `begin` within the inner loop would be catastrophic 😁 

```suggestion
                    begin += n
```

tiftran: eek

merino/providers/adm.py
quiiver: This is the kind of scenario where feature flags can be useful:

* have a downstream dependency
* branch in code with feature flags to account for downstream dependency
* merge the branched code with confidence that you won't currently break anything and you can move on
  * this is generally good because you don't want PRs sitting around and bit rotting
* downstream dep gets resolved
* push a config change to flip the new path on (either for a % or 100%)
* validate that it works as expected
* remove branch in code and clean up flag

tiftran: yup understood!

quiiver: (i'm not saying we have to use feature flags here, just calling out that this is a good use case for them)

merino/providers/adm.py
tiftran: i think `id` is reserved in python, was trying to not use it, result_id makes more sense you're right

merino/providers/adm.py
ncloudioj: Hmm, did this break anything in the contract tests? If not, I think we should add some tests to cover it.

tests/contract/volumes/kinto/data-02.json
```diff
@@ -11,6 +11,10 @@
         "keywords": [
             "coffee",
             "kaffee"
+        ],
+        "full_keywords": [
+           [ "coffee",1],
```
ncloudioj, line 16:
```suggestion
           ["coffee",1],
```

merino/providers/adm.py
```diff
@@ -100,7 +101,7 @@ class NonsponsoredSuggestion(BaseSuggestion):
 class Provider(BaseProvider):
     """Suggestion provider for adMarketplace through Remote Settings."""
 
-    suggestions: dict[str, int] = {}
+    suggestions: dict[str, Tuple[int, str]] = {}
```
ncloudioj, line 104: Throwing this idea for discussion: alternatively, we can keep all the full keywords in a separate list (like what we do for suggestions with `results`), then we don't need to save a copy for each query, rather we just need an index that points to an element of the full keyword list (i.e. `Tuple[int, str] -> Tuple[int, int]`). Doing so could save some memory for Merino as most full keywords are pretty long.

merino/providers/adm.py
tiftran: had the same thinking

tiftran: tests didn't pick up on this since all the tuples had values of 1, so the loop only go through once, changed contract test data for [this](https://github.com/mozilla-services/merino-py/pull/90/files#diff-39fd904b576c25a9c97637fd95124f3dab01b928e759b621930a33fde9d711d0R104-R107)

merino/providers/adm.py
hackebrot: I second this! Would be great to use this as an opportunity for all of us to test out feature flags in Merino. 🚀 

(FWIW my PR for integrating feature flags in metrics will be up later today)

merino/providers/adm.py
Trinaa: This kind of logical coverage should be done in unit tests ideally. Could some be added?

ncloudioj: Yes, I think we can add some unit tests to `tests/unit/providers/test_adm.py` with some changes to the `FakeBackend` defined in there.

merino/providers/adm.py
```diff
@@ -181,14 +185,28 @@ async def _fetch(self) -> None:
...
             for suggestion in res.json():
-                id = len(results)
-                for kw in suggestion.pop("keywords", []):
-                    # Note that for adM suggestions, each keyword can only be mapped to
-                    # a single suggestion.
-                    suggestions[kw] = id
+                result_id = len(results)
+                keywords = suggestion.pop("keywords")
+                full_keywords_tuples = suggestion.pop("full_keywords")
+                full_keywords_list.append(full_keywords_tuples[0][0])
```
ncloudioj, line 196: Hmm, why insert the first full_keyword to the list over here? I think we can just do everything inside the loop below.

```py
fkw_index = 0
for full_keyword, n in full_keywords_tuples:
    for query, fkw in zip(...):
        suggestion[query] = (result_id, fkw_index)
    begin += n
    full_keywords_list.append(full_keyword)
    fkw_index += 1
results.append(suggestion)
```

merino/providers/adm.py
tiftran: okie doke, made some changes to `test_adm.py` for more realistic scenarios

merino/providers/adm.py
```diff
@@ -160,7 +162,9 @@ async def _fetch(self) -> None:
         """Fetch suggestions, keywords, and icons from Remote Settings."""
         # A dictionary keyed on suggestion keywords, each value stores an index
         # (pointer) to one entry of the suggestion result list.
-        suggestions: dict[str, int] = {}
+        suggestions: dict[str, Tuple[int, int]] = {}
+        # A list of full keywords
+        full_keywords_list: list[str] = []
```
ncloudioj, line 167: Nit: to be consistent with other variables, let's drop the suffix `list`.
```suggestion
        full_keywords: list[str] = []
```

merino/providers/adm.py
```diff
@@ -181,14 +185,25 @@ async def _fetch(self) -> None:
...
             res = await done_task
+
             for suggestion in res.json():
-                id = len(results)
-                for kw in suggestion.pop("keywords", []):
-                    # Note that for adM suggestions, each keyword can only be mapped to
-                    # a single suggestion.
-                    suggestions[kw] = id
+                result_id = len(results)
+                keywords = suggestion.pop("keywords")
```
ncloudioj, line 194:
```suggestion
                keywords = suggestion.pop("keywords", [])
```

merino/providers/adm.py
```diff
@@ -181,14 +185,25 @@ async def _fetch(self) -> None:
...
+                full_keywords_tuples = suggestion.pop("full_keywords")
```
ncloudioj, line 195:
```suggestion
                full_keywords_tuples = suggestion.pop("full_keywords", [])
```

merino/providers/adm.py
```diff
@@ -209,12 +225,14 @@ def hidden(self) -> bool:  # noqa: D102
     async def query(self, srequest: SuggestionRequest) -> list[BaseSuggestion]:
         """Provide suggestion for a given query."""
         q = srequest.query
-        if (id := self.suggestions.get(q)) is not None:
-            res = self.results[id]
+        if (suggest_look_ups := self.suggestions.get(q)) is not None:
+            results_id = suggest_look_ups[0]
+            fkw_id = suggest_look_ups[1]
```
ncloudioj, line 230: Nit: result_id, fkw_id = suggest_look_ups

ncloudioj (approved):
LGTM, r+ with comments.

The changes in quicksuggest-rs are ready to land as well, so we don't need to wrap this inside a featureflag, let's get that landed first, and then merge this one.

merino/providers/adm.py
```diff
@@ -182,14 +186,25 @@ async def _fetch(self) -> None:
...
-                    # a single suggestion.
-                    suggestions[kw] = id
+                result_id = len(results)
+                keywords = suggestion.pop("keywords", [])
+                full_keywords_tuples = suggestion.pop("full_keywords", [])
+                begin = 0
+                for full_keyword, n in full_keywords_tuples:
+                    for query, fkw in zip(
+                        islice(keywords, begin, begin + n), repeat(full_keyword, n)
+                    ):
```
pjenvey, line 201: `fkw` isn't used so I don't think the `zip` necessary? also `keywords` should always be a list so do we need `islice`?
```suggestion
                    for query in keywords[begin : begin + n]:
```

pjenvey (changes requested):

pjenvey (approved):

PR #96, opened 2022-10-27
[CONSVC-2015] feat: lower down score for adm wikipedia suggestions

ncloudioj:
This enforces the ranking criteria of "Sponsored Suggestions (0.3)" > "Top Picks (0.25)" > "Wikipedia (0.2)" >= "Local Remote Settings (0.2)". It's a bit messy to juggle those many scores, but we will clean those up once more sophisticated ranking gets added to Merino.

This fixes [CONSVC-2015](https://mozilla-hub.atlassian.net/browse/CONSVC-2015).

### DIFF
--- a/tests/unit/providers/test_adm_wikipedia.py
+++ b/tests/unit/providers/test_adm_wikipedia.py
@@ -1,0 +1,122 @@
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+import json
+from typing import Any
+
+import httpx
+import pytest
+
+from merino.config import settings
+from merino.providers.adm import NonsponsoredSuggestion, Provider
+from tests.unit.web.util import srequest
+
+
+class FakeBackend:
+    """Fake Remote Settings backend that returns suggest data for tests."""
+
+    async def get(self, bucket: Any, collection: Any) -> list[dict[Any, Any]]:
+        """Return fake records."""
+
+        return [
+            {
+                "type": "data",
+                "schema": 111,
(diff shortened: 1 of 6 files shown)


tests/unit/providers/test_adm_wikipedia.py
```diff
@@ -0,0 +1,122 @@
...
+
+from merino.config import settings
+from merino.providers.adm import NonsponsoredSuggestion, Provider
+from tests.unit.web.util import srequest
+
+
+class FakeBackend:
+    """Fake Remote Settings backend that returns suggest data for tests."""
+
+    async def get(self, bucket: Any, collection: Any) -> list[dict[Any, Any]]:
```
taddes, line 18: `list[dict[str, Any]] instead`?

tests/unit/providers/test_adm_wikipedia.py
```diff
@@ -0,0 +1,122 @@
...
+                    "location": "main-workspace/quicksuggest/icon-01",
+                    "mimetype": "application/octet-stream",
+                },
+                "content_type": "image/png",
+                "id": "icon-01",
+                "last_modified": 123,
+            },
+        ]
+
+    async def fetch_attachment(self, attachment_uri: Any) -> httpx.Response:
```
taddes, line 51: Might we want this URI to be a `str`, or is there a chance another value would be passed in?

taddes (approved):
Looks great @ncloudioj , just insignificant optional mypy suggestions for some `str` types. Tested and works on my end. You'll also likely just have to do a little rebase in the config docs as the Top Pick provider was merged in.

PR #105, opened 2022-11-07
[CONSVC-2100] feat: index domains under limit

### LINKED ISSUE #104: Index domains below the minimal index threshold
Per Product, we also wanted to include those top domains (e.g msn.com) with less than 4 characters (i.e. query_char_limit). The results should be labelled as TopPicks as usual. Note that since search queries sent from Firefox are at least 2 characters long, we don’t need to index anything below that.

Recommended that a separate index (e.g. short_domain_index) for those exceptions so that the existing primary & secondary indices can be kept as is. This index would include all prefixes from 2 to query_char_limit - 1 for those short domains. And then, given a search query from Firefox, if its length is between 2 and query_char_limit - 1, Merino would search from this index.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/CONSVC-2103)
┆Link To Issue: https://mozilla-hub.atlassian.net/browse/CONSVC-2103

taddes:
Implemented a new index to store query results that are between the FIrefox minimum of 2 characters, and those that are less than the set limit of 4 (though it is established this limit could change in the future).  An index is built to point to the same result set and the short index also supports similars for these domains, keeping their similars and short character variants in their own index.

Added unit, contract and integration tests

Closes #104


taddes:
Thanks for discussing today @ncloudioj , I added contract test scenarios to cover this too, after our meeting. Lmk how things look! @Trinaa , if you had any thoughts on the test setup for this different scenario, you're welcome to add any thoughts you have!

merino/providers/top_picks.py
```diff
@@ -73,6 +76,13 @@ async def query(self, srequest: SuggestionRequest) -> list[BaseSuggestion]:
         # Ignore https:// and http://
         if srequest.query.startswith("http"):
             return []
+        # Suggestions between Firefox char min of 2 and query limit - 1 for short domains
+        if len(srequest.query) >= FIREFOX_CHAR_LIMIT and len(srequest.query) <= (
+            QUERY_CHAR_LIMIT - 1
+        ):
```
Trinaa, line 82: Could this condition be simplified to the following?

```suggestion
    if FIREFOX_CHAR_LIMIT <= len(srequest.query) <= (QUERY_CHAR_LIMIT - 1):
```

tests/unit/providers/test_top_picks.py
```diff
@@ -154,3 +154,60 @@ async def test_query(srequest: SuggestionRequestFixture, top_picks: Provider) ->
...
+        Suggestion(
+            block_id=0,
+            title="Abc",
+            url="https://abc.test",
+            provider="top_picks",
+            is_top_pick=True,
+            is_sponsored=False,
+            icon="",
+            score=settings.providers.top_picks.score,
+        )
```
Trinaa, line 212: Is this a duplicate of lines 187-199?

tests/unit/providers/test_top_picks.py
```diff
@@ -154,3 +154,60 @@ async def test_query(srequest: SuggestionRequestFixture, top_picks: Provider) ->
             score=settings.providers.top_picks.score,
         )
     ]
+
+
+@pytest.mark.asyncio
+async def test_short_domain_query(
+    srequest: SuggestionRequestFixture, top_picks: Provider
+) -> None:
```
Trinaa, line 162: I suggest to parameterize this test to run each query, or case, in isolation.

tests/integration/api/v1/suggest/test_top_picks.py
```diff
@@ -52,3 +52,26 @@ def test_top_picks_no_result(client: TestClient, query: str):
...
+    """Test if Top Picks provider responds with a short domain or similar"""
+    response = client.get(f"/api/v1/suggest?q={query}")
+    assert response.status_code == 200
+
+    result = response.json()
+    assert result
+    assert result["suggestions"][0]["url"] == url
+    assert result["suggestions"][0]["title"] == title
+    assert result["suggestions"][0]["is_top_pick"]
+    assert not result["suggestions"][0]["is_sponsored"]
```
Trinaa, line 77: I suggest asserting against an expected list of suggestions, instead of suggestion attributes. So that if Suggestion is updated, the developer making the change will have a better understanding of the impact. Example:

```suggestion
    """Test if Top Picks provider responds with a short domain or similar"""
    expected_suggestions: [Suggestion] = [Suggestion(
        title=title,
        url=url,
        provider="top_picks",
        is_sponsored=False,
        score=0.25,
        icon="",
        block_id=0,
        is_top_pick=True,
    )]

    response = client.get(f"/api/v1/suggest?q={query}")

    assert response.status_code == 200
    result = response.json()
    assert result
    assert result["suggestions"] == expected_suggestions
```

Trinaa (commented):
Love to see the 3 tier test strategy! ❤️ 

I left a few comments, let me know what you think!

merino/providers/top_picks.py
```diff
@@ -14,6 +14,7 @@
 SCORE: float = settings.providers.top_picks.score
 LOCAL_TOP_PICKS_FILE: str = settings.providers.top_picks.top_picks_file_path
 QUERY_CHAR_LIMIT: int = settings.providers.top_picks.query_char_limit
+FIREFOX_CHAR_LIMIT: Final = 2
```
ncloudioj, line 17:
```suggestion
# The minimal characters that Firefox Urlbar would send to Merino 
URLBAR_MIN_CHAR_LIMIT: Final(int) = 2
```

merino/providers/top_picks.py
```diff
@@ -133,6 +145,17 @@ def build_index(domain_list: dict[str, Any]) -> dict[str, Any]:
                 score=SCORE,
             )
 
+            # Insertion of short keys between Firefox limit of 2 and QUERY_CHAR_LIMIT - 1
+            if (
+                len(domain) >= FIREFOX_CHAR_LIMIT
+                and len(domain) <= QUERY_CHAR_LIMIT - 1
```
ncloudioj, line 151:
```suggestion
                FIREFOX_CHAR_LIMIT <= len(domain) < QUERY_CHAR_LIMIT
```

merino/providers/top_picks.py
ncloudioj: Or `FIREFOX_CHAR_LIMIT <= len(domain) < QUERY_CHAR_LIMIT`

dev/top_picks_for_ci.json
```diff
@@ -45,6 +45,21 @@
...
+            "domain": "abc",
+            "url": "https://abc.test",
+            "icon": "",
+            "categories": [
+                "web-browser"
+            ],
+            "similars": [
+                "aa",
+                "ab",
+                "acb"
```
ncloudioj, line 61: We should also consider the case where short domains might have long similar variants whose lengths are great or equal to `QUERY_CHAR_LIMIT`. For example, we should include "abcd" to `similars`. In such case, I think we should index "abcd" into the `self.secondary_index` as usual.

merino/providers/top_picks.py
```diff
@@ -133,6 +145,17 @@ def build_index(domain_list: dict[str, Any]) -> dict[str, Any]:
...
+            ):
+                for chars in range(FIREFOX_CHAR_LIMIT, len(domain) + 1):
+                    short_domain_index[domain[:chars]].append(index_key)
+                for variant in record.get("similars", []):
```
ncloudioj, line 155: As mentioned above, we should check the length of `variant`. If it's greater than or equal to `QUERY_CHAR_LIMIT`, it should be added to the secondary index. Otherwise, it will break the invariant of all queries in `short_domain_index` should be in the range `[FIREFOX_CHAR_LIMIT, QUERY_CHAR_LIMIT)`.

tests/unit/providers/test_top_picks.py
taddes: Good catch, intent was `ab` for that one.

merino/providers/top_picks.py
taddes: Of course. I am a fan of imperative paradigms when handling these since, at least to my eyes, I can read and understand what is going on faster as I think it takes a bit longer to understand that expression, but this is a pretty simple case. I'll change it.

merino/providers/top_picks.py
taddes: Excellent! Good comment. Added type annotation to Final as well

merino/providers/top_picks.py
taddes: Sounds good. What are your thoughts on retaining `(QUERY_CHAR_LIMIT - 1)` instead of just `QUERY_CHAR_LIMIT` ? I know the `-1` is redundant when removing `<=`, but given this handles the limits a bit differently, I liked the `-1`.  Whatever you think is best, I am happy to do so let me know your thoughts!

ncloudioj: They're the same, so it's up to you :). I chose the half-open interval because it saved the parentheses and -1

tests/unit/providers/test_top_picks.py
taddes: 👍

taddes:
Hi @ncloudioj @Trinaa ! I have made the requested changes from your reviews. Here is a summary:
- Parameterized several test functions
- Matching against returned suggestion constructed by parameterized inputs
- Refactored some logic in tests
- Refactored the bounds checking for query, while retaining `QUERY_CHAR_LIMIT - 1` foe clarity between it and the standard `QUERY_CHAR_LIMIT`, I feel it reads more easily for someone reading the code, however I am happy to alter if you feel it is needless still.
- @ncloudioj , to your concern regarding similars being longer than the short domain in the first place, here are my thoughts in response to your commentary so feel free to discuss:
As it is, I have the logic segmented to the short index if any domain hits the threshold of 2 char to 3 char,  our current limits. Given that we iterate though with the 2 char minimum and check against both the domain length, or the similar domain length (variant), any similar domain that is longer than its parent domain will produce a matching result.  I kind of like that these are all segmented from the primary and secondary index, because it isolates the logic for short domains which are triggered by their parent domain. If we were to try and append to the secondary domain, I feel this would clutter the logic and pollute the second_index.  As it stands, both the indexing and querying is triggered by the limitation of these characters, so whether it is the actual term or the similar, shorter or longer, all tests verify that the results return as expected. Lmk what you think, we can discuss further.

dev/top_picks_for_ci.json
taddes: I have tests verifying that this works successfully already.  I had a longer comment in PR after these changes stating the rationale from keeping this segmented from `secondary_index`, which would complicate things and not keep the shorter domains segmented out from the logic of the other indexes.

merino/providers/top_picks.py
```diff
@@ -133,6 +144,16 @@ def build_index(domain_list: dict[str, Any]) -> dict[str, Any]:
...
 
+            # Insertion of short keys between Firefox limit of 2 and QUERY_CHAR_LIMIT - 1
+            # QUERY_CHAR_LIMIT can be modified, therefore suggestions in this category
+            # should be less than QUERY_CHAR_LIMIT, not equal.
+            if FIREFOX_CHAR_LIMIT <= len(domain) <= (QUERY_CHAR_LIMIT - 1):
+                for chars in range(FIREFOX_CHAR_LIMIT, len(domain) + 1):
+                    short_domain_index[domain[:chars]].append(index_key)
+                for variant in record.get("similars", []):
+                    for chars in range(FIREFOX_CHAR_LIMIT, len(variant) + 1):
+                        short_domain_index[variant[:chars]].append(index_key)
```
ncloudioj, line 155: There should have a `continue` at the end of this if block, otherwise, it will follow through and insert the keywords into both primary and secondary indices again. I think that's why all the tests passed down below.

merino/providers/top_picks.py
```diff
@@ -73,6 +77,11 @@ async def query(self, srequest: SuggestionRequest) -> list[BaseSuggestion]:
         # Ignore https:// and http://
         if srequest.query.startswith("http"):
             return []
+        # Suggestions between Firefox char min of 2 and query limit - 1 for short domains
+        if FIREFOX_CHAR_LIMIT <= len(srequest.query) <= (QUERY_CHAR_LIMIT - 1):
```
ncloudioj, line 81: @taddes Regarding your last comment on putting all the keywords for short domains and their similars into the `short_doamin_index` rather than `secondary_index`, that will be questionable because this check would only be true if the query length is less than `QUERY_CHAR_LIMIT`, queries for those long similars will never be able to enter here and do the lookup.

ncloudioj:
>  If we were to try and append to the secondary domain, I feel this would clutter the logic and pollute the second_index. As it stands, both the indexing and querying is triggered by the limitation of these characters, so whether it is the actual term or the similar, shorter or longer, all tests verify that the results return as expected. Lmk what you think, we can discuss further.

@taddes As pointed out in the review comments, looks like Merino already saves keywords for those long simiars in the secondary index, which is why all the tests passed ;).

I'd argue that if we were to store all keywords for short domains as well as their similars into the `short_domain_index` regardless of their length, that would violate the invariant of "all keywords in short_domain_index should be at least (FIREFOX_CHAR_LIMIT) characters and no more than (QUERY_CHAR_LIMIT - 1)".

Does that make sense?

merino/providers/top_picks.py
taddes: Hey @ncloudioj , yes that makes sense now, sorry for my confusion. That query check does eliminate the possible smiliars results should the term be shorter. I will add a check for that condition of a short domain similar in the second index.

merino/providers/top_picks.py
taddes: Yup, that's it. That's why!

taddes:
Thanks for catching that missing `continue`. Lmk if the change as is is ok, I did have it now add all similars to the secondary index and added a check for the secondary index if the short domain query is triggered. Also added that missing `continue`! Happy to make whatever minor tweaks. I did add a check in the case that if there were no similars in the short character limit, to continue without attempting to iterate though the secondary indexes. Seems paranoid, I suppose, but I thought it could be useful, esp if there are no similars.

merino/providers/top_picks.py
```diff
@@ -73,7 +77,15 @@ async def query(self, srequest: SuggestionRequest) -> list[BaseSuggestion]:
         # Ignore https:// and http://
         if srequest.query.startswith("http"):
             return []
-        # Ignore requests below or above character minimums
+        # Suggestions between Firefox char min of 2 and query limit - 1 for short domains
+        if FIREFOX_CHAR_LIMIT <= len(srequest.query) <= (QUERY_CHAR_LIMIT - 1):
+            if ids := self.short_domain_index.get(srequest.query):
+                res = self.results[ids[0]]
+                return [res]
+            elif ids := self.secondary_index.get(srequest.query):
```
ncloudioj, line 85: This else branch is not needed and will be unreachable, because of this check `if FIREFOX_CHAR_LIMIT <= len(srequest.query) <= (QUERY_CHAR_LIMIT - 1)`. So we can just drop it.

merino/providers/top_picks.py
```diff
@@ -133,6 +147,19 @@ def build_index(domain_list: dict[str, Any]) -> dict[str, Any]:
...
+            if FIREFOX_CHAR_LIMIT <= len(domain) <= (QUERY_CHAR_LIMIT - 1):
+                for chars in range(FIREFOX_CHAR_LIMIT, len(domain) + 1):
+                    short_domain_index[domain[:chars]].append(index_key)
+                if not record.get("similars"):
+                    continue
+                else:
+                    for variant in record.get("similars", []):
+                        for chars in range(FIREFOX_CHAR_LIMIT, len(variant) + 1):
+                            secondary_index[variant[:chars]].append(index_key)
+                        continue
```
ncloudioj, line 161: I think this can be simplified as well. Will comment in more detail in a bit.

merino/providers/top_picks.py
taddes: Actually removing that breaks the tests because a little below, we do the general bounds checking for max and min and the min case is triggered. This subsequent check is needed so overly short or excessively long queries don't trigger a lookup.

merino/providers/top_picks.py
taddes: Yeah, I can likely just remove the if check if there is no record, I was just being explicit because this was easy to miss, proven by my missing the continue altogether and the test passing.

merino/providers/top_picks.py
ncloudioj: I guess that's because you insert the *entire* set of keywords for similars into the secondary index below. I think we should do this instead,

```
for similar in similars:
    if IREFOX_CHAR_LIMIT <= len(srequest.query) <= (QUERY_CHAR_LIMIT - 1):
        # Insert keywords into short_domain_index
    else:
        # Insert keywords into secondary_index. Note that keywords in this case must be at least QUERY_CHAR_LIMIT
```

merino/providers/top_picks.py
ncloudioj: I think this should just work.

```suggestion
                for variant in record.get("similars"):
                    if len(variant) >= QUERY_CHAR_LIMIT:
                        # Long variants will be indexed later into `secondary_index`
                        continue
                        
                    for chars in range(FIREFOX_CHAR_LIMIT, len(varaint) + 1):
                        short_domain_index[varinat[:chars]].append(index_key)
```

ncloudioj: Note that `continue` is intentionally omitted as we need the following indexer to index the long variant. Both short domain and short variants will be ignored because that `for chars in range(QUERY_CHAR_LIMIT, len(domain) + 1` will be no-op for them.

taddes:
Thanks again @ncloudioj for the revisions and change in thinking about how the logic works, esp letting the secondary index be the source for longer variants of the shorter domains (a mouthful, that sentence!) Let me know how this looks now, should be good.

merino/providers/top_picks.py
taddes: Good call here.  I also amended tests for the longer variants as their shorter keywords of 2 - 3 chars will not trigger a lookup, just as it works with the standard domains that show at 4 char and >.

merino/providers/top_picks.py
```diff
@@ -133,12 +147,25 @@ def build_index(domain_list: dict[str, Any]) -> dict[str, Any]:
                 score=SCORE,
             )
 
+            # Insertion of short keys between Firefox limit of 2 and QUERY_CHAR_LIMIT - 1
+            # For similars, the values are added to the secondary index.
```
ncloudioj, line 151:
```suggestion
            # For similars equal or longer than QUERY_CHAR_LIMIT, the values are added to the secondary index.
```

ncloudioj (approved):
Looks great, just one minor note on code comment.

🚢

tests/integration/api/v1/suggest/test_top_picks.py
```diff
@@ -52,3 +54,36 @@ def test_top_picks_no_result(client: TestClient, query: str):
 
     result = response.json()
     assert result["suggestions"] == []
+
+
+@pytest.mark.parametrize(
+    "query,title,url",
```
Trinaa, line 60:
```suggestion
    ["query", "title", "url"],
```

tests/unit/providers/test_top_picks.py
```diff
@@ -154,3 +154,86 @@ async def test_query(srequest: SuggestionRequestFixture, top_picks: Provider) ->
             score=settings.providers.top_picks.score,
         )
     ]
+
+
+@pytest.mark.parametrize(
+    "query,title,url",
```
Trinaa, line 160:
```suggestion
    ["query", "title", "url"],
```

tests/unit/providers/test_top_picks.py
```diff
@@ -154,3 +154,86 @@ async def test_query(srequest: SuggestionRequestFixture, top_picks: Provider) ->
...
+) -> None:
+    """Test invalid inputs for Top Pick Provider when providing short domain queries."""
+    await top_picks.initialize()
+
+    res = await top_picks.query(srequest(query))
+    assert res == []
+
+
+@pytest.mark.parametrize(
+    "query,title,url",
```
Trinaa, line 208:
```suggestion
    ["query", "title", "url"],
```

Trinaa (approved):
I put 3 quick suggestions to consolidate multi-parameter style with the rest of the code base.

merino/providers/top_picks.py
taddes: 👍

tests/integration/api/v1/suggest/test_top_picks.py
taddes: Excellent, thanks @Trinaa !

Issue #85, opened 2022-10-20
Extract middleware unit tests


data-sync-user:
Refactor middleware tests into unit and integration test layers to clearly delineate test responsibilities

* Tests using the FAST API TestClient and responsible for verifying system interactions as a group, can be classified as integration tests
* Tests checking logical operations within middleware modules can be classified as unit tests

*Example* 
[https://github.com/mozilla-services/merino-py/pull/74|https://github.com/mozilla-services/merino-py/pull/74|smart-link]

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/CONSVC-2063)
┆epic: Add integration test layer to merino-py

data-sync-user:
➤ Katrina Anderson commented:

Update (2022/11/08)

  * Update acceptance criteria to include more specifics
  * Refactored tests for UserAgent & FeatureFlags modules
  * Update integration test module names
  * Update documentation
  * Create draft pull requesthttps://github.com/mozilla-services/merino-py/pull/106 ( https://github.com/mozilla-services/merino-py/pull/106|smart-link )
    * Asked for behaviour verification in comments

data-sync-user:
➤ Katrina Anderson commented:

Update (2022/11/09)

  * Refactor tests for LoggingMiddleware and update draft PR

data-sync-user:
➤ Katrina Anderson commented:

Update (2022/11/10)

  * Refactored LoggingMiddleware logic into a utility for better testability

data-sync-user:
➤ Katrina Anderson commented:

Update (2022/11/14)

  * Refactored MetricsMiddleware integration tests into endpoint modules
  * Added unit test module

PR #106, opened 2022-11-08
[CONSVC-2063] test: extract middleware unit tests

Trinaa:
# Description
- rename suggest integration test files to `test_suggest_top_picks.py` and `test_suggest_wiki_fruit.py`
- move `test_user_agent.py` to the `util` directory
- refactor `UserAgentMiddleware` tests to unit level or move to `test_suggest.py` module
- refactor `FeatureFlagMiddleware` tests to unit level and add unit coverage
- create `log_data_creators.py` utility module to allow for unit level testing
- refactor `LoggingMiddleware` tests to unit level or move to endpoint modules
- refactor `MetricsMiddleware` integration tests to endpoint modules and add unit coverage
- update unit test documentation to include middleware and log_data test fixtures

# Issue(s)
#85 | [CONSVC-2063](https://mozilla-hub.atlassian.net/browse/CONSVC-2063)


merino/middleware/user_agent.py
```diff
@@ -42,7 +42,7 @@ async def __call__(self, scope: Scope, receive: Receive, send: Send) -> None:
         """Parse user agent information through "User-Agent" and store the result
         to `scope`.
         """
-        if scope["type"] != "http":  # pragma: no cover
+        if scope["type"] != "http":
             await self.app(scope, receive, send)
             return
```
Trinaa, line 48: @ncloudioj I am wondering why the ua parsing is done in a utility instead of directly in the middleware? Was it for testability?

tests/unit/middleware/test_featureflags.py
```diff
@@ -1,86 +1,63 @@
...
+    scope: Scope,
+    receive_mock: Receive,
+    send_mock: Send,
+) -> None:
+    """Test that SID is assigned `None` is `sid` query parameter is not available."""
+    scope["query_string"] = "q=nope"
+
+    await featureflags_middleware(scope, receive_mock, send_mock)
+
+    assert session_id_context.get() is None
```
Trinaa, line 49: @ncloudioj Can you confirm that in the unlikely case that the sid is not in the query, that it's alright that the `session_id_context` is set to `None`?

merino/middleware/user_agent.py
ncloudioj: Yes, and also for modularity so that the middleware can be more cohesive. Just like we have a utility module for feature flags and a separate feature flags middleware.

tests/unit/middleware/test_featureflags.py
ncloudioj: Yes, it's fine to have `session_id_context` as `None`. As the `sid` was added since Firefox 103, so it's totally possible that Merino receives queries without sid from old Firefox versions.

tests/unit/middleware/test_user_agent.py
```diff
@@ -0,0 +1,123 @@
...
+"""Unit tests for the middleware user_agent module."""
+
+import pytest
+from pytest_mock import MockerFixture
+from starlette.types import ASGIApp, Receive, Scope, Send
+
+from merino.middleware import ScopeKey
+from merino.middleware.user_agent import UserAgent, UserAgentMiddleware
+
+Scenario = tuple[str, str, str, str]
```
quiiver, line 14: I'm confused by this type definition. it seem like `Scenario` should be `tuple[str, UserAgent]` what am i missing?

quiiver: oh, i just saw this was still a draft. sorry for the noise! feel free to ignore

Trinaa: Not at all! You are right and it is fixed in [e2cf7a2](https://github.com/mozilla-services/merino-py/pull/106/commits/e2cf7a22414eacdaef90826d6065bce10b7c3d13).

unit tests aren't in the mypy scope atm, otherwise I think I would have caught it sooner.

tests/unit/middleware/test_logging.py
```diff
@@ -0,0 +1,237 @@
...
+    receive_mock: Receive,
+    send_mock: Send,
+) -> None:
+    """Test that no logging action takes place for an unexpected Scope type."""
+    caplog.set_level(logging.INFO)
+    scope: Scope = {"type": "not-http"}
+
+    await logging_middleware(scope, receive_mock, send_mock)
+
+    assert len(caplog.messages) == 0
```
Trinaa, line 237: @hackebrot I'd like your opinion on this test module.

I've attempted to move some logging test cases from integration to unit here; however, the result makes me feel uneasy.

In particular that I extract the `send_wrapper` callable from the mocked instance of `LoggerMiddleware.app` seems less than ideal.

hackebrot: Notes

Hi @Trinaa! 👋🏻 

Sure, no problem! I share your concerns about the `extract_send_wrapper` fixture.

To me, this unit test design indicates that there's opportunity to improve testability of the current implementation of the `LoggingMiddleware`.

Let's take a step back.

I don't remember where I read this, but I've found the following guidance quite helpful with problems like this one: The difference between a library and a framework is that with libraries we control the flow of the program and our code calls library code, whereas with frameworks the framework controls the flow and we fill in the blanks so that the framework code calls our code.

What does this mean for testing?

We don't have control over how FastAPI or Starlette calls our middleware code. But we know that if our class implements the interface that the framework expects and we add the middleware correctly, the framework will call our code as documented.

My recommendation is to test the middleware itself in integration tests, but extract our code into functions that make it possible to unit test them. That way we don't have to rely on framework implementation details to effectively test our code.

I think we can achieve that by making the following design changes to the `LoggingMiddleware`:

1. (Optionally) Add a `typing.Protocol` that defines the inputs and outputs for a log data creator function (inputs/outputs that we can create in tests):

    ```python
    class LogDataCreator(Protocol):
        """Protocol for Callables that generate log data for the given inputs."""

        def __call__(
            self, message: Message, request: Request, dt: datetime
        ) -> dict[str, Any]:
            ...
    ```

1. Extract the body of the `if PATTERN.match(request.url.path):` branches into functions that implement the above protocol

    ```python
    def create_suggest_log_data(
        message: Message, request: Request, dt: datetime
    ) -> dict[str, Any]:
        """Create log data for the suggest API endpoint."""
        location = request.scope[ScopeKey.GEOLOCATION]
        ua = request.scope[ScopeKey.USER_AGENT]

        general_data = {
            "sensitive": True,
            "errno": 0,
            "time": dt.isoformat(),
        }

        request_data = {
            "path": request.url.path,
            "method": request.method,
            "query": request.query_params.get("q"),
            "code": message["status"],
            # Provided by the asgi-correlation-id middleware.
            "rid": Headers(scope=message)["X-Request-ID"],
            "session_id": request.query_params.get("sid"),
            "sequence_no": int(seq) if (seq := request.query_params.get("seq")) else None,
            "client_variants": request.query_params.get("client_variants", ""),
            "requested_providers": request.query_params.get("providers", ""),
        }

        location_data = {
            "country": location.country,
            "region": location.region,
            "city": location.city,
            "dma": location.dma,
        }

        ua_data = {
            "browser": ua.browser,
            "os_family": ua.os_family,
            "form_factor": ua.form_factor,
        }

        return {**general_data, **request_data, **location_data, **ua_data}
    ```

1. Update the middleware to select a logger and one of these functions based on the existing if-statement.
1. Pass in the created data to the according logger when emitting an `INFO` message with it.

In summary, with some refactoring we can add unit test coverage for code we own (extract info from the request and add more that we care about it) and complement that by integration test coverage to ensure our code is correctly integrated with the FastAPI framework (`LoggingMiddleware` is added to app, uses different loggers with different data based on request URL path).

Hope this helps! Please let me know if you have any questions. 🦊

Trinaa: **Warning: I rebased**

Thank you for the detailed explanation @hackebrot 

I took a stab in [e8d2001](https://github.com/mozilla-services/merino-py/pull/106/commits/e8d2001832de708ff1642b3a6662df66cb50cff2), keeping in mind the organization of `UserAgentMiddleware` and `user_agent_parsing.py`. I feel this structure is similar...

Adding on: I noticed some warnings associated with using pytest-freezgun. Do you know off-hand what that is about and how it can be solved?

hackebrot: You're welcome, @Trinaa!

I'm not familiar with that warning, but found this upstream issue and pull request with a fix:
* https://github.com/ktosiek/pytest-freezegun/issues/35
* https://github.com/ktosiek/pytest-freezegun/pull/38

Both have been open for some time. So, most likely a combination of bit rot and maintainers not having the bandwidth to actively work on the project right now.

From what I can tell, we only use [pytest-freezegun](https://pypi.org/project/pytest-freezegun/) here for `@pytest.mark.freeze_time("1998-03-31")`. The plugin is a wrapper for the [freezegun](https://pypi.org/project/freezegun/) library. The pytest plugin recognizes the marker, then injects a pytest fixture for marked tests. The fixture calls freezegun:

```python
freezer = freeze_time(*args, ignore=ignore, **kwargs)
frozen_time = freezer.start()
yield frozen_time
freezer.stop()
```

See [pytest-freezegun plugin implementation](https://github.com/ktosiek/pytest-freezegun/blob/3202256486ea405904ded68b8bd61a409b856a75/pytest_freezegun.py#L43-L47).

My suggestion is to install and use freezegun directly:

```python
# Freeze time for a pytest style test:

@freeze_time("2012-01-14")
def test():
    assert datetime.datetime.now() == datetime.datetime(2012, 1, 14)
```

See [freezegun README](https://github.com/spulec/freezegun/blob/master/README.rst).

If you'd rather continue to use a pytest marker or fixture we can do that too. It won't be a lot of code.

Hope that helps!

Trinaa: Change applied [a13fe46](https://github.com/mozilla-services/merino-py/pull/106/commits/a13fe463171bf64231573c741282ea55fe297fd3)

merino/util/log_data_creators.py
```diff
@@ -0,0 +1,71 @@
...
+from starlette.datastructures import Headers
+from starlette.requests import Request
+from starlette.types import Message
+
+from merino.middleware import ScopeKey
+from merino.middleware.geolocation import Location
+from merino.middleware.user_agent import UserAgent
+
+
+def create_log_data(request: Request, message: Message, dt: datetime) -> dict[str, Any]:
```
ncloudioj, line 14: By MozLog's spec, this log is called "request-summary", so we can also call it `create_request_summary_log_data(...)`.

tests/integration/api/conftest.py
```diff
@@ -29,3 +33,26 @@ def fixture_test_client_with_events() -> Iterator[TestClient]:
     """
     with TestClient(app) as client:
         yield client
+
+
+@pytest.fixture(name="log_data")
```
ncloudioj, line 38: I'd recommend renaming this fixture to `extract_log_data()` or `get_log_data()`

tests/integration/api/test_heartbeat.py
```diff
@@ -20,25 +25,49 @@ def test_heartbeats(client: TestClient, endpoint: str) -> None:
     assert response.status_code == 200
 
 
-def test_non_suggest_request_logs_contain_required_info(
-    mocker: MockerFixture,
+@freeze_time("1998-03-31")
```
ncloudioj, line 28: Have seen this magic date in various places, is there a way for us to simplify it by not having to freeze time for those tests?

ncloudioj:
@Trinaa I just skimmed over the PR, and it looks good. I will take another closer look at it tomorrow :)

tests/integration/api/test_heartbeat.py
Trinaa: For the verification of log data to be comprehensive, we need to include the time entry both for presence and format. This is hard to do with time constantly changing across systems. I have had better success patching with a library like freezegun, than custom patching, but am open to other suggestions?

Would it help if the time was different (less random/more recent)? Would a clarifying comment alongside the decorator help?

tests/integration/api/conftest.py
Trinaa: Change applied [2a15e6f](https://github.com/mozilla-services/merino-py/pull/106/commits/2a15e6fd492dd9ef93bc1fc43a9f2854853c0aeb).
I opted for `extract_request_summary_log_data` to follow with previous recommendation.

merino/util/log_data_creators.py
Trinaa: Change applied [2a15e6f](https://github.com/mozilla-services/merino-py/pull/106/commits/2a15e6fd492dd9ef93bc1fc43a9f2854853c0aeb).

tests/integration/api/test_heartbeat.py
ncloudioj: I don't have any better recommendations off hands. One thing we could do is drop the `time` field from `extract_*_log()` so that we don't need to check it in tests, but that's obviously flawed. I think we can keep it as is.

ncloudioj (approved):
Looks great, thanks!

@Trinaa Out of curiosity, why 1998-03-31?

tests/unit/util/test_log_data_creator.py
```diff
@@ -0,0 +1,155 @@
...
+from merino.middleware.geolocation import Location
+from merino.middleware.user_agent import UserAgent
+from merino.util.log_data_creators import (
+    create_request_summary_log_data,
+    create_suggest_log_data,
+)
+
+
+@pytest.mark.parametrize(
+    "headers,expected_agent,expected_lang",
```
hackebrot, line 24: It's a personal preference, but I find the following syntax more readable:

```suggestion
    ["headers", "expected_agent", "expected_lang"],
```

hackebrot (approved):
Hi @Trinaa! 👋🏻 

I've added a few suggestions for your consideration.

tests/unit/util/test_log_data_creator.py
Trinaa: Change applied: [9162333](https://github.com/mozilla-services/merino-py/pull/106/commits/91623339988fd4e41fb20b4beef14ba3f6edf2fc)

tests/integration/api/v1/suggest/test_suggest.py
```diff
@@ -78,62 +108,105 @@ def test_client_variants(client: TestClient) -> None:
...
+        "dma": 819,
+        "browser": "Firefox(103.0)",
+        "os_family": "macos",
+        "form_factor": "desktop",
+    }
+
     client.get(
-        f"{root_path}?q={query}&sid={sid}&seq={seq}"
-        f"&client_variants={client_variants}&providers={providers}"
+        f"{expected_log_data['path']}?q={expected_log_data['query']}"
```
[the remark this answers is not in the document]
Trinaa: Change applied: [9162333](https://github.com/mozilla-services/merino-py/pull/106/commits/91623339988fd4e41fb20b4beef14ba3f6edf2fc)
This is much better!

merino/middleware/logging.py
```diff
@@ -24,61 +26,24 @@ class LoggingMiddleware:
     """An ASGI middleware for logging."""
 
     def __init__(self, app: ASGIApp) -> None:
-        """Initilize."""
+        """Initialize."""
```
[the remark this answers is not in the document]
Trinaa: Change applied: [9162333](https://github.com/mozilla-services/merino-py/pull/106/commits/91623339988fd4e41fb20b4beef14ba3f6edf2fc)

Trinaa:
> Looks great, thanks!
> 
> @Trinaa Out of curiosity, why 1998-03-31?

image

PR #112, opened 2022-11-14
[CONSVC-2052] feat: add timeout handling for query tasks

ncloudioj:
This adds a query handling timeout for the `suggest` endpoint. The rationale is that since Firefox uses a timeout (currently 200ms) for each query request to Merino, Merino should do the same to ensure slow providers would not prevent Merino from serving suggestions from other providers. When timeout occurs, Merino will cancel all ongoing query tasks and log/record metrics accordingly. 

Note that:
- this works at the request-level rather than the provider-level, meaning that once the timeout gets triggered, all the unfinished query tasks will be cancelled and optionally processed by the timeout callback. It does what Merino needs for now. We can extend this to be provider-evel timeout if needed in the future.  
- `asyncio.gather()` also supports timeout, but it doesn't provide fine-grained control on timeout handling. Therefore, I ended up using the asyncio wait primitives. The benchmark shows that the overhead is negligible compared to `asyncio.gather()`

This fixes [CONSVC-2052](https://mozilla-hub.atlassian.net/browse/CONSVC-2052).


tests/unit/test_task_runner.py
```diff
@@ -0,0 +1,167 @@
...
+
+    assert len(done_tasks) == 2
+    for task in done_tasks:
+        match task.get_name():
+            case "normal-task":
+                assert await task
+            case "raised-task":
+                with pytest.raises(RuntimeError) as exc_info:
+                    await task
+                assert str(exc_info.value) == "error"
```
Trinaa, line 75: I worry long term using a loop and decision logic like this in test, because it doesn't detect changes, like if a task name changes. Something like the following seems more secure:

```suggestion
    normal_task = next(task for task in done_tasks if task.get_name() == "normal-task")
    assert await normal_task

    raised_task = next(task for task in done_tasks if task.get_name() == "raised-task")
    with pytest.raises(RuntimeError) as exc_info:
        await raised_task
    assert str(exc_info.value) == "error"
```

quiiver:
> meaning that once the timeout gets triggered, all the unfinished query tasks will be cancelled and optionally processed by the timeout callback. It does what Merino needs for now. We can extend this to be provider-evel timeout if needed in the future.

I haven't read the code yet but does this mean that completed providers will be returned in the response?

ncloudioj:
> I haven't read the code yet but does this mean that completed providers will be returned in the response?

Correct.

merino/web/api_v1.py
```diff
@@ -73,16 +79,30 @@ async def suggest(
...
 
-    lookups = [
-        metrics_client.timeit_task(
-            p.query(srequest),
-            f"providers.{p.name}.query",
+    lookups: list[Task] = []
+    for p in search_from:
+        task = metrics_client.timeit_task(
+            p.query(srequest), f"providers.{p.name}.query"
         )
```
quiiver, line 86: So technically we are cancelling this task when we cancel the pending tasks upon timeout. does task cancellation propagate?

ncloudioj: When the wait timeout occurs, we're cancelling all the pending tasks (one per each provider) one-by-one [here](https://github.com/mozilla-services/merino-py/pull/112/files#diff-480bef6343fe53a72b16091b8899dc189e037e4d427e44c3b10e427b39571902R50-R52). So there will no cancellation propagation involved here.

quiiver: Ok makes sense, I have another couple questions. 

1. the aiodogstatsd library uses a `done_callback` to record the timing of the underyling coroutine. does the done callback get called when a task is cancelled?
2. according to the `Task` docs calling `Task.cancel` should throw a `CancelledError` exception but i'm not seeing any exception catching. Whats going on there?

tests/integration/api/v1/suggest/test_timeout_handling.py
```diff
@@ -0,0 +1,62 @@
...
+        "get.api.v1.suggest.timing",
+        "get.api.v1.suggest.status_codes.200",
+        "providers.timedout-sponsored.query.timeout",
+        "providers.sponsored.query",
+        "providers.timedout-sponsored.query",
+        "response.status_codes.200",
+    ]:
+        report.assert_any_call(
+            metric_key, mocker.ANY, mocker.ANY, mocker.ANY, mocker.ANY
+        )
```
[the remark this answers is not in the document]
quiiver: 🤯 that's great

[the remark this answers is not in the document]
Trinaa: Credit to @hackebrot , I saw him do something like this when testing feature flags.

merino/web/api_v1.py
ncloudioj: > the aiodogstatsd library uses a done_callback to record the timing of the underyling coroutine. does the done callback get called when a task is cancelled?

Yes, it will be called as usual, meaning that it will time a cancelled task in this case. I thought about clearing the done_callback before it gets cancelled, but then figured we might want to keep it for completeness. I am open to other options, too.

>  according to the Task docs calling Task.cancel should throw a CancelledError exception but i'm not seeing any exception catching. Whats going on there?

We only call cancel on those pending tasks, which should be fine and no `CancelledError` will be raised. Are you concerning about calling cancel but somehow the task is completed at that point? In that case, it might raise I think we can add a check for that.

tests/unit/test_task_runner.py
ncloudioj: Good catch. Would this do it as well?

```suggestion
    for task in done_tasks:
        match task.get_name():
            case "normal-task":
                assert await task
            case "raised-task":
                with pytest.raises(RuntimeError) as exc_info:
                    await task
                assert str(exc_info.value) == "error"
            case _:
                raise AssertionError(f"Unexpected task: {task.get_name()}")
```

merino/web/api_v1.py
quiiver: > Are you concerning about calling cancel but somehow the task is completed at that point?
Not particularly, the docs just make it seem like `CancelledError`s are automatically thrown when you call `cancel` on a `Task` instance regardless of its completion status.

https://docs.python.org/3.11/library/asyncio-task.html#asyncio.Task.cancel

> This arranges for a [CancelledError](https://docs.python.org/3.11/library/asyncio-exceptions.html#asyncio.CancelledError) exception to be thrown into the wrapped coroutine on the next cycle of the event loop.

ncloudioj: Ah, I see what you mean. I think it will raise a `CancelledError` if you await it or call `task.result()` once it gets cancelled, but neither is the case here as it just cancels them and returns them as a separate task list along with the done tasks. If the consumer decides to use them regardless, the yeah `CancelledError` will be thrown.

Does that make sense?

quiiver: Yes thanks 😅

tests/integration/api/v1/suggest/test_timeout_handling.py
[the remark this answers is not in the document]
hackebrot: 😁

tests/unit/test_task_runner.py
Trinaa: 🤔 Yes. I think that works.

Trinaa:
@ncloudioj I've merged https://github.com/mozilla-services/merino-py/pull/106, I apologize for the conflicts this creates. I'll note two things:

1. Please move the `tests/unit/test_task_runner.py` module to `tests/unit/util/test_task_runner.py`. Perhaps propagate the `util` to `utils` directory name change to the unit test directory.

2. (**Optional**) The `test_with_timedout_provider` could be included in the `test_suggest.py`, since it's a generalized suggest endpoint behaviour.

ncloudioj:
> Please move the tests/unit/test_task_runner.py module to tests/unit/util/test_task_runner.py. Perhaps propagate the util to utils directory name change to the unit test directory.

Yep, sounds good, will do.

> The test_with_timedout_provider could be included in the test_suggest.py, since it's a generalized suggest endpoint behaviour.

If you don't strongly disagree, I'd like to keep it in a separate test file as I am almost certain that we will add more related functionalities and test cases in the future, which might mess up those generic tests in `test_suggest.py`.

Trinaa (approved):
Things LGTM on the test side.

pjenvey (approved):

PR #114, opened 2022-11-17
[CONSVC-2105] added v1 domain file and set it as the default for the provider 

taddes:
Added the domain file without icons and similar/variants currently. Set it as default for the provider.


merino/configs/default.toml
```diff
@@ -86,4 +86,4 @@ score_wikipedia = 0.2
 enabled_by_default = false
 score = 0.25
 query_char_limit = 4
-top_picks_file_path = "dev/top_picks.json"
```
ncloudioj, line 89: I think we can just replace the old file with the new one for each update to keep the repo slim as we don't need to track the history here for now.

taddes: You got it. Removed old file and renamed the file we just got with the actual 1k domains as `top_picks` so it will be the same.

taddes:
Good call @ncloudioj , removed those earlier as a suggestion from Phil as it would return `None`, but in these cases where there are no similars, we indeed want that guard. We previously didn't have that situation where `similars` were empty and we definitely need it, as you originally pointed out. Changes made, thank you as always. Tested with new domain file and looks good.

ncloudioj (approved):
Looks good. Could you please give it a spin locally to see it goes? Thanks!

taddes:
Looking good so far, posted a few screen captures of api responses in Jira tkt

PR #115, opened 2022-11-18
[CONSVC-1510] feat: dynamic Wikipedia provider

ncloudioj:
This fixes [CONSVC-1510](https://mozilla-hub.atlassian.net/browse/CONSVC-1510)


merino/providers/wikipedia.py
```diff
@@ -0,0 +1,153 @@
...
+                self.build_suggestion(doc)
+                for doc in res["suggest"][SUGGEST_ID][0]["options"]
+            ]
+        else:
+            return []
+
+    @staticmethod
+    def build_suggestion(doc: dict[str, Any]) -> dict[str, Any]:
+        """Build a suggestion dict based on the ES result."""
+        title = str(doc["text"])
```
quiiver, line 101: you need the `title` source field here as that will be the canonical page title. we're currently not indexing what wikipedia calls `redirects` those are misspellings, alternate spellings or synonyms that redirect to the canonical page but those would be the value of `doc['text']` here when we do start including them

merino/providers/wikipedia.py
```diff
@@ -0,0 +1,153 @@
...
+                "completion": {
+                    "field": "suggest",
+                    "size": MAX_SUGGESTIONS,
+                },
+            }
+        }
+
+        try:
+            res = await self.client.search(
+                index=INDEX_ID, suggest=suggest, timeout=TIMEOUT_MS, source=False
```
quiiver, line 84: you can do `source_includes=['title']` here to just have the `title` field in the source of the response object

merino/providers/wikipedia.py
ncloudioj: Sounds good. 👍

merino/providers/wikipedia.py
```diff
@@ -0,0 +1,169 @@
...
+            logger.warning(f"Failed to search from ES: {e}")
+            return []
+
+        if "suggest" in res:
+            return [
+                self.build_suggestion(doc)
+                for doc in res["suggest"][SUGGEST_ID][0]["options"]
+            ]
+        else:
+            return []
```
Trinaa, line 112: Would moving more decision logic/responsibility out of the ElasticBackend and into the Wikipedia provider be possible? Like catching exceptions, logging and building the suggestion.

My thinking comes from looking at the tests. I wasn't expecting unit testing against the ElasticBackend and having to patch the AsyncElasticsearch object. If we move some of these responsibilities to the Provider we can test these behaviours with the more sustainable injected backend.

Trinaa (commented):
@ncloudioj 
Will we add the integration tests in [CONSVC-2055](https://mozilla-hub.atlassian.net/browse/CONSVC-2055)?
Perhaps we can discuss the need and feasibility of contract tests at the next engineering meeting.

ncloudioj:
> Will we add the integration tests in [CONSVC-2055](https://mozilla-hub.atlassian.net/browse/CONSVC-2055)?
Perhaps we can discuss the need and feasibility of contract tests at the next engineering meeting.

I did try adding the integration tests but wound up dropping those as they looked more like unit tests without using the actual Elasticsearch backend. I'm particularly interested in adding contract tests for this by using an ES docker container. Yeah, we can discuss this in detail in the next meeting.

tests/unit/providers/test_init_providers.py
```diff
@@ -14,7 +14,7 @@ async def test_init_providers() -> None:
 
     providers, default_providers = get_providers()
 
-    assert len(providers) == 4
+    assert len(providers) == 5
     assert ProviderType.ADM in providers
     assert ProviderType.WIKI_FRUIT in providers
     assert ProviderType.TOP_PICKS in providers
```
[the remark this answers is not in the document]
ncloudioj: Interestingly, I've updated this test in https://github.com/mozilla-services/merino-py/pull/115/commits/953b9243839c6c21902ffb4d3001ac7776e312c6. Not exactly like yours, but let me know if that looks reasonable to you.

merino/providers/wikipedia.py
ncloudioj: I think the motivation of having this `ElaticBackend` is twofold:
* To support DI for better testability
* To abstract away those ES specifics, such as: making requests to the backend, parsing responses for relevant data, exception handling, etc, so that the provider doesn't need to handle those details and the end implementation can be more cohesive and compact.
  * It'll be straightforward to change the backend in the future if needed. For instance, we can add another backend (e.g. `LiveAPIWikipediaBackend`) using the same `WikipediaBackend` protocol.
  * Admittedly, this would require more tests for those backends e.g. `ElasticBackend` including patching the underlying ES client.
  * BTW, `build_suggestion()` is an unfortunate misnomer, `build_wikipidia_article()` is perhaps more accurate for what it does.
 
If we were to move those responsibilities from the backend to the provider, I can't see anything else this backend could offer, and all ES-related details will likely be leaked to the provider.

Let me know what you think about the above points.

tests/unit/providers/test_init_providers.py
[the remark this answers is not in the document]
Trinaa: I like the advantage the for loop brings that the test won't break with provider additions or subtractions.

It's less comprehensive in that it doesn't check the key-value mapping. Is it a concern that 'adm' could be assigned a provider other than `adm.Provider` for instance? 

Another thing to consider is, with the for loop we assert against each provider instead of once against the collection. This means a less comprehensive error message. If mapping isn't a concern, what about:
```
    expected_providers = {p.value for p in ProviderType}
    assert providers.keys() == expected_providers
```

Trinaa:
> I did try adding the integration tests but wound up dropping those as they looked more like unit tests without using the actual Elasticsearch backend. I'm particularly interested in adding contract tests for this by using an ES docker container. Yeah, we can discuss this in detail in the next meeting.

I think there is value in adding at least a happy path integration test, where you make the call from API level and check the response. In doing this you will verify that the internal interfaces communicate properly, that the provider works as expected with the middleware layer for instance.

tests/unit/providers/test_init_providers.py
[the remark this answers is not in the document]
ncloudioj: > It's less comprehensive in that it doesn't check the key-value mapping. Is it a concern that 'adm' could be assigned a provider other than adm.Provider for instance?

That should be very unlikely since Merino doesn't programmatically set providers, rather providers are hardcoded.

> Another thing to consider is, with the for loop we assert against each provider instead of once against the collection. This means a less comprehensive error message. If mapping isn't a concern, what about:

Yep, we can do this for sure.

merino/providers/wikipedia.py
Trinaa: I understand your backend vision better now thank you for taking the time to detail it. 

My vision differed in that I looked at the backend object as a very thin wrapper around IO calls with little to no complexity. The value comes from the isolation of our logic into testable modules and the clarity a minimalist wrapper provides into how we depend on external services.

Consistency on backend object responsibility and vision is perhaps something to look out for. When I look at the remotesettings.py::LiveBackend for instance, it has a reduced set of responsibilities (no logging or exception handling).

Is injecting the AsyncElasticsearch client into the ElasticBackend too much? 😛 I'm also okay to see how the patch does over time.

ncloudioj: I see your points as well. Yeah, the backend context here is broader than that of RS. Let's see how this works over time.

> Is injecting the AsyncElasticsearch client into the ElasticBackend too much?

Lol, I've seriously thought about injecting AsyncElasticsearch into ElasticBackend. But that extra complexity made me opt for other options.

Trinaa (approved):

Trinaa:
Thank you for this work @ncloudioj! It will be cool to see this provider in action! 🚀

merino/providers/__init__.py
```diff
@@ -66,6 +70,16 @@ async def init_providers() -> None:
                 providers["wiki_fruit"] = WikiFruitProvider(
                     name=provider_type, enabled_by_default=setting.enabled_by_default
                 )
+            case ProviderType.WIKIPEDIA:
+                providers["wikipedia"] = wikipediaProvider(
+                    backend=(
+                        ElasticBackend(setting.es_hosts)  # type: ignore [arg-type]
+                        if setting.backend == "elasticsearch"
+                        else WikipediaTestBackend()
```
mtbottle, line 78: I'm curious to why you chose to use the `TestBackend` (which always return no results) as the default _test_ behaviour. I saw that you also implemented `TestEchoBackend`, which will return a correctly formatted response (perhaps invalid, but it gives some sense of what the response will look like, without necessarily digging in the code). 

Also, what's your thoughts on bootstrapping a local Elasticsearch with some data already preloaded for testing?

ncloudioj: > I'm curious to why you chose to use the TestBackend (which always return no results) as the default test behaviour. I saw that you also implemented TestEchoBackend, which will return a correctly formatted response (perhaps invalid, but it gives some sense of what the response will look like, without necessarily digging in the code).

Because this provider is now disabled by default, though we might enable it in the future, if it returns anything (e.g. `TestEchoBackend`) for any query request, that could break other tests unexpectedly. That's why it, along with other test backends, returns nothing in tests. Likewise, `TestEchoBackend` was used by test cases for this particular provider where it expects the backend to return non-empty results.

> Also, what's your thoughts on bootstrapping a local Elasticsearch in Docker with some data already preloaded for testing?

Yeah, that's exactly what we plan to do in [CONSVC-2055](https://mozilla-hub.atlassian.net/browse/CONSVC-2055). My current thought is that we will spin off an ES container and then index some test pages in Merino's contract tests. We can discuss the final plan in our next meeting.

merino/providers/__init__.py
```diff
@@ -77,6 +83,20 @@ async def init_providers() -> None:
                 providers["wiki_fruit"] = WikiFruitProvider(
                     name=provider_type, enabled_by_default=setting.enabled_by_default
                 )
+            case ProviderType.WIKIPEDIA:
+                providers["wikipedia"] = wikipediaProvider(
+                    backend=(
+                        ElasticBackend(
+                            cloud_id=setting.es_cloud_id,
```
ncloudioj, line 90: N.B: this requires an Elastic cloud ID to init the backend (an empty string will make it complain so we provide a fake ID for development), we might want to make it more flexible to be able to use other ES instances in future.

merino/providers/wikipedia/backends/elastic.py
```diff
@@ -0,0 +1,72 @@
...
+    @staticmethod
+    def build_article(doc: dict[str, Any]) -> dict[str, Any]:
+        """Build a wikipedia article based on the ES result."""
+        title = str(doc["_source"]["title"])
+        quoted_title = quote(title.replace(" ", "_"))
+        return {
+            "full_keyword": title,
+            "title": title,
+            "url": f"https://en.wikipedia.org/wiki/{quoted_title}",
+        }
```
mtbottle, line 72: Can probably a subsequent task, but we should probably make an Article model.

merino/providers/wikipedia/backends/elastic.py
```diff
@@ -0,0 +1,72 @@
...
+                source_includes=["title"],
+            )
+        except Exception as e:
+            logger.warning(f"Failed to search from ES: {e}")
+            return []
+
+        if "suggest" in res:
+            return [
+                self.build_article(doc)
+                for doc in res["suggest"][SUGGEST_ID][0]["options"]
```
mtbottle, line 58: Hmm... the nested `res` data access scares me. We only check that "suggest" is in the response, but not the other options. There is potential for any of the subsequent access to fail too... so maybe we should wrap this with a `try` instead?

like:

```
try:
    articles = res["suggest"][SUGGEST_ID][0]["options"]
except KeyError:
    return []

return [ self.build_article(doc) for doc in articles ]
```

or something?

Trinaa (approved):

tests/unit/providers/test_wikipedia.py
```diff
@@ -0,0 +1,173 @@
...
+from tests.types import FilterCaplogFixture
+from tests.unit.types import SuggestionRequestFixture
+
+# Fake Elastic Cloud ID and password for testing.
+TEST_CLOUD_ID: str = (
+    "testing:dXMtZWFzdC0xLmF3cy5mb3VuZC5pbyRjZWM2ZjI2MWE3NGJmMjRjZTMzYmI4ODExYj"
+    "g0Mjk0ZiRjNmMyY2E2ZDA0MjI0OWFmMGNjN2Q3YTllOTYyNTc0Mw=="
+)
+TEST_USER: str = "elastic"
+TEST_PASSWORD: str = "PASSWORD"
```
mtbottle, line 27: Can we put these in the `configs/testing.toml` instead of hard coding them here?

merino/providers/wikipedia/backends/elastic.py
```diff
@@ -0,0 +1,72 @@
...
+        try:
+            res = await self.client.search(
+                index=INDEX_ID,
+                suggest=suggest,
+                timeout=TIMEOUT_MS,
+                source_includes=["title"],
+            )
+        except Exception as e:
+            logger.warning(f"Failed to search from ES: {e}")
+            return []
```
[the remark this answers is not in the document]
taddes: @Trinaa is there a way to also do that while explicitly calling the logger warning? May be in some cases it's hard to chase down the exception if it bubbles up. Maybe I'm missing something, though!

merino/providers/__init__.py
```diff
@@ -15,6 +15,11 @@
 from merino.providers.weather.backends.accuweather import AccuweatherBackend
 from merino.providers.weather.provider import Provider as WeatherProvider
 from merino.providers.wiki_fruit import WikiFruitProvider
+from merino.providers.wikipedia.backends.elastic import ElasticBackend
+from merino.providers.wikipedia.backends.test_backends import (
+    TestBackend as WikipediaTestBackend,
+)
+from merino.providers.wikipedia.provider import Provider as wikipediaProvider
```
taddes, line 22: Capitalize here? 
```suggestion
from merino.providers.wikipedia.provider import Provider as WikipediaProvider
```

tests/unit/providers/test_wikipedia.py
Trinaa: I'm not sure about this idea, because different `*.toml` files are used depending on environment.

The values could be put in the `fixture_es_backend` though, or have their own fixture.

merino/providers/wikipedia/backends/protocol.py
```diff
@@ -0,0 +1,20 @@
...
+    might define additional methods and attributes which this provider doesn't
+    directly depend on.
+    """
+
+    async def shutdown(self) -> None:  # pragma: no cover
+        """Shut down connection to the backend"""
+        ...
+
+    async def search(self, q: str) -> list[dict[str, Any]]:  # pragma: no cover
+        """Search suggestions for a given query from the backend."""
```
mtbottle, line 19: I had kind of a hard time understanding the docstring. How about: 

```        
Search Wikipedia and return articles relevant to query.
```

?

merino/providers/wikipedia/backends/elastic.py
```diff
@@ -0,0 +1,72 @@
...
+            return [
+                self.build_article(doc)
+                for doc in res["suggest"][SUGGEST_ID][0]["options"]
+            ]
+        else:
+            return []
+
+    @staticmethod
+    def build_article(doc: dict[str, Any]) -> dict[str, Any]:
+        """Build a wikipedia article based on the ES result."""
```
mtbottle, line 65: nit: `s/wikipedia/Wikipedia/`

merino/providers/wikipedia/backends/elastic.py
ncloudioj: Yeah, Elastic really likes using deeply nested objects in the response 🤷‍♂️

Per its API doc, if `suggest` is present in the response, meaning it has found some hits for this query, that's why I only check that in the if statement. As for other lookups, I opted to not be super defensive because that could silently ignore those API change failures in the future. When that happens at some point, this will fail fast and loudly so that our tests or even Sentry will capture it right away.

Does that make sense? If so, I can add some comments here to clarify.

merino/providers/__init__.py
```diff
@@ -15,6 +15,11 @@
 from merino.providers.weather.backends.accuweather import AccuweatherBackend
 from merino.providers.weather.provider import Provider as WeatherProvider
 from merino.providers.wiki_fruit import WikiFruitProvider
+from merino.providers.wikipedia.backends.elastic import ElasticBackend
+from merino.providers.wikipedia.backends.test_backends import (
+    TestBackend as WikipediaTestBackend,
+)
+from merino.providers.wikipedia.wikipedia import Provider as wikipediaProvider
```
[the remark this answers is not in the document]
ncloudioj: Oops.

merino/providers/wikipedia/backends/elastic.py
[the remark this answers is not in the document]
ncloudioj: Yeah, I remember that, here is the [convo](https://github.com/mozilla-services/merino-py/pull/115#discussion_r1030884289). :)

But now I think it makes more sense to propagate a generic backend exception (e.g. `ElasticSearchError`) to the provider. Will fix it up.

tests/unit/providers/test_wikipedia.py
ncloudioj: At least a fake `TEST_CLOUD_ID` is needed for all environments because when Merino instantiates `AsyncElasticsearch(cloud_id=cloud_id,...)`, it will validate the cloud_id right away and throws if it's invalid. So if we move this over to `testing.toml`, we have to do the same for other environments. Other than lazy instantiation for Elastic client, any better way to deal with this?

merino/providers/wikipedia/backends/elastic.py
mtbottle: > I opted to not be super defensive because that could silently ignore those API change failures in the future. 

Good point! Let's leave it less defensive.

mtbottle (approved):
🚀

PR #120, opened 2022-11-21
chore: pin node to LTS to avoid bustages

ncloudioj:
Sigh~, this is the actual fix to the `docs-publish-to-prod` bustage. As verified in #119 .


hackebrot (approved):
LGTM! 🤖

PR #123, opened 2022-11-23
[CONSVC-1603] feat: add example feature flag decision

hackebrot:
# Description

This is adding a feature flag decision to the API, so we can learn about using StatsD tags as query filters in our Grafana dashboard, while we don't use any actual feature flags in the suggest API.

Note: I enabled `dynaconf_merge` for feature flags the dynaconf `testing` environment, so that we can use the feature flag both in a deployed environment and in tests.

# Issue(s)
[CONSVC-1603](https://mozilla-hub.atlassian.net/browse/CONSVC-1603)

### DIFF
--- a/merino/configs/flags/default.toml
+++ b/merino/configs/flags/default.toml
@@ -1,16 +1,23 @@
 # Example feature flag configuration block. Replace `example` with the name of your feature.
 # i.e. for a feature called `geo_location` the toml header will be `[default.flags.geo_location]`
 # and it will be checked in code with the name `geo_location` (e.g. `flags.is_enabled("geo_location")`)
-# 
+#
 # [default.flags.example]
 # `scheme` - Bucketing scheme for the flag. Allowed values are 'random' and 'session',
 #   defaults to `session`. Random generates a random bucketing id for every flag check.
 #   Session bucketing uses the session id of the request as the bucketing key so that feature
 #   checks within a given search session would be consistent.
 # scheme="session"
-# 
+#
 # `enabled` - Required. This represents the % enabled for the
 #   flag and must be a float between 0 and 1. A value of 0 indicates that
-#   the flag is 'off' (meaning that a check for the flag will return False 100% 
+#   the flag is 'off' (meaning that a check for the flag will return False 100%
 #   of the time) and a value of 1 indicates that the flag is 'on' for all.
 # enabled=0.5
+
+# We define the following feature flag in the default config, so we can learn
+# about using StatsD tags as query filters in our Grafana dashboard, while we
+# don't use any actual feature flags in the suggest API.
+[default.flags.hello_world]
+scheme = 'random'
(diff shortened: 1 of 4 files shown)


ncloudioj:
@hackebrot How long do you believe we will keep this feature flag around? If this is meant for long-term usage, I'd recommend using a dedicated endpoint for it. Because while the overall overhead is low, it just feels a little wasteful to me to do this over here.

hackebrot:
Hi @ncloudioj! 👋🏻 

Totally agree! My intention is to deploy this to stage and prod, then create a query in Grafana that includes a filter for the StatsD tag, and then once we have confirmed that this is working as expected, deploy a change deleting that feature flag again. I don't expect this to take longer than 2 days.

ncloudioj:
Gotcha! Thanks for the context, Raphael!

I think it's fine to have it for this short-term learning exercise.

merino/configs/flags/default.toml
```diff
@@ -1,16 +1,23 @@
...
 #   flag and must be a float between 0 and 1. A value of 0 indicates that
-#   the flag is 'off' (meaning that a check for the flag will return False 100% 
+#   the flag is 'off' (meaning that a check for the flag will return False 100%
 #   of the time) and a value of 1 indicates that the flag is 'on' for all.
 # enabled=0.5
+
+# We define the following feature flag in the default config, so we can learn
+# about using StatsD tags as query filters in our Grafana dashboard, while we
+# don't use any actual feature flags in the suggest API.
+[default.flags.hello_world]
```
ncloudioj, line 21: Nit: perhaps use a more meaningful name here, e.g "test_flight_01"? 😉

merino/configs/flags/testing.toml
```diff
@@ -1,3 +1,5 @@
+[testing]
+dynaconf_merge = true
```
ncloudioj, line 2: If it makes sense, you can also permanently add this feature flag here so that you don't need to modify `test_suggest.py` below once the feature flag gets taken out from default.toml.

ncloudioj (approved):
Looks good. r+ wc.

merino/configs/flags/default.toml
hackebrot: Changed applied in 492053eab9ec590f288680979b17e174c13d8d48

merino/configs/flags/testing.toml
hackebrot: Change applied in d98c8dfa546b01eacfa17d61aeaa7b1e3a40afc5

Trinaa (approved):

Trinaa:
[nit] Is there a JIRA item associated with this change?

hackebrot:
> [nit] Is there a JIRA item associated with this change?

Yes, CONSVC-1564. I've updated the description.

PR #127, opened 2022-11-29
[CONSVC-2122] fix: handle empty subdivisions in geolocation parsing

ncloudioj:
This fixes [CONSVC-2122](https://mozilla-hub.atlassian.net/browse/CONSVC-2122)

### DIFF
--- a/merino/middleware/geolocation.py
+++ b/merino/middleware/geolocation.py
@@ -63,7 +63,7 @@ class GeolocationMiddleware:
         scope[ScopeKey.GEOLOCATION] = (
             Location(
                 country=record.country.iso_code,
-                region=record.subdivisions[0].iso_code,
+                region=record.subdivisions[0].iso_code if record.subdivisions else None,
                 city=record.city.names["en"],
                 dma=record.location.metro_code,
                 postal_code=record.postal.code if record.postal else None,


merino/middleware/geolocation.py
```diff
@@ -63,7 +63,7 @@ async def __call__(self, scope: Scope, receive: Receive, send: Send) -> None:
         scope[ScopeKey.GEOLOCATION] = (
             Location(
                 country=record.country.iso_code,
-                region=record.subdivisions[0].iso_code,
+                region=record.subdivisions[0].iso_code if record.subdivisions else None,
```
ncloudioj, line 66: FWIW, `record.subdivisions` is of `tuple` type.

Trinaa, line 66: Should we expand this mitigation to country, city and dma record values?

I think the most sustainable way to test this, would be to find IP addresses in maxmind DB that doesn't resolve a subdivision. Patching has risk with maxmind DB updates.

Trinaa (approved):
Do you create tasks for fixes like this in JIRA?

ncloudioj:
> Do you create tasks for fixes like this in JIRA?

Yes, I did. See the PR comment.

merino/middleware/geolocation.py
ncloudioj: > Should we expand this mitigation to country, city and dma record values?

I've double-checked the other fields, they are all present in `record` and having `None` as fallback value if the info is unavailable. Only the `record.subdivisions` is of `tuple` type so the check is necessary.

Re: the test case for this, agreed, I real IP that doesn't resolve subdivisions would be ideal. I will poke around to see if I can find some for this.

PR #129, opened 2022-11-30
[CONSVC-2042] Add suggestions-per metric

mtbottle:
This change adds some metrics to get the number of suggestions per request (and per provider, in each request). 

More details on the feature can be found in [CONSVC-2042](https://mozilla-hub.atlassian.net/browse/CONSVC-2042).

Notes:
- I refactored a bunch of the test providers to output a Suggestion object where the `provider` field is the same as the `provider.name`. I noticed that the "real" providers all returned the Suggestion objects with the provider's actual name, but the fake objects were using a name that was different from what is attached to the Provider class. I'm not sure we want this decoupling, so I refactored all of them to return the `provider.name`.
- `test_suggest_feature_flags_tags_in_metrics` was too stringent as it ensures that the tags set for each metric is _exactly_ of a particular size (which obviously doesn't work for me because I'm adding to some of the tag sets). I stubbed it out so that we just need the tag to exist in tag set, rather than have an exact check. I hope I captured the intent of the test properly with the refactoring.


tests/integration/api/v1/suggest/test_suggest.py
```diff
@@ -321,4 +325,7 @@ def test_suggest_feature_flags_tags_in_metrics(
     tags_per_metric = {
         call.args[0]: [*call.args[3].keys()] for call in report.call_args_list
     }
-    assert tags_per_metric == expected_tags_per_metric
+    for metric in expected_metric_keys:
+        tags = tags_per_metric.get(metric)
+        assert tags is not None
+        assert should_contain_ff_tag == (feature_flag_tag in tags)
```
Trinaa, line 331: I think you did capture the intent of this test, but this implementation will result in the assertions nested in the for loop being executed multiple times. In the event of failure, we will have less information regarding the cause.

Instead what if we filtered `tags_per_metric` to only include feature flags. Then we can still have a collection assertion and minimize the change to test parameters. Something like:
```
    tags_per_metric = {
        call.args[0]: [
            metric
            for metric in call.args[3].keys()
            if metric.startswith("feature_flag.")
        ]
        for call in report.call_args_list
    }
    assert tags_per_metric == expected_tags_per_metric
```
The naming convention is mentioned in the [feature_flag.md](https://github.com/mozilla-services/merino-py/blob/main/docs/dev/feature_flags.md#metrics).

mtbottle: > but this implementation will result in the assertions nested in the for loop being executed multiple times.

Good call! I like your approach better, so I'll update this!

merino/web/api_v1.py
```diff
@@ -114,6 +118,25 @@ async def suggest(
     return JSONResponse(content=jsonable_encoder(response))
 
 
+def emit_suggestions_per_metrics(
+    metrics_client: Client, suggestions: list, active_providers: list
```
pjenvey, line 122: Nit: we could avoid the `list()` call above w/ `Iterable`:

```suggestion
    metrics_client: Client, suggestions: list, active_providers: Iterable[str]
```

`suggestions` could also type to `list[BaseSuggestion]` but I'm not sure how strict we care to be with such arguments

merino/web/api_v1.py
```diff
@@ -114,6 +118,25 @@ async def suggest(
...
+) -> None:
+    """Emit metrics for suggestions per request and suggestions per request by provider."""
+    metrics_client.histogram("suggestions-per.request", value=len(suggestions))
+
+    provider_suggestion_count = {p: 0 for p in active_providers}
```
pjenvey, line 127: Are `0` values useful for this histogram (I'm not sure)?

merino/web/api_v1.py
```diff
@@ -106,6 +106,10 @@ async def suggest(
         )
     )
 
+    emit_suggestions_per_metrics(
+        metrics_client, suggestions, list(active_providers.keys())
```
pjenvey, line 110:
```suggestion
        metrics_client, suggestions, active_providers.keys()
```

merino/web/api_v1.py
mtbottle: If the main point of this metric is to get an idea of what the typical number of Suggestions that a provider returns, then I think logging 0 will be useful. Excluding 0 will likely make the summary statistics for the histogram, percentiles and mean, skew higher than the typical observation, which will then make it look like the provider is returning more suggestions than it really is.

However, there are some limitations to my current implementation. For example, I don't differentiate when a provider might have not returned something due to a timeout vs. when a provider legitimately doesn't have any suggestions. If we timed out, then it might make sense to actually exclude this provider from emitting 0 (as it technically didn't give us back a response, rather than a response with 0 suggestions).

merino/web/api_v1.py
mtbottle: This is much nicer, thanks! 

> suggestions could also type to list[BaseSuggestion] but I'm not sure how strict we care to be with such arguments

I think being explicit with `BaseSuggestion` is a good idea because the `emit_suggestions_per_metrics` function introspects into the Suggestion object. So I'll also add this change!

merino/web/api_v1.py
```diff
@@ -106,6 +107,8 @@ async def suggest(
         )
     )
 
+    emit_suggestions_per_metrics(metrics_client, suggestions, active_providers.keys())
```
ncloudioj, line 110: I think we should use `search_from` here since each request could specify its interested provider(s) instead of querying all the `active_providers`.

merino/web/api_v1.py
ncloudioj: I think it makes sense to record 0 values for the sake of completeness.

We could also use `Collections.Counter`.

```
counter = Counter([suggest.provider for suggestion in suggestions])
for provider in active_providers:
    suggestion_count = counter[provider]  # count of a missing element is zero
    ...
```

mtbottle: > We could also use Collections.Counter.

This makes the counting so elegant! TIL

tests/integration/api/v1/suggest/test_timeout_handling.py
```diff
@@ -63,6 +63,11 @@ def test_with_timedout_provider(
     expected_metric_keys = [
         "providers.sponsored.query",
         "providers.timedout-sponsored.query.timeout",
+        "suggestions-per.request",
+        # suggestions-per.provider gets called twice
```
taddes, line 67: Good call here @mtbottle ! As discussed in our 1 on 1, I'm just adding the suggestion to add a comment here that one should check the configuration toml files. It is possible that in the future more than 2 providers are set to active as more providers roll out as `enabled_by_default`.  The configs for tests reside in `merino/configs/testing.toml`, for contract tests it's `merino/configs.ci.toml`  so we can just point to that first one. I think documenting this provides a backup for the fact that this isn't currently scalable.  That said, no need to refactor or change anything significantly here.

tests/integration/api/v1/suggest/test_suggest.py
mtbottle: Ah, I saw that the feature flag was removed in #131 . Does it make sense to remove this test altogether (I don't think it's been removed in the original PR). cc/ @Trinaa

taddes (commented):
Great work @mtbottle ! Just left the one comment we discussed regarding the default active providers.

merino/web/api_v1.py
pjenvey: No need for the list construction also:

```python
Counter(suggest.provider for suggestion in suggestions)
```

merino/web/api_v1.py
mtbottle: Sounds good!

tests/integration/api/v1/suggest/test_timeout_handling.py
mtbottle: Updated the comment to reference where the 2 provider reference is! The integration test have the providers injected, so it's a bit easier to control them than looking at the configuration file. But definitely having a comment on where to verify that there's 2 providers will be useful!

merino/web/api_v1.py
```diff
@@ -114,6 +117,28 @@ async def suggest(
...
+
+    suggestion_counter = Counter(suggestion.provider for suggestion in suggestions)
+
+    logging.error(f"{searched_providers}")
+
+    for provider in searched_providers:
+        provider_name = provider.name
+        suggestion_count = suggestion_counter[provider_name]
+        metrics_client.histogram(
+            "suggestions-per.provider",
```
ncloudioj, line 136: By convention, we normally write the provider id as part of the metric name.

 ```suggestion
            f"suggestions-per.provider.{provider_name}",
```

File: merino/web/api_v1.py [hunk 8]
```diff
@@ -114,6 +117,28 @@ async def suggest(
...
+    metrics_client: Client,
+    suggestions: list[BaseSuggestion],
+    searched_providers: list[BaseProvider],
+) -> None:
+    """Emit metrics for suggestions per request and suggestions per request by provider."""
+    metrics_client.histogram("suggestions-per.request", value=len(suggestions))
+
+    suggestion_counter = Counter(suggestion.provider for suggestion in suggestions)
+
+    logging.error(f"{searched_providers}")
```
ncloudioj, line 130: Debugging leftover?

merino/web/api_v1.py
```diff
@@ -106,6 +107,8 @@ async def suggest(
         )
     )
 
+    emit_suggestions_per_metrics(metrics_client, suggestions, search_from)
```
ncloudioj, line 110: Not for this PR, just a note. To reduce response latency, we might consider recording metrics via [background-tasks](https://fastapi.tiangolo.com/tutorial/background-tasks/) if this takes too long.

merino/web/api_v1.py
mtbottle: Ah, if we're providing this in the metric name, then we should remove it from the tags? I agree that this is more conventional, and would prefer this over using tags for the provider. The only downside is that cross comparisons will be more difficult, but I'm not sure this is useful to look at.

quiiver: I also prefer this. for context, i wrote the ticket based on the way this metric was implemented in the rust version of merino.

merino/web/api_v1.py
mtbottle: Whoops! Thanks for the catch!

merino/web/api_v1.py
ncloudioj: I think either format is fine as long as it's consistent across the code base. If we prefer tags, could you file a follow-up ticket to convert the existing metrics as well (and update the Grafana dashboard)? Thank!

ncloudioj (approved):
Looks great. Thanks!

🚢

merino/web/api_v1.py
mtbottle: By popular opinion, I've renamed the metric to use the suggested naming scheme. I'll update the Grafana dashboard once this change is out!

mtbottle:
> @mtbottle Hmm, looks like ./idea got sneaked into the PR, maybe mis-checkins?

Ah, how embarrassing 🤦 Removed now.

PR #130, opened 2022-11-30
Updated v1 domain file with icon urls

abhi-agg:
The new `top_picks.json` file generated by taking the old `top_picks.json` file and feeding it to the [modified tippy top](https://github.com/abhi-agg/tippy-top-sites/tree/d52c4c4afdc094645e119f37cf3f37d4e3b89bc8)


ncloudioj:
If we are using the original icon urls instead of copying & hosting them behind the Mozilla-managed CDN, that might cause some privacy concerns in Firefox. @abhi-agg Do we have any plan to host those icons?

quiiver:
@ncloudioj yes the plan is still to host them but we wanted to include the original source urls here to make sure we were not blocking anyone

abhi-agg:
@ncloudioj Yes, as Wil already shared, we plan to host them behind mozilla managed CDN 👍🏾

taddes:
Hi @quiiver and @abhi-agg , so shall we hold off until the urls are hosted behind the managed CDN and we have a file that points to said internal icons?

quiiver:
I think we can use this for testing and evaluation purposes, we just need to be careful about not enabling the provider until we can move the icons to a cdn

taddes:
Ok, thanks @quiiver , WFM. As it stands, the provider is not enabled by default so should be good there.

taddes (approved):
Looks good @abhi-agg , we'll just proceed for testing and make sure to provide the updated file with CDN internal urls in the future.

ncloudioj:
@taddes Are we waiting for anything else to merge this?

Issue #116, opened 2022-11-18
Include 'tests' in mypy linting


data-sync-user:
Include the `tests` directory in the _mypy_ linting scope in merino-py.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/DISCO-2118)
┆Epic: Merino: maintenance

data-sync-user:
➤ Katrina Anderson commented:

Update (2022\12\02)

  * PR Opened → https://github.com/mozilla-services/merino-py/pull/135 ( https://github.com/mozilla-services/merino-py/pull/135 )
  * Setting task status to 'In Review'

data-sync-user:
➤ Katrina Anderson commented:

Update (2022\12\05)
Merged pull request
Set task to 'Done'

PR #122, opened 2022-11-23
[CONSVC-2085] feat: migrate to python 3.11

ncloudioj:
Took a crack at trying out Python 3.11. Approx. 20% uplift in throughput and 8% uplift in p95 latency. Holly Smokes! 💨 

Preliminary benchmark results (running 100VU for 60 seconds.):

### Main branch (Python 3.10)
- RPS: 2,858
- Latency (P95): 47.20ms

### This branch (Python 3.11)
- RPS: 3,418 (20% ⬆️)
- Latency (P95): 43.07ms (8% ⬇️)

This closes [CONSVC-2085](https://mozilla-hub.atlassian.net/browse/CONSVC-2085).

PR #135, opened 2022-12-02
[CONSVC-2118] test: include 'tests' in mypy linting

Trinaa:
# Description
- Update the `Makefile` to include the `tests\unit` and `tests\load` directories in pydocstyle linting
- Update the `.pre-commit-config.yaml` file to including the `tests\unit`, `tests\integration` and `tests\load` directories in pydocstyle linting
- Remediate errors in `tests` directory

# Issue(s)
#116 | [CONSVC-2118](https://mozilla-hub.atlassian.net/browse/CONSVC-2118)


tests/load/locust_tests/locustfile.py
```diff
@@ -12,11 +12,11 @@
 import faker
 import kinto_http
 from client_info import DESKTOP_FIREFOX, LOCALES
-from kinto import download_suggestions
+from kinto_client import download_suggestions
 from locust import HttpUser, events, task
 from locust.clients import HttpSession
 from locust.runners import MasterRunner
-from models import ResponseContent
+from suggest_models import ResponseContent
```
taddes, line 19: Nice, I think it's good to have clearer identifiers like this, faster to read!

.pre-commit-config.yaml
```diff
@@ -44,7 +44,7 @@ repos:
     rev: v0.971
     hooks:
       - id: mypy
-        files: '^merino|^tests/contract'
+        files: '^merino|^tests'
         args: [
           --python-version=3.10,
```
taddes, line 49: Should we update the Python version to `3.11` here, given the recent upgrade?

tests/unit/providers/test_accuweather.py
```diff
@@ -174,10 +174,10 @@
 
 
 def set_response_bodies(
-    location: dict = default_location_body,
-    current_conditions: dict = default_current_conditions_body,
-    forecast: dict = default_forecast_body,
-):
+    location=default_location_body,
```
taddes, line 177: Can I ask why the type annotations were removed here? I'm guessing implicit typing of some sort, or perhaps some context regarding the globals below? Just curious!  🕵️‍♂️

taddes (approved):
Great stuff, and rational call so we check everything @Trinaa . I just had 2 questions and a thought on updating the Python version. 🚀

.pre-commit-config.yaml
Trinaa: I see in the https://github.com/mozilla-services/merino-py/pull/122 pull request that this change is included. If that PR is merged first, I will be sure the added test directories are compliant.

tests/unit/providers/test_accuweather.py
Trinaa: Good question. Essentially it's a judgment call on my part, since the types provided were not specific enough for mypy and the true types are complex and inconsistent.

Normally I might declare a type alias or model to rectify the error, but since there is an Accuweather Epic [CONSVC-2020](https://mozilla-hub.atlassian.net/browse/CONSVC-2020) next on the docket and it looks as if there will be changes that may drastically alter this code [CONSVC-2057](https://mozilla-hub.atlassian.net/browse/CONSVC-2057), I opted to conserve the effort.

PR #137, opened 2022-12-03
[CONSVC-2139] feat: add support for provider-level timeouts

ncloudioj:
In certain cases, providers might be able to tolerate higher query latency than others, for instance, for zero-prefix weather suggestions, Firefox sends the query request to Merino in the background, hence no timeout is enforced on the client side. This patch allows us to specify provider-level query timeout to overwrite the top-level timeout. When multiple providers are specified in the request, Merino will use the maximum timeout among those providers as the maximum query waiting time for that request. 

Particularly, now the Accuweather provider uses a 1-second timeout rather than the default 200ms.

This closes [CONSVC-2139](https://mozilla-hub.atlassian.net/browse/CONSVC-2139).

cc: @0c0w3

### DIFF
--- a/merino/config.py
+++ b/merino/config.py
@@ -10,6 +10,7 @@ _validators = [
     Validator("metrics.dev_logger", is_type_of=bool),
     Validator("deployment.canary", is_type_of=bool),
     Validator("providers.accuweather.enabled_by_default", is_type_of=bool),
+    Validator("providers.accuweather.query_timeout_sec", is_type_of=float, gte=0),
     Validator("providers.adm.enabled_by_default", is_type_of=bool),
     Validator("providers.adm.cron_interval_sec", gt=0),
     Validator("providers.adm.resync_interval_sec", gt=0),
--- a/merino/web/api_v1.py
+++ b/merino/web/api_v1.py
@@ -14,7 +14,7 @@ from starlette.requests import Request
 from merino.config import settings
 from merino.metrics import Client
 from merino.middleware import ScopeKey
-from merino.providers import get_providers
+from merino.providers import get_max_timeout, get_providers
 from merino.providers.base import BaseProvider, BaseSuggestion, SuggestionRequest
 from merino.utils import task_runner
 from merino.web.models_v1 import ProviderResponse, SuggestResponse
@@ -89,7 +89,9 @@ async def suggest(
 
     completed_tasks, _ = await task_runner.gather(
         lookups,
-        timeout=QUERY_TIMEOUT_SEC,
+        timeout=get_max_timeout(
+            frozenset(p.name for p in search_from), QUERY_TIMEOUT_SEC
+        ),
         timeout_cb=partial(task_runner.metrics_timeout_handler, metrics_client),
     )
     suggestions = list(
--- a/tests/unit/providers/test_init_providers.py
+++ b/tests/unit/providers/test_init_providers.py
@@ -9,7 +9,12 @@ from pytest_mock import MockerFixture
 
 from merino.config import settings
 from merino.exceptions import InvalidProviderError
-from merino.providers import ProviderType, get_providers, init_providers
+from merino.providers import (
+    ProviderType,
+    get_max_timeout,
+    get_providers,
+    init_providers,
+)
 
 
 @pytest.mark.asyncio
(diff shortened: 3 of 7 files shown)


docs/ops.md
```diff
@@ -85,7 +85,8 @@ tooling.
 - `runtime.query_timeout_sec` (`MERINO_RUNTIME__QUERY_TIMEOUT_SEC`) - A floating
   point (in seconds) indicating the maximum waiting period for queries issued within
   the handler of the `suggest` endpoint. All the unfinished query tasks will be
-  cancelled once the timeout gets triggered.
+  cancelled once the timeout gets triggered. Note that this timeout could be overwritten
+  by individual providers with the same configuration.
```
mtbottle, line 89: nit: confusing sentence,  specifically `overwritten by individual providers with the same configuration` is unclear. I would suggest:

`Note that this timeout can also be configured by specific providers. The provider configured timeout takes precedence over this value`.

mtbottle (approved):
Change LGTM! I just have a nitpick, and a non-blocking comment.

docs/ops.md
ncloudioj: Good catch, thanks!

merino/web/api_v1.py
```diff
@@ -89,7 +89,7 @@ async def suggest(
 
     completed_tasks, _ = await task_runner.gather(
         lookups,
-        timeout=QUERY_TIMEOUT_SEC,
+        timeout=get_max_timeout([p.name for p in search_from], QUERY_TIMEOUT_SEC),
```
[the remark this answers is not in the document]
ncloudioj: Correct. The current timeout mechanism is not truly enforced at the provider level, rather it picks the longest one (if specified) and uses that as the timeout for the entire request (which in turn might involve multiple providers). I think this is fine for now as Firefox will specify the provider(s) in interest, but we might want to implement the true provider-level timeout in the future if needed.

hackebrot, line 92: We populate `provider_timeouts` when calling `init_providers` which maps from a provider ID or `name` to a `float` (we hardcode the ID, but it's effectively identical to the name).

The local variable `search_from` is of type `list[BaseProvider]`. The provider instances it holds are filtered by `active_providers` even if specified via request query parameters (`active_providers` is a local variable that points to `providers: dict[str, BaseProvider]` in `merino/providers/__init__.py`).

We collect the names of providers from `search_from` and pass that as an argument to `get_max_timeout` (`providers: list[str]`). The interator in `get_max_timeout` skips providers which don't appear in `provider_timeouts` which in turn is created from `providers` in `merino/providers/__init__.py`.

We might be able to follow a more direct approach:

```suggestion
        timeout=max(
            [provider.query_timeout_sec for provider in search_from],
            default=QUERY_TIMEOUT_SEC,
        ),
```

My concern is specifically about the added complexity of converting IDs to provider instances and vice versa and using that to call helper functions which depend on module attributes.

tests/unit/providers/test_init_providers.py
```diff
@@ -27,6 +32,26 @@ async def test_init_providers() -> None:
...
+    [
+        ([ProviderType.ADM.value], 0.1, settings.runtime.query_timeout_sec),
+        (
+            [ProviderType.ACCUWEATHER.value, ProviderType.ADM.value],
+            0.1,
+            settings.providers.accuweather.query_timeout_sec,
+        ),
+        (["a-missing-provider"], 0.1, 0.1),
+    ],
+)
```
hackebrot, line 46: I suggest assigning IDs for test items explicitly here for the sake of comprehensibility.

merino/config.py
```diff
@@ -10,6 +10,7 @@
     Validator("metrics.dev_logger", is_type_of=bool),
     Validator("deployment.canary", is_type_of=bool),
     Validator("providers.accuweather.enabled_by_default", is_type_of=bool),
+    Validator("providers.accuweather.query_timeout_sec", is_type_of=float, gte=0),
```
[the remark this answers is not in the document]
mtbottle: AFAICT, the only provider that actually overrides the `query_timeout_sec` is Accuweather. The implementation doesn't allow other providers to override it yet. If we decide to extend overriding of `query_timeout_sec` to other providers, then it makes sense to make the Validation more generic.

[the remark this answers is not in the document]
ncloudioj: Yes, this configuration is currently only overridden by this provider. In fact, I'd be more inclined to have each provider define its own validator to suit the actual need. For instance, the default 200ms timeout is used for low-latency scenarios, whereas for the zero-prefix Accuweather provider, Firefox won't timeout if Merino can't get back to it in 200ms. Therefore, the valid timeout range for this kind of provider would be wider than that of others.

That said, I think ideally we should have the following validators (assuming we allow Merino to use up to 5 sec to fetch forecasts from Accuweather):
```
Validator("providers.runtime.query_timeout_sec", is_type_of=float, gte=0, lte=0.2),
Validator("providers.accuweather.query_timeout_sec", is_type_of=float, gte=0, lte=5),
```

merino/web/api_v1.py
ncloudioj: Good point. I think originally I applied `@functools.cache` to `get_max_timeout()` but later ended up dropping it due to its overhead. Yeah, we can get the maximum timeout directly as you mentioned here. I will file a follow-up to simplify it.

PR #140, opened 2022-12-05
refactor: simplify timeout handling for providers

ncloudioj:
This is a follow-up for #137, in which I missed @hackebrot's review comments.

### DIFF
--- a/merino/config.py
+++ b/merino/config.py
@@ -10,7 +10,11 @@ _validators = [
     Validator("metrics.dev_logger", is_type_of=bool),
     Validator("deployment.canary", is_type_of=bool),
     Validator("providers.accuweather.enabled_by_default", is_type_of=bool),
-    Validator("providers.accuweather.query_timeout_sec", is_type_of=float, gte=0),
+    # Set the upper bound of query timeout to 5 seconds as we don't want Merino
+    # to wait for responses from Accuweather indefinitely.
+    Validator(
+        "providers.accuweather.query_timeout_sec", is_type_of=float, gte=0, lte=5.0
+    ),
     Validator("providers.adm.enabled_by_default", is_type_of=bool),
     Validator("providers.adm.cron_interval_sec", gt=0),
     Validator("providers.adm.resync_interval_sec", gt=0),
@@ -18,7 +22,23 @@ _validators = [
     Validator("providers.adm.backend", is_in=["remote-settings", "test"]),
     Validator("providers.adm.score_wikipedia", gte=0, lte=1),
     Validator("providers.wikifruit.enabled_by_default", is_type_of=bool),
-    Validator("runtime.query_timeout_sec", is_type_of=float, gte=0),
+    # Since Firefox will time out the request to Merino if it takes longer than 200ms,
+    # the default query timeout of Merino should not be greater than that 200ms.
+    Validator(
+        "runtime.query_timeout_sec",
+        is_type_of=float,
+        gte=0,
+        lte=0.2,
+        env=["production", "ci"],
+    ),
+    # Allow a longer timeout for testing & development
+    Validator(
+        "runtime.query_timeout_sec",
+        is_type_of=float,
+        gte=0,
+        lte=1.0,
+        env=["testing", "development"],
+    ),
     Validator("sentry.mode", is_in=["disabled", "release", "debug"]),
     Validator("sentry.env", is_in=["prod", "stage", "dev"]),
     Validator("sentry.traces_sample_rate", gte=0, lte=1),
--- a/tests/integration/api/v1/fake_providers.py
+++ b/tests/integration/api/v1/fake_providers.py
@@ -158,3 +158,13 @@ class TimeoutSponsoredProvider(SponsoredProvider):
         """Query against the TimeoutSponsoredProvider."""
         await asyncio.sleep(settings.runtime.query_timeout_sec * 2)
         return await super().query(srequest)
+
+
+class TimeoutTolerantSponsoredProvider(TimeoutSponsoredProvider):
+    """A timeout tolerant sponsored provider."""
+
+    def __init__(self, enabled_by_default) -> None:
+        super().__init__(enabled_by_default=enabled_by_default)
+        self._name = "timedout-tolerant-sponsored"
+        # It can tolerate for 4x of the default query timeout
+        self._query_timeout_sec = settings.runtime.query_timeout_sec * 4
--- a/tests/integration/api/v1/suggest/test_timeout_handling.py
+++ b/tests/integration/api/v1/suggest/test_timeout_handling.py
@@ -5,6 +5,7 @@
 """Integration tests for the Merino v1 suggest API endpoint focusing on time out
 behavior.
 """
+from collections import namedtuple
 
 import aiodogstatsd
 import pytest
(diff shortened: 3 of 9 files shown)


tests/integration/api/v1/suggest/test_timeout_handling.py
```diff
@@ -73,3 +73,38 @@ def test_with_timedout_provider(
...
+    "providers",
+    [
+        {
+            "timeout-tolerant-sponsored": TimeoutTolerantSponsoredProvider(
+                enabled_by_default=True
+            )
+        }
+    ],
+)
+def test_with_provider_customized_timeout(
```
mtbottle, line 88: It feels like this test can be parameterized with the test above with a bit of a refactor. In particular, if we add and `expected_timeout_log` and `expected_metrics` (and modify the test itself to use those lists/sets), then we should be able to reuse the test above to add this case.

Short POC:

```
@pytest.mark.parametrize(
    ["providers", "expected_logs", "expected_metrics"],
    [
        (
            {
                "sponsored": SponsoredProvider(enabled_by_default=True),
                "timedout-sponsored": TimeoutSponsoredProvider(enabled_by_default=True),
            },
            [
                "Timeout triggered in the task runner",
                "Cancelling the task: timedout-sponsored due to timeout",
            ],
            {
                "providers.timedout-sponsored.query.timeout",
                "response.status_codes.200",
            },
        ),
    ],
    ids=["timeout_with_provider"],
)
def test_with_timedout_provider(
    mocker: MockerFixture,
    client: TestClient,
    caplog: LogCaptureFixture,
    filter_caplog: FilterCaplogFixture,
    expected_logs: list[str],
    expected_metrics: set[str],
) -> None:

    ...

    # Check logs for the timed out query
    records = filter_caplog(caplog.records, "merino.utils.task_runner")
    logs = [r.__dict__["msg"] for r in records]
    assert logs == expected_logs

    # Check metrics for the timed out query
    metric_keys: list[str] = {call.args[0] for call in report.call_args_list}
    assert expected_metrics.issubset(metric_keys)
```

merino/config.py
```diff
@@ -10,15 +10,35 @@
...
+        env=["production", "ci"],
+    ),
+    # Allow a longer timeout for testing & development
+    Validator(
+        "runtime.query_timeout_sec",
+        is_type_of=float,
+        gte=0,
+        lte=1.0,
+        env=["testing", "development"],
+    ),
```
mtbottle, line 41: Hmm... I'm wondering if we will be missing a potential production performance issue if dev or testing environments are consistently going over 200ms? I am also somewhat tempted to make validation simpler by _not_ adding this unless we're actually seeing lots of performance issues with development.

tests/integration/api/v1/suggest/test_timeout_handling.py
hackebrot: I like that idea, @mtbottle! +1 🚀

hackebrot (changes requested):

merino/config.py
ncloudioj: > I'm wondering if we will be missing a potential production performance issue if dev or testing environments are consistently going over 200ms?

FWIW, we use a larger timeout in the testing environment (not including contract tests) to avoid potential intermittent test failures in CI due to its instability in computing resources. As for development, I think using a relaxed timeout would actually allow us to understand the real query duration, otherwise, the task runner will time out the request before it gets a chance to finish it.

In addition, since we use the same timeout for stage & production and CI (i.e. contract tests), I believe it's very unlikely to get performance issues sneaked into production. Even if that ever happens somehow, the timeout metrics would capture those right away.

Does that make sense?

mtbottle: > we use a larger timeout in the testing environment (not including contract tests) to avoid potential intermittent test failures in CI due to its instability in computing resources.

Oh interesting. Did we have a history of test instability in CI? I think unit tests and integration tests run on 4CPU/8Gb Docker containers on CircleCI, which _feels_ like it would stable enough (and also these tests have the network calls all mocked out, which reduces instability even more). 

I think the points make sense 👍 . I just wanted to make sure that we're not starting to differentiate the configurations unnecessarily (YAGNI), as maintaining 2 validation code paths, although not super complex, is still introducing complexity to the system.

ncloudioj: Yes, we did run into that several times in CI or even running locally inside the container, and we ended up introducing a couple of delays to avoid the flaky test failures (see [here](https://github.com/mozilla-services/merino-py/blob/main/tests/contract/volumes/client/scenarios.yml#L157)).

> I just wanted to make sure that we're not starting to differentiate the configurations unnecessarily (YAGNI), as maintaining 2 validation code paths, although not super complex, is still introducing complexity to the system.

Agreed. I am also fine with dropping this treatment until the problem arises in the future.

tests/integration/api/v1/suggest/test_timeout_handling.py
mtbottle: I have a preference for the parameterized version for a few reasons:

1. Consistency of tests. We can be sure that the same things are checked, which increases confidence in correctness.
2. Maintainability. Tests tend to share a lot of similarities, like how the testee function is invoked, or having similar assert statements. If we make a change to how the provider is invoked (for example), then this will require changes to all the tests. This parameterized version will only need the change in one place. 

>  Though, the downside of doing so is that it might hurt the readability of those two test cases. For example, it may take some time for the readers to figure out the following parameters are used to test

Ya, it's a bit more indirect and requires a bit more time to read and understand.

ncloudioj: Yeah, I agree with you about the benefits of parametrization, I will refactor this to it perhaps with more parameters to cover some other cases.

One thing that I found awkward about the parametrization is that for the 1st case, we wanted to ensure `providers.timedout-sponsored.query.timeout` was recorded in metrics, whereas for the 2nd case, we wanted to verify `providers.timedout-tolerant-sponsored.query.timeout` should *not* be recorded. Since we can't do `expected_metrics.issubset()` and `not expected_metrics.issubset()`, we will have to list all the metrics for each case.

hackebrot:
Looks like CI failed for the most recent push due to an issue around dependencies:

```
make: *** [Makefile:19: .install.stamp] Error 1
```

tests/integration/api/v1/fake_providers.py
```diff
@@ -158,3 +158,13 @@ async def query(self, srequest: SuggestionRequest) -> list[BaseSuggestion]:
         """Query against the TimeoutSponsoredProvider."""
         await asyncio.sleep(settings.runtime.query_timeout_sec * 2)
         return await super().query(srequest)
+
+
+class TimeoutTolerantSponsoredProvider(TimeoutSponsoredProvider):
```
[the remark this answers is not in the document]
ncloudioj: @hackebrot Thanks for the insights! I understand what your concerns are now, and yeah, I for sure trust your judgment and look forward to seeing the fix :)

tests/integration/api/v1/suggest/test_timeout_handling.py
```diff
@@ -15,61 +15,162 @@
...
-def fixture_providers() -> Providers:
-    """Define providers for this module which are injected automatically."""
-    return {
-        "sponsored": SponsoredProvider(enabled_by_default=True),
-        "timedout-sponsored": TimeoutSponsoredProvider(enabled_by_default=True),
-    }
-
-
-def test_with_timedout_provider(
+@pytest.mark.parametrize(
```
hackebrot, line 23: We could probably increase the readability of this by using a `collections.namedtuple`:

```python
Scenario = namedtuple(
    "Scenario",
    [
        "providers",
        "expected_suggestion_count",
        "expected_logs_on_task_runner",
        "expected_metric_keys",
    ],
)
```

Then define a dict (order should be guaranteed since Python 3.7) for the scenarios:

```python

SCENARIOS: dict[str, Scenario] = {
    # Case I: Only one provider that will time out.
    #
    #   - Expects:
    #     - No suggestions
    #     - Timeout logs recorded in the task runner
    #     - Timeout metrics recorded in the task runner
    "A-timed-out-provider": Scenario(
        providers={
            "timedout-sponsored": TimeoutSponsoredProvider(enabled_by_default=True),
        },
        expected_suggestion_count=0,
        expected_logs_on_task_runner={
            "Timeout triggered in the task runner",
            "Cancelling the task: timedout-sponsored due to timeout",
        },
        expected_metric_keys={
            "providers.timedout-sponsored.query",
            "providers.timedout-sponsored.query.timeout",
            "suggestions-per.request",
            "suggestions-per.provider.timedout-sponsored",
            "get.api.v1.suggest.timing",
            "get.api.v1.suggest.status_codes.200",
            "response.status_codes.200",
        },
    ),
}
```

The decorator would change to:

```python
@pytest.mark.parametrize(
    argnames=[
        "providers",
        "expected_suggestion_count",
        "expected_logs_on_task_runner",
        "expected_metric_keys",
    ],
    ids=SCENARIOS.keys(),
    argvalues=SCENARIOS.values(),
)
```

The benefit of using namedtuples for this is that it's easier to spot what the values for each test parameter for a given test scenario because. What do you think @ncloudioj?

ncloudioj: Fancy 👖. TIL💡, thanks!

hackebrot (approved):
LGTM! 🚀

ncloudioj:
Thank you both for the review!

PR #141, opened 2022-12-05
[DISCO-2044] Put the log metadata fields into their own objects

mtbottle:
This will make it easier to do data validation on fields we send to logger and also encapsulates the logging metadata into extendable objects.

Fixes DISCO-2044

### DIFF
--- a/merino/utils/log_data_creators.py
+++ b/merino/utils/log_data_creators.py
@@ -1,7 +1,8 @@
 """A utility module for lag data creation"""
 from datetime import datetime
-from typing import Any
+from typing import Any, Optional
 
+from pydantic import BaseModel
 from starlette.datastructures import Headers
 from starlette.requests import Request
 from starlette.types import Message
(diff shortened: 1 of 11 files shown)


merino/utils/log_data_creators.py
```diff
@@ -11,63 +12,98 @@
...
+    time: datetime
+    path: str
+    method: str
+
+    def dict(self, **kwargs) -> dict[str, Any]:
+        """Override the dict method to convert the datetime type to iso-formatted string."""
+        d: dict[str, Any] = super(LogDataModel, self).dict(**kwargs)
+        if d.get("time"):
+            d["time"] = d["time"].isoformat()
+        return d
```
mtbottle, line 30: Not a fan of this, but I wanted to preserve the `time` as a `datetime` rather than `str`, but it doesn't export nicely. Any suggestions on more elegant ways to do this instead will be appreciated.

tests/unit/utils/test_log_data_creator.py
```diff
@@ -57,11 +62,10 @@ def test_create_request_summary_log_data(
         }
     )
     message: Message = {"type": "http.response.start", "status": "200"}
-    dt: datetime = datetime(1998, 3, 31)
 
     log_data: dict[str, Any] = create_request_summary_log_data(request, message, dt)
 
-    assert log_data == expected_log_data
+    assert log_data == expected_log_data.dict()
```
Trinaa, line 68: The type of `log_data` should be `RequestSummaryLogDataModel` now I believe and I think then we can keep the assert as is.

```
log_data: RequestSummaryLogDataModel = create_request_summary_log_data(request, message, dt)

assert log_data == expected_log_data
```

tests/unit/utils/test_log_data_creator.py
```diff
@@ -158,4 +163,4 @@ def test_create_suggest_log_data(
 
     log_data: dict[str, Any] = create_suggest_log_data(request, message, dt)
 
-    assert log_data == expected_log_data
+    assert log_data == expected_log_data.dict()
```
Trinaa, line 166: Similar to above, I think `log_data` is now type `SuggestLogDataModel`

```
log_data: SuggestLogDataModel = create_suggest_log_data(request, message, dt)

assert log_data == expected_log_data
```

Trinaa (changes requested):
Thank you for picking up this task @mtbottle!

There are a couple more integration tests where we can use these models:
test_error.py::test_error_request_log_data
test_heartbeat.py::test_heartbeat_request_log_data
test_version.py::test_version_request_log_data
test_unsupported::test_unsupported_endpoint_request_log_data
test_providers.py::test_providers_request_log_data

**Note** merging with `main` may be helpful, since it will add unit tests to the mypy scope.

merino/utils/log_data_creators.py
Trinaa: Being Devil's advocate... Why preserve the `datetime` type if it creates this overhead?

mtbottle: Good question! I wanted to keep the `time` as `datetime` for type checking (ie. ensuring that we don't accept any random string that's not `datetime` formatted when passed into the object). I _think_ it's worth keeping.

hackebrot: We can enforce valid datetime `str` values on the pydantic model, I think:
https://docs.pydantic.dev/usage/types/#datetime-types

Attempts to create a new model instance with a `str` value for `time` that doesn't validate against the format should result in an exception.

Trinaa (approved):

merino/utils/log_data_creators.py
hackebrot: Hi @mtbottle! 👋🏻 

Just wanted to check if you've seen my above comment. From what I can tell we can rely on pydantic to validate input parameters for model fields on model instance creation. We can use a `datetime` model field for `time` and then use `.json()` rather than `.dict()` when attaching that to the log message.

See https://docs.pydantic.dev/usage/exporting_models/#modeljson

mtbottle: Oh sorry! I saw your comment and thought it was just adding context to my comment. 

> We can use a datetime model field for time and then use .json() rather than .dict() when attaching that to the log message.

Ah ya, this works and is much nicer! I'll update this change to use `.json` instead. Thanks for following up on this!

hackebrot: No worries! Thank you, @mtbottle! 🚀

PR #149, opened 2022-12-13
Logging Object to use Separate `datetime` and `str` fields for timestamps

mtbottle:
As discussed over Slack with @hackebrot , I think it's worth being explicit about having separate fields for `datetime` object and `str` representation. This will make it clearer in the `dict()` export that we care about the `str` object, but also preserves the implicit validation of the `datetime` Pydantic object.

### DIFF
--- a/merino/utils/log_data_creators.py
+++ b/merino/utils/log_data_creators.py
@@ -2,7 +2,7 @@
 from datetime import datetime
 from typing import Any, Optional
 
-from pydantic import BaseModel
+from pydantic import BaseModel, parse_obj_as, validator
 from starlette.datastructures import Headers
 from starlette.requests import Request
 from starlette.types import Message
@@ -18,16 +18,19 @@ class LogDataModel(BaseModel):
     """
 
     errno: int
-    time: datetime
+    time: str
     path: str
     method: str
 
-    def dict(self, **kwargs) -> dict[str, Any]:
-        """Override the dict method to convert the datetime type to iso-formatted string."""
-        d: dict[str, Any] = super().dict(**kwargs)
-        if d.get("time"):
-            d["time"] = d["time"].isoformat()
-        return d
+    @validator("time", pre=True)
+    def validate_time(cls, v, values, **kwargs):
+        """Attempt to parse the provided value for `time` into a datetime
+        model which will check that the value is a valid datetime representation.
+        Then ensure that we output the datetime in a consistent isoformat,
+        if it wasn't already in that format.
+        """
+        dt: datetime = parse_obj_as(datetime, v)
+        return dt.isoformat()
 
 
 class RequestSummaryLogDataModel(LogDataModel):
--- a/tests/integration/api/v1/suggest/test_suggest.py
+++ b/tests/integration/api/v1/suggest/test_suggest.py
@@ -5,7 +5,6 @@
 """Integration tests for the Merino v1 suggest API endpoint."""
 
 import logging
-from typing import Any
 
 import aiodogstatsd
 import pytest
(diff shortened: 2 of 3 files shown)


merino/utils/log_data_creators.py
```diff
@@ -18,16 +18,17 @@ class LogDataModel(BaseModel):
...
-    def dict(self, **kwargs) -> dict[str, Any]:
-        """Override the dict method to convert the datetime type to iso-formatted string."""
-        d: dict[str, Any] = super().dict(**kwargs)
-        if d.get("time"):
-            d["time"] = d["time"].isoformat()
-        return d
+    @validator("time", always=True)
+    def set_time(cls, v, values, **kwargs):
+        """Set the time field based on the value of time_input."""
+        return values["time_input"].isoformat()
```
hackebrot, line 31: There are a few cases that we need to guard against:
1. `time_input` is required but not provided and the `str` value for `time` is provided but not a valid datetime representation
2. `time_input` is provided as a `str` value that is not a valid datetime representation
3. `time_input` is provided, but so is `time`. However we need to infer `time` from `time_obj`

Example code:
1. `LogDataModel(errno=0, time="nope", path="/hello", method="GET")`
2. `LogDataModel(errno=0, time_input="nope", path="/hello", method="GET")`
3. `LogDataModel(errno=0, time_input=datetime.now(), time="nope", path="/hello", method="GET")`

The current implementation:
1. fails with a `KeyError: 'time_input'` rather than the regular `field required (type=value_error.missing)` validation error and `invalid datetime format (type=value_error.datetime)` validation error
2. fails with a `KeyError: 'time_input'` rather than the regular `invalid datetime format (type=value_error.datetime)` validation error
3. does not raise an exception, but ignores the value for `time` that was provided

Note on 1: From what I can tell pydantic doesn't expose the specific datetime validation functionality, but we can use `parse_obj_as(datetime, v)` to attempt to load the provided value into a temporary model and that will run validation.

The code below covered all of these cases when I tested it locally:

```python
@validator("time", always=True, pre=False)
def set_time(cls, v, values, **kwargs):
    """Set the value for the time field based on the value of time_obj."""
    if time_input := values.get("time_input"):
        # A value for `time_input` was provided and pydantic validated and
        # parsed that value into a datetime object.

        if v is not None:
            raise ValueError("time and time_input both specified")

        # Set the value for `time` based on `time_input`
        return time_input.isoformat()

    if v is not None:
        # Attempt to parse the provided value for `time` into a datetime
        # model which will check that the value is a valid representation.
        parse_obj_as(datetime, v)

    return v
```

hackebrot (changes requested):
Thank you for submitting a pull request for this @mtbottle!

I've added one comment inline. 📝

merino/utils/log_data_creators.py
```diff
@@ -18,16 +18,19 @@ class LogDataModel(BaseModel):
...
-        return d
+    @validator("time", always=True, pre=True)
+    def validate_time(cls, v, values, **kwargs):
+        """Attempt to parse the provided value for `time` into a datetime
+        model which will check that the value is a valid datetime representation.
+        Then ensure that we output the datetime in a consistent isoformat,
+        if it wasn't already in that format.
+        """
+        dt: datetime = parse_obj_as(datetime, v)
+        return dt.isoformat()
```
mtbottle, line 33: @hackebrot The last conditional clause of your last comment inspired me to simplify this more. We never _really_ need to use the object, so I think using it as an object only during the validation is sufficient. This avoids the complexity of keeping around an unused `datetime` object, as well as ensuring that both string and object are in sync. When I add `pre=True`, the field doesn't actually gets string validated yet, which makes it easier to treat as if it's a `datetime` field type. I've added some tests below to make sure that we get a `ValidationError` for incorrect times, but also parse the `datetime.isoformat` from the date specified by the input. Let me know what you think!

hackebrot (changes requested):

merino/utils/log_data_creators.py
mtbottle: Ah, good catch! We don't need `always=True`, so removing it!

hackebrot (changes requested):
Hi @mtbottle! 👋🏻 

I've added suggestions for the tests. Let me know what you think!

Aside from that this is good to go!

hackebrot (approved):
Great work, @mtbottle! 🚀

PR #150, opened 2022-12-15
chore: set accuweather query timeout to 5 seconds

ncloudioj:
A followup of [DISCO-2139](https://mozilla-hub.atlassian.net/browse/DISCO-2139?focusedCommentId=627176).


0c0w3 (approved):
Thanks!

PR #151, opened 2022-12-16
[DISCO-2138] fix: guard against malformed seq no

tiftran:
We could have requests where `seq` is not a valid integer string. These requests will most likely be sent from non-Firefox user agents but we should still handle the conversion exceptions in Merino to avoid errors.

### DIFF
--- a/merino/utils/log_data_creators.py
+++ b/merino/utils/log_data_creators.py
@@ -81,6 +81,7 @@ def create_suggest_log_data(
     """Create log data for the suggest API endpoint."""
     location: Location = request.scope[ScopeKey.GEOLOCATION]
     user_agent: UserAgent = request.scope[ScopeKey.USER_AGENT]
+    seq: str = request.query_params.get("seq", "")
 
     return SuggestLogDataModel(
         # General Data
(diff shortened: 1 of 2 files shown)


merino/utils/log_data_creators.py
```diff
@@ -94,7 +95,7 @@ def create_suggest_log_data(
         code=message["status"],
         rid=Headers(scope=message)["X-Request-ID"],
         session_id=request.query_params.get("sid"),
-        sequence_no=int(seq) if (seq := request.query_params.get("seq")) else None,
+        sequence_no=int(seq) if (seq.isnumeric() and float(seq).is_integer()) else None,
```
ncloudioj, line 98: Given that `seq` should be any natural number. We can use `isdecimal()` here within the oneliner.
```suggestion
        sequence_no=int(seq) if (seq := request.query_params.get("seq", "")) and seq.isdecimal() else None,
```

tiftran: oooh clever

ncloudioj (approved):
Looks good, thanks!

PR #152, opened 2022-12-18
Add ADR template and metrics in testing ADR.

linabutler:
This PR:

1. Adds a template for our Architecture Decision Records, using the same format as Firefox Accounts. It's more structured than Nygard's [status-context-decision-consequences template](https://cognitect.com/blog/2011/11/15/documenting-architecture-decisions), but explicitly enumerates the alternatives considered and the pros and cons of each, which has been helpful in other projects (FxA, Application Services, UniFFI) that use this format. How do we feel about trying out this format?
2. Adds our first ADR for [DISCO-2089](https://mozilla-hub.atlassian.net/browse/DISCO-2089) and [DISCO-2090](https://mozilla-hub.atlassian.net/browse/DISCO-2090)!

### DIFF
--- a/docs/adr/0000-local-and-test-metrics.md
+++ b/docs/adr/0000-local-and-test-metrics.md
@@ -1,0 +1,108 @@
+# Reporting metrics in local development and tests
+
+* Status: Proposed
+* Deciders: raphael, lina
+* Date: 2022-12-16
+
+## Context and Problem Statement
+
+merino-py emits counts, histograms, and timing metrics as it runs. In production, these metrics are sent to Datadog, using an extension of the [StatsD](https://github.com/statsd/statsd) protocol called [DogStatsD](https://docs.datadoghq.com/developers/dogstatsd). For local development, merino-py provides the `metrics.dev_logger` config option, which logs outgoing metrics to the console instead of sending them to Datadog. merino-py also has integration tests that cover recording and emitting metrics. How can we avoid having the `metrics.dev_logger` setting and integration tests rely on implementation details of the `aiodogstatsd` client library to suppress reporting to Datadog?
+
+## Decision Drivers
+
+* Maintain test coverage for merino-py's metrics.
+* Make it easy for engineers to see what metrics are emitted while working on the code.
+* Avoid potential future compatibility issues when upgrading `aiodogstatsd`.
+
+## Considered Options
+
+* A. Do nothing.
+* B. Always send metrics to a live DogStatsD agent.
+* C. Change merino-py's `metrics.Client` proxy to not call into `aiodogstatsd`.
+* D. Subclass `aiodogstatsd.Client` to suppress sending metrics to DogStatsD.
+* E. Extend merino-py's `metrics` module with pluggable backends for production and testing.
+* F. Set up a mock DogStatsD endpoint that collects and checks metrics.
+* G. Ask the `aiodogstatsd` authors to expose a reporting hook as part of their public API.
+
+## Decision Outcome
+
+TODO: Fill me in once we decide!
+
+## Pros and Cons of the Options
+
+### A. Do nothing.
+
+This option would keep the code as it's written today:
+
+1. When the `metrics.dev_logger` config option is set, `metrics.configure_metrics()` sets the pseudo-private `aiodogstatsd.Client._protocol` property to an instance of `metrics._LocalDatagramLogger`. `_LocalDatagramLogger` subclasses `aiodogstatsd.client.DatagramProtocol` to log metrics instead of sending them to DogStatsD. ([DISCO-2089](https://mozilla-hub.atlassian.net/browse/DISCO-2089))
+2. merino-py's integration tests patch the pseudo-private `aiodogstatsd.Client._report()` method with a mock implementation. ([DISCO-2090](https://mozilla-hub.atlassian.net/browse/DISCO-2090))
+
+* Good, because this approach currently works!
+* Good, because this option doesn't require any extra work.
+* Good, because `aiodogstatsd` is in maintenance mode, and unlikely to break merino-py's usage in the future.
+* Bad, because relying on non-public APIs could break merino-py if `aiodogstatsd` does change its implementation in the future.
+* Bad, because merino-py's metrics reporting is tightly coupled to `aiodogstatsd`. Sending metrics to a different backend would require substantial changes to the tests.
+
+### B. Send metrics to a live DogStatsD agent.
+
+This option would remove `metrics.dev_logger`. Instead, developers would need to run a DogStatsD agent locally, and configure merino-py to use that agent.
+
+* Good, because there would be a single path for metrics reporting in local development, testing, and production.
+* Good, because this means less code in merino-py for metrics logging.
+* Good, because unofficial local DogStatsD agents exist, written in [Go](https://github.com/jonmorehouse/dogstatsd-local) and [Ruby](https://github.com/drish/dogstatsd-local).
+* Good, because local DogStatsD agents can log metrics in different formats.
+* Bad, because developers would need to configure and run a separate service when working locally.
+* Bad, because this approach requires an unofficial local DogStatsD agent. The [official DogStatsD agent](https://hub.docker.com/r/datadog/dogstatsd) is only offered as a Docker container that reports metrics to Datadog, and doesn't support custom or local-only backends. The plain StatsD agent, which supports [custom backends](https://github.com/statsd/statsd/blob/master/docs/backend.md), doesn't support the DogStatsD extensions (histograms) that merino-py uses.
+* Bad, because this is only suitable for local development. Setting up a live DogStatsD agent would be more cumbersome to do in the integration tests.
+* Bad, because this option assumes that merino-py will only ever report metrics to DogStatsD.
+
+### C. Change merino-py's `metrics.Client` proxy to not call into `aiodogstatsd` in testing.
+
+merino-py's `metrics.Client` class proxies calls to `aiodogstatsd.Client`. The proxy keeps track of all metrics calls for testing, and adds custom tags for feature flags. We can extend the proxy to only log metrics to the console if `metrics.dev_logger` is set, without forwarding them to the underlying `aiodogstatsd.Client`.
+
+* Good, because this needs minimal changes to support metrics logging for gauges, counters, histograms, and one-off timings.
+* Good, because proxying calls avoids relying on implementation details of `aiodogstatsd`.
+* Bad, because this approach doesn't work for `aiodogstatsd.Client.{timeit, timeit_task}()`, which return thunks that call `aiodogstatsd.Client.timing()`. Since the proxy can't intercept these calls, it would need to reimplement these methods.
+* Bad, because this option assumes that merino-py will only ever report metrics to DogStatsD.
+
+### D. Subclass `aiodogstatsd.Client` to suppress sending metrics to StatsD in tests.
+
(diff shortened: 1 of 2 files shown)


docs/adr/0000-local-and-test-metrics.md
```diff
@@ -0,0 +1,108 @@
...
+
+* Good, because this needs minimal changes to support metrics logging for gauges, counters, histograms, and one-off timings.
+* Good, because proxying calls avoids relying on implementation details of `aiodogstatsd`.
+* Bad, because this approach doesn't work for `aiodogstatsd.Client.{timeit, timeit_task}()`, which return thunks that call `aiodogstatsd.Client.timing()`. Since the proxy can't intercept these calls, it would need to reimplement these methods.
+* Bad, because this option assumes that merino-py will only ever report metrics to DogStatsD.
+
+### D. Subclass `aiodogstatsd.Client` to suppress sending metrics to StatsD in tests.
+
+This option would change `metrics.Client` to be a subclass of `aiodogstatsd.Client`, instead of proxying calls to it. The subclass would override the `{gauge, increment, decrement, histogram, distribution, timing}()` methods to log metrics if the f
```
ncloudioj, line 70: Oops, looks like the paragraph was truncated or unfinished.

docs/adr/0000-local-and-test-metrics.md
```diff
@@ -0,0 +1,108 @@
...
+* B. Always send metrics to a live DogStatsD agent.
+* C. Change merino-py's `metrics.Client` proxy to not call into `aiodogstatsd`.
+* D. Subclass `aiodogstatsd.Client` to suppress sending metrics to DogStatsD.
+* E. Extend merino-py's `metrics` module with pluggable backends for production and testing.
+* F. Set up a mock DogStatsD endpoint that collects and checks metrics.
+* G. Ask the `aiodogstatsd` authors to expose a reporting hook as part of their public API.
+
+## Decision Outcome
+
+TODO: Fill me in once we decide!
```
ncloudioj, line 29: Have we identified any winner(s) or favorite(s) yet? Or we will fill that in at a later time?

mtbottle: Another related question: are we using the PR as a discussion where we come to a decision? or is the intent that we have a discussion and decision made elsewhere, and this PR is purely for recording the already made decision? The reason I ask this is because the way I might go about reviewing such a PR will be different based on whether I am meant to discuss and sway a decision, vs a review on its quality as technical documentation.

docs/adr/template.md
```diff
@@ -0,0 +1,72 @@
...
+
+## Context and Problem Statement
+
+[Describe the context and problem statement, e.g., in free form using two to three sentences. You may want to articulate the problem in form of a question.]
+
+## Decision Drivers <!-- optional -->
+
+* [driver 1, e.g., a force, facing concern, …]
+* [driver 2, e.g., a force, facing concern, …]
+* … <!-- numbers of drivers can vary -->
```
mtbottle, line 17: Decision drivers can sometimes be ranked, as they aren't all equally as important. I'm wondering if there's a way we can capture it here. 🤔 (or maybe we can have it justified in `Decision Outcome`?)

docs/adr/template.md
```diff
@@ -0,0 +1,72 @@
...
+## Pros and Cons of the Options <!-- optional -->
+
+### [option 1]
+
+[example | description | pointer to more information | …] <!-- optional -->
+
+* Good, because [argument a]
+* Good, because [argument b]
+* Bad, because [argument c]
+* … <!-- numbers of pros and cons can vary -->
```
mtbottle, line 49: Personal preference nit: I would prefer that we format this with separate Pro/Cons list.

```
#### Pros:

* [argument a]
* [argument b]

#### Cons:

* [argument c]
```

I think the titles and visual separation helps me with scanning the document (quicker to find what I'm looking for). But if there's a standard that we're trying to adhere to, then that can override my personal preference!

docs/adr/0000-local-and-test-metrics.md
```diff
@@ -0,0 +1,108 @@
+# Reporting metrics in local development and tests
+
+* Status: Proposed
+* Deciders: raphael, lina
+* Date: 2022-12-16
+
+## Context and Problem Statement
+
+merino-py emits counts, histograms, and timing metrics as it runs. In production, these metrics are sent to Datadog, using an extension of the [StatsD](https://github.com/statsd/statsd) protocol called [DogStatsD](https://docs.datadoghq.com/developers/dogstatsd). For local development, merino-py provides the `metrics.dev_logger` config option, which logs outgoing metrics to the console instead of sending them to Datadog. merino-py also has integration tests that cover recording and emitting metrics. How can we avoid having the `metrics.dev_logger` setting and integration tests rely on implementation details of the `aiodogstatsd` client library to suppress reporting to Datadog?
```
mtbottle, line 9: To keep the documentation more maintainable, I would suggest breaking down paragraphs with [semantic linefeeds](https://rhodesmill.org/brandon/2012/one-sentence-per-line/).

mtbottle:
Thanks for starting this! 🌟 The first ADR is definitely an interesting topic!

docs/adr/0000-local-and-test-metrics.md
ddurst: I second this. It's a little disorienting between the RFC list and an ADR PR. I'm fine either way, but ADR in my mind is a decision to be recorded whereas an RFC is an idea to be discussed. We should make whatever decision consistent across App Svcs.

docs/adr/0000-local-and-test-metrics.md
linabutler: Oooh, I like that a lot—thank you!

docs/adr/template.md
linabutler: Let's do it! I copy-pasted this template from FxA/UniFFI, but think it's more important to have a structure that works for us, and I totally agree that it makes scanning a lot easier—and we can inspire changes the other way, too! 🚀

docs/adr/template.md
linabutler: Oooh, yes—maybe we can represent this as a numbered list, and expand the blurb in the outcome section about the relative importance of the decision drivers?

docs/adr/0000-local-and-test-metrics.md
linabutler: That's a great question—I'm curious what y'all think! 😄 Reading our [RFC process guidelines](https://mozilla-hub.atlassian.net/wiki/spaces/FJT/pages/11469735/RFC+Process+for+Firefox+Services), I got the impression that it's geared toward significant changes (new features, cross-team projects, design reviews, substantial refactors) seeking feedback from multiple stakeholders, and I wasn't sure that this counted as either.

The line between the RFC list and ADR PRs feels a little fuzzy to me, too—for example, @mhammond (hi Mark! 👋🏼) just put up https://github.com/mozilla/application-services/pull/5302, which is a draft ADR with a chosen solution sent to the RFC list for feedback.

For this one specifically, @hackebrot and I started a discussion on the [Jira ticket](https://mozilla-hub.atlassian.net/browse/DISCO-2089) to talk through some of these options, and we're meeting this week to chat more about them and make a decision. After that, we can update the PR with the outcome. Do we feel like that's a good approach for these kinds of smaller discussions, or would going through the RFC process be valuable—for example, if other folks want to chime in?

docs/adr/0000-local-and-test-metrics.md
linabutler: 😳 Oops, thanks for catching that!

docs/adr/0000-local-and-test-metrics.md
mhammond: I agree this is a bit messy - it's a classic "if the only tool you have is a hammer, everything looks like a nail". I did an ADR because app-services already has ADRs, but doesn't really have RFCs, and I didn't really want to invent a separate RFC process just for this. The ADR template is also a little too prescriptive IMO and doesn't really suit all such discussions, so my ADR kinda ignored the template a little. It was sent to the RFC list because that was suggested by karloff as the appropriate place to reach the other stakeholders for remote-settings in particular.

So yeah, my ADR was probably closer to an RFC :) I don't really think the distinction is particularly useful - in practice, someone writing an RFC intends making a decision, just like an ADR.

docs/adr/template.md
mtbottle: Ah a simple numbered list sound like a good start! I was originally thinking of maybe adding some sort of ranking system where you add a number (from 1-5, for instance) next to each point to indicate importance, but that might be more complicated than it's worth. We can change this later if we find that a numbered list doesn't exactly gives us the nuance that we need for the ranking.

docs/adr/0000-local-and-test-metrics.md
mtbottle: Thanks for all the feedback! What I'm understanding from what has been mentioned is that we have a way to solicit feedback on an idea (the RFC process), and that we can link the ADR PR for discussion or some other document. For this particular PR, discussion on the decision is being made in JIRA so this ADR is _not_ the place to discuss the decision. I think this process is fine, and I don't see a strong reason to change it. I think that ADR reviews tend to solicit another round of evaluating the decision, so knowing the process for decision making is important for not delaying actually having a decision made!

docs/adr/0000-local-and-test-metrics.md
```diff
@@ -0,0 +1,182 @@
+# Reporting metrics in local development and tests
+
+* Status: Proposed
+* Deciders: raphael, lina
+* Date: 2022-12-20
+
+## Context and Problem Statement
+
+merino-py emits counts, histograms, and timing metrics as it runs.
+In production, these metrics are sent to Datadog, using an extension of the
```
hackebrot, line 10: My current understanding is that we do not send merino-py metrics to Datadog.

The data flow is as follows:
1. merino-py uses aiodogstatsd to record and submit StatsD metrics
2. the aiodogstatsd client is configured to send metrics to a telegraf agent
3. the telegraf agent uses a StatsD Input and InfluxDB Output
4. the telegraf agent sends metrics to InfluxDB
5. Grafana queries InfluxDB and visualizes metrics

Does that sound accurate to you, @dlactin?

docs/adr/0000-local-and-test-metrics.md
```diff
@@ -0,0 +1,182 @@
...
+[StatsD](https://github.com/statsd/statsd) protocol called [DogStatsD](https://docs.datadoghq.com/developers/dogstatsd).
+For local development, merino-py provides the `metrics.dev_logger` config option,
+which logs outgoing metrics to the console instead of sending them to Datadog.
```
hackebrot, line 13: See above

docs/adr/0000-local-and-test-metrics.md
```diff
@@ -0,0 +1,182 @@
...
+merino-py also has integration tests that cover recording and emitting metrics.
+How can we avoid relying on implementation details of  the `aiodogstatsd` client library
+to suppress reporting to Datadog?
```
hackebrot, line 16:
```suggestion
to suppress reporting to InfluxDB?
```

hackebrot (commented):
Hi @linabutler! 👋🏻 

Thank you for creating this ADR! 🚀 

I've added a comment about the data flow of metrics in merino-py.

docs/adr/0000-local-and-test-metrics.md
quiiver: For context, we use the aiodogstatd package because it supports the Datadog extensions to the StatsD protocol which are enabled on telegraf.

PR #153, opened 2022-12-19
Suggest endpoint provider limitation on duplicate query params and client variants DISCO-2152

taddes:
Simple change to avoid potential DDoS where multiple instances of same provider are passed into the `providers` query param.  Wrapped in a set to eliminate duplicates. Didn't need additional filtering due to `if` check, which ensures the passed in `provider` is among those defined. Closes out one low severity finding in recent pen test. 

UPDATED: Added integration testing of api and also expanded functionality to limit number of `client_variants` that can be passed in to 5 to prevent low-risk reflection of client_variants. From the security assessment, it was found that because a list of `client_variants` would be constructed by splitting the `client_variant` query parameter, it was possible to pass in thousands of junk values or duplicates. This has now been limited to 5 and also filtered out.  If this is not desirable, we can modify the value, or have it set as a config value, as opposed to a hard coded limit. Is there a world in which we'd see more than 5 `client_variants`? Either way, I added testing to ensure only unique values return given duplicate `client_variants` and also that only a limit of values return.

### DIFF
--- a/merino/configs/default.toml
+++ b/merino/configs/default.toml
@@ -40,6 +40,9 @@ format = "mozlog"
 # Setting to contol the limit of optional client variants passed
 # to suggest endpoint as part of experiments or rollouts.
 client_variant_max = 25
+# Value passed into Query object for FastAPI query parameter validator.
+# Sets limitation on the maximum string length of a query.
+query_max_length = 100
 
 [default.metrics]
 dev_logger = false
--- a/merino/configs/default.toml
+++ b/merino/configs/default.toml
@@ -29,6 +29,9 @@ debug = false
 # timeout with the same name `query_timeout_sec`. See `accuweather` as an
 # example.
 query_timeout_sec = 0.2
+# Setting to contol the limit on optional client variants passed
+# to suggest endpoint as part of experiments or rollouts.
+client_variant_max = 25
 
 [default.logging]
 # Any of "DEBUG", "INFO", "WARNING", "ERROR", "CRITICAL"
--- a/merino/configs/default.toml
+++ b/merino/configs/default.toml
@@ -40,9 +40,10 @@ format = "mozlog"
 # Setting to contol the limit of optional client variants passed
 # to suggest endpoint as part of experiments or rollouts.
 client_variant_max = 25
-# Value passed into Query object for FastAPI query parameter validator.
+# Values passed into Query object for FastAPI query parameter validator.
 # Sets limitation on the maximum string length of a query.
-query_max_length = 100
+client_variant_character_max = 100
+query_character_max = 500
 
 [default.metrics]
 dev_logger = false
--- a/merino/configs/default.toml
+++ b/merino/configs/default.toml
@@ -39,7 +39,7 @@ format = "mozlog"
 [default.web.api.v1]
 # Setting to contol the limit of optional client variants passed
 # to suggest endpoint as part of experiments or rollouts.
-client_variant_max = 25
+client_variant_max = 10
 # Values passed into Query object for FastAPI query parameter validator.
 # Sets limitation on the maximum string length of a query.
 client_variant_character_max = 100
--- a/merino/configs/default.toml
+++ b/merino/configs/default.toml
@@ -29,15 +29,17 @@ debug = false
 # timeout with the same name `query_timeout_sec`. See `accuweather` as an
 # example.
 query_timeout_sec = 0.2
-# Setting to contol the limit on optional client variants passed
-# to suggest endpoint as part of experiments or rollouts.
-client_variant_max = 25
 
 [default.logging]
 # Any of "DEBUG", "INFO", "WARNING", "ERROR", "CRITICAL"
 level = "INFO"
 # Any of "mozlog" (i.e. JSON) or "pretty"
 format = "mozlog"
+
+[default.web.api.v1]
+# Setting to contol the limit of optional client variants passed
+# to suggest endpoint as part of experiments or rollouts.
+client_variant_max = 25
 
 [default.metrics]
 dev_logger = false
--- a/merino/web/api_v1.py
+++ b/merino/web/api_v1.py
@@ -32,6 +32,11 @@ SUGGEST_RESPONSE = {
 # Timeout for query tasks.
 QUERY_TIMEOUT_SEC = settings.runtime.query_timeout_sec
 
+# Client Variant Maximum - used to limit the number of
+# possible client variants for experiments.
+# See https://mozilla-services.github.io/merino/api.html#suggest
+CLIENT_VARIANT_MAX = settings.runtime.client_variant_max
+
 
 @router.get(
     "/suggest",
@@ -114,9 +119,9 @@ async def suggest(
     response = SuggestResponse(
         suggestions=suggestions,
         request_id=correlation_id.get(),
-        # client_variant restriction: set to remove duplicates, [0:10] to
+        # client_variant restriction: set to remove duplicates, [0:CLIENT_VARIANT_MAX] to
         # limit potential reflection back to client.
-        client_variants=list(set(client_variants.split(",")))[:5]
+        client_variants=list(set(client_variants.split(",")))[:CLIENT_VARIANT_MAX]
         if client_variants
         else [],
     )
--- a/merino/web/api_v1.py
+++ b/merino/web/api_v1.py
@@ -35,7 +35,7 @@ QUERY_TIMEOUT_SEC = settings.runtime.query_timeout_sec
 # Client Variant Maximum - used to limit the number of
 # possible client variants for experiments.
 # See https://mozilla-services.github.io/merino/api.html#suggest
-CLIENT_VARIANT_MAX = settings.runtime.client_variant_max
+CLIENT_VARIANT_MAX = settings.web.api.v1.client_variant_max
 
 
 @router.get(
--- a/merino/web/api_v1.py
+++ b/merino/web/api_v1.py
@@ -69,7 +69,10 @@ async def suggest(
     active_providers, default_providers = sources
     if providers is not None:
         search_from = [
-            active_providers[p] for p in providers.split(",") if p in active_providers
+            active_providers[p]
+            # Set used to filter out possible duplicate provider names.
+            for p in set(providers.split(","))
+            if p in active_providers
         ]
     else:
         search_from = default_providers
--- a/merino/web/api_v1.py
+++ b/merino/web/api_v1.py
@@ -6,7 +6,7 @@ from functools import partial
 from itertools import chain
 
 from asgi_correlation_id.context import correlation_id
-from fastapi import APIRouter, Depends
+from fastapi import APIRouter, Depends, Query
 from fastapi.encoders import jsonable_encoder
 from fastapi.responses import JSONResponse
 from starlette.requests import Request
@@ -36,6 +36,7 @@ QUERY_TIMEOUT_SEC = settings.runtime.query_timeout_sec
 # possible client variants for experiments.
 # See https://mozilla-services.github.io/merino/api.html#suggest
 CLIENT_VARIANT_MAX = settings.web.api.v1.client_variant_max
+QUERY_MAX_LENGTH = settings.web.api.v1.query_max_length
 
 
 @router.get(
@@ -46,7 +47,7 @@ CLIENT_VARIANT_MAX = settings.web.api.v1.client_variant_max
 )
 async def suggest(
     request: Request,
-    q: str,
+    q: str = Query(max_length=QUERY_MAX_LENGTH),
     providers: str | None = None,
     client_variants: str | None = None,
     sources: tuple[dict[str, BaseProvider], list[BaseProvider]] = Depends(
@@ -121,7 +122,10 @@ async def suggest(
         request_id=correlation_id.get(),
         # client_variant restriction: set to remove duplicates, [0:CLIENT_VARIANT_MAX] to
         # limit potential reflection back to client.
-        client_variants=list(set(client_variants.split(",")))[:CLIENT_VARIANT_MAX]
+        # [:CLIENT_VARIANT_MAX] filter at end to drop any trailing string.
+        client_variants=client_variants.split(",", maxsplit=CLIENT_VARIANT_MAX)[
+            :CLIENT_VARIANT_MAX
+        ]
         if client_variants
         else [],
     )
--- a/merino/web/api_v1.py
+++ b/merino/web/api_v1.py
@@ -70,7 +70,7 @@ async def suggest(
     if providers is not None:
         search_from = [
             active_providers[p]
-            # Set used to filter out possible duplicate provider names.
+            # Set used to filter out possible duplicate providers passed in.
             for p in set(providers.split(","))
             if p in active_providers
         ]
@@ -114,7 +114,11 @@ async def suggest(
     response = SuggestResponse(
         suggestions=suggestions,
         request_id=correlation_id.get(),
-        client_variants=client_variants.split(",") if client_variants else [],
+        # client_variant restriction: set to remove duplicates, [0:10] to
+        # limit potential reflection back to client.
+        client_variants=list(set(client_variants.split(",")))[:5]
+        if client_variants
+        else [],
     )
     return JSONResponse(content=jsonable_encoder(response))
 
--- a/merino/web/api_v1.py
+++ b/merino/web/api_v1.py
@@ -122,9 +122,7 @@ async def suggest(
     response = SuggestResponse(
         suggestions=suggestions,
         request_id=correlation_id.get(),
-        # client_variant restriction: set to remove duplicates, [0:CLIENT_VARIANT_MAX] to
-        # limit potential reflection back to client.
-        # [:CLIENT_VARIANT_MAX] filter at end to drop any trailing string.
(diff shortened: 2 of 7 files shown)


ncloudioj (changes requested):
Looks good. Can you also add a test for this?

taddes:
Certainly @ncloudioj ! A contract or integration test to ensure the api response is valid as we expect, in the case several operators of the same name are passed in? I know the DDoS side of things is more a load test and a measure of the load on the service. Thanks

ncloudioj:
I think an integration test for `tests/integration/api/v1/suggest/test_suggest.py` should be sufficient.

taddes:
Done @ncloudioj from updated PR comment:

UPDATED: Added integration testing of api and also expanded functionality to limit number of `client_variants` that can be passed in to 5 to prevent low-risk reflection of client_variants. From the security assessment, it was found that because a list of `client_variants` would be constructed by splitting the `client_variant` query parameter, it was possible to pass in thousands of junk values or duplicates. This has now been limited to 5 and also filtered out.  If this is not desirable, we can modify the value, or have it set as a config value, as opposed to a hard coded limit. Is there a world in which we'd see more than 5 `client_variants`? Either way, I added testing to ensure only unique values return given duplicate `client_variants` and also that only a limit of values return.

taddes:
Note: `#nosec` annotation was added as I was getting an inaccurate security warning from black, which warned me of using `random` in cryptographic/security implementations, which this is not. For context, this was the message: https://bandit.readthedocs.io/en/1.7.4/blacklists/blacklist_calls.html#b311-random

merino/web/api_v1.py
```diff
@@ -111,7 +114,11 @@ async def suggest(
     response = SuggestResponse(
         suggestions=suggestions,
         request_id=correlation_id.get(),
-        client_variants=client_variants.split(",") if client_variants else [],
+        # client_variant restriction: set to remove duplicates, [0:10] to
```
ncloudioj, line 117: Shouldn't it be `[0:5]`? Also, can you share the source of the 5 maximum client variants in Firefox?

tests/integration/api/v1/suggest/test_suggest.py
```diff
@@ -99,7 +100,61 @@ def test_client_variants(client: TestClient) -> None:
 
     result = response.json()
     assert len(result["suggestions"]) == 1
-    assert result["client_variants"] == ["foo", "bar"]
+    # Both the client_variants and test data are converted to sets to check membership,
+    # irrespective of the order because set() is used to remove duplicates.
+    assert set(result["client_variants"]) == set(["foo", "bar"])
+
+
+def test_client_variants_duplicated_variant(client: TestClient) -> None:
```
ncloudioj, line 108: Nice tests. Could you plz also add one for `providers`?

merino/web/api_v1.py
taddes: Yeah, good catch. I had it set to 10 originally but it seemed somewhat excessive. I couldn't find the actual source of the possible values for client variants, but given it is referenced in the urlbar preference docs (https://firefox-source-docs.mozilla.org/browser/urlbar/preferences.html ), it's mostly used for experimentation. I just thought 5 would be a sufficient limitation. Thoughts on if it should be more?

tests/integration/api/v1/suggest/test_suggest.py
taddes: Totally, will add! Also, you ok with the `set()` filtering implementation for removing duplicates? In the wild I do imagine some requests could potentially have differing order, so we'd not order to be tightly coupled. That said, if precedence of the possible `client_variants` matters, that is something that we could take into account. `set()` is just the fastest, cleanest way to filter. I could do other logic if we want to maintain the order.

ncloudioj: yeah, `set()` is fine because Neither Firefox nor Merino rely on the order of the variants.

merino/web/api_v1.py
ncloudioj: Unless it's enforced by the client (which is not at the moment), let's not cap those in Merino to avoid unexpected truncations. Deduping is good to be done here though :)

tests/integration/api/v1/suggest/test_suggest.py
taddes: gr8!

merino/web/api_v1.py
taddes: Sure, happy to remove. However, my only concern is a large scale reflection attack / request can still get around 8 responses per junk request (even though doesn't expose any data).  Can we still cap it, just at a higher number maybe? 100? I'd eather not leave it in the case it leaves the endpoint susceptible to undesirable results given an indefinite number of client variants. Lmk what you think and I'll go with it, then push that final `provider` test.

ncloudioj: Of course, go for it if you find it necessary (plz define a constant for this). My understanding is that setting the maximum variants alone won't be able to eliminate all the threats, for example, the attacker can send one variant with a massive number of characters, which still could drain the resources of Merino.

I think the better place to protect Merino from this kind of attack is on either LB or reverse proxies. For example, we can set the maximum URL size via [large_client_header_buffers](http://nginx.org/en/docs/http/ngx_http_core_module.html#large_client_header_buffers) for NGNIX.

taddes:
Revisions completed as requested. Added configurable default in both testing and default along with a provider test. Set to 25. Can easily change, good call on server configuration, let's do that as well. Is it SRE request dependent or can we mod those server details?

taddes:
Found an occasional bug in the contract tests when reviewing CI and the contract test code to see if client_variants were checked. I added sorting logic before the comparison, which is invoked in the wiki_fruit__apple_with_client_variants test.

taddes:
Hey @Trinaa ! I know it's still holiday times so please disregard until you have time to look, but I just thought I'd bring your attention to this one since it has a contract test. Just wanted to get your thoughts. I just call a sorting function and ensure the values inside the keys are sorted before comparison. This way, we keep the simpler assertion statement. I know generally HTTP requests will have preferred orders for params, but it is still possible things are received or possibly returned in orders we can't guarantee. Thanks so much!

Trinaa:
> Hey @Trinaa ! I know it's still holiday times so please disregard until you have time to look, but I just thought I'd bring your attention to this one since it has a contract test. Just wanted to get your thoughts. I just call a sorting function and ensure the values inside the keys are sorted before comparison. This way, we keep the simpler assertion statement. I know generally HTTP requests will have preferred orders for params, but it is still possible things are received or possibly returned in orders we can't guarantee. Thanks so much!

Looks okay to me!

merino/web/api_v1.py
```diff
@@ -111,7 +119,11 @@ async def suggest(
     response = SuggestResponse(
         suggestions=suggestions,
         request_id=correlation_id.get(),
-        client_variants=client_variants.split(",") if client_variants else [],
+        # client_variant restriction: set to remove duplicates, [0:CLIENT_VARIANT_MAX] to
+        # limit potential reflection back to client.
+        client_variants=list(set(client_variants.split(",")))[:CLIENT_VARIANT_MAX]
+        if client_variants
+        else [],
```
mtbottle, line 126: Thoughts on validating `client_variant`? It is a user input, so ideally, we would have some sort of feedback to the user that they've inputted too many `client_variants` with a 400 error.

I also noticed that we aren't _actually_ using the `client_variants` (yet?). So, currently it's only being input into the API, then just returned as a list again. If we were to use it in the API for processing, then I would recommend pulling the validation up to an earlier point so that we can return the `client_variants` that were actually used in processing.

taddes: Thanks for the valuable thoughts @mtbottle ! I agree, we should have some input feedback should the `client_variants` exceed some limit. This is probably something we can look at in addition to  the server config as @ncloudioj  had mentioned, putting a hard limitation on the text that is allowed.  I also see when I was working on this that the `client_variants` are essentially used for experiments and there is not a hard and fast use case we can easily define. 

 I think it may be valuable to table these features and concerns for another ticket so that we can address that processing earlier on, since there may be some desired feature that gets passed on that we aren't fully aware of.  Once we can have more clarity on that, we can likely better understand how the `client_variants` are used. Whether it is a doc or some subsequent tasks, we can have these down as something separate, since as it stands this remediates the immediate security concerns and all this additional `client_variant` modification is a nice to have. Besides, this was one of those 'informational disclosure' issues that wasn't actually dumping actually internal implementation details, just reflecting things back as you had said.

merino/configs/default.toml
```diff
@@ -29,6 +29,9 @@ debug = false
 # timeout with the same name `query_timeout_sec`. See `accuweather` as an
 # example.
 query_timeout_sec = 0.2
+# Setting to contol the limit on optional client variants passed
+# to suggest endpoint as part of experiments or rollouts.
+client_variant_max = 25
```
ncloudioj, line 34: Let's add an entry for this to the doc [here](https://github.com/mozilla-services/merino-py/blob/main/docs/ops.md#runtime-configurations).

ncloudioj: Can you also create a new config section for this as it doesn't sound like a runtime thing? How about `default.web.api.v1`?

merino/web/api_v1.py
ncloudioj: Those are good points. Let's file a follow-up to define the validation rules and equip all the API parameters with them. FastAPI comes with a bunch of [validators](https://fastapi.tiangolo.com/tutorial/query-params-str-validations/#additional-validation) that we can take advantage of.

merino/web/api_v1.py
```diff
@@ -111,7 +119,11 @@ async def suggest(
     response = SuggestResponse(
         suggestions=suggestions,
         request_id=correlation_id.get(),
-        client_variants=client_variants.split(",") if client_variants else [],
+        # client_variant restriction: set to remove duplicates, [0:CLIENT_VARIANT_MAX] to
+        # limit potential reflection back to client.
+        client_variants=list(set(client_variants.split(",")))[:CLIENT_VARIANT_MAX]
```
ncloudioj, line 124:
```suggestion
        client_variants=client_variants.split(",", maxsplit=CLIENT_VARIANT_MAX))[:-1]  # `[:-1]` to drop the trailing string
```
`str.split()` takes a second argument `maxsplit` to specify that at most `maxsplit` splits are done, i.e. the list will have at most `maxsplit+1` elements. Also, I don't think we need to check for duplicates here as the API spec demands that all variants should be unique. Duplicates can only be caused by either program errors or malicious requests, which should be handled by `CLIENT_VARIANT_MAX`. What do you think?

taddes: Solid recommendation, I had considered the kwarg `maxsplit` as I just worked with it on another task. 

Since this recommended implementation would require `[:CLIENT_VARIANT_MAX]` tacked on anyways as an additional filter to the list, using it both in `maxsplit` and then on the list afterward I think decreases readability.  Also, `maxsplit` would already place a limitation on the number of elements so calling that after wouldn't have an effect on individual trailing str, just the number of elements in the list.

 I am also with you that only malicious requests or errors could result in duplicates being passed, but I opted for the option to apply the same logic to filter them as for providers. Was thinking in terms of defense in depth for this, though the risk is very very low.`CLIENT_VARIANT_MAX` will just set the minimum number of possibilities, but does not take into consideration duplicates, or am I confused? What do you think the advantages and disadvantages are to leaving the list(set()) explicit coercion followed by filtering the list length? Happy to do whatever.

merino/configs/default.toml
taddes: Great, very good rec. I will add a section as `default.web.api.v1`. Does that translate to `MERINO_WEB_API_V1__CLIENT_VARIANT_MAX`?

ncloudioj: I think it should be `MERINO_WEB__API__V1__CLIENT_VARIANT_MAX`, each `.` (except for the one after `default`) needs a double underscore as a level separator.

merino/web/api_v1.py
ncloudioj: I think using `maxsplit` here could at least save a list creation if the `client_variants` contains an excessive amount of CSVs. Yeah, that `[:CLIENT_VARIANT_MAX]` is ugly, but without it, the trailing str (if any) will be included. Do you have better way to handle that?

ncloudioj: > CLIENT_VARIANT_MAX will just set the minimum number of possibilities, but does not take into consideration duplicates, or am I confused?

Correct, my comment completely ignored duplicates.

> What do you think the advantages and disadvantages are to leaving the list(set()) explicit coercion followed by filtering the list length?

So if don't care about duplicates, then we don't need that `set()`, also `str.split()` already returns a list so we don't need the outer `list()` either. Lastly, if we just do `str.split(",", maxsplit=CLIENT_VARIANT_MAX)`, given a string with more than CLIENT_VARIANT_MAX variants, it will return the trailing str as the last item of the list (i.e. for `1,2,3, ..., {CLIENT_VARIANT_MAX}, ..., {CLIENT_VARIANT_MAX+N}`, it returns `['1', '2', '3', ... 'CLIENT_VARIANT_MAX', 'CLIENT_VARIANT_MAX+1,...CLIENT_VARIANT_MAX+N]`). But the point of having this constant is to grab up to `CLIENT_VARIANT_MAX` of variants, so that's why I put the `[:CLIENT_VARIANT_MAX]` bit for that. Does that make sense?

ncloudioj: I am still not convinced that `list(set(client_variants.split(",")))[:CLIENT_VARIANT_MAX]` is the most effective way to guard against those malformed values.

How about we leave it as is for now and use another followup to handle API parameter validations as I mentioned in [here](https://github.com/mozilla-services/merino-py/pull/153/files#r1060950041)? For example, if we apply a max string length to `client_variants`, then we don't need to limit the variant number anymore. Let me know what you think.

taddes: Ach @ncloudioj , yeah agreed, wasn't ready for full review there as I was looking at some alternatives to that which I would push. I was intending, baring any additional insights or approaches, to go with what you recommend, which looks like what I will do now as I couldn't think of something better.

Thanks for the clarification on API parameter validations, I just thought you were referring to the Dynaconf Validators for the defined `client_variants_max` which I added, which too is useful!

 I'll get this adjusted as you recommend. Thank you!

taddes:
Added the `Query` validator for basic query and am also adding this limitation to the `client_variant`, however it broke a lot of tests dealing with timeouts, so just addressing that regression to see why. Throws a 400, which makes sense. Will update.

taddes:
@ncloudioj made corrections:
- Figured out the test regression caused when I applied the minimum count to `client_variants`, which was that it needed a `default = None` setting.
- Separated out the two validators and defined as constants when used, one for the query string itself (set to 500 chars, thoughts?) and 100 chars for client variants. Tests reflect this. I can make whatever tweaks you think are ideal given our api spec.  Counts seemed reasonable given the use of each parameter.
- Updated testing to account for this maximums so both query param and client_variant param are accounted for
- Updated docs.

merino/web/api_v1.py
```diff
@@ -111,7 +122,14 @@ async def suggest(
     response = SuggestResponse(
         suggestions=suggestions,
         request_id=correlation_id.get(),
-        client_variants=client_variants.split(",") if client_variants else [],
+        # client_variant restriction: set to remove duplicates, [0:CLIENT_VARIANT_MAX] to
```
ncloudioj, line 125: Nit: please update the comment as we no longer dedup here.

As an aside, now that we have a max character limit for `client_variants`, do we still need to cap the client variants here? I am inclined to drop it as that check feels somewhat redundant to me. For example, if the max length of `client_variant` is 100, one can only pack up to 50ish variants to it as there will be 49 commas in the string. @taddes I will defer this to you :)

tests/contract/client/tests/test_merino.py
```diff
@@ -160,7 +160,14 @@ def assert_200_response(
     """Check that the content for a 200 OK response is what we expect."""
     expected_content_dict = step_content.dict(exclude=CONTENT_EXCLUDE)
     merino_content_dict = merino_content.dict(exclude=CONTENT_EXCLUDE)
-    assert expected_content_dict == merino_content_dict
+
```
ncloudioj, line 163: Is this still necessary as the deduping is dropped?

ncloudioj (approved):
LGTM, just a couple of nits for your consideration. 

Thanks for hardening the suggest API! 🙌

tests/contract/client/tests/test_merino.py
taddes: Yeah, not necessary. I was gonna leave it as the order can be reversed, but since this is a contract test and now only a malformed request would trigger this, I can remove the ordering logic.

merino/web/api_v1.py
taddes: Yeah good point for sure and it most cases the cap on the parameter length will throw a 400 which will cover us for any serious invalid request.  I do kind of like the added layer there as it does do a little bit of extra input validation and cleaning, especially in the cases where a short enough request could still have a lot of values. Since I added some extra tests for it and it isn't causing any problems, I think we can leave it and down the road remove it if need be. I've also lowered the threshold to 10 for `client_variants`, which probably makes sense. Thank you so much @ncloudioj !

PR #154, opened 2022-12-20
[DISCO-2147] refactor: use composition for fake providers in int tests

hackebrot:
# Description

This pull request refactors the fake providers we use in integration tests to use composition instead of inheritance. This architecture is optimized for comprehensibility and long-term maintainability.

For additional context, please see https://github.com/mozilla-services/merino-py/pull/140#discussion_r1046977863.

# Issue(s)
[DISCO-2147](https://mozilla-hub.atlassian.net/browse/DISCO-2147)

### DIFF
--- a/tests/integration/api/v1/fake_providers.py
+++ b/tests/integration/api/v1/fake_providers.py
@@ -95,6 +95,17 @@ async def raise_error(srequest: SuggestionRequest) -> list[BaseSuggestion]:
     raise RuntimeError(srequest.query)
 
 
+def add_latency(query_callable: QueryCallable, *, latency: float) -> QueryCallable:
+    """Add latency to a QueryCallable."""
+
+    async def latent_call(srequest: SuggestionRequest) -> list[BaseSuggestion]:
+        """Sleep before running the query callable."""
+        await asyncio.sleep(latency)
+        return await query_callable(srequest=srequest)
+
+    return latent_call
+
+
 class FakeProvider(BaseProvider):
     """Fake provider for integration tests."""
 
@@ -106,7 +117,6 @@ class FakeProvider(BaseProvider):
         hidden: bool,
         query_callable: QueryCallable,
         query_timeout_sec: float = settings.runtime.query_timeout_sec,
-        sleep_before_sec: float = 0.0,
     ) -> None:
         super().__init__()
         self._name = name
@@ -114,7 +124,6 @@ class FakeProvider(BaseProvider):
         self._query_callable = query_callable
         self._enabled_by_default = enabled_by_default
         self._query_timeout_sec = query_timeout_sec
-        self._sleep_before_sec = sleep_before_sec
 
     async def initialize(self) -> None:
         """Initialize method for the fake provider."""
@@ -122,9 +131,6 @@ class FakeProvider(BaseProvider):
 
     async def query(self, srequest: SuggestionRequest) -> list[BaseSuggestion]:
         """Run the query callable for this fake provider."""
-        if self._sleep_before_sec:
-            await asyncio.sleep(self._sleep_before_sec)
-
         return await self._query_callable(srequest=srequest)
 
     def hidden(self) -> bool:
@@ -192,8 +198,10 @@ class FakeProviderFactory:
             name=provider_name,
             enabled_by_default=enabled_by_default,
             hidden=False,
-            query_callable=query_sponsored(provider_name),
-            sleep_before_sec=settings.runtime.query_timeout_sec * 2,
+            query_callable=add_latency(
+                query_sponsored(provider_name),
+                latency=settings.runtime.query_timeout_sec * 2,
+            ),
         )
 
     @staticmethod
@@ -206,7 +214,9 @@ class FakeProviderFactory:
             name=provider_name,
             enabled_by_default=enabled_by_default,
             hidden=False,
-            query_callable=query_sponsored(provider_name),
-            sleep_before_sec=settings.runtime.query_timeout_sec * 2,
+            query_callable=add_latency(
+                query_sponsored(provider_name),
+                latency=settings.runtime.query_timeout_sec * 2,
+            ),
             query_timeout_sec=settings.runtime.query_timeout_sec * 4,
         )
--- a/tests/integration/api/v1/fake_providers.py
+++ b/tests/integration/api/v1/fake_providers.py
@@ -5,54 +5,19 @@
 """Provider fakes for the v1 integration test modules."""
 
 import asyncio
+from typing import Protocol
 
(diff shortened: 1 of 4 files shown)


ncloudioj:
@hackebrot Thanks for sharing your solution to simplify those fake providers used for testing!

I have a few questions and would like to hear your thoughts on those.

- Now that we have a single fake provider, `FakeProvider`,  with various configurable handles to customize its behaviours within the `FakeProvider.query()` function, how should we expand it for other future test scenarios? For example, if we want to mock a sponsored provider with rate limiting, i.e. a provider would time out queries if it gets called more than X times within a timeframe. Implementing that whole logic inside of `FakeProvider.query()` seems irrelevant for other fake providers. Will we create another fake provider class for this instead?
- The other question I have, which is probably a non-issue, is that it might confuse folks when they realize that we're using inheritance to implement regular providers for Merino, whereas we're using provider factory to do the same for testing. Which approach should I use?

mtbottle (commented):
I really like this approach!

The only concern I have is that `FakeProvider` is now taking on the brunt of all the configuration changes for different scenarios that we would want to use. The complexity of all these configurations will become another maintainability issue if we're not careful and selective about what configurations actually belongs in `FakeProvider`.

I've proposed offloading some configurations to the `query_callable` (in a comment below), but this might not be super intuitive.

tests/integration/api/v1/fake_providers.py
```diff
@@ -144,27 +87,126 @@ async def query(self, srequest: SuggestionRequest) -> list[BaseSuggestion]:
...
-        self._name = "timedout-tolerant-sponsored"
-        # It can tolerate for 4x of the default query timeout
-        self._query_timeout_sec = settings.runtime.query_timeout_sec * 4
+    @property
+    def enabled_by_default(self) -> bool:
+        """Return a boolean indicating whether the provider is enabled."""
+        return self._enabled_by_default
+
+
+class ProviderFactory:
```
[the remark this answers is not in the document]
hackebrot: Renamed `ProviderFactory` to `FakeProviderFactory`. 👍🏻

tests/integration/api/v1/fake_providers.py
```diff
@@ -144,27 +87,126 @@ async def query(self, srequest: SuggestionRequest) -> list[BaseSuggestion]:
...
+        query_timeout_sec: float = settings.runtime.query_timeout_sec,
+        sleep_before_sec: float = 0.0,
+    ) -> None:
+        super().__init__()
+        self._name = name
+        self._hidden = hidden
+        self._query_callable = query_callable
+        self._enabled_by_default = enabled_by_default
+        self._query_timeout_sec = query_timeout_sec
+        self._sleep_before_sec = sleep_before_sec
```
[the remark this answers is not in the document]
hackebrot: You're not wrong! I agree that the latency really is a property of the query rather than the fake provider.

That reminds me a lot of how I would implement something like that in Go:
https://github.com/hackebrot/turtle/blob/30f91c6aef1da02d854012060eb6a7e450167738/cmd/turtle/writer.go#L16-L27

tests/integration/api/v1/fake_providers.py
```diff
@@ -144,27 +87,136 @@ async def query(self, srequest: SuggestionRequest) -> list[BaseSuggestion]:
...
-        self._name = "timedout-tolerant-sponsored"
-        # It can tolerate for 4x of the default query timeout
-        self._query_timeout_sec = settings.runtime.query_timeout_sec * 4
+    @property
+    def enabled_by_default(self) -> bool:
+        """Return a boolean indicating whether the provider is enabled."""
+        return self._enabled_by_default
+
+
+class FakeProviderFactory:
```
Trinaa, line 146: @hackebrot I know you were questioning the need for the factory. Seeing it here, I like the organization it brings.

Trinaa (approved):
I think this is a step in the right direction (composition vs inheritance).

hackebrot:
Hi @ncloudioj! 👋🏻 

We discussed some of this already on Slack, but I wanted to comment here as well to close the loop.

We would **not** be adding a new class and inherit from `FakeProvider` to overwrite the existing `query` implementation. Instead we would develop new wrappers for the `QueryCallable` objects similar to `add_latency` in this pull request.

You bring up a very good point there! When should we use composition and when should we use inheritance. I think it's OK to use inheritance for code-reuse, but we need to pay attention to how many levels deep we go.

Currently all providers inherit directly from `BaseProvider`, they all exist in their own module, and they don't rely on shared backends or duplicate a lot of functionality. That's still manageable. Providers in integration tests need to be much more flexible since we want test a variety of different scenarios.

```text
BaseProvider
├── accuweather.Provider
├── adm.Provider
├── top_picks.Provider
└── wiki_fruit.WikiFruitProvider
```

The providers for integration tests use many more levels:

```text
BaseProvider
├── CorruptProvider
├── HiddenProvider
├── NonsponsoredProvider
└── SponsoredProvider
    └── TimeoutSponsoredProvider
        └── TimeoutTolerantSponsoredProvider
```

My recommendation is to merge the refactoring in this pull request and start a separate conversation about composition vs. inheritance for providers in general. What do you think?

tests/integration/api/v1/fake_providers.py
```diff
@@ -5,55 +5,20 @@
...
-class HiddenProvider(BaseProvider):
-    """A provider fake intended for test with a 'hidden' property set to return True."""
-
-    def __init__(self, enabled_by_default) -> None:
-        self._enabled_by_default = enabled_by_default
-        self._name = "hidden"
-
-    async def initialize(self) -> None:
-        """Initialize method for the HiddenProvider."""
+    async def __call__(self, srequest: SuggestionRequest) -> list[BaseSuggestion]:
```
ncloudioj, line 18: I think it will be useful to pass the provider in here so that the callable could use it if needed, but we can do that in a follow-up.

tests/integration/api/v1/fake_providers.py
```diff
@@ -144,27 +87,136 @@ async def query(self, srequest: SuggestionRequest) -> list[BaseSuggestion]:
...
+                query_sponsored(provider_name),
+                latency=settings.runtime.query_timeout_sec * 2,
+            ),
+        )
+
+    @staticmethod
+    def timeout_tolerant_sponsored(
+        enabled_by_default: bool = True,
+    ) -> FakeProvider:
+        """Return a new sponsored fake provider with a higher timeout that sleeps."""
```
ncloudioj, line 211:
```suggestion
        """Return a new sponsored fake provider with its own custom query timeout."""
```

ncloudioj (approved):

Issue #82, opened 2022-10-20
Encapsulate AccuWeather API communication in backend object


data-sync-user:
In the AccuWeather provider, extract the `httpx.AsyncClient` object creation and usage to a new backend object and inject backend object as a dependency, in order to:

* Have structural cohesion between providers, such as the ADM Provider
* Simplify future maintenance with clear delineation of communication interface
* Facilitate unit and integration testing 

**Background**
Injection of 3rd party communications is already achieved with the AdmProvider and can be used as an example.

merino/providers/__init__.py

!image-20221020-144014.png|width=528,height=351!

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/DISCO-2057)
┆Attachments: <a href="https://mozilla-hub.atlassian.net/rest/api/2/attachment/content/29906">image-20221020-144014.png</a>
┆Epic: New suggestion provider: AccuWeather (Zero-prefix)

data-sync-user:
➤ Katrina Anderson commented:

Update (2022/12/21)
Status: On track 
Blockers: None
Notes: Completed first pass of prod code refactor, next step is to refactor and add tests. Branch ( https://github.com/mozilla-services/merino-py/tree/DISCO-2057_accuweather_backend ).

data-sync-user:
➤ Katrina Anderson commented:

Update (2022/12/28)
Status: On track 
Blockers: None
Notes: 

  * Refactor assignment of settings to Provider initialization
  * Add Value checks for Accuweather settings

data-sync-user:
➤ Katrina Anderson commented:

Update (2022/12/29)
Status: On track 
Blockers: None
Notes:

  * Add unit tests for Temperature
  * Opened PR ( https://github.com/mozilla-services/merino-py/pull/156 )
  * Set status to 'In Review'

PR #156, opened 2022-12-29
[DISCO-2057] Encapsulate AccuWeather API communication in backend object

Trinaa:
# Description
Extract the AccuWeather API communications into a backend object in order to have structural cohesion between providers, simplify future maintenance with clear delineation of AccuWeather API interface and facilitate unit and integration testing. 

**Changes include:**
* Redefining the AccuWeather provider as the weather provider, which uses a `WeatherBackend` protocol to describe behaviour from partners.
* Refactoring the AccuWeather API logic into a backend object.
* Injecting the AccuWeather backend into the weather Provider
* Injecting `merino.config.settings` into `Provider` and `Accuweather` from a central location, `providers\__init__.py`
* Grouping `remotesettings.py` and `accuweather.py` with their corresponding providers in directories
* Refactoring existing and adding unit tests
* Adding integration tests

**Notes:**
* Renaming in [c93753a](https://github.com/mozilla-services/merino-py/pull/156/commits/c93753aeba0de3487705999dcd67605e976374b9) exaggerates scope of changes, I suggest reviewing from [e4bfdee](https://github.com/mozilla-services/merino-py/pull/156/commits/e4bfdee20719bbe3f38ab17003ef47a7073d4de1)
* Overall test coverage has increased from 95% to 97%. 
* Contract testing task remains [DISCO-2032](https://mozilla-hub.atlassian.net/browse/DISCO-2032)
* Follow-up Items created: [DISCO-2174](https://mozilla-hub.atlassian.net/browse/DISCO-2174) & [DISCO-2175](https://mozilla-hub.atlassian.net/browse/DISCO-2175)

# Issue(s)
#82 | [DISCO-2057](https://mozilla-hub.atlassian.net/browse/DISCO-2057)

[DISCO-2032]: https://mozilla-hub.atlassian.net/browse/DISCO-2032?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


Trinaa:
@0c0w3 I'd appreciate if you could take a look at these changes, in case I've introduced an incompatible behaviour. 

In particular I created a departure by reducing the scope of the try/except context and by changing the broad `Exception` clause to`HTTPError` (I understand this to be the top level exception from `AsyncClient`).

tests/unit/providers/test_weather.py
```diff
@@ -0,0 +1,194 @@
...
+        query_timeout_sec=0.2,
+    )
+
+
+def test_enabled_by_default(provider: Provider) -> None:
+    """Test for the enabled_by_default method."""
+    assert provider.enabled_by_default is False
+
+
+def test_hidden(provider: Provider) -> None:
```
mtbottle, line 63: nit: `test_not_hidden_by_default` to be consistent with L58

merino/providers/weather.py
```diff
@@ -0,0 +1,130 @@
...
+    high: Temperature
+    low: Temperature
+
+
+class WeatherReport(BaseModel):
+    """Model for weather conditions."""
+
+    city_name: str
+    current_conditions: CurrentConditions
+    forecast: Forecast
```
mtbottle, line 51: Hmm... the tricky thing with refactoring something to make it general is that we might end up generalizing stuff prematurely (and end up setting up a standard that is too specific to the way it's built right now and  hard to change later). Are we sure that these objects are the _right_ ones that we want to generalize for the `WeatherProvider`? I'm looking at the `url`, and feel like that field (for both forecast and current conditions) is very specific to Accuweather.

mtbottle (commented):
Code generally LGTM! I also like the approach here to decouple the Accuweather calls from the provider. I just had a few naming nits and a question about some of the generalizations.

tests/unit/providers/test_weather.py
Trinaa: Change applied: [5257292](https://github.com/mozilla-services/merino-py/pull/156/commits/5257292496516b82143fc78c4abd6156e14b8a81)

merino/providers/weather.py
Trinaa: I'll poke @ncloudioj and @0c0w3 for this one, because I don't have context on suggestion design. The `CurrentConditions` model and `Forecast` model are part of the `Suggestion` model definition in `merino.providers.weather`, so I'm assuming Firefox is dependent on this scheme?

ncloudioj:
Now I think I know better about the confusion around backends.

I feel like `backends` should be part of the `providers` module, actually, I was thinking about having this layout instead:

```
|_ providers
  |_ weather
    |_ provider.py
    |_ backend.py
  |_ adm
    |_ provider.py
    |_ backend.py
  |_ wikipedia
    |_ provider.py
    |_ backend.py
  |_ ...
```
Apparently, `backend.py` can be optional for certain providers. What do you think? @Trinaa @mtbottle

merino/providers/__init__.py
```diff
@@ -7,13 +7,14 @@
 from merino import metrics
 from merino.config import settings
 from merino.exceptions import InvalidProviderError
-from merino.providers.accuweather import Provider as AccuWeatherProvider
-from merino.providers.adm import Provider as AdmProvider
-from merino.providers.adm import TestBackend
+from merino.providers.adm.adm import Provider as AdmProvider
+from merino.providers.adm.adm import TestBackend
```
mtbottle, line 11: Rename the `providers/adm/adm.py` -> `providers/adm/provider.py` (and split backend?) for consistency?

merino/providers/weather/backends/accuweather.py
```diff
@@ -0,0 +1,243 @@
...
+    localized_name: str
+
+
+class AccuweatherError(BackendError):
+    """Error during interaction with the AccuWeather API."""
+
+
+class AccuweatherBackend:
+    """Backend that connects to the AccuWeather API."""
+
```
ncloudioj, line 34: Let's declare all the instance variables here.

merino/providers/weather/provider.py
```diff
@@ -0,0 +1,130 @@
...
+
+    async def get_weather_report(
+        self, geolocation: Location
+    ) -> Optional[WeatherReport]:  # pragma: no cover
+        """Get weather information from partner.
+
+        Raises:
+            BackendError: Category of error specific to provider backends.
+        """
+        ...
```
mtbottle, line 78: I think we discussed moving this protocol into `backend`. I think it would be good to have it there as it's the interface for the backends for this provider.

merino/providers/weather/backends/accuweather.py
```diff
@@ -0,0 +1,243 @@
...
+        async with AsyncClient(base_url=self.url_base) as client:
+            if not (location := await self.get_location(client, country, postal_code)):
+                return None
+
+            if not (current := await self.get_current_conditions(client, location.key)):
+                return None
+
+            if not (forecast := await self.get_forecast(client, location.key)):
+                return None
```
ncloudioj, line 92: Once the location key is fetched, we can issue two tasks to fetch the current condition and the forecast concurrently, which might speed up this. Also, since we're now at 3.11, we can leverage [asyncio.TaskGroups](https://docs.python.org/3/library/asyncio-task.html#task-groups) for this. Can be done in a follow up though as it's unrelated to this PR.

merino/providers/weather/provider.py
mtbottle: If we move the protocol, this would also mean that a couple of the models above should probably move as well.

merino/providers/weather/provider.py
```diff
@@ -0,0 +1,130 @@
...
+    current_conditions: CurrentConditions
+    forecast: Forecast
+
+
+class Suggestion(BaseSuggestion):
+    """Model for weather suggestions."""
+
+    city_name: str
+    current_conditions: CurrentConditions
+    forecast: Forecast
```
mtbottle, line 59: Nothing we have to do for this PR, but I think that sharing the `Suggestion` object with models that are used in the `Backend` (ie. `CurrentConditions` and `Forecast`) is going to implicitly make a tight coupling with the `provider` and `backend`. If we want to go for proper loose coupling, then we'll likely want to create separate models for 1. what is exposed through our APIs and 2. how the data is passed from the `backend` to `provider`. Anyways, at this point, I don't think the tight coupling is a big issue as we don't have a concrete case for extension (and given the scope, can be more easily refactorable).

merino/providers/weather/provider.py
```diff
@@ -0,0 +1,130 @@
...
+
+
+class WeatherBackend(Protocol):
```
ncloudioj, line 62: I wonder if the reason for defining the protocol inside the provider is to avoid the cyclic import, no? It feels more natural to me to define protocols in `backends/__init__.py` for each provider.

merino/providers/__init__.py
Trinaa: I changed the module name and created a task to do the subsequent changes, see [DISCO-2174](https://mozilla-hub.atlassian.net/browse/DISCO-2174) (please modify as you see fit).
Change applied: [73b3566](https://github.com/mozilla-services/merino-py/pull/156/commits/73b356666017d40ce5965534267c9ce090d514a8)

merino/providers/weather/backends/accuweather.py
Trinaa: Change applied: [e556ced](https://github.com/mozilla-services/merino-py/pull/156/commits/e556cedea21945a1392d47fc1468f89e61ee621f)

merino/providers/weather/backends/accuweather.py
Trinaa: Great ideas! I created a follow-up item [DISCO-2175](https://mozilla-hub.atlassian.net/browse/DISCO-2175) (please modify as you see fit)

merino/providers/weather/provider.py
Trinaa: I see that @ncloudioj created the task [DISCO-2031; Revisit the response schema for the suggest endpoint](https://mozilla-hub.atlassian.net/browse/DISCO-2031), perhaps this is a good place to document your thoughts?

merino/providers/weather/provider.py
Trinaa: I opted to move the protocol and it's related models into `weather\backends\protocol.py`. This way we avoid the circular reference and the protocol location has higher visiblity (and can easily map to the unit tests 😸).
Change applied: [0aa5f7f](https://github.com/mozilla-services/merino-py/pull/156/commits/0aa5f7f604c7e0a6f93f2a57ea2b43e0b319b9fb)

Let me know if that works?

Note: Relates to @mtbottle's comment above https://github.com/mozilla-services/merino-py/pull/156#discussion_r1061825185

merino/providers/weather/provider.py
ncloudioj: Yeah, we need to find a better way to organize/home those models. Some of them are defined along with providers, the others are located in `protocols.py` which could cause confusion and other trouble over time. Let's handle that in a separate refactor.

ncloudioj (approved):
Nice refactor. Also, thanks for cranking up the test coverage! 📈

0c0w3 (approved):
Thanks, lgtm overall! I'll defer to Nan and others for details.

mtbottle (approved):
🚀

PR #162, opened 2023-01-03
[Disco-2135] feat use git revision as sentry release tag

taddes:
Sentry events should include the version of merino-py they were emitted from for diagnosing purposes. It is best practice to use the git SHA-1 hash.  One can view this hash that matches the current HEAD by typing `git rev-parse HEAD` or by accessing the SHA has in the `.git/refs/heads` file that matches the current branch.

Sentry’s cli tool can set up releases to the sentry instance directly but it’s simpler to have merino-py fetch the current git SHA1 itself and set it as the `release` argument to sentry_sdk.init(). Investigating the cli version would result in a lot of overhead.  

After reviewing the various snippets and options available -see [here](https://forum.sentry.io/t/fetch-git-sha-equivalent-in-the-unified-python-sdk/5521/2) e.g.), the option to make use of the `raven` SDK as a model to build our own was the best option, as this would work outside of stage/prod, such as during local development when pointing to the merino-local sentry project. It also prevents the resource leakage that could come as a result of using the `gitpython` module.  

The implementation differs here in that I have split the functionality into 3 discrete functions:
-`read_git_head_file` first reads the current HEAD, validates that the root of the project is a valid git repo and returns a file reference to where the SHA hash is contained, stripping out extraneous values.
- `check_packed_refs` handles a possible edge case after git has run garbage collection. The note in the code explains this, but explanation and matching documentation is presented here: 
    In the case git has run "auto gc" (garbage collection), loose objects could be moved into .git/packed-refs. This is where branch references could be moved, making it important to check if the hash is stored in packed-refs. Git automatically writes and updates references in refs/heads, but if that reference is not there packed-refs is checked. See Git Docs [here](https://git-scm.com/book/en/v2/Git-Internals-Maintenance-and-Data-Recovery)
- `fetch_git_sha` is the central function of the module that calls 'read_git_head_file`, then calls `check_packed_refs` to check if the SHA is contained there and then finally reads and captures the SHA-1 hash at the correct path, returning the string of the HEAD reference that is passed in as the Sentry Release Tag.

A few notes regarding testing and design:
- I opted to change the structure from the internal Sentry raven implementation since it was old-school Python 2 and contained a lot of logic in a single function, which is hard to test and hard to follow.  Separating out concerns in this way makes for easier testing, however testing presents a few issues I discuss below. I'd hate to break this out further just because it would then slide into equally unreadable territory if there is too much decoupling.  I opted for a central function that calls the initial check and construction of a path and calls  a function to check an edge case, finally then resolving the product of the two. 
- In review, I would like to discuss a few more ideas in mocking and testing this. It is easy enough to validate the correct files by calling the `subprocess` module to use the shell to read the files to test against.  However, it is harder to concoct a situation where the `packed_ref` path takes place. 
- I added some `# nosec` annotations as there are risks to the use of the `subprocess` module, however we are not taking unsanitized input but rather passing an explicit command to verify in testing, not the actual initialization of the merino service.  It is for this reason too that I did not simply call `subprocess.call()` to use git to get this value as that would be riskier in the instantiation of the service, though that would be nice as it would be a one-liner!


ncloudioj:
Since Merino runs as a container in production, the repo metadata won't be available there.

I think we can grab the commit hash from the packaged `version.json` (produced by [CI upon image creation](https://github.com/mozilla-services/merino-py/blob/main/.circleci/config.yml#L309-L316)). The `commit` field is what we can use. Here is an [example](https://github.com/mozilla-services/merino-py/blob/main/merino/web/dockerflow.py#L21-L26) of how we use it in Merino.

taddes:
Ah dang, I did not know the `.git` data won't be there, thought the container would include this.  Too bad, I can change the code to read the `version.json` then. Regardless, glad I got this experience.  Keeping you looped in on this @pjenvey , guess we have to do the `version.json`.

pjenvey:
Oops, sorry about that @taddes and good catch @ncloudioj.

taddes:
No prob! @pjenvey So is there extra that should/needs to be done with the `version.json` given that local dev branches will just have "TBD" for build and commit, as well as "dev" for the version.  I see in prod it's the same, so should we be doing something different here and updating the `version.json` as it's not actively happening. Thanks!

pjenvey:
@taddes stage/prod have their version.json overwritten via circleci as you can see on the [/_\_version__ endpoint](https://merino.services.mozilla.com/__version__).

For local dev, we can either say we don't care about it and send "TBD" or whatever, or, since you already wrote the git version, we could keep it and use it for the sake of local dev and fall back to the value in version.json when there's no `.git` dir. That shouldn't require too much more effort.

taddes:
Hey @pjenvey . Yes, I was just picking up a task regarding that endpoint as well. Seems like `version` remains empty and it's likely wise we find a value we'd want it to be populated with. Thoughts? What is done on other projects?

Also, I want to explore the possibility of finding some way to actually have `version.json` in git source control be tethered to the current deployed version. As it stands, it's all just placeholder values.

Finally, and a great idea I must say on your part, to have the function pick up the environment and given that run one function or another. the `.git` reading for dev and then just a read of `version.json` for stage and prod.  I can just use `os.environ` to pick that up from the local config, or do you have another thought? I know with `merino.settings` I could use the method `using_env()` in a context manager to access a different environment, though that might not be the best approach. 

Can you also share any thoughts you have on why we don't have git on the Docker image.  Is it for security reasons we don't have the `.git` dir? Too much sensitive data exposure?

Thanks again!

ncloudioj:
After a second thought, I'd recommend that let's drop the `.git` part for this PR, and just use the `version.json` file for stage/prod. For two reasons:
- All the IO and file parsing to get the commit hash from .git would almost certainly slow down the app startup, which is not ideal even for dev
- It comes with complexities around implementation and maintenance, which, I don't believe, justifies the value it could bring to us

Since Merino is configured to log all the exceptions using this [handler](https://github.com/mozilla-services/merino-py/blob/main/merino/config_logging.py#L48-L53) in the dev environment, exceptions should be easily located in the logs. In case you do want to use Sentry in dev, I don't think it'll be too hard to locate exceptions on Sentry due to its much lower usage in dev. So lack of version identifier should not be too terrible IMHO.

@taddes What do you think?

taddes:
> 

Hey @ncloudioj ! Yes, I agree, that is a very sensible justification. I was doing a little benchmarking and while it didn't seem all that degrading, the cost is not free. I agree it's not worth the overhead and we don't want to risk adding any needless latency, especially if it isn't for much benefit. The test mocking for this was becoming a bit onerous as well, and for it to only be of benefit for dev doesn't make sense.I also agree maintenance and possible problems with this justify this as well. The original Sentry `raven` SDK for python did basically this, but it's just simpler to read the `version` file.

All the same, I'll save the functionality as a gist and we can always look at it down the road should we want to integrate this in the future. Thanks for your thoughts. I'll revise the code.

ncloudioj:
> All the same, I'll save the functionality as a gist and we can always look at it down the road should we want to integrate this in the future. Thanks for your thoughts. I'll revise the code.

@taddes Thank you for taking the spike of exploring all those potential solutions to this! Now you're the expert of .git on the team :)

taddes:
Thanks so much @ncloudioj and @pjenvey ! Set for review. LMK if anything else needs to change as is for Sentry init. Sure did learn a lot about `.git` and how to deal with refs, HEADS, she SHA values. Very valuable information. Ok, good to go for review. Putting link to gist for `.git` [here](https://gist.github.com/taddes/188c6637cbbb9e6337164f65ab57324d)

.circleci/config.yml
```diff
@@ -214,7 +214,7 @@ jobs:
             fi
   docs-build:
     docker:
-      - image: cimg/rust:1.63
+      - image: cimg/rust:1.64
```
ncloudioj, line 217: This version was already updated in main, can you do a rebase?

merino/config_sentry.py
```diff
@@ -1,27 +1,56 @@
...
         debug="debug" == settings.sentry.mode,
         environment=settings.sentry.env,
         # Set traces_sample_rate to 1.0 to capture 100%
         # of transactions for performance monitoring.
         # We recommend adjusting this value in production,
         traces_sample_rate=settings.sentry.traces_sample_rate,
     )
+
+
+def fetch_sha_hash_from_version_file(
```
ncloudioj, line 38: This looks pretty much identical to the one in [dockerflow.py](https://github.com/mozilla-services/merino-py/blob/main/merino/web/dockerflow.py#L21-L25). Could you extract this as a helper function (e.g. add a new function `fetch_app_version` to `utils/version.py`)?

Also, since `version.py` is always at the root directory, `with open("version.json") as f:` should be sufficient.

tests/unit/conftest.py
```diff
@@ -22,3 +23,12 @@ def srequest(query: str) -> SuggestionRequest:
         return SuggestionRequest(query=query, geolocation=Location())
 
     return srequest
+
+
+@pytest.fixture
+def project_root() -> str:
```
ncloudioj, line 29: Forgot to drop?

version.json
```diff
@@ -2,5 +2,5 @@
   "source": "https://github.com/mozilla-services/merino-py",
   "version": "dev",
   "commit": "TBD",
-  "build": "TBD"
-}
+  "build": "https://app.circleci.com/pipelines/github/mozilla-services/merino-py"
```
ncloudioj, line 5: What do we use this for? FWIW, the `build` tag is also populated in CI with different build URLs.

merino/config_sentry.py
hackebrot: I recommend to replace the use of `os.path` with `pathlib` for the improved ergonomics.

With `pathlib`, this method could be:

```python
def fetch_sha_hash_from_version_file(
    merino_root_path: pathlib.Path,
) -> str | None:
    """Fetch the SHA hash from the version.json file.

    During deployment, this file is written and values are
    populated for the current version of Merino in production and staging.
    """
    version_file: pathlib.Path = merino_root_path / "version.json"
    version_file_content: dict = json.loads(version_file.read_text())
    return version_file_content.get("commit")
```

See https://docs.python.org/3/library/pathlib.html

version.json
taddes: Yup, I know it gets populated in CI, but `"source"` is already filled out with the repo url and it seemed sensible to have the `"build"` populated as well so that it appears in source control. I can take it out, but it seemed sensible given the source value @ncloudioj,

tests/unit/conftest.py
taddes: Nope, left it as it's referenced as the first argument in the test functions.

merino/config_sentry.py
taddes: Thanks so much for your review @hackebrot ! I hadn't thought to use `pathlib` and it had been a while since using it, since I've often defaulted to the old `os` module usages.  That is indeed a very efficient way of handling such issues and I'll be sure to think of it when handling tasks like this going forward.  I made a couple revisions to your recommended code (including a comment for possible developers unfamiliar with pathlib and the possible alarm of not seeing a context manager) and would appreciate your feedback:

- I left the lines handling the possibility of an invalid path intact, though quite unlikely, I felt better if there were some form of error handling in case the `version.json` were not present. Your thoughts?
- I had to wrap the `merino_root_path` argument in `pathlib.Path` to allow operator overloading to support adding the `"version.json"` string to the path.  I do like this function having the argument of a path passed into it, though I also realize it may not be necessary. It does constrain the function and possibly make testing harder, because it will internally always read the `cwd`, opposed to a value being passed in. 

`def fetch_app_version(
    merino_root_path: pathlib.Path,
) -> str | None:
    """Fetch the SHA hash from the version.json file.
    During deployment, this file is written and values are
    populated for the current version of Merino in production and staging.
    """
    version_file: pathlib.Path = pathlib.Path(merino_root_path) / "version.json"
    if not version_file.exists():
        error_message = (
            f"version.json file does not exist at file path: {merino_root_path}"
        )
        raise FileNotFoundError(error_message)

    # pathlib has the 'read_text()' function that opens the file, reads it
    # and closes the file like a context manager. Emulates open() function.
    version_file_content: dict = json.loads(version_file.read_text())
    return version_file_content.get("commit")
`

My last thought is that if the [dockerflow.py](https://github.com/mozilla-services/merino-py/blob/main/merino/web/dockerflow.py#L21-L25) example Nan provided does this same task using the `os` module, is it better to match the functionality, given it does a similar thing yet in a very different context. Or, do you propose `pathlib` regardless? It does take a bit of reading for someone who is unfamiliar to get a sense of the few differences, but it is indeed much more pythonic.

hackebrot: Hi @taddes! 👋🏻 

> * I left the lines handling the possibility of an invalid path intact, though quite unlikely, I felt better if there were some form of error handling in case the `version.json` were not present. Your thoughts?

The `Path.read_text()` method calls `io.open()` under the hood, which is an alias for the built in `open()` function. 

That function raises the following error if the file does not exist:

```text
FileNotFoundError: [Errno 2] No such file or directory: 'nope/version.json'
```

I feel like this error message along with traceback (which I've omitted here) is providing sufficient detail about what went wrong. We don't need the extra check and can rely on `read_text()` to do the right thing.

You can check this for yourself using the following snippet:

```python
import pathlib
import json


def fetch_sha_hash_from_version_file(
    merino_root_path: pathlib.Path,
) -> str | None:
    """Fetch the SHA hash from the version.json file.

    During deployment, this file is written and values are
    populated for the current version of Merino in production and staging.
    """
    version_file: pathlib.Path = merino_root_path / "version.json"
    version_file_content: dict = json.loads(version_file.read_text())
    return version_file_content.get("commit")


if __name__ == "__main__":
    commit_sha = fetch_sha_hash_from_version_file(pathlib.Path("nope"))
    print(commit_sha)
```

> * I had to wrap the `merino_root_path` argument in `pathlib.Path` to allow operator overloading to support adding the `"version.json"` string to the path, otherwise Python was getting mad at me for trying to perform division on strings.

That sounds like you were trying to pass in a string value for `merino_root_path`. With the type hint for the argument in place, I'd expect mypy to raise an error.

Do you still see that issue with `configure_sentry()` passing in a `pathlib.Path` value?

> * I do like this function having the argument of a path passed into it, though I also realize it may not be necessary. It does constrain the function and possibly make testing harder, because it will internally always read the `cwd`, opposed to a value being passed in.

I'm a fan of explicitely defining interfaces/dependencies. It drastically improves comprehensibility and testability. My recommendation is to have `fetch_app_version` accept the `merino_root_path`. For convenience, we can change this to be a keyword-argument and set the default value to the cwd, so it's optional for callers to provide a value.

> My last thought is that if the [dockerflow.py](https://github.com/mozilla-services/merino-py/blob/main/merino/web/dockerflow.py#L21-L25) example Nan provided does this same task using the `os` module, is it better to match the functionality, given it does a similar thing yet in a very different context. Or, do you propose `pathlib` regardless? It does take a bit of reading for someone who is unfamiliar to get a sense of the few differences, but it is indeed much more pythonic.

I agree with @ncloudioj that we should extract this functionality into a utility function and use it for configuring the sentry SDK as well as the `/version` API endpoint. I generally recommended always using `pathlib.Path` objects for file paths within the application context and converting from/to strings when reading paths from files or returning an HTTP response etc. (similar to `datetime` objects).

Having said that, I would create a new pydantic model that holds the information in `version.json`:

```json
{
  "source": "https://github.com/mozilla-services/merino-py",
  "version": "dev",
  "commit": "TBD",
  "build": "TBD"
}
```

1. The new utility function optionally accepts the merino root path as `pathlib.Path` object and defaults to the CWD and then returns an instance of the new `Version` model.
2. The `/version` API endpoint can use the `Version` model for `response_model`. We don't even need to convert it to JSON for the response. FastAPI handles that for us.
3. The Sentry SDK configuration routine can read the `commit` information from that model instance using the `.commit` attribute.

Hope that helps!

version.json
ncloudioj: Right, `source` is prepopulated since it's a constant, whereas the `build` will be populated by CI pointing to that particular build, e.g. `https://circleci.com/gh/mozilla-services/merino-py/4014`. So using this constant here could be confusing.

tests/unit/conftest.py
```diff
@@ -22,3 +23,12 @@ def srequest(query: str) -> SuggestionRequest:
...
+    """Define the project root of merino for testing. Used for the reading of
+    the .git/refs/heads file that contains the SHA-1 hash. This is passed to
```
ncloudioj, line 31: I mean this `... reading of the .git/refs.heads file` is no longer true, so let's drop it.

taddes: Agh! Thanks for catching that 🤦

version.json
taddes: Fair enough, I will take it out!

taddes:
Thanks for all the great points @hackebrot and @ncloudioj  , definitely picked up some good models here and new information. I've made the revisions you recommended and made some adjustments in the tests to provide `pathlib.Path` objects where applicable.  I took care of of configuring the `/__version__` endpoint  to use the new util function.

I have made this function reside in `utils/version.py`, moved matching tests and renamed to `utils/test_version.py` now and it accepts the root path as a `pathlib.Path` with cwd default. I updated Dockerflow to use `pathlib` and modified the api tests to test using it as well. Also created the `Version` model that is constructed and returned, which allows access of the various attributes.

Hopefully that does it, just let me know if there are any other things that catch your eye.

merino/utils/version.py
```diff
@@ -0,0 +1,26 @@
...
+    version_file: pathlib.Path = pathlib.Path(merino_root_path) / "version.json"
+    if not version_file.exists():
+        error_message = (
+            f"version.json file does not exist at file path: {merino_root_path}"
+        )
+        raise FileNotFoundError(error_message)
+
+    # pathlib has the 'read_text()' function that opens the file, reads it
+    # and closes the file like a context manager. Uses built-in open() function.
+    version_file_content: dict = json.loads(version_file.read_text())
```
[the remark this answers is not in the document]
taddes: Yup!

merino/utils/version.py
```diff
@@ -0,0 +1,33 @@
...
+    commit value, repo source url, version, and CI build values .
+    During deployment, this file is written and values are populated for
+    the current version of Merino in production and staging.
+    """
+    version_file: pathlib.Path = merino_root_path / "version.json"
+
+    # pathlib has the 'read_text()' function that opens the file, reads it
+    # and closes the file like a context manager. Uses built-in open() function.
+    version_file_content: dict = json.loads(version_file.read_text())
+    return Version.parse_obj(version_file_content)
```
ncloudioj, line 33: Nit: while they seem similar in functionality, we tend to use `Version(**version_file_content)` more often for Merino.

merino/config_sentry.py
```diff
@@ -1,23 +1,32 @@
...
 
 import sentry_sdk
 from sentry_sdk.integrations.fastapi import FastApiIntegration
 from sentry_sdk.integrations.starlette import StarletteIntegration
 
 from merino.config import settings
+from merino.utils.version import fetch_app_version_file
+
+logger = logging.getLogger(__name__)
+MERINO_PATH = pathlib.Path.cwd()
```
ncloudioj, line 13: Since `fetch_app_version_file()` already provides the default path,  we can use that one.

merino/web/dockerflow.py
```diff
@@ -1,14 +1,16 @@
...
 
 from fastapi import APIRouter, HTTPException
 from fastapi.encoders import jsonable_encoder
 from fastapi.responses import JSONResponse, Response
 
+from merino.utils.version import fetch_app_version_file
+
 router = APIRouter()
 logger = logging.getLogger(__name__)
+MERINO_PATH = pathlib.Path.cwd()
```
ncloudioj, line 13: Ditto.

ncloudioj (approved):
r+ wc.

Thanks for adding this metadata to Sentry!

merino/utils/version.py
taddes: Got it! I had read in the docs for Pydantic [here](https://docs.pydantic.dev/usage/models/#helper-functions)  to use the helper function, but totally can use the `**` unpack operators.

merino/config_sentry.py
taddes: Sure

tests/unit/utils/test_version.py
```diff
@@ -0,0 +1,46 @@
...
+    """Test that the 'version.json' file in the root merino directory
+    can be read. Used to populate values in __version__ endpoint and to
+    access the individual values held within the json file.
+    This is the method implemented in the staging and production environments
+    as there is no populated version.json file in dev.
+    """
+    version_file = fetch_app_version_file(project_root)
+    assert version_file
+    assert type(version_file) == Version
+    assert hasattr(version_file, attribute)
```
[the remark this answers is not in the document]
hackebrot: Good call! +1 from me. 👍🏻 

Pydantic supports equality checks on models, so you can do the following:

```python
def test_version(project_root: pathlib.Path) -> None:
    """Happy path test for fetch_app_version_file()."""

    expected_version = Version(
        source="https://github.com/mozilla-services/merino-py",
        version="dev",
        commit="TBD",
        build="TBD",
    )

    assert fetch_app_version_file(project_root) == expected_version
```

hackebrot (changes requested):

merino/web/dockerflow.py
```diff
@@ -18,12 +20,11 @@
 )
 async def version() -> JSONResponse:
     """Dockerflow: Query service version."""
-    if not os.path.exists("version.json"):
+    if not pathlib.Path("version.json").exists():
```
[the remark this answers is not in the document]
taddes: Thanks @hackebrot ! Your comment about coupling got me thinking that it was perhaps a better approach to define the file itself as a parameter with the `version.json` as a default.  Internally it will still work the same, however we can pass in a different path and file, which resolves the coupling concern.  I know it is highly unlikely that there is ever a need for this since it does one thing very specifically, but it at least makes it a little more flexible and resolves your concern about coupling, I hope!

As for the check, I am fine with the above approach using a try/except block, however patching for that was problematic. Perhaps the change to the function itself to accept 2 parameters resolves this? We can discuss that further if you'd really prefer this approach, but `mocker.patch.object(pathlib.Path, "exists", return_value=False)` works very well in the integration tests and the try/except approach wouldn't allow the same patching, so we'd need to reassess that as patching the specific function was cumbersome and problematic for me, though I may just be going about it the wrong way.

For readability, I always prefer things to be explicit, so in many cases we can just call the function with no arguments and the defaults will apply, but for Dockerflow, defining two constants doesn't seem too bad, right? I'm open to leaving them out, I just feel it makes the code more readable, and since Dockerflow is quite lightweight, it makes it easier to sub values in those constants were they ever to change (though again, unlikely).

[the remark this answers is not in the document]
taddes: Also, thx @ncloudioj for the approve and the little comments, made the changes.  Lmk if how I've slightly shifted this is ok with you, or if it's not the right direction I can reverse course.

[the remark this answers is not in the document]
ncloudioj: IMHO adding another argument for this is kinda overkill, as we won't use this helper function to read other files other than "version.json". In fact, I'd be inclined to drop the first argument as well since "version.json" is required to be located in the root directory of any Dockerflow-compliant project. If that's not the case, the app should fail immediately by all means. I thought we kept it as an argument solely to facilitate testing and I think that's fine.

That being said, I think we can check the presence of this "version.json" in the root directory at application [startup](https://github.com/mozilla-services/merino-py/blob/main/merino/main.py#L20). If missing, Merino should refuse to boot. With that, we can drop all the checks elsewhere including this one since it's already enforced at startup.

What do you think?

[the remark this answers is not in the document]
hackebrot: Getting mocks right can be quite tricky! 🤖 

The `fetch_app_version_file()` method does not call `exists()` directly, so that's probably why your mock is ineffective. For integration tests, you can raise a `FileNotFoundError` when attempting to read text from a file.

Try this:

```python
mocker.patch.object(pathlib.Path, "read_text", side_effect=FileNotFoundError)
```

What you're really testing in `test_version_error` is that our FastAPI app turns that error into a 500 response. Best to add the following check to the test for the error message:

```python
assert response.json() == {"detail": "Version file does not exist"}
```

I don't think we should be adding an additional argument. My concern about coupling was that the API handler code performs a check for the existence of a very specific file (`version.json`), but then calls the utility function without specifying the file. The file check is not guaranteed to look for the same file as `fetch_app_version_file()` does.

We don't need to call `exists()` in the API handler and can catch the exception raised by `version_file.read_text()` instead. The code in `dockerflow.py` can rely on the default value in the utility and doesn't need to define constant variables for the merino root directory.

It's a good idea to check for the version information at app startup. In the interest of completing this work and merging this pull request, I suggest opening a new ticket for this.

merino/web/dockerflow.py
```diff
@@ -1,29 +1,32 @@
...
-from fastapi.encoders import jsonable_encoder
-from fastapi.responses import JSONResponse, Response
+from fastapi.responses import Response
+
+from merino.utils.version import Version, fetch_app_version_file
 
 router = APIRouter()
 logger = logging.getLogger(__name__)
+MERINO_PATH = pathlib.Path.cwd()
+VERSION_FILENAME = "version.json"
```
hackebrot, line 13: Please remove these two constants.

taddes: 👍

merino/web/dockerflow.py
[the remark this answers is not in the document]
taddes: Thanks @hackebrot ! I've made those changes now and I will then open something new to handle the checks at at startup. Had some good feedback fron @ncloudioj about that and he defined a good way to go about it.  I made that change in the integration tests.  It wasn't calling `exists()` on `fetch_app_version_file()` that was the issue, but rather having issues just getting `mocker.patch("fetch_app_version_file", result=None)` to function, for example. The one you offered above works just fine, but I will have to snag some time to discuss some nuances of more custom mocks in the near future.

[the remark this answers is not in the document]
taddes: Thanks @hackebrot ! I've made those changes now and I will then open something new to handle the checks at at startup. Had some good feedback fron @ncloudioj about that and he defined a good way to go about it.  I made that change in the integration tests.  It wasn't calling `exists()` on `fetch_app_version_file()` that was the issue, but rather having issues just getting `mocker.patch("fetch_app_version_file", result=None)` to function, for example. The one you offered above works just fine, but I will have to snag some time to discuss some nuances of more custom mocks in the near future.

hackebrot (changes requested):

merino/config_sentry.py
```diff
@@ -1,23 +1,30 @@
...
+import logging
 
 import sentry_sdk
 from sentry_sdk.integrations.fastapi import FastApiIntegration
 from sentry_sdk.integrations.starlette import StarletteIntegration
 
 from merino.config import settings
+from merino.utils.version import fetch_app_version_file
+
+logger = logging.getLogger(__name__)
```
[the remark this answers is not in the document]
taddes: We were initially with the original error handling, I'll take it out

version.json
```diff
@@ -3,4 +3,4 @@
   "version": "dev",
   "commit": "TBD",
   "build": "TBD"
-}
+}
```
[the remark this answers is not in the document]
taddes: I think that's incorrect, it's a total of 6 lines and the auto format leaves no newline

merino/utils/version.py
```diff
@@ -0,0 +1,33 @@
...
+import json
+import logging
+import pathlib
+
+from pydantic import BaseModel, HttpUrl
+
+logger = logging.getLogger(__name__)
+
+
+class Version(BaseModel):
```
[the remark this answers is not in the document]
taddes: Yes, that's sensible, I saw that in the docs. Good call! I have some future ideas about implementing some of the pydantic validators throughout the code for additional coverage, much like what you describe.

taddes:
Thanks for the detailed test revisions and details @hackebrot , I have made the requested changes.

tests/unit/utils/test_version.py
```diff
@@ -0,0 +1,62 @@
...
+import json
+import pathlib
+
+import pytest
+from pydantic import ValidationError
+
+from merino.utils.version import fetch_app_version_from_file
+
+# NOTE:  argument is defined in the  fixture in
+# conftest.py, at the root of this directory.
```
hackebrot, line 15: We can remove this note.

hackebrot (approved):
LGTM! 📦 🚀 

Please squash your commits when merging.

PR #163, opened 2023-01-03
[DISCO-2141] Build merinopy in separate CI step

mtbottle:
This allows the built artifact to be reused in various downstream steps.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -156,7 +156,7 @@ jobs:
             docker_layer_caching: true
       - run:
             name: Build image
-            command: docker build -t app:build .
+            command: make docker-build
       - run:
           name: Save image into workspace
           command: |
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -22,10 +22,13 @@ workflows:
           requires:
             - unit-tests
             - integration-tests
+      - docker-image-build:
+          <<: *pr-filters
       - contract-tests:
           <<: *pr-filters
           requires:
             - checks
+            - docker-image-build
       - docs-build:
           <<: *pr-filters
 
@@ -44,10 +47,13 @@ workflows:
           requires:
             - unit-tests
             - integration-tests
+      - docker-image-build:
+          <<: *main-filters
       - contract-tests:
           <<: *main-filters
           requires:
             - checks
+            - docker-image-build
       - docs-build:
           <<: *main-filters
       - docs-publish-github-pages:
@@ -62,6 +68,7 @@ workflows:
             - integration-tests
             - test-coverage-check
             - contract-tests
+            - docker-image-build
       # The following job will require manual approval in the CircleCI web application.
       # Once provided, and when all the requirements are fullfilled (e.g. tests)
       - unhold-to-deploy-to-prod:
@@ -73,6 +80,7 @@ workflows:
             - integration-tests
             - test-coverage-check
             - contract-tests
+            - docker-image-build
       # On approval of the `unhold-to-deploy-to-prod` job, any successive job that requires it
       # will run. In this case, it's manually triggering deployment to production.
       - docker-image-publish-prod:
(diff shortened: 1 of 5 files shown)


.circleci/config.yml
```diff
@@ -139,13 +147,38 @@ jobs:
...
     working_directory: ~/merino
     steps:
       - checkout
+      - attach_workspace:
+            at: /tmp/workspace
+      - run:
+            name: Load Docker image from workspace
+            command: docker load -i /tmp/workspace/merinopy.tar.gz
       - run:
           name: Contract tests
```
Trinaa, line 183: From the [CircleCI log](https://app.circleci.com/pipelines/github/mozilla-services/merino-py/762/workflows/17360159-c4f4-4e8d-9ca7-e1e648444845/jobs/3279?invite=true#step-104-84), I think the merino image is still being built by the contract tests? 

In which case I'm thinking some changes in the contract `docker-compose.yaml` and `Makefile` may also be needed.

mtbottle: Thanks for the catch! I _think_ this is no longer built now in the [latest run](https://app.circleci.com/pipelines/github/mozilla-services/merino-py/765/workflows/adaeb6d8-3414-4fe7-b90d-7e47f45333a4/jobs/3299?invite=true#step-104-39).

Trinaa: Yep! Now the contract-tests re-use the image. 

But for someone who doesn't have the image built locally, the `make` command now fails:
image

What do you think of including a docker build step in the `Makefile` and making the contract-tests step composite? Something like (with better names 😹):

```
.PHONY: build
build:  ##  Build the merino docker image
	docker build -t app:build .

.PHONY: contract-tests-only
contract-tests:  ##  Run contract tests using docker compose (no merino build)
	docker-compose \
      -f $(CONTRACT_TEST_DIR)/docker-compose.yml \
      -p merino-py-contract-tests \
      up --abort-on-container-exit

.PHONY: contract-tests
contract-tests:  build contract-tests-only ##  Build merino image and run contract tests using docker compose
```

We could then use the non-composite step in the CI config? Maybe the `build` command can be used?
 
_If we do this we have to update the [make command doc](https://github.com/mozilla-services/merino-py/blob/main/docs/dev/index.md#general-commands)_

mtbottle: Arg! Thanks for this catch!

How about we simplify this to just add the docker build command right before the contract test? We don't really need the docker build anywhere else, and the tagging is awkwardly specific to the one used in the `docker-compose`.

mtbottle: NVM about this, just realized why it won't work -.-

Makefile
```diff
@@ -75,12 +75,19 @@ integration-tests: $(INSTALL_STAMP)  ##  Run integration tests
...
 	docker-compose \
       -f $(CONTRACT_TEST_DIR)/docker-compose.yml \
       -p merino-py-contract-tests \
-      up --abort-on-container-exit --build
+      up --abort-on-container-exit
+
+.PHONY: local-contract-tests
+local-contract-tests: docker-build contract-tests
 
 .PHONY: contract-tests-clean
```
ncloudioj, line 92: Let's rename `contract-tests-clean` to `local-contract-tests-clean` to be consistent.

Makefile
```diff
@@ -75,12 +75,19 @@ integration-tests: $(INSTALL_STAMP)  ##  Run integration tests
 integration-test-fixtures: $(INSTALL_STAMP)  ##  List fixtures in use per integration test
 	MERINO_ENV=testing $(POETRY) run pytest $(INTEGRATION_TEST_DIR) --fixtures-per-test
 
+.PHONY: docker-build
+docker-build:
```
ncloudioj, line 79:
```suggestion
docker-build:  ## Build the docker image for Merino named "app:build"
```
The comment after "##" will be printed out when typing "make help"

ncloudioj (approved):
LGTM, r+ wc.

Makefile
mtbottle: I've decided to rename the original `make contract-tests` to `make run-contract-tests`, and reverted the `make local-contract-tests` to `make contract-tests`. I figured that this would probably be less overhead on developers to learn new commands to run them locally. The CircleCI job now runs the `make run-contract-tests` instead. This means that the `contract-tests-clean` can probably be left as is then.

.circleci/config.yml
```diff
@@ -139,32 +147,58 @@ jobs:
...
+  docker-image-build:
+    docker:
+      - image: cimg/base:2022.08
+    steps:
+      - checkout
+      - setup_remote_docker:
+            docker_layer_caching: true
+      - run:
+            name: Build image
+            command: docker build -t app:build .
```
Trinaa, line 159: Could this be?

```suggestion
            command: make docker-build
```

.circleci/config.yml
Trinaa: All good!

Trinaa (approved):
LGTM! I don't know if you saw the stats on CircleCI, but these changes make the build ~23% faster which is cool! 🚀 

I left one observation, but I leave it to you to take or leave.

PR #165, opened 2023-01-04
Write the version.json file before building image

mtbottle:
This ensures that the `version.json` file exists so that it can be available during the build process.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -155,6 +155,8 @@ jobs:
       - setup_remote_docker:
             docker_layer_caching: true
       - write-version
+      - store_artifacts:
+            path: version.json
       - run:
             name: Build image
             command: make docker-build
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -154,6 +154,7 @@ jobs:
       - checkout
       - setup_remote_docker:
             docker_layer_caching: true
+      - write-version
```
mtbottle, line 157: I just also want to confirm that this doesn't need to be turned into an artifact? The other usages of this also includes:

```
- store_artifacts:
path: version.json
```

mtbottle: Justification for why it's not needed: Docker should pick up the file as it's in the working directory during the actual build. I'm pretty sure that the artifact store is not needed at this point.

ncloudioj: I think we make it an artifact mostly for future references. So, yeah, let's also do that here, and then you can remove the `write-version` steps for `publish-images-to-stage/prod` as they're no longer needed there.

mtbottle: Sure, I originally wanted fewer changes (just in case this breaks more things). But I am up for trying this out as well.

ncloudioj (approved):
🚢

Issue #155, opened 2022-12-21
Switch to ruff for linting


data-sync-user:
[Ruff](https://github.com/charliermarsh/ruff) is a new, very fast (written in rust) linter that could likely replace our uses of isort, flake8 and pydocstyle.

However our current linting checks are very quick as the merino-py codebase isn’t very large, so this is a low priority.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/DISCO-2165)
┆Epic: 2024 Q1 Maintenance

data-sync-user:
➤ Bastian Gruber commented:

From what I gathered, we could also replace black with ruff:
https://github.com/astral-sh/ruff/blob/main/docs/faq.md#how-does-ruffs-formatter-compare-to-black ( https://github.com/astral-sh/ruff/blob/main/docs/faq.md#how-does-ruffs-formatter-compare-to-black|smart-link )

PR #166, opened 2023-01-04
Split the `lint` command to be able to run all in parallel

mtbottle:
We have multiple lint commands that are currently run sequentially. However, this makes it inefficient, as we will need to re-run the linter to get errors for subsequent commands. 

This change will allow us to use the `Makefile`'s `--keep-going` (`-k`) so that it can run all the dependant targets in the `lint` target regardless of whether it fails the checks. I opted for this approach over using multiple CircleCI jobs as it feels kind of overkill to bootstrap another job container to do all these small checks.  

Devs will also need to update their workflow to run `make -k lint` rather than `make lint` to get the advantage of parallelism (I'll make an announcement on Slack).

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -96,7 +96,7 @@ jobs:
       - checkout
       - run:
           name: Code linting
-          command: make lint
+          command: make -k lint
   unit-tests:
     docker:
       - image: cimg/python:3.11
--- a/Makefile
+++ b/Makefile
@@ -19,14 +19,32 @@ $(INSTALL_STAMP): pyproject.toml poetry.lock
 	$(POETRY) install
 	touch $(INSTALL_STAMP)
 
-.PHONY: lint
-lint: $(INSTALL_STAMP)  ##  Run various linters
+.PHONY: isort
+isort: $(INSTALL_STAMP)  ##  Run isort
 	$(POETRY) run isort --check-only $(APP_AND_TEST_DIRS)
+
+.PHONY: black
+black: $(INSTALL_STAMP)  ##  Run black
 	$(POETRY) run black --quiet --diff --check merino $(APP_AND_TEST_DIRS)
+
+.PHONY: flake8
+flake8: $(INSTALL_STAMP)  ##  Run flake8
 	$(POETRY) run flake8 $(APP_AND_TEST_DIRS)
+
+.PHONY: bandit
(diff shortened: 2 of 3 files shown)


mtbottle:
Sample failure on CircleCI (failed on isort): 
Screen Shot 2023-01-04 at 16 06 51

.circleci/config.yml
```diff
@@ -96,7 +96,7 @@ jobs:
       - checkout
       - run:
           name: Code linting
-          command: make lint
+          command: make -k lint
```
ncloudioj, line 99: IIRC, `-k` means "keep-on-going" when encountering failures. `-j` makes make to execute many recipes simultaneously. Shall we use `-k` and `-j` both as the intent of this PR is for parallelism?

mtbottle: These commands don't take too long to run, plus running them in parallel makes it maybe a bit difficult to see which command the error is associated with. So I'm kind of tempted to keep all the errors associated with a command together by having them run sequentially. If they do take a long time, I would then probably push for splitting it out into a new CircleCI job instead.

ncloudioj: Sounds good, let's update [this comment](https://github.com/mozilla-services/merino-py/pull/166/files#diff-536f1e6cf1be5c6864dca4b8e50669398c5977362a9a9782a7f852ae8e369f5cR53) to avoid confusion.

ncloudioj (approved):
r+ wc.

pjenvey:
FYI see the related #155 -- ruff can replace a few of these linters and should be faster

mtbottle:
> FYI see the related https://github.com/mozilla-services/merino-py/issues/155 -- ruff can replace a few of these linters and should be faster

Ooo, definitely looks like an interesting project! Adding this would mean having Rust dependencies in a Python project though, which will increase the complexity of the codebase/deploy pipeline (for the tradeoff of faster tests).

pjenvey:
> Adding this would mean having Rust dependencies in a Python project though, which will increase the complexity of the codebase/deploy pipeline (for the tradeoff of faster tests).

@mtbottle AFAIK this shouldn't be an issue: it's easily installed by pip as it provides binary wheels for the major platforms. The `cryptography` package we already use uses Rust internally and works similarly, as will the future 2.0 release of `pydantic` (which will begin using its Rust based `pydantic-core` package).

PR #174, opened 2023-01-10
[DISCO-2144] Tests to verify logging handlers 

taddes:
Added testing to verify happy path of logging configuration for both `mozlog` and `pretty`.  Had to dig into logging src code and docs to find the` logging.root.manager`, which provides a lot of methods and attributes on a given environment's logging config. From there, calling loggerDict["merino"].handlers[0].name on this object provided the handler specified when passed into the `configure_logging()` method. 

I also added two values to `.gitigore` since most of our workflows are on macOS and some folks use VS Code. This covers possible `DS_Store` files and `vs code` local configs and prevents them from being tracked.


tests/unit/test_config_logging.py
```diff
@@ -39,3 +43,23 @@ def test_configure_logging_mozlog_production():
...
+
+
+def test_log_handler_assigned_mozlog() -> None:
+    """Test that the log handler is assigned as expected when the configured format is 'mozlog'"""
+    settings.logging.format = "mozlog"
+    configure_logging()
+
+    merino_log_manager: Any = logging.root.manager
+    handler_name = merino_log_manager.loggerDict["merino"].handlers[0].name
+    assert handler_name == "console-mozlog"
```
Trinaa, line 55: If you'd like to specify 'Manager' instead of 'Any', you can do the following, but I'm not really sure if that's better...

```suggestion
    merino_log_manager: Manager = logging.root.manager
    merino_logger: Logger | PlaceHolder = merino_log_manager.loggerDict["merino"]
    assert isinstance(merino_logger, Logger)
    assert merino_logger.handlers[0].name == "console-mozlog"
```
Note: Logger, Manager & Placeholder need to be imported from `logging`

Trinaa (changes requested):
Unit test coverage increased from 79% to 80%! 🚀 
I've noted some minor fixes and made some suggestions for you to address at your own discretion.

tests/unit/test_config_logging.py
taddes: Thanks so much for the review @Trinaa ! I'll look over and apply the changes. Good recs

taddes: Yeah, as for this, MyPy isn't smart enough to pick up `PlaceHolder` and `Logger`. Even though the code works, it gets mad as it things there is no `handlers` attribute. I mucked around with that a while trying to get it to work, so I think let's leave it as `Any` if you're comfortable. It'd be nice to get it right, but I don't think it'll work with MyPy. Plus, not as important since we at least have the `Manager` type correctly bound.

Trinaa (approved):

PR #178, opened 2023-01-16
Add architecture diagram for Merino-py [do not deploy]

mtbottle:
The architecture diagram is coded with [Mermaid](https://mermaid.js.org/).

I put the image on the landing page of our documentation. If this isn't the right place for it, then we can create a new page. 

Also, I only included the `/api/v1/suggest` API for now. There are other APIs that we _might_ want to also add a flow diagram for. If we should add them, then let me know and I'll update the diagram.

### DIFF
--- a/docs/intro.md
+++ b/docs/intro.md
@@ -43,6 +43,66 @@ dependencies required for Merino.
 [9]: ./dev/testing.md
 [10]: ./dev/profiling.md
 
+## Architecture
+
+```mermaid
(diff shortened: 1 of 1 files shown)


ncloudioj (approved):
This is great, thanks!

Happy Diagramming!

ncloudioj:
Plz amend the commit message with `[do not deploy]` so it wont trigger a deploy.

PR #180, opened 2023-01-19
Add mbbook-mermaid to compile Mermaid diagrams in mdbook [do not deploy]

mtbottle:
This should fix the issue with the architecture diagram not compiling in mdbook.


ncloudioj (approved):
Thanks!

PR #181, opened 2023-01-19
doc: update archtecture diagram [do not deploy]

ncloudioj:
Added some FontAwesome icons :)


mtbottle (approved):
The icons don't show up in GitHub preview, but I've confirmed that it shows up when compiled with mdbook.

PR #182, opened 2023-01-20
fix: round weather temperatures to nearest integer

0c0w3:
We've gotten feedback from foxfooders that showing one decimal place for weather temperatures in C is unnecessary and uncommon. AccuWeather does include one decimal place in C temperatures but not in F.

This commit makes sure we end up with rounded temperatures in both C and F regardless of whether we're using AccuWeather's values or computing them ourselves.


ncloudioj (approved):
LGTM, thanks!

PR #185, opened 2023-01-25
Configure Merino to talk to cluster

mtbottle:
This change enables Merino to get configured to communicate with the Elasticsearch cluster. The cluster details will be grabbed from the environment variables.

The staging configuration depends on: https://github.com/mozilla-services/cloudops-infra/pull/4646

### DIFF
--- a/merino/configs/default.toml
+++ b/merino/configs/default.toml
@@ -113,6 +113,12 @@ top_picks_file_path = "dev/top_picks.json"
 enabled_by_default = false
 # The backend of the provider. Either "elasticsearch" or "test".
 backend = "elasticsearch"
+# Elasticsearch Cloud ID. The following one is *fake* and will be
+# overridden in production.
+# es_cloud_id = """
+#   testing:dXMtZWFzdC0xLmF3cy5mb3VuZC5pbyRjZWM2ZjI2MWE3NGJmMjRjZ
+#   TMzYmI4ODExYjg0Mjk0ZiRjNmMyY2E2ZDA0MjI0OWFmMGNjN2Q3YTllOTYyNTc0Mw==
+# """
 es_url = ""
 # Elasticsearch index
 es_index = "enwiki_merino_test"
--- a/merino/configs/default.toml
+++ b/merino/configs/default.toml
@@ -113,18 +113,9 @@ top_picks_file_path = "dev/top_picks.json"
 enabled_by_default = false
 # The backend of the provider. Either "elasticsearch" or "test".
 backend = "elasticsearch"
-# Elasticsearch Cloud ID. The following one is *fake* and will be
-# overridden in production.
-es_cloud_id = """
-  testing:dXMtZWFzdC0xLmF3cy5mb3VuZC5pbyRjZWM2ZjI2MWE3NGJmMjRjZ
-  TMzYmI4ODExYjg0Mjk0ZiRjNmMyY2E2ZDA0MjI0OWFmMGNjN2Q3YTllOTYyNTc0Mw==
-"""
-# Elasticsearch user
-es_user = ""
-# Elasticsearch password
-es_password = ""
+es_url = ""
 # Elasticsearch index
-es_index = "enwiki"
+es_index = "enwiki_merino_test"
 # The maximum suggestions for each search request.
 es_max_suggestions = 3
 # The timeout (in millisecond) for each request to ES
--- a/merino/providers/wikipedia/backends/elastic.py
+++ b/merino/providers/wikipedia/backends/elastic.py
@@ -1,5 +1,5 @@
 """The Elasticsearch backend for Dynamic Wikipedia."""
-from typing import Any, Final
+from typing import Any, Final, Optional
 from urllib.parse import quote
 
 from elasticsearch import AsyncElasticsearch
@@ -19,9 +19,15 @@ class ElasticBackend:
 
     client: AsyncElasticsearch
 
-    def __init__(self, *, cloud_id: str, user: str, password: str) -> None:
+    def __init__(self, *, url: str, cloud_id: Optional[str]) -> None:
         """Initialize."""
-        self.client = AsyncElasticsearch(cloud_id=cloud_id, basic_auth=(user, password))
+        # Only one of `cloud_id` or `hosts` can be passed.
+        # If the `cloud_id` is specified, then use that.
+        # Otherwise fallback to using the specified URL.
+        if cloud_id:
+            self.client = AsyncElasticsearch(cloud_id=cloud_id)
+        else:
+            self.client = AsyncElasticsearch(url)
 
     async def shutdown(self) -> None:
         """Shut down the connection to the ES cluster."""
--- a/merino/providers/wikipedia/backends/elastic.py
+++ b/merino/providers/wikipedia/backends/elastic.py
@@ -14,20 +14,28 @@ TIMEOUT_MS: Final[str] = f"{settings.providers.wikipedia.es_request_timeout_ms}m
 MAX_SUGGESTIONS: Final[int] = settings.providers.wikipedia.es_max_suggestions
 
 
+class ElasticBackendError(BackendError):
+    """Error during interactions with Elasticsearch"""
+
+
 class ElasticBackend:
     """The client that works with the Elasticsearch backend."""
 
     client: AsyncElasticsearch
 
-    def __init__(self, *, url: str, cloud_id: Optional[str]) -> None:
+    def __init__(
+        self, *, url: Optional[str] = None, cloud_id: Optional[str] = None
+    ) -> None:
         """Initialize."""
         # Only one of `cloud_id` or `hosts` can be passed.
-        # If the `cloud_id` is specified, then use that.
-        # Otherwise fallback to using the specified URL.
-        if cloud_id:
+        if cloud_id and not url:
             self.client = AsyncElasticsearch(cloud_id=cloud_id)
-        else:
+        elif url and not cloud_id:
             self.client = AsyncElasticsearch(url)
+        else:
+            raise ElasticBackendError(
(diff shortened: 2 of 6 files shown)


merino/configs/default.toml
```diff
@@ -113,18 +113,9 @@ top_picks_file_path = "dev/top_picks.json"
...
-  testing:dXMtZWFzdC0xLmF3cy5mb3VuZC5pbyRjZWM2ZjI2MWE3NGJmMjRjZ
-  TMzYmI4ODExYjg0Mjk0ZiRjNmMyY2E2ZDA0MjI0OWFmMGNjN2Q3YTllOTYyNTc0Mw==
-"""
-# Elasticsearch user
-es_user = ""
-# Elasticsearch password
-es_password = ""
+es_url = ""
 # Elasticsearch index
-es_index = "enwiki"
```
mtbottle, line 127: I've created the index under new name to test functionality in staging. We'll need to change it back once the ingest job is done.

ncloudioj: You can also use [local setting files](https://www.dynaconf.com/settings_files/#local-settings-files) for temporary settings.

For example, you can create a file "default.local.toml" under "merino/configs/" 
```
[default]
debug = true
dynaconf_merge = true

# [default.providers.wikipedia]
es_index = "enwiki_merino_test"
```

Local config files won't be checked into the repo.

ncloudioj (approved):
LGTM 🚢

merino/providers/wikipedia/backends/elastic.py
```diff
@@ -19,9 +19,15 @@ class ElasticBackend:
 
     client: AsyncElasticsearch
 
-    def __init__(self, *, cloud_id: str, user: str, password: str) -> None:
+    def __init__(self, *, url: str, cloud_id: Optional[str]) -> None:
```
quiiver, line 22: Shouldn't Url here be optional as well? What will the value be if cloud_id is passed?

merino/providers/wikipedia/backends/elastic.py
```diff
@@ -19,9 +19,15 @@ class ElasticBackend:
...
         """Initialize."""
-        self.client = AsyncElasticsearch(cloud_id=cloud_id, basic_auth=(user, password))
+        # Only one of `cloud_id` or `hosts` can be passed.
+        # If the `cloud_id` is specified, then use that.
+        # Otherwise fallback to using the specified URL.
+        if cloud_id:
+            self.client = AsyncElasticsearch(cloud_id=cloud_id)
```
Trinaa, line 28: This condition isn't covered by tests according to the coverage report. I suggest to add a unit test.

tests/unit/providers/wikipedia/test_provider.py
```diff
@@ -0,0 +1,68 @@
+"""Unit tests for the Merino v1 suggest API endpoint for the Wikipedia Provider."""
```
mtbottle, line 1: These were moved from `backend/test_elastic.py`. They were testing the provider, so didn't belong in that file.

merino/providers/wikipedia/backends/elastic.py
mtbottle: > Shouldn't Url here be optional as well? 

Yep! Updated!

> What will the value be if cloud_id is passed?

The URL should be None or empty if the cloud id is passed. Currently we can't get to the case with `cloud_id` because it's gated with the URL: https://github.com/mozilla-services/merino-py/blob/f4f311766111518405ef301eb6e4fc6ab6b10690/merino/providers/__init__.py#L98 I think we can change this when we get the Elastic cloud cluster (as there's probably some other things we'll need to update, like SSL and auth). I'll create a ticket for this.

tests/unit/providers/wikipedia/backends/test_elastic.py
```diff
@@ -1,86 +1,59 @@
...
 
-
-@pytest.fixture(name="wikipedia")
-def fixture_wikipedia() -> Provider:
-    """Return a Wikipedia provider that uses a test backend."""
-    return Provider(backend=FakeEchoWikipediaBackend())
+FAKE_ELASTIC_CLOUD_ID = (
+    "cluster:dXMtZWFzdC0xLmF3cy5mb3VuZC5pbyQ0ZmE4ODIxZTc1NjM0MDMyYmVk"
+    "MWNmMjIxMTBlMmY5NyQ0ZmE4ODIxZTc1NjM0MDMyYmVkMWNmMjIxMTBlMmY5Ng=="
+)
```
Trinaa, line 19: Globals can cause tests to become unisolated. Perhaps fixturize the test values for backend init parameters?

tests/unit/providers/wikipedia/backends/test_elastic.py
```diff
@@ -1,86 +1,59 @@
...
 
 
-def test_enabled_by_default(wikipedia: Provider) -> None:
-    """Test for the enabled_by_default method."""
-    assert wikipedia.enabled_by_default
-
+def test_initialize_es_backend_cloud_id_success():
+    """Test that the ElasticBackend initializes successfully with a cloud_id."""
+    es = ElasticBackend(cloud_id=FAKE_ELASTIC_CLOUD_ID)
+    assert es
```
Trinaa, line 34: Since units are white-box tests, we can spy on the AsyncElasticsearch object to make sure it's being initialized as expected:

```suggestion
def test_initialize_es_backend_cloud_id_success(mocker: MockerFixture) -> None:
    """Test that the ElasticBackend initializes successfully with a cloud_id."""
    es_client_spy: Any = mocker.spy(AsyncElasticsearch, "__init__")

    ElasticBackend(cloud_id=FAKE_ELASTIC_CLOUD_ID)

    es_client_spy.assert_called_once()
    assert es_client_spy.call_args.kwargs["cloud_id"] == FAKE_ELASTIC_CLOUD_ID
```

tests/unit/providers/wikipedia/backends/test_elastic.py
```diff
@@ -1,86 +1,59 @@
...
-@pytest.mark.asyncio
-async def test_shutdown(wikipedia: Provider, mocker: MockerFixture) -> None:
-    """Test for the shutdown method."""
-    spy = mocker.spy(FakeEchoWikipediaBackend, "shutdown")
-    await wikipedia.shutdown()
-    spy.assert_called_once()
+def test_initialize_es_backend_url_success():
+    """Test that the ElasticBackend initializes successfully with a URL."""
+    es = ElasticBackend(url="http://localhost:9200")
+    assert es
```
Trinaa, line 40: Similar to above:

```suggestion
def test_initialize_es_backend_url_success(mocker: MockerFixture) -> None:
    """Test that the ElasticBackend initializes successfully with a URL."""
    url: str = "http://localhost:9200"
    es_client_spy: Any = mocker.spy(AsyncElasticsearch, "__init__")

    ElasticBackend(url=url)

    es_client_spy.assert_called_once()
    assert url in es_client_spy.call_args[0]
```

tests/unit/providers/wikipedia/backends/test_elastic.py
```diff
@@ -1,86 +1,59 @@
...
-    ["query", "expected_title"],
-    [("foo", "foo"), ("foo bar", "foo_bar"), ("foØ bÅr", "fo%C3%98_b%C3%85r")],
+    ["url", "cloud_id"],
+    [
+        (None, None),
+        (
+            "https://localhost:9200",
+            FAKE_ELASTIC_CLOUD_ID,
+        ),
+    ],
```
Trinaa, line 51: You could add ids if you think it adds clarity?
```suggestion
    ],
    ids=["without_url_and_cloud_id", "with_url_and_cloud_id"]
```

mtbottle: After turning `FAKE_ELASTIC_CLOUD_ID` into a fixture, I decided to not parameterize this function, so split this into 2 test functions instead.

merino/configs/default.toml
```diff
@@ -113,18 +113,9 @@ top_picks_file_path = "dev/top_picks.json"
 enabled_by_default = false
 # The backend of the provider. Either "elasticsearch" or "test".
 backend = "elasticsearch"
-# Elasticsearch Cloud ID. The following one is *fake* and will be
-# overridden in production.
-es_cloud_id = """
```
ncloudioj, line 118: Since `default.toml` is the base of all the config variants, removing it from here would make its future reference kinda confusing. I know that TOML doesn't have a concept of "nil" or "null", can you add a new item to determine which mode ES should be in? For example:
```
es_mode = "[local|cloud]"
```

And then add this validation to [here](https://github.com/mozilla-services/merino-py/blob/main/merino/config.py#L24-L33).
```
Validator("providers.wikipedia.es_mode", is_in=["url", "cloud"]),
```

mtbottle: Hmm... I was toying around with having a mode config, but wasn't sure how that would help with the settings discovery (ie. we'll probably want `es_cloud_id` to be empty, even if we set the mode). Since we have only a `url` version right now, it might be kind of early to really consider adding `cloud_id`. There's probably a whole bunch of other configs that we'll need at that point (like SSL and auth). So I have [a ticket](https://mozilla-hub.atlassian.net/browse/DISCO-2217) to fix the configuration for Elastic Cloud once we get the settings. I think it might make more sense to figure out the configuration organization at that point, and just get this merged so that we can test to see if it can talk to the temporary cluster. What do you think?

ncloudioj: Sure, sounds good. Can you leave it commented out rather than deleting it?

Also, please add a new entry for `es_url` to [here](https://github.com/mozilla-services/merino-py/blob/main/docs/ops.md#wikipedia-dynamic-match-provider).

Trinaa (approved):
TY! 🚀

PR #186, opened 2023-01-25
DISCO-2216 Merino Jobs - Introduce a CLI interface for managing jobs within the merino repository

quiiver:
After some discussion with a few of you, I wanted to explore creating a CLI container that could be used to run jobs and have it live within the merino application.

Some benefits to having this code live here:
* Can leverage existing team norms (code review, linting, etc)
* Shared dependencies between API code and CLI code. This means we don't have to maintain a separate set of dependencies like elasticsearch clients.
* Code colocation. The code that generates the datasets live next to the code that uses the datasets and should reduce context switching.

However, it is not without drawbacks. It complicates the deployment by essentially bundling two deployment targets into the same codebase (jobs and api). I feel as though thats a reasonable trade off given the above benefits but welcome discussion on this topic as it affects all of us.

Separation of concerns when deploying is achieved through docker build contexts (`job_runner` and `web_api`). Using `docker build` we can specify the final layer of the image to use the context we want.

### TODO
I haven't included any changes to CI in this because it requires some coordination with SRE and I wanted to get this up first. Essentially, we would push a new image name something like `mozilla/merino-jobs` and have that image be specified in the deployment of the job code


quiiver:
Tagging @mozilla-services/context-services-test-eng as I would like some help with testing as this is mostly interfacing with external services

mtbottle (commented):
> Separation of concerns when deploying is achieved through docker build contexts (job_runner and web_api). Using docker build we can specify the final layer of the image to use the context we want.

TIL: [Multistage builds in Docker](https://docs.docker.com/build/building/multi-stage/) 🤯 

I see that the test coverage has decreased from 98->86, so I think we'll probably want to increase the test coverage.

Also, I think the behaviour is a bit interesting, as we download to GCS then process the download from GCS (rather than downloading it directly). Since this is going to live in Merino-py from now on, can we add 1. some documentation on the CLI job (maybe a README in `merino/jobs` or `merino/jobs/wikipedia_indexer`) and 2. update the [architecture diagram](https://github.com/mozilla-services/merino-py/blob/main/docs/intro.md#architecture) to include the indexing workflow?

merino/jobs/wikipedia_indexer/__init__.py
```diff
@@ -0,0 +1,73 @@
...
+
+indexer_cmd = typer.Typer(
+    name="wikipedia-indexer",
+    help="Commands for indexing wikipedia exports into elasticsearch",
+)
+
+
+@indexer_cmd.command()
+def index(
+    elasticsearch_hostname: str = "http://35.192.164.92:9200/",
```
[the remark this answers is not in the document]
quiiver: Yeah I think we can leverage the configs that currently exist within in jobs. I'll look into that

merino/jobs/wikipedia_indexer/suggestion.py
```diff
@@ -0,0 +1,131 @@
...
+            "title": title,
+            "suggest": {
+                "input": inputs,
+                "weight": score,
+            },
+            "suggest-stop": {
+                "input": inputs,
+                "weight": score,
+            },
+        }
```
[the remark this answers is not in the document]
quiiver: yeah the full doc or a subset should be returned in the response via the `_source` field. we can use pydantic models to model that, i will look into what that might look like

merino/jobs/wikipedia_indexer/settings/v1.py
```diff
@@ -0,0 +1,122 @@
+"""Version 1 of the index mapping & settings"""
+
+SUGGEST_MAPPING = {
+    "dynamic": False,
+    "properties": {
```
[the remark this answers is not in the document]
quiiver: as a property of the document?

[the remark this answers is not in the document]
mtbottle: Yep! I think we discussed having it in the document as a property (so add a `version` field), as well as having it in the index alias (ie. `enwiki_v1`). Either way works for me (or we can even have both if you think it's necessary). I think the important thing is that we have a way to specify versioning for non-backwards compatible changes.

merino/jobs/wikipedia_indexer/filemanager.py
```diff
@@ -0,0 +1,142 @@
...
+        return None
+
+    def _parse_date(self, filename: str) -> dt:
+        """Parse datestring out of filename"""
+        date_match = re.match(self.file_pattern, filename)
+        if date_match:
+            try:
+                return dt.strptime(date_match.group(1), "%Y%m%d")
+            except ValueError:
+                pass
```
Trinaa, line 82: I'm surprised this is needed! Having the regex, seems preventative.

quiiver: the regexp doesn't define a valid date format but it could and then this wouldn't be necessary. good call

Trinaa:
@quiiver were you doing some verification (manual, semi-automated) while producing the code for this job? Do you have data sets you used for verification or use cases you repeated for sanity? My thinking is that automating existing scenarios that gave you confidence would be a good jumping off point. 

Otherwise my thinking would be to start bottom-up, with some unit tests, targeting more isolated and decision heavy areas, for example the Scorer.score() method.

merino/jobs/wikipedia_indexer/__init__.py
```diff
@@ -0,0 +1,73 @@
...
+from merino.jobs.wikipedia_indexer.filemanager import FileManager
+from merino.jobs.wikipedia_indexer.indexer import Indexer
+
+logger = logging.getLogger(__name__)
+
+# Shared options
+gcs_path_option = typer.Option(
+    "merino-jobs-dev/wikipedia-exports",
+    "--gcs-path",
+    help="Full gcs path to folder containing wikipedia exports on gcs",
```
ncloudioj, line 17:
```suggestion
    help="Full GCS path to the folder containing Wikipedia exports",
```

merino/jobs/wikipedia_indexer/__init__.py
```diff
@@ -0,0 +1,73 @@
...
+)
+
+gcp_project_option = typer.Option(
+    "contextual-services-dev",
+    "--gcp-project",
+    help="GCP project to use for gcs",
```
ncloudioj, line 23:
```suggestion
    help="GCP project to use for GCS",
```

merino/jobs/wikipedia_indexer/__init__.py
```diff
@@ -0,0 +1,73 @@
...
+)
+
+indexer_cmd = typer.Typer(
+    name="wikipedia-indexer",
+    help="Commands for indexing wikipedia exports into elasticsearch",
```
ncloudioj, line 28:
```suggestion
    help="Commands for indexing Wikipedia exports into Elasticsearch",
```

merino/jobs/wikipedia_indexer/__init__.py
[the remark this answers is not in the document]
ncloudioj: I think we will use environment variables in production, so those job specific settings can be added to Merino config files as usual.

merino/jobs/wikipedia_indexer/__init__.py
```diff
@@ -0,0 +1,73 @@
...
+
+
+@indexer_cmd.command()
+def index(
+    elasticsearch_hostname: str = "http://35.192.164.92:9200/",
+    elasticsearch_alias: str = "enwiki",
+    elasticsearch_username: Optional[str] = None,
+    elasticsearch_password: Optional[str] = None,
+    index_version: str = "v1",
+    total_docs: int = 6_400_00,
```
ncloudioj, line 39: Did you mean `6_400_000`?

merino/jobs/wikipedia_indexer/__init__.py
```diff
@@ -0,0 +1,73 @@
...
+    """Copy file from wikimedia to gcs"""
+    file_manager = FileManager(gcs_path, gcp_project, export_base_url)
+
+    logger.info(
+        "Ensuring latest dump is on GCS",
+        extra={"gcs_path": gcs_path, "gcp_project": gcp_project},
+    )
+    latest = file_manager.stream_latest_dump_to_gcs()
+    if not latest.name:
+        raise RuntimeError("Unable to ensure latest dump on gcs or missing filename.")
```
ncloudioj, line 73:
```suggestion
        raise RuntimeError("Unable to ensure latest dump on GCS or missing file name.")
```

merino/jobs/wikipedia_indexer/filemanager.py
```diff
@@ -0,0 +1,142 @@
...
+
+    def handle_starttag(self, tag: str, attrs: List[Tuple[str, Optional[str]]]) -> None:
+        """When the parser encounters a start tag check for Anchor and push into list."""
+        if tag.lower() == "a":
+            hrefs = [(k, v) for k, v in attrs if v is not None and self._is_href(k, v)]
+            self.file_paths.extend([v for _, v in hrefs])
+
+
+class FileManager:
+    """Tools for managing files on wikimedia export directory and copying into GCS"""
```
ncloudioj, line 38:
```suggestion
    """Tools for managing files on Wikimedia export directory and copying them into GCS"""
```

merino/jobs/wikipedia_indexer/filemanager.py
```diff
@@ -0,0 +1,142 @@
...
+        self.base_url = export_base_url
+        if "/" in gcs_bucket:
+            self.gcs_bucket, self.object_prefix = gcs_bucket.split("/", 1)
+        else:
+            self.gcs_bucket = gcs_bucket
+            self.object_prefix = ""
+
+    def get_latest_dump(self, latest_gcs: Blob) -> str | None:
+        """Find the latest export that's older than the latest on gcs."""
+        if self.base_url is None:
```
ncloudioj, line 60: If `self.base_url` must be present, shall we change its type from `Optional[str]` to `str`?

merino/jobs/wikipedia_indexer/filemanager.py
```diff
@@ -0,0 +1,142 @@
...
+    def __init__(self, filter_wildcard: Pattern[str]) -> None:
+        super().__init__()
+        self.filter = re.compile(filter_wildcard)
+
+    def _is_href(self, k: str, v: str) -> bool:
+        return k == "href" and re.search(self.filter, v) is not None
+
+    def handle_starttag(self, tag: str, attrs: List[Tuple[str, Optional[str]]]) -> None:
+        """When the parser encounters a start tag check for Anchor and push into list."""
+        if tag.lower() == "a":
```
ncloudioj, line 32: Per [API doc](https://docs.python.org/3/library/html.parser.html#html.parser.HTMLParser.handle_starttag), `tag` should already be lower case.
```suggestion
        if tag == "a":
```

merino/jobs/wikipedia_indexer/filemanager.py
```diff
@@ -0,0 +1,142 @@
...
+from google.cloud.storage.fileio import BlobReader, BlobWriter
+
+logger = logging.getLogger(__name__)
+
+
+class DirectoryParser(HTMLParser):
+    """Parse the directory listing to find the specified file."""
+
+    filter: Pattern[str]
+    file_paths: List[str] = []
```
ncloudioj, line 21: Using `[]` as the default value for an instance variable will make all instances of `DirectoryParser` share the same list (see more details [here](https://stackoverflow.com/questions/1132941/least-astonishment-and-the-mutable-default-argument)), which could lead to surprising behaviors when working with multiple `DirectoryParser`s. We can move it to `__init__()` to avoid that:
```
def __init__(self, ...) -> None:
    self.file_paths = []
```

merino/jobs/wikipedia_indexer/filemanager.py
```diff
@@ -0,0 +1,142 @@
...
+        self.filter = re.compile(filter_wildcard)
+
+    def _is_href(self, k: str, v: str) -> bool:
+        return k == "href" and re.search(self.filter, v) is not None
+
+    def handle_starttag(self, tag: str, attrs: List[Tuple[str, Optional[str]]]) -> None:
+        """When the parser encounters a start tag check for Anchor and push into list."""
+        if tag.lower() == "a":
+            hrefs = [(k, v) for k, v in attrs if v is not None and self._is_href(k, v)]
+            self.file_paths.extend([v for _, v in hrefs])
```
ncloudioj, line 34:
```suggestion
            hrefs = [v for k, v in attrs if v is not None and self._is_href(k, v)]
            self.file_paths.extend(hrefs)
```

merino/jobs/wikipedia_indexer/filemanager.py
```diff
@@ -0,0 +1,142 @@
...
+    base_url: Optional[str]
+    gcs_bucket: str
+    object_prefix: str
+    file_pattern: Pattern
+    client: Client
+
+    def __init__(
+        self, gcs_bucket: str, gcs_project: str, export_base_url: Optional[str] = None
+    ) -> None:
+        self.file_pattern = re.compile("^enwiki-(\\d+)-cirrussearch-content.json.gz")
```
ncloudioj, line 49:
```suggestion
        self.file_pattern = re.compile(r"^enwiki-(\d+)-cirrussearch-content.json.gz")
```

merino/jobs/wikipedia_indexer/filemanager.py
```diff
@@ -0,0 +1,142 @@
...
+            self._stream_dump_to_gcs(latest_dump_url)
+        else:
+            logger.info("Currently up to date")
+
+        return latest_gcs
+
+    def stream_latest_from_gcs(self) -> Generator:
+        """Stream latest file from GCS"""
+        for chunk in self._stream_from_gcs(self.get_latest_gcs()):
+            yield chunk
```
ncloudioj, line 109:
```suggestion
        yield from self._stream_from_gcs(self.get_latest_gcs())
```

merino/jobs/wikipedia_indexer/filemanager.py
quiiver: so i looked at this and even with a proper regexp there are situations that an incorrect date could be present so i think its worthwhile to keep the exception catching here

merino/jobs/wikipedia_indexer/filemanager.py
```diff
@@ -0,0 +1,138 @@
...
+            self.gcs_bucket = gcs_bucket
+            self.object_prefix = ""
+
+    def get_latest_dump(self, latest_gcs: Blob) -> Optional[str]:
+        """Find the latest export that's older than the latest on gcs."""
+        resp = requests.get(self.base_url)
+        parser = DirectoryParser(self.file_pattern)
+        parser.feed(str(resp.content))
+        links = parser.file_paths
+        if len(links) == 1:
```
ncloudioj, line 63: What happens if we get more than one link here? Should we grab the most recent one from them?

merino/jobs/wikipedia_indexer/filemanager.py
```diff
@@ -0,0 +1,138 @@
...
+                pass
+        # return a zero date if nothing is found
+        return dt(1, 1, 1)
+
+    def get_latest_gcs(self) -> Blob:
+        """Find the most recent file on GCS"""
+        bucket = self.client.bucket(self.gcs_bucket)
+        blobs: list[Blob] = [b for b in bucket.list_blobs(prefix=self.object_prefix)]
+        blobs.sort(key=lambda b: self._parse_date(str(b.name)))
+        return blobs[-1]
```
ncloudioj, line 88: Just in case that the bucket is empty.
```suggestion
def get_latest_gcs(self) -> Optional[Blob]:
    ...
    return blobs[-1] if blobs else None
```

merino/jobs/wikipedia_indexer/indexer.py
```diff
@@ -0,0 +1,165 @@
...
+# Maximum queue length
+MAX_LENGTH = 5000
+
+
+class Indexer:
+    """Index documents from wikimedia search exports into Elasticsearch"""
+
+    QUEUE_MAX_LENGTH = 5000
+
+    queue: List[Mapping[str, Any]] = []
```
ncloudioj, line 24: Let's init this in the initializer to avoid surprising behavior.

merino/jobs/wikipedia_indexer/indexer.py
```diff
@@ -0,0 +1,165 @@
...
+            raise Exception("could not create the index")
+
+    def _enqueue(self, index_name: str, tpl: Tuple[Mapping[str, Any], ...]):
+        op, doc = self._parse_tuple(index_name, tpl)
+        self.queue.append(op)
+        self.queue.append(doc)
+
+    def _index_docs(self, force: bool):
+        qlen = len(self.queue)
+        if qlen > 0 and (qlen >= self.QUEUE_MAX_LENGTH or force is True):
```
ncloudioj, line 107:
```suggestion
        qlen >= self.QUEUE_MAX_LENGTH or force:
```

merino/jobs/wikipedia_indexer/indexer.py
```diff
@@ -0,0 +1,165 @@
...
+            res = None
+            try:
+                res = self.es_client.bulk(operations=self.queue)
+                if res["errors"] is not False:
```
ncloudioj, line 111:
```suggestion
                if not res["errors"]:
```

merino/jobs/wikipedia_indexer/indexer.py
```diff
@@ -0,0 +1,165 @@
...
+        # operation idempotent from our side)
+        id = op["index"]["_id"]
+        # TODO make this more generic
+        op = {"index": {"_index": index_name, "_id": id}}
+        suggestion = self.suggestion_builder.build(id, dict(doc))
+        return op, suggestion
+
+    def _get_index_name(self, file_name) -> str:
+        if "/" in file_name:
+            file_name = file_name.split("/")[-1]
```
ncloudioj, line 134:
```suggestion
            _, file_name = file_name.rsplit("/", 1)
```

merino/jobs/wikipedia_indexer/indexer.py
```diff
@@ -0,0 +1,165 @@
...
+        base_name = "-".join(file_name.split("-")[:2])
+        return f"{base_name}-{self.index_version}"
+
+    def _ensure_index(self, index_name: str):
```
ncloudioj, line 138:
```suggestion
    def _ensure_index(self, index_name: str) -> dict[str, bool]:
```

merino/configs/default.toml
```diff
@@ -131,3 +133,16 @@ es_max_suggestions = 3
...
 score = 0.23
+
+
+[default.jobs.wikipedia_indexer]
+# Index version that will be written.
+index_version = "v1"
+# Estimate of the total documents in the elasticsearch index.
+total_docs = 6_400_000
+# GCS path. Combined bucket and object prefix (folders).
+gcs_path = "merino-jobs-dev/wikipedia-exports"
```
ncloudioj, line 144: Could you move those dev settings to `development.toml` and leave them as empty strings here?

merino/jobs/wikipedia_indexer/indexer.py
```diff
@@ -0,0 +1,165 @@
...
+    def _index_docs(self, force: bool):
+        qlen = len(self.queue)
+        if qlen > 0 and (qlen >= self.QUEUE_MAX_LENGTH or force is True):
+            res = None
+            try:
+                res = self.es_client.bulk(operations=self.queue)
+                if res["errors"] is not False:
+                    raise Exception(res["errors"])
+            except Exception as e:
+                print(res)
```
ncloudioj, line 114:
```suggestion
                log.warning("failed to index docs", extra=res)
```

merino/jobs/wikipedia_indexer/indexer.py
```diff
@@ -0,0 +1,165 @@
...
+        if not latest.name:
+            raise RuntimeError("No exports available on gcs")
+
+        # parse the index name out of the latest file name
+        print(latest.name)
+        index_name = self._get_index_name(latest.name)
+        logger.info("Ensuring index exists", extra={"index": index_name})
+
+        # TODO add configuration of optional indexing strategies
+        if self._ensure_index(index_name)["acknowledged"]:
```
ncloudioj, line 56: Wonder what would happen if any of the following actions fails, will it leave some bad states behind that might prevent subsequent retries from making progress? I noticed that `_parse_tuple()` was made idempotent, but can "index.refresh()" and "_flip_aliast_to_latest()" achieve the same? If not, shall we retry those post-index failed actions within this job?

merino/jobs/wikipedia_indexer/indexer.py
```diff
@@ -0,0 +1,165 @@
...
+            prior: Optional[Mapping[str, Any]] = None
+            logger.info("Start indexing", extra={"index": index_name})
+            for i, line in enumerate(self.file_manager._stream_from_gcs(latest)):
```
ncloudioj, line 59: Non-issue; just for discussion: If it takes too long to index the entire 6M docs sequentially, I guess we can spawn multiple processes to index them in parallel as this job seems embarrassingly parallel.

ncloudioj (approved):
Looks good, r+ wc.

I think we will probably tune ES mapping and settings later, but that doesn't block us from giving this spin. EXCITING! :shipit:

ncloudioj:
Ah, it passed that 95% test coverage threshold, whew 😼.

Pretty sure @Trinaa and @hackebrot will come to the rescue again to lay out the updated test strategy for Merino Jobs!

hackebrot (commented):
Hi @quiiver! 👋🏻 

As I've mentioned in our previous engineering meeting, I think managing jobs in the **merino-py** repository is a great idea! This makes it easier for us to share code between the web application and the CLI application, such as config and model definitions, and also allows us to use libraries and frameworks that we're already familiar with, such as pydantic, pytest, typing etc. This boosts productivity and long-term maintainability! 🚀 

I checked out your branch locally and familiarized myself with the architecture.  Because it's our very first CLI application for merino-py, there are a few things I'd like us to consider. I'd really like to hear your thoughts on this, too, @ncloudioj!

Right now the amount of shared code is rather negligible, but I think we need to create better separation between the FastAPI web application, the Typer CLI application and code that we want to share between them in the future.

We should strive to make it easy to do the right thing, which in this case means for example **not** import FastAPI app code (routes, middlewares, response models) in the Typer app and vice versa. We currently create the FastAPI web application in `merino/main.py` which is on the same level as the directory containing the CLI application `merino/jobs/`.

A better code organization could look something like this (this is a subset of directories and files we would have - for demonstration purposes):

```text
merino
├── backends
│   ├── accu_weather.py
│   └── remote_settings.py
├── cli
│   └── jobs
│       ├── app.py
│       └── wikipedia_indexer
│           └── commands.py
├── config
│   ├── logging.py
│   └── sentry.py
├── providers
│   ├── adm.py
│   └── weather.py
├── utilities
│   ├── cron.py
│   └── user_agent_parsing.py
└── web
    ├── app.py
    ├── middleware
    │   ├── geo_location.py
    │   └── user_agent.py
    └── routers
        ├── api_v1.py
        └── dockerflow.py
```

I need to think about this a bit more, but wanted to give some feedback here while I'm working on the architecture proposal. I'll note that we had a similar layout in the old merino repository with reusable Rust crates (see https://github.com/mozilla-services/merino).

Because we're already at more than 44 comments on this pull request, I don't think we should be blocking merging this branch on my feedback on code organization and architecture. I'll create a doc and share that with you for feedback.

I do want to draw your attention to @Trinaa's comment about meaningful code coverage above (https://github.com/mozilla-services/merino-py/pull/186#issuecomment-1409468976). I would like to see us increase unit test coverage for the wikipedia-indexer functionality. Without unit tests, we lack the ability to narrow down and debug issues with the code. Let's add more unit tests so we can be more confident in our implementation! 🤖

ncloudioj:
@hackebrot I like that proposal and believe that's the right direction to go.

As we've decided to adopt the monorepo approach for Merino, I guess we will go through some trial and error to get to a desirable state for many things (code structure, testing, CI/CD, documentation, etc). There is also work in progress for Poetry to better support monorepo workflows. So yeah, I am quite hopeful to see this work out for Merino in the long run.

Issue #100, opened 2022-11-01
Refactor Top Picks Provider and Encapsulate Top Picks Backend


data-sync-user:
* Refactor the Top Picks provider into distinct  `TopPicksBacked`, `Protocol` and `Provider` classes so that the structure conforms to the new architecture. This improves extensibility and diminishes tight coupling between the data source and the provider.
* Modify unit tests to accord with division of responsibility.  The provider will query the data that is generated from the backend, proving more robust and less coupled.

Originally: 

In the TopPicks provider, extract the `os` and `json` file IO usage to a new backend object and inject backend object as a dependency, in order to:

* Have structural cohesion between providers, such as the ADM Provider
* Simplify future maintenance with clear delineation of communication interface
* Facilitate unit testing

**Background**
Injection of 3rd party communications is already achieved with the AdmProvider and can be used as an example.

merino/providers/*init*.py

!image-20221020-144014.png|width=66%!

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/DISCO-2084)
┆Attachments: <a href="https://mozilla-hub.atlassian.net/rest/api/2/attachment/content/30338">image-20221020-144014.png</a>
┆Epic: Post MVP for Navigational Queries

data-sync-user:
➤ Taddes Korris commented:

I have a seperate class defined to support inversion on control and further decoupling of the ingestion of the domain list from the provider. Will test extracting this out to either a utility or may just have the class reside at the same level. Should this be useful for other providers, it can be exposed to a wider set of them down the road. For now, keeping it where it is seems wise.

PR #187, opened 2023-01-27
[DISCO-2084] Restructure Top Pick Provider with Backend

taddes:
# Description

Refactor of Top Picks provider to confirm to new architectural standard, with separate 'backend' that handles indexing of local json file and hands this data to a provider, all defined by a `TopPicksBackend` protocol that defines structure and behavior. Emulates the `fetch` function used by AdM provider.

**Changes include:**
- Refactor of Top Pick Provider to segment out logic to an injectable backend, protocol class defining `TopPicksBacked` and `TopPicksData`, and `Provider`.
- Restructure static methods into callable module functions.
- Grouping files into their corresponding provider/backend directories.
- Injection of settings into `Provider` at `providers/__init__.py` call.
- Improved error handling thanks to @Trinaa 's valuable feedback and research.
- Refactor unit tests and add additional unit tests
- Refactor integration tests and add additional integration tests.

# Issue(s)
#100  | [DISCO-2084](https://mozilla-hub.atlassian.net/browse/DISCO-2084)

[DISCO-2084]: https://mozilla-hub.atlassian.net/browse/DISCO-2084?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


taddes:
Have to make some minor modifications to the integration testing suite as the parametrized version of some previous tests doesn't fit the new architecture.

merino/providers/top_picks/backends/top_picks.py
```diff
@@ -0,0 +1,142 @@
...
+        try:
+            if not os.path.exists(file):
+                logger.warning("Local file does not exist")
+                raise FileNotFoundError
+            with open(file, "r") as readfile:
+                domain_list: dict = json.load(readfile)
+                return domain_list
+        except Exception as e:
+            logger.warning("Cannot Process File: {e}")
+            raise e
```
Trinaa, line 64: I notice this PR is still in draft, so please forgive this comment if it's premature. I'd like to visit the Exception handling with an example, following our conversation on Monday.

1. The backends are not responsible for logging. Instead exceptions can be used to propagate messages to the loggers in providers. 

2. In the method `read_domain_list` 
     * When the file doesn't exist, raise the TopPicksError immediately, this will be more efficient than invoking the try/catch context
     * Instead of catching the highest level `Exception`, we can be more specific. According to documentation the [open()](https://docs.python.org/3/library/functions.html#open) method should raise an `OSError` and the [json.load()](https://docs.python.org/3/library/json.html#json.load) method should raise a `JSONDecodeError`. 

3. In the method `fetch`, the try/catch will be catching Exceptions we raised in downstream methods and will be re-raising them, nesting information further in the chain. In this case better to let the exceptions bubble.
     
**Example**
```suggestion
    async def fetch(self) -> TopPicksData:
        """Fetch Top Picks suggestions from domain list."""
        return await asyncio.to_thread(self.build_indices)

    @staticmethod
    def read_domain_list(file: str) -> dict[str, Any]:
        """Read local domain list file"""
        if not os.path.exists(file):
            raise TopPicksError("Local file does not exist")

        try:
            with open(file, "r") as readfile:
                domain_list: dict = json.load(readfile)
                return domain_list
        except OSError as os_error:
            raise TopPicksError(f"Cannot open file '{file}'") from os_error
        except JSONDecodeError as json_error:
            raise TopPicksError(f"Cannot decode file '{file}'") from json_error
```

taddes:
Hey @Trinaa , no not premature at all, was going to continue that discussion with you regarding the exception handling. I'll add what you provided and we'll get a time to discuss it further. Meanwhile, I can at least make these changes.

merino/configs/default.toml
```diff
@@ -107,6 +107,7 @@ score_wikipedia = 0.2
 enabled_by_default = false
 score = 0.25
 query_char_limit = 4
+firefox_char_limit = 2
```
ncloudioj, line 110: Just noticed that we didn't document those settings for this provider, could you add those to this [section](https://mozilla-services.github.io/merino-py/ops.html#provider-configuration)?

ncloudioj: Another ask for you :), while you're here, can you also add validations for those settings? You can reference existing validators [here](https://github.com/mozilla-services/merino-py/blob/main/merino/config.py#L12-L33). Thanks!

merino/providers/top_picks/backends/protocol.py
```diff
@@ -0,0 +1,25 @@
...
+from pydantic import BaseModel
+
+
+class TopPicksData(BaseModel):
+    """Class that holds Top Pick Suggestion Content."""
+
+    primary_index: defaultdict
+    secondary_index: defaultdict
+    short_domain_index: defaultdict
+    results: list[dict]
```
ncloudioj, line 14: Nit: I think we can now help the type analyzer (and the readers) with more detailed annotations.
```suggestion
    primary_index: defaultdict[str, list[int]]
    secondary_index: defaultdict[str, list[int]]
    short_domain_index: defaultdict[str, list[int]]
    results: list[dict[str, Any]]
```

merino/providers/top_picks/backends/top_picks.py
```diff
@@ -0,0 +1,132 @@
...
+class TopPicksBackend:
+    """Backend that indexes and provides Top Pick data."""
+
+    top_picks_file_path: str
+    query_char_limit: int
+    firefox_char_limit: int
+    score: float
+
+    def __init__(
+        self, top_picks_file_path, query_char_limit, firefox_char_limit, score
```
ncloudioj, line 26: This could save the following check on `top_picks_file_path` if it wanted to ensure the path is always specified.  
```suggestion
        self, top_picks_file_path: str, query_char_limit: int, firefox_char_limit: int, score: float
```

merino/providers/top_picks/backends/top_picks.py
```diff
@@ -0,0 +1,132 @@
...
+
+        self.top_picks_file_path = top_picks_file_path
+        self.query_char_limit = query_char_limit
+        self.firefox_char_limit = firefox_char_limit
+        self.score = score
+
+    async def fetch(self) -> TopPicksData:
+        """Fetch Top Picks suggestions from domain list."""
+        top_picks_data: TopPicksData = await asyncio.to_thread(self.build_indices)
+        return top_picks_data
```
ncloudioj, line 40: Nit:
```suggestion
        return await asyncio.to_thread(self.build_indices)
```

merino/providers/top_picks/backends/top_picks.py
```diff
@@ -0,0 +1,132 @@
...
+
+    @staticmethod
+    def read_domain_list(file: str) -> dict[str, Any]:
+        """Read local domain list file"""
+        try:
+            if not os.path.exists(file):
```
ncloudioj, line 46: `open(file, "r")` will throw if the requested file is missing, so maybe we don't need to check its existence here?

merino/providers/top_picks/provider.py
```diff
@@ -0,0 +1,97 @@
...
+    is_top_pick: bool
+
+
+class Provider(BaseProvider):
+    """Top Pick Suggestion Provider."""
+
+    top_picks_data: TopPicksData
+    primary_index: defaultdict = defaultdict(list)
+    secondary_index: defaultdict = defaultdict(list)
+    short_domain_index: defaultdict = defaultdict(list)
```
ncloudioj, line 28: Hmm, why do we duplicate those here? Can we just use those indices from `top_picks_data` instead?

merino/providers/top_picks/provider.py
```diff
@@ -0,0 +1,97 @@
...
+    async def initialize(self) -> None:
+        """Initialize the provider."""
+        try:
+            # Fetch Top Picks suggestions from domain list.
+            self.top_picks_data: TopPicksData = await self.backend.fetch()
+            self.primary_index = self.top_picks_data.primary_index
+            self.secondary_index = self.top_picks_data.secondary_index
+            self.short_domain_index = self.top_picks_data.short_domain_index
+            # Convert result dictionaries into Suggestion objects.
+            self.results: list[Suggestion] = [
```
ncloudioj, line 52: This duplication also seems a little wasteful to me as now we will keep them both in memory. Is it possible to do the conversion at the index building phase? If not, I think we can also consider dropping `top_picks_data.results` once the conversion is done here.

merino/providers/top_picks/provider.py
```diff
@@ -0,0 +1,97 @@
...
+                Suggestion(**result) for result in self.top_picks_data.results
+            ]
+            self.query_min = self.top_picks_data.index_char_range[0]
+            self.query_max = self.top_picks_data.index_char_range[1]
+            self.query_char_limit = self.top_picks_data.query_char_limit
+            self.firefox_char_limit = self.top_picks_data.firefox_char_limit
+
+        except Exception as e:
```
ncloudioj, line 60: Since `backend.fetch()` will only raise `TopPicksError` errors, let's capture those here instead. Also, apart from logging, we might also want to re-throw the exception to let provider initializer handle it rather than silencing it.

tests/integration/api/v1/suggest/test_suggest_top_picks.py
```diff
@@ -5,46 +5,180 @@
...
     assert response.status_code == 200
-
     result = response.json()
-    assert len(result["suggestions"]) == 1
-    assert result
-    assert result["suggestions"][0]["url"] == url
-    assert result["suggestions"][0]["title"] == title
-    assert result["suggestions"][0]["is_top_pick"]
-    assert not result["suggestions"][0]["is_sponsored"]
+    assert Suggestion(**result["suggestions"][0]) == expected_suggestion[0]
```
ncloudioj, line 180:
```suggestion
    assert [Suggestion(**suggestion) for suggestion in result["suggestions"]] == expected_suggestion
```

merino/providers/top_picks/backends/top_picks.py
Trinaa: This or maybe the check can be part of the initialization as part of the value check?

For sure the subsequent action of raising TopPicksError within the try-catch loop is not desirable, as it will trigger the loop context unnecessarily.

tests/integration/api/v1/suggest/test_suggest_top_picks.py
```diff
@@ -5,46 +5,180 @@
...
+        index_char_range=(4, 7),
+        query_char_limit=4,
+        firefox_char_limit=2,
+    )
+
+
+@pytest.fixture(name="backend_mock")
+def fixture_backend_mock(mocker: MockerFixture, top_picks_data: TopPicksData) -> Any:
+    """Create a Top Pick backend mock object for test."""
+    backend_mock: Any = mocker.AsyncMock(spec=TopPicksBackend)
```
ncloudioj, line 142: Why not use a dedicated JSON file for this rather than mocking this with those hardcoded variables? Looks like we already have one in `tests/data/top_picks.json`, we can just add another one for integration tests.

tests/unit/providers/top_picks/conftest.py
```diff
@@ -0,0 +1,153 @@
...
+def fixture_top_picks_parameters() -> dict[str, Any]:
+    """Define Top Pick provider parameters for test."""
+    return {"name": "top_picks", "enabled_by_default": config.enabled_by_default}
+
+
+@pytest.fixture(name="backend_mock")
+def fixture_backend_mock(mocker: MockerFixture, top_picks_data: TopPicksData) -> Any:
+    """Create a Top Pick backend mock object for test."""
+    backend_mock: Any = mocker.AsyncMock(spec=TopPicksBackend)
+    backend_mock.fetch.return_value = top_picks_data
```
ncloudioj, line 144: Ditto: https://github.com/mozilla-services/merino-py/pull/187/files#r1097776152

tests/integration/api/v1/suggest/test_suggest_top_picks.py
```diff
@@ -5,46 +5,180 @@
...
 
 from merino.config import settings
-from merino.providers.top_picks import Provider as TopPicksProvider
-from merino.providers.top_picks import Suggestion
+from merino.providers.top_picks.backends.protocol import TopPicksData
+from merino.providers.top_picks.backends.top_picks import TopPicksBackend
+from merino.providers.top_picks.provider import Provider, Suggestion
 from tests.integration.api.v1.types import Providers
 
+config = settings.providers.top_picks
```
Trinaa, line 20: I suggest not to set config as a global value, but to set it within a test fixture. Globals can break test isolation. Furthermore, you may consider setting these values outright in the test, as opposed to using the configs.

_\* This also applies to `unit/providers/top_picks/conftest.py`_

tests/integration/api/v1/suggest/test_suggest_top_picks.py
```diff
@@ -5,46 +5,180 @@
...
+
+
+@pytest.fixture(name="top_picks_data")
+def fixture_top_picks_data() -> TopPicksData:
```
Trinaa, line 24: Is this fixture a duplication of the one in `unit/providers/top_picks/conftest.py`? If yes, then it's declaration can be moved to a `conftest.py` with visibility to both unit and integration tests, for better re-usability.

merino/providers/top_picks/provider.py
```diff
@@ -0,0 +1,97 @@
...
+            len(srequest.query) < self.firefox_char_limit
+            or len(srequest.query) > self.query_max
+        ):
+            return []
+        if ids := self.primary_index.get(srequest.query):
+            res = self.results[ids[0]]
+            return [res]
+        elif ids := self.secondary_index.get(srequest.query):
+            res = self.results[ids[0]]
+            return [res]
```
ncloudioj, line 95: I feel like this block is getting harder to follow and take a crack at refactoring it:
```py
ids: Optional[int]
query: str = srequest.query

match len(query):
    case qlen if self.firefox_char_limit <= qlen <= (self.query_char_limit - 1):
        ids = self.short_domain_index.get(query)
    case qlen if self.query_char_limit <= qlen <= self.query_max:
        ids = self.primary_index.get(query) or self.secondary_index.get(query)
    case _:
        ids = None 

return [self.results[ids[0]]] if ids else []
```
Let me know what do you think?

tests/integration/api/v1/suggest/test_suggest_top_picks.py
Trinaa: Good call Nan, to take it further and avoid the assertion loop, you could do:
```suggestion
    actual_suggestions: list[Suggestion] = [
        Suggestion(**suggestion) for suggestion in result["suggestions"]
    ]
    assert actual_suggestions == expected_suggestion
```

tests/integration/api/v1/suggest/test_suggest_top_picks.py
```diff
@@ -5,46 +5,180 @@
 """Integration tests for the Merino v1 suggest API endpoint configured with the
```
Trinaa, line 5: I suggest to add a test to cover the unhappy path case, where the `TopPicksBackend` raises a `TopPicksError`. What should the API return? Status 200 with no suggestions?

merino/providers/top_picks/backends/top_picks.py
```diff
@@ -0,0 +1,132 @@
...
+                raise TopPicksError("Local file does not exist")
+            with open(file, "r") as readfile:
+                domain_list: dict = json.load(readfile)
+                return domain_list
+        except OSError as os_error:
+            raise TopPicksError(f"Cannot open file '{file}'") from os_error
+        except JSONDecodeError as json_error:
+            raise TopPicksError(f"Cannot decode file '{file}'") from json_error
```
Trinaa, line 54: Lines 52 and 54 are not covered by testing. Please add a unit test.

tests/unit/providers/top_picks/backends/test_top_picks.py
```diff
@@ -0,0 +1,75 @@
...
+
+def test_local_file_exists() -> None:
+    """Test that the Top Picks Nav Query file exists locally"""
+    assert os.path.exists(settings.providers.top_picks.top_picks_file_path)
+
+
+def test_local_file_not_found(mocker) -> None:
+    """Test that the Top Picks Nav Query file exists locally"""
+    mocker.patch("os.path.exists", return_value=False)
+    assert not os.path.exists(settings.providers.top_picks.top_picks_file_path)
```
Trinaa, line 53: I suggest to remove `test_local_file_exists` and `test_local_file_not_found`, since they do not test the logic within the `TopPicksBackend` module.

tests/unit/providers/top_picks/backends/test_top_picks.py
```diff
@@ -0,0 +1,75 @@
...
+    """Test that the JSON file containing the domain list can be processed"""
+    with pytest.raises(TopPicksError):
+        top_picks.read_domain_list("./wrongfile.json")
+
+
+@pytest.mark.asyncio
+async def test_fetch(top_picks: TopPicksBackend, top_picks_data: TopPicksData):
+    """Test the fetch method returns TopPickData."""
+    result = await top_picks.fetch()
+    assert result == top_picks_data
```
Trinaa, line 75: Nit: missing types
```suggestion
async def test_fetch(top_picks: TopPicksBackend, top_picks_data: TopPicksData) -> None:
    """Test the fetch method returns TopPickData."""
    result: TopPicksData = await top_picks.fetch()
    assert result == top_picks_data
```

tests/unit/providers/top_picks/backends/test_top_picks.py
```diff
@@ -0,0 +1,75 @@
...
+    assert not os.path.exists(settings.providers.top_picks.top_picks_file_path)
+
+
+def test_read_domain_list(top_picks: TopPicksBackend) -> None:
+    """Test that the JSON file containing the domain list can be processed"""
+    domain_list = top_picks.read_domain_list(
+        settings.providers.top_picks.top_picks_file_path
+    )
+    assert domain_list["domains"][0]["domain"] == "example"
+    assert len(domain_list["domains"][1]["similars"]) == 5
```
Trinaa, line 62: In the interest of avoiding IO in unit tests. Consider mocking the behaviour of `open`.
Also pytest is good about comparing dictionaries, so you can probably reduce to a single assertion.

Trinaa (changes requested):
Hey Taddes, thank you for tackling this refactor!
Send me a ping if you'd like to follow-up on any of the comments I left.

merino/configs/default.toml
taddes: Yah, great call! Adding validators for all and documenting them.

merino/providers/top_picks/provider.py
taddes: I'm keeping naming consistency between the two and while it is repeated between the two, it makes accessing the attribute locally simpler. It avoids the awkwardness of longer lines that result clean statements like this:
```python
       if (
            len(srequest.query) < self.firefox_char_limit
            or len(srequest.query) > self.query_max
        ):
            return []
        if ids := self.primary_index.get(srequest.query):
            res = self.results[ids[0]]
            return [res]
```
Turning into:
```python
       if (
            len(srequest.query) < self.top_picks_data.firefox_char_limit
            or len(srequest.query) > self.top_picks_data.query_max
        ):
            return []
        if ids := self.top_picks_data.primary_index.get(srequest.query):
            res = self.top_picks_data.results[ids[0]]
            return [res]

```
which I find more annoying to read. I can change it if you prefer, I just thought it's more explicit this way.

merino/providers/top_picks/provider.py
taddes: Gotcha, only thing is we'd create an anti-pattern because the Backend does not have access to `Suggestion` and it shouldn't, right, given the separation of responsibilities? The construction of Suggestions in other providers happens at the provider level after the 'raw' (in this case, indexed but not Suggestion-ified) data is given to the provider. The alternative is to convert to a suggestion during the query call, what do you think? Is the memory impact sizable? I was going for efficiency in that I wanted to value to be clear and easily referenced in the class so that there's no question what it is, but also that it's ready to provide the suggestion with no conversion like it was in the initial spec.

merino/providers/top_picks/provider.py
ncloudioj: I see your point, though the duplication still seems unnecessary to me. I think you can either split `top_picks_data` into pieces and store them as instance variables (like what you do here but without storing `top_picks_data` as an instance variable) or just use `top_picks_data`, which I'd prefer, with simplified code paths like I proposed [here](https://github.com/mozilla-services/merino-py/pull/187/files#r1097816817).

merino/providers/top_picks/provider.py
ncloudioj: Good call. Actually, I was wondering about the motivation of `TopPicksError`, it doesn't seem to have any specific exceptional behaviors for this backend. Should we just use `BackendError` instead?

tests/integration/api/v1/suggest/test_suggest_top_picks.py
ncloudioj: Let's not hardcode the test dataset this way. Any new change to be made in the future would require lots of manual updates. If the existing test dataset is not sufficient for integration tests, feel free to add a new one as I mentioned [here](https://github.com/mozilla-services/merino-py/pull/187/files#r1097776152).

tests/integration/api/v1/suggest/test_suggest_top_picks.py
```diff
@@ -81,13 +217,12 @@ def test_top_picks_short_domains(
             is_top_pick=True,
             is_sponsored=False,
             icon="",
-            score=settings.providers.top_picks.score,
+            score=0.25,
```
ncloudioj, line 220: If we update the score later, all those tests with `0.25` will break. Let's use the configuration value instead. Please update all other hardcoded scores in this PR as well.

taddes: That's sensible, I'll catch anything that seems like a simple change like this. There are instances already in our other provider tests where there are hard-coded values that could break if changes are made, so we'll just have to look out overall. I'll change whatever I see that seems applicable.

merino/providers/top_picks/backends/top_picks.py
taddes: Yup, got it, just slightly altered assertion and called `del` on the `top_picks_file_path` property in the params fixture to make the `__init__` call still break. Saves a few lines, thanks for the idea!

merino/providers/top_picks/provider.py
taddes: Yeah, solid points. I'll just access all properties off `top_picks_data` then so we only need one `await` call for the main data, should be clear enough. I just changed the schema for `TopPicksData` so the query min and max are separate attributes instead of a tuple, so we're not doing weird index checking that could confuse a reader. Coversion taking place when recommendation returned now.

ncloudioj: Sounds good.

Since now we do conversion at the query time. I think we can drop the `score` field from the raw dict to solve this [review commeint](https://github.com/mozilla-services/merino-py/pull/187/files#r1097775373).

merino/providers/top_picks/provider.py
Trinaa: You could also do that. I think the name/type of the error brings clarity 'TopPicksError', you know exactly which backend raised that, but it's not essential.

merino/providers/top_picks/backends/top_picks.py
taddes: Yeah, I will remove the check since the init will fail if the file is not present. If that's ok with you both. The actual reading of the domain file would throw the TopPicksError anyways, meaning this is probably overkill. We have big problems if the configs point to no file or the wrong file anyway!

merino/providers/top_picks/provider.py
taddes: Ok, I will catch the `BackendError here then raise the Error. Will update.

tests/integration/api/v1/suggest/test_suggest_top_picks.py
taddes: 👍 makes sense, the 0th index is the only piece of data anyway but this is indeed easier to read.

merino/providers/top_picks/backends/top_picks.py
```diff
@@ -0,0 +1,132 @@
...
+    """Error during interaction with Top Picks data."""
+
+
+class TopPicksBackend:
+    """Backend that indexes and provides Top Pick data."""
+
+    top_picks_file_path: str
+    query_char_limit: int
+    firefox_char_limit: int
+    score: float
```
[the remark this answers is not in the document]
taddes: Sure, I know we talked about the merits of this and while it wasn't a big problem to have it in the backend, I've moved it out to provider.

tests/unit/providers/top_picks/backends/test_top_picks.py
taddes: I know there isn't a specific block that this tests, but in the case the test file gets removed or something happens with structure, I don't think it actively hurts anything. Ok if I leave it?

merino/providers/top_picks/backends/top_picks.py
```diff
@@ -0,0 +1,132 @@
...
+
+    def __init__(
+        self, top_picks_file_path, query_char_limit, firefox_char_limit, score
+    ) -> None:
+        """Initialize Top Picks backend."""
```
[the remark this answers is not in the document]
taddes: 👍

tests/integration/api/v1/suggest/test_suggest_top_picks.py
taddes: As we discussed, realized we don't query on demand like the accuweather provider, so no need. Thanks for the discussion on this one!

tests/integration/api/v1/suggest/test_suggest_top_picks.py
taddes: Definitely makes sense, pulled this out and referenced `settings` instead. Good call!

taddes:
@Trinaa Thanks a million again for discussing the pending test questions I had, glad we got to resolve them! As for the idea of putting all the shared fixtures in `conftest.py` at the root of the `tests` dir, I can definitely do it, though I do see the benefit to leaving them be in their own scopes. I know I'd have to change `backend_mock`, which is the fixture name within each provider, and in this case it would be `top_picks_mock`. Maybe easier to leave each scoped to its own context because otherwise I'm really filling up that shared conftest file purely for the benefit of the top picks provider.  I'm ok either way with this, but this just occurred to me after we talked.

merino/providers/top_picks/provider.py
taddes: Hey @ncloudioj . This is a very slick solution, nicely done. I was also playing around with something I thought could be equally readable yet more concise, which is arguably tricky. 

I'm not opposed to changing it to this at all, and though there are a few drawbacks to the existing method at present, it is relatively readable. I have to say given I've now pulled `score` out of the backend and am passing it in after destructuring the TopPicksData, the solution you propose gets a little but more unwieldy. For the sake of accomplishing this refactor, are you comfortable with us tabling this and I can think about it some more? Then we can just change it to this or if I find a solution that is a variation on this theme.  I can make a note and gist of this so we actually revisit this. That or if you think this is the time and place, given you put some serious thought and work into this, we can do it, it just needs a little more tweaking. LMK!

tests/unit/providers/top_picks/conftest.py
taddes: Kat and I discussed this and since we are using mock data objects throughout the other providers, we're opting to not use the file, as it's also nice to avoid the I/O calls. However, what's more important is that it's not so much constructing the mock, but rather having the comparison data to verify validity. We spent a good time on this and think it's probably wisest to keep the test data to compare against segmented in the tests, if you're ok with it.

tests/integration/api/v1/suggest/test_suggest_top_picks.py
taddes: See comment below:

tests/unit/providers/top_picks/conftest.py
ncloudioj: I am not against using mock objects in general, though I do think this particular mock object would be really hard to maintain if we want to update it in the future. Imagine how many fields we need to manually update if we want to add a new site to it. Let's not do that.

taddes: Totally, I get that, but for consistency we'd then need to open a ticket to do this for the other providers that create the same mocks, as they manually too set values to compare against in the same way. Also, Kat and I spent a long time trying to debug when converting the mock backend into one that was read by the file to no avail. I know more time could resolve the issue, but I just want to get this refactor over the finish line and optimize after in subsequent tasks. I'm just concerned about over-optimizing at this point and would rather change this in a task after this is done, if that's ok.

taddes: Ach! @ncloudioj ! I forgot to mention - and I feel this is important so sorry I forgot it - when testing this, I wanted the mocking of objects to emulate how the testing is done across all providers for one. Secondly, I didn't want the 'read' source data from the JSON file to be indexed only to then be compared against itself (the data file), which is why I originally didn't go for reading the file to create the objects. It created a tight coupling as I am both instantiating and testing the same thing, which I felt defeated the purpose of the test.

tests/integration/api/v1/suggest/test_suggest_top_picks.py
Trinaa: I think this data is needed to verify the `top_picks.fetch()` method at unit test level. It is the 'expected' value.
Once it's there... I don't see harm in re-use (we are already stuck maintaining it)?

It is still duplicated though... Why not move it to top level conftest.py?

merino/providers/top_picks/backends/top_picks.py
```diff
@@ -0,0 +1,137 @@
...
+            TopPicksError: If the top picks file path cannot be opened or decoded.
+        """
+        try:
+            with open(file, "r") as readfile:
+                domain_list: dict = json.load(readfile)
+                return domain_list
+        except OSError as os_error:
+            raise TopPicksError(f"Cannot open file '{file}'") from os_error
+        except JSONDecodeError as json_error:
+            raise TopPicksError(f"Cannot decode file '{file}'") from json_error
```
Trinaa, line 59: This case is showing as uncovered in test metrics. Is it possible you removed a test by accident?

tests/unit/providers/top_picks/test_provider.py
```diff
@@ -32,100 +31,56 @@
...
+    assert top_picks.top_picks_data == top_picks_data
 
 
 @pytest.mark.asyncio
 async def test_initialize_exception(top_picks: Provider, mocker) -> None:
-    """Test that proper exception is thrown if initialization is unsuccessful"""
+    """Test that exception is thrown if initialization is unsuccessful."""
     mocker.patch.object(top_picks, "initialize", side_effect=Exception)
     with pytest.raises(Exception):
         await top_picks.initialize()
```
Trinaa, line 57: Remove test.
This test is validating a mock, not the Provider.

Trinaa (commented):
It's a bit lost in the above comments, but the `fixture_top_picks_data()` is still duplicated
[ref](https://github.com/mozilla-services/merino-py/pull/187#discussion_r1106145793)

tests/unit/providers/top_picks/test_provider.py
taddes: Got it! Thx!

merino/providers/top_picks/backends/top_picks.py
taddes: Yeah, I easily added back the `JSONDecodeError` test, thx! Must have removed it during the refactor.

tests/unit/providers/top_picks/backends/test_top_picks.py
```diff
@@ -0,0 +1,68 @@
...
+
+@pytest.fixture(name="top_picks")
+def fixture_top_picks(top_picks_parameters: dict[str, Any]) -> TopPicksBackend:
+    """Create a Top Picks object for test."""
+    return TopPicksBackend(**top_picks_parameters)
+
+
+def test_init_failure_no_domain_file(
+    top_picks_parameters: dict[str, Any]
+) -> TopPicksBackend:
```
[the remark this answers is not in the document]
taddes: Just ended up removing that data entirely. @ncloudioj made great points about just building data from test json file.

taddes:
Ok @ncloudioj @Trinaa , thanks for all the refactor feedback, excited to be nearly done with this! We should now be good to go. I removed the mock test data, the mocked backends and based everything off of the real backend reading the test data. There were more than enough ways to verify the data and suggestions given what is produced by provider and backend instantiation. Removes a lot of needless bloat.

tests/unit/providers/top_picks/backends/test_top_picks.py
```diff
@@ -0,0 +1,88 @@
...
+def test_read_domain_list_os_error(top_picks_backend: TopPicksBackend) -> None:
+    """Test that read domain fails and raises exception with invalid file path."""
+    with pytest.raises(TopPicksError):
+        top_picks_backend.read_domain_list("./wrongfile.json")
+
+
+def test_read_domain_list_json_decode_err(
+    top_picks_backend: TopPicksBackend, mocker
+) -> None:
+    """Test that the read function fails, raising TopPicksError when OSError captured."""
```
Trinaa, line 64:
```suggestion
    """Test that the read function fails, raising TopPicksError when a JSONDecodeError is captured."""
```

tests/integration/api/v1/suggest/test_suggest_top_picks.py
```diff
@@ -5,46 +5,76 @@
...
+    return {
+        "name": "top_picks",
+        "enabled_by_default": settings.providers.top_picks.enabled_by_default,
+        "score": settings.providers.top_picks.score,
+    }
+
+
 @pytest.fixture(name="providers")
-def fixture_providers() -> Providers:
+def fixture_providers(backend: Any, top_picks_parameters: dict[str, Any]) -> Providers:
```
Trinaa, line 49:
```suggestion
def fixture_providers(
    backend: TopPicksBackend, top_picks_parameters: dict[str, Any]
) -> Providers:
```

Trinaa (approved):
LGTM 🚀 ! (two small things I leave to you)

tests/integration/api/v1/suggest/test_suggest_top_picks.py
taddes: 👍

tests/unit/providers/top_picks/backends/test_top_picks.py
taddes: 👍

taddes:
Awesome, thanks @Trinaa ! If there's nothing more @ncloudioj , I will merge after those small changes Kat mentioned.

merino/providers/top_picks/provider.py
```diff
@@ -0,0 +1,86 @@
...
+        self.score = score
+        self._name = name
+        self._enabled_by_default = enabled_by_default
+
+    async def initialize(self) -> None:
+        """Initialize the provider."""
+        try:
+            # Fetch Top Picks suggestions from domain list.
+            self.top_picks_data: TopPicksData = await self.backend.fetch()
+
```
ncloudioj, line 45:
```suggestion
```

merino/providers/top_picks/provider.py
```diff
@@ -0,0 +1,86 @@
...
+        except BackendError as backend_error:
+            logger.warning(
```
ncloudioj, line 47: If a `BackendError` gets raised, this will leave `self.top_picks_data` undefined. We need to how to handle it.
- We can propagate the exception and let the application fail to start
- Or we can ignore it and retry `backend.fetch()` later, though we need to set an empty value for `this.top_picks_data`, otherwise, the subsequent calls on `query()` will fail

Since we're using a local file now, retrying is perhaps not gonna help that much. I think we can go with the first option. If the fetch fails, let's just crash the app.

Does that make sense?

ncloudioj (changes requested):
One more thing we need to handle, other than that, it looks great!

merino/providers/top_picks/provider.py
taddes: Gotcha. Yeah, so shall I simply raise an exception (re raise) within the except block to crash the app? Conversely, I suppose having an empty list of suggestions instantiated,  should the backend data return noting seems decent too, we'd still get the log, but not crash.

ncloudioj: Yeah, re-raise the exception and crash sounds reasonable in this case. If the JSON file is problematic, we should let it fail fast and hopefully we can capture it before it hits production. If this somehow happens in production, the crash is also fine as another merino instance will be spawned later.

taddes: Done @ncloudioj . Re -aises and passes context of `backend_error`, traceback very clear:

``` 
 File "/Users/taddeskorris/Documents/code/mozilla/merino-py/merino/providers/top_picks/provider.py", line 51, in initialize
    raise BackendError(
merino.exceptions.BackendError: Failed to fetch data from Top Picks Backend. Cannot decode file 'dev/top_picks.json'
```

merino/providers/top_picks/provider.py
```diff
@@ -0,0 +1,89 @@
...
+                "Failed to fetch data from Top Picks Backend.",
+                extra={"error message": f"{backend_error}"},
+            )
+            raise BackendError(
```
ncloudioj, line 51: Nit: you can just do `raise backend_error`.

ncloudioj (approved):
Looks great! Just two minor nits.

Thanks! :shipit:

PR #191, opened 2023-02-09
DISCO-2205 fix: decouple accuweather and weather settings

tiftran:


docs/ops.md
```diff
@@ -212,30 +230,15 @@ These are production providers that generate suggestions.
     of Wikipedia suggestions for this provider as a floating point number.
     Defaults to 0.2.
 
-#### Accuweather Provider
+#### Weather Provider
 - Accuweather - Providers weather suggestions & forecasts from Accuweather.
-  - `enabled_by_default` (`MERINO_PROVIDERS__ACCUWEATHER__ENABLED_BY_DEFAULT`) - Whether
+  - `enabled_by_default` (`MERINO_PROVIDERS__WEATHER__ENABLED_BY_DEFAULT`) - Whether
     or not this provider is enabled by default.
```
Trinaa, line 236: We should add a `backend` entry

```
  - `backend` (`MERINO_PROVIDERS__ACCUWEATHER__backend`) - The backend of the provider.
    Either `accuweather` or `test`.
```

docs/ops.md
```diff
@@ -212,30 +230,15 @@ These are production providers that generate suggestions.
     of Wikipedia suggestions for this provider as a floating point number.
     Defaults to 0.2.
 
-#### Accuweather Provider
+#### Weather Provider
 - Accuweather - Providers weather suggestions & forecasts from Accuweather.
```
Trinaa, line 234: Nit
```suggestion
- AccuWeather - Provides weather suggestions & forecasts.
```

docs/ops.md
```diff
@@ -212,30 +230,15 @@ These are production providers that generate suggestions.
...
-  - `enabled_by_default` (`MERINO_PROVIDERS__ACCUWEATHER__ENABLED_BY_DEFAULT`) - Whether
+  - `enabled_by_default` (`MERINO_PROVIDERS__WEATHER__ENABLED_BY_DEFAULT`) - Whether
     or not this provider is enabled by default.
-  - `score` (`MERINO_PROVIDERS__ACCUWEATHER__SCORE`) - The ranking score for this provider
+  - `score` (`MERINO_PROVIDERS__WEATHER__SCORE`) - The ranking score for this provider
     as a floating point number. Defaults to 0.3.
-  - `query_timeout_sec` (`MERINO_PROVIDERS_ACCUWEATHER__QUERY_TIMEOUT_SEC`) - A floating
+  - `query_timeout_sec` (`MERINO_PROVIDERS__WEATHER__QUERY_TIMEOUT_SEC`) - A floating
     point (in seconds) indicating the maximum waiting period when Merino queries Accuweather
     for weather forecasts. This will override the default query timeout
```
Trinaa, line 241: Nit
```suggestion
  - `query_timeout_sec` (`MERINO_PROVIDERS__WEATHER__QUERY_TIMEOUT_SEC`) - A floating
    point (in seconds) indicating the maximum waiting period when Merino queries
    for weather forecasts. This will override the default query timeout.
```

merino/providers/__init__.py
```diff
@@ -29,7 +30,7 @@
 class ProviderType(str, Enum):
     """Enum for provider type."""
 
-    ACCUWEATHER = "accuweather"
+    WEATHER = "weather"
```
Trinaa, line 33: I think this may change the query path that Firefox has to use... 
Running locally I had to change the `provider` to equal "weather" instead of "accuweather" in the URL. Maybe a good idea to coordinate on this.

mtbottle (approved):
LGTM!

Trinaa (changes requested):
I detect a breaking change with the URL, just want to make sure that's checked on.
Thank you for doing this work Tif, the settings make more sense now.

merino/providers/__init__.py
ncloudioj: Let's keep the provider ID unchanged for now as this feature is still under foxfooding.

Trinaa: Yes true, we would still have to remove 'ACCUWEATHER' from Merino, when/if the change is made in Firefox. Leaving it for another day is probably easier. Sry @tiftran Code by Committee...

merino/providers/__init__.py
```diff
@@ -46,16 +47,20 @@ async def init_providers() -> None:
...
-                    backend=AccuweatherBackend(
-                        api_key=setting.api_key,
-                        url_base=setting.url_base,
-                        url_param_api_key=setting.url_param_api_key,
-                        url_postalcodes_path=setting.url_postalcodes_path,
-                        url_postalcodes_param_query=setting.url_postalcodes_param_query,
-                        url_current_conditions_path=setting.url_current_conditions_path,
-                        url_forecasts_path=setting.url_forecasts_path,
+            case ProviderType.WEATHER:
+                providers["weather"] = WeatherProvider(
```
ncloudioj, line 51: Looks like we have two options here:

- Option A: make a duplicate record to cover both cases, and then drop the "accuweather" one once Firefox is done using it. I guess this duplication will break some unit tests, so we will have to handle those as well.
```
provider = WeatherProvider(...)
providers["weather"] = provider
providers["accuweather"] = provider
``` 

- Option B. Just name it "accuweather" (i.e. `providers["accuweather"] = WeatherProvider(...)`) for now and rename it to "weather" later once Firefox changes that provider name.

I'd prefer B, what do you think? @tiftran @Trinaa

ncloudioj (approved):
LGTM, just one comment to be resolved.

merino/providers/__init__.py
tiftran: sooo given that we want to keep "accuweather" as the provider name.
Then we would need to call the configs settings something else? (or i can be completely wrong about this...." `*.providers.accuweather` and `*.accuweather_backend`, how does that sound? @Trinaa @ncloudioj

ncloudioj: I think we can leave those two config sections as is, we can name them whatever we want, only the keys (i.e. the actual provider name) of the `merino.providers.providers` dict matter.

merino/configs/testing.toml
```diff
@@ -30,14 +30,15 @@ client_variant_max = 5
...
-[testing.providers.accuweather]
-# Whether or not this provider is enabled by default.
-enabled_by_default = false
+[testing.accuweather_be]
 api_key = "test"
 # The URL in `url_base` isn't actually pinged during testing but we set it
 # anyway to make sure the production server isn't accidentally pinged.
 url_base = "test://test"
 
+[test.providers.accuweather]
```
ncloudioj, line 39:
```suggestion
[testing.providers.accuweather]
```

merino/configs/default.toml
```diff
@@ -77,9 +77,12 @@ env = "dev"
 traces_sample_rate = 0.1
 
 [default.providers.accuweather]
+backend = "accuweather"
 enabled_by_default = false
 score = 0.3
 query_timeout_sec = 5.0
+
+[default.accuweather_be]
```
ncloudioj, line 85: Noticed that you didn't update those in `docs/op.md`, maybe we just call it `default.accuweather`?

tiftran: eek you're right

merino/providers/__init__.py
Trinaa: I'll leave it up to @tiftran ;)

Trinaa (approved):

ncloudioj:
@tiftran It just occurred to me that before we merge this change, we will need to update the environment variable [here](https://github.com/mozilla-services/cloudops-infra/blob/master/projects/merino/k8s/charts/merinopy/templates/secret.yaml#L15).

PR #192, opened 2023-02-09
Update Merino to grab the talk to the Elastico Elasticsearch

mtbottle:
Elastico requires a Cloud ID and API key to be passed to the client.

The provider times out when I try to connect to it locally, so I've increased the provider's timeout. Once we get an idea of what the distribution for the request looks like, we can reduce the timeout. The connection _should_ be faster in GCP, but not insignificant (around 170ms).

### DIFF
--- a/merino/configs/default.toml
+++ b/merino/configs/default.toml
@@ -119,18 +119,14 @@ es_cloud_id = """
   testing:dXMtZWFzdC0xLmF3cy5mb3VuZC5pbyRjZWM2ZjI2MWE3NGJmMjRjZ
   TMzYmI4ODExYjg0Mjk0ZiRjNmMyY2E2ZDA0MjI0OWFmMGNjN2Q3YTllOTYyNTc0Mw==
 """
-# Elasticsearch url
-es_url = ""
-# Elasticsearch user
-es_user = ""
-# Elasticsearch password
-es_password = ""
+es_api_key = ""
 # Elasticsearch index
-es_index = "enwiki"
+es_index = "enwiki-v1"
 # The maximum suggestions for each search request.
 es_max_suggestions = 3
 # The timeout (in millisecond) for each request to ES
-es_request_timeout_ms = 200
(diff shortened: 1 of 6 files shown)


merino/configs/default.toml
```diff
@@ -119,18 +119,14 @@ es_cloud_id = """
...
+es_api_key = ""
 # Elasticsearch index
-es_index = "enwiki"
+es_index = "enwiki-v1"
 # The maximum suggestions for each search request.
 es_max_suggestions = 3
 # The timeout (in millisecond) for each request to ES
-es_request_timeout_ms = 200
+es_request_timeout_ms = 5000
+query_timeout_sec = 5.0
```
mtbottle, line 129: Setting this arbitrarily to 5 seconds.

tests/unit/providers/wikipedia/test_provider.py
```diff
@@ -0,0 +1,71 @@
+"""Unit tests for the Merino v1 suggest API endpoint for the Wikipedia provider."""
```
mtbottle, line 1: These were moved from backend/test_elastic.py. They were testing the provider, so didn't belong in that file.

Similar to https://github.com/mozilla-services/merino-py/pull/185/files#r1091029958

merino/providers/wikipedia/backends/elastic.py
```diff
@@ -43,7 +43,6 @@ async def search(self, q: str) -> list[dict[str, Any]]:
             res = await self.client.search(
                 index=INDEX_ID,
                 suggest=suggest,
-                timeout=TIMEOUT_MS,
```
ncloudioj, line 46: Now that the ES query timeout is 5sec, did you still want to drop the timeout here?

mtbottle: Ah whoops! I meant to make it the same as the provider. Will add it back!

ncloudioj (approved):
LGTM! :shipit:

PR #194, opened 2023-02-10
[DISCO-2238] refactor: Add a manager for provider creation and registration

ncloudioj:
This adds a new module for provider management. It also adds a `type` field to the provider configuration so that Merino can have multiple providers with the same type, which would be useful for composite providers (e.g. cache providers) in the future. 

This fixes [DISCO-2238](﻿https://mozilla-hub.atlassian.net/browse/DISCO-2238).

[DISCO-2238]: https://mozilla-hub.atlassian.net/browse/DISCO-2238?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


ncloudioj:
> I'm guessing that we have a concrete use case soon (weather provider?).

Yup. For example, we can make a Redis provider by which any other Merino providers can be wrapped.

```
[default.providers.weather]
type = "redis"
[default.providers.weather.inner_provider]
type = "accuweather"
... 
```

merino/providers/manager.py
```diff
@@ -0,0 +1,114 @@
...
+from merino.providers.weather.backends.fake_backends import FakeWeatherBackend
+from merino.providers.weather.provider import Provider as WeatherProvider
+from merino.providers.wiki_fruit import WikiFruitProvider
+from merino.providers.wikipedia.backends.elastic import ElasticBackend
+from merino.providers.wikipedia.backends.fake_backends import FakeWikipediaBackend
+from merino.providers.wikipedia.provider import Provider as WikipediaProvider
+
+
+@unique
+class ProviderType(str, Enum):
```
taddes, line 24: Not sure if we are wanting to use the `auto()` method provided from `enum` to create values for the enum members, but we could technically call `auto()` when assigning each of the provider types in this enum. That way, the name will be the uppercase and the value will be as provided. I don't have any strong attachment and admit the `auto()` can make things harder to read if one doesn't know what it is, but I thought I'd throw that out here. 
```python
    ACCUWEATHER =  auto()
    ADM = auto()
    TOP_PICKS = auto()
    WIKI_FRUIT = auto()
    WIKIPEDIA = auto()

File: tests/integration/test_setup.py [hunk 2]
```diff
@@ -15,12 +15,13 @@
...
+    mocker.patch.dict(
+        settings.providers, values={"unknown-provider": {"type": "unknown-type"}}
+    )
 
     with pytest.raises(InvalidProviderError) as excinfo:
         # This will run all the FastAPI startup event handlers.
         with TestClient(app):
-            # should never make to here
-            assert False
+            ...
```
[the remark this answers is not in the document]
taddes: Hm, weird. I had a similar case with a `return None` being required by MyPy, where it didn't like the implicit return...

merino/config.py
```diff
@@ -15,13 +15,15 @@
     Validator(
         "providers.accuweather.query_timeout_sec", is_type_of=float, gte=0, lte=5.0
     ),
+    Validator("providers.accuweather.type", is_type_of=str, must_exist=True),
```
taddes, line 18: Sorry in advance, but with the Top Picks changes landing in the last merge, you'll need a rebase to pull in the Validators for Top Picks and will just need to add the `providers.top_picks.type` validator.

taddes (commented):
LGTM, added a few comments regarding the enum and Top Picks additions that will be required after recent merge. Happy to approve when those pull in, of course. Are there going to be more changes beyond that to this PR? Good concept, having a manager module!

merino/config.py
ncloudioj: No worries, rebase is cheap :)

tests/integration/test_setup.py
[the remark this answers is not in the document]
ncloudioj: Yup, it's weird. Mypy used to be happy with this, but not any more :/

merino/providers/manager.py
ncloudioj: Hmm, we still need those lowercase IDs since that's what Firefox uses when specifying providers for each request.

mtbottle (approved):
LGTM

PR #195, opened 2023-02-10
Tag and push a `latest` image to docker hub.

quiiver:
I talked to jbuck about creating a new docker hub repo for `merino-jobs` however after some back and forth we decided to test how awkward it is to override the entrypoint in the job. To do that we need a stable tag for the docker image so this pr should tag and push a `latest` when we push to prod


ncloudioj (approved):
Let's give it try then! :shipit:

PR #196, opened 2023-02-15
Add flags to indexing job to help prevent index appending

mtbottle:
Currently, if we run the indexing job on the same GCS file with the same version, it will just the entries to an existing index if one exists. This might not be the behaviour that we want, so I've added 2 things:

1. A flag to determine if the job should allow index append (defaults to False, and fails to index)
2. Version configurable via CMD (so that it's easier to change the version)

@quiiver I'm kind of unsure if this solves the problem that we've discussed, or if it makes some things inconvenient. Let me know what you think!


Makefile
```diff
@@ -164,3 +164,7 @@ docker-compose-down:  ## Run `docker-compose down` in `./dev`
 .PHONY: help
 help:
 	@grep -E '^[a-zA-Z_-]+:.*?## .*$$' $(MAKEFILE_LIST) | sort | awk 'BEGIN {FS = ":.*?## "}; {printf "\033[36m%-20s\033[0m %s\n", $$1, $$2}'
+
+.PHONY: wikipedia-indexer
+wikipedia-indexer:
+	$(POETRY) run python merino/jobs/cli.py $@ ${job}
```
quiiver, line 170: `make` is kind of annoying to interact with cli apps. I recommend either `poetry run merino-jobs` or `poetry shell` to activate the env and `merino-jobs` to run the cli

merino/jobs/wikipedia_indexer/__init__.py
```diff
@@ -26,6 +26,22 @@
     help="GCP project to use for GCS",
 )
 
+version_option = typer.Option(
+    job_settings.index_version, "--version", help="Version of the index"
+)
```
quiiver, line 31: I think its nice having the extra help in the options list but just so you're aware, `typer` will parse the command arguments & types and those will be used as command arguments in the cli app

quiiver (commented):
Append isn't quite right because we use the page_id to have stable document_ids. so a re-index in a live index updates the index in place but doesn't append new data. In ES updates are two ops (delete, index).

My original thought was we'd want a way to create new immutable indexes and should not re-index a live index if possible.

First we'd need to change the index name to include a run sequence identifier:
`enwiki-{date}-{version}-{seq}`

Then the flow would be something like:
```
if index does not exists:
  create index
  run indexing code
else if flag to force new index
  create new index by incrementing the run seq
  run indexing code
else 
  bail on indexing process
```

mtbottle:
> Append isn't quite right because we use the page_id to have stable document_ids. so a re-index in a live index updates the index in place but doesn't append new data.

Ah, I see now. I see that you use the same Wikipedia ID to ensure this. 👍 I'll change the flag to something like `force-new-index` or something then.

> First we'd need to change the index name to include a run sequence identifier: `enwiki-{date}-{version}-{seq}`

Hmm... the sequence identifier feels kind of complicated. I initially wrote this with using the epoch timestamp. So: `enwiki-{timestamp}-{version}`. I opted out of this because the timestamp is kind of hard to understand and if we wanted to have the runs "themed" (ie. experimental... want to try to run with different settings etc.), then the version name gives us more flexibility and readibility. @quiiver what do you think about using timestamp instead of a sequence identifier?

quiiver:
> what do you think about using timestamp instead of a sequence identifier?
Right now the date part of the index name has a 1-1 correlation with the wikipedia export. This was done to make it easy to correlate which index goes to which export. I would want to keep that however if we want to use a timestamp as a run identifier i think that's great.

mtbottle:
Updated this PR to use the timestamp in the index name (so that we don't reuse the same index per run). I've also removed the flag as the chance of name collision should be much lower if we're using timestamp for sequence (and erroring out when there _is_ a collision seems sensible). I'll also update the PR description to reflect the new changes.

quiiver (approved):
nice, lgtm

PR #198, opened 2023-02-15
DISCO-2175 Optimize fetching current conditions and forecast data

tiftran:


merino/providers/weather/backends/accuweather.py
```diff
@@ -92,18 +93,19 @@ async def get_weather_report(
         async with AsyncClient(base_url=self.url_base) as client:
             if not (location := await self.get_location(client, country, postal_code)):
                 return None
+            async with asyncio.TaskGroup() as tg:
+                current = tg.create_task(
+                    self.get_current_conditions(client, location.key)
+                )
+                forecast = tg.create_task(self.get_forecast(client, location.key))
```
Trinaa, line 100: **1. Weak References**
In the [documentation](https://docs.python.org/3/library/asyncio-task.html#creating-tasks) there is a warning about task assignments using _weakref_. I'm not sure if that applies here? Do we need to save references to these tasks because we use the result?

**2. Exception Groups**
In the [documentation](https://docs.python.org/3/library/asyncio-task.html#task-groups) it indicates that multiple exceptions will be combined in an ExceptionGroup. When I force `get_current_conditions` and `get_forecast` to both raise exceptions I get an "Internal Server Error". I think we need to handle this specifically and I think adding a unit test to cover this case would be a good idea.

Pic of my console blowing up 😆 
image

Trinaa (changes requested):

merino/providers/weather/backends/accuweather.py
tiftran: 🙈

ncloudioj: No worries, `except*` comes to the rescue to handle exception groups.

```py
try:
    async with asyncio.TaskGroup() as tg:
    task_current = tg.create_task(self.get_current_conditions(client, location.key))
    task_forecast = tg.create_task(self.get_forecast(client, location.key))
except* AccuweatherError as ae:  # looks like this is the only error we need to handle
    # some error handling
else:
    # happy path, let's grab the results
    current = await current
    forecast = await forecast 
```
Per its doc, if any exception happened to a task, TG will cancel all the ongoing tasks in the group, which is exactly what we

pyproject.toml
```diff
@@ -78,6 +78,7 @@ wrapt = "^1.14.1"
 elasticsearch = {extras = ["async"], version = "^8.5.0"}
 google-cloud-storage = "^2.6.0"
 typer = "^0.7.0"
+asyncio = "^3.4.3"
```
ncloudioj, line 81: `asyncio` is part of python's standard library, no need to install separately.

merino/providers/weather/backends/accuweather.py
Trinaa: No worries @tiftran. I'm learning here along with you and it's great stuff!

tiftran: okie doke, this has been fun. I don't think we want `except* AccuweatherError`, since we're handling an AccuweatherError to raise an AccuweatherError? It also doesn't "disolve"(?) the ExceptionGroup either.
```
>>> 
>>> def f():
...     raise BaseExceptionGroup("unhandled errors in a TaskGroup", [KeyError(), KeyError()])
... 
>>> try:
...     f()
... except* KeyError as e:
...     raise KeyError("hello")
... 
  | ExceptionGroup:  (1 sub-exception)
  +-+---------------- 1 ----------------
    | Exception Group Traceback (most recent call last):
    |   File "<stdin>", line 2, in <module>
    |   File "<stdin>", line 2, in f
    | ExceptionGroup: unhandled errors in a TaskGroup (2 sub-exceptions)
    +-+---------------- 1 ----------------
      | KeyError
      +---------------- 2 ----------------
      | KeyError
      +------------------------------------
    | 
    | During handling of the above exception, another exception occurred:
    | 
    | Traceback (most recent call last):
    |   File "<stdin>", line 4, in <module>
    | KeyError: 'hello'
    +------------------------------------

```

merino/providers/weather/backends/accuweather.py
```diff
@@ -92,18 +94,26 @@ async def get_weather_report(
...
-                return None
-
-        return WeatherReport(
-            city_name=location.localized_name,
-            current_conditions=current,
-            forecast=forecast,
-        )
+            current = await task_current
+            forecast = await task_forecast
+            if current and forecast:
```
Trinaa, line 110: Is this allowed?
```suggestion
if (current := await task_current) and (forecast := await task_forecast):
```

What was the verdict on _weakref_? Does it apply?

merino/providers/weather/backends/accuweather.py
Trinaa: My console doesn't blow up anymore! 🚀 
image

Trinaa (approved):
I don't want to block. Please update the test before merge.

merino/providers/weather/backends/accuweather.py
```diff
@@ -92,18 +94,26 @@ async def get_weather_report(
...
                 return None
+            try:
+                async with asyncio.TaskGroup() as tg:
+                    task_current = tg.create_task(
+                        self.get_current_conditions(client, location.key)
+                    )
+                    task_forecast = tg.create_task(
+                        self.get_forecast(client, location.key)
+                    )
+            except BaseExceptionGroup as e:
```
ncloudioj, line 105: `ExceptionGroup` should be sufficient here.
```suggestion
            except ExceptionGroup as e:
```

merino/providers/weather/backends/accuweather.py
```diff
@@ -92,18 +94,26 @@ async def get_weather_report(
...
+                raise AccuweatherError(f"err:{e.exceptions}")
```
ncloudioj, line 106:
```suggestion
                raise AccuweatherError(f"Failed to fetch weather report: {e.exceptions}")
```

tests/unit/providers/weather/backends/test_accuweather.py
```diff
@@ -5,6 +5,7 @@
 """Unit tests for the AccuWeather backend module."""
 
 import json
+from builtins import BaseExceptionGroup
```
ncloudioj, line 8: s/BaseExceptionGroup/ExceptionGroup/. Also, not need to import it.

merino/providers/weather/backends/accuweather.py
taddes: Very snazzy Pythonic approach @Trinaa 👏

ncloudioj: I think it is. We can do:
```py
    return (
        WeatherReport(
            city_name=location.localized_name,
            current_conditions=current,
            forecast=forecast,
        )
        if (current := await task_current) and (forecast := await task_forecast)
        else None
    )
```

tests/unit/providers/weather/backends/test_accuweather.py
```diff
@@ -399,6 +409,48 @@ async def test_get_weather_report_failed_current_conditions_query(
...
+            ),
+        ),
+    ]
+    mocker.patch.object(AsyncClient, "get", side_effect=side_effects)
+    expected_error_value: str = "unhandled errors in a TaskGroup'"
+
+    with pytest.raises(AccuweatherError) as accuweather_error:
+        await accuweather.get_weather_report(geolocation)
+
+    assert expected_error_value in str(accuweather_error.value)
```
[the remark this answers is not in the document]
tiftran: okie doke

tests/unit/providers/weather/backends/test_accuweather.py
tiftran: oh pycharm lol

PR #202, opened 2023-02-16
[DISCO-2028] Cache weather reports in Redis

linabutler:
This PR adds logic to cache `WeatherReport` responses from AccuWeather in Redis, expiring after 15 minutes by default.

### DIFF
--- a/merino/cache/protocol.py
+++ b/merino/cache/protocol.py
@@ -1,0 +1,33 @@
+"""Protocol for cache adapters."""
+
+from datetime import timedelta
+from typing import Optional, Protocol
+
+
+class CacheAdapter(Protocol):
+    """A protocol describing a cache backend."""
+
+    async def get(self, key: str) -> Optional[bytes]:  # pragma: no cover
+        """Get the value associated with the key. Returns `None` if the key isn't in the cache.
+
+        Raises:
+            - `CacheAdapterError` for cache backend errors.
+        """
+        ...
+
+    async def set(
+        self,
+        key: str,
+        value: bytes,
+        ttl: Optional[timedelta] = None,
+    ) -> None:  # pragma: no cover
+        """Store a key-value pair in the cache, with an optional time-to-live.
+
+        Raises:
+            - `CacheAdapterError` for cache backend errors.
+        """
+        ...
+
+    async def close(self) -> None:  # pragma: no cover
+        """Close the adapter and release any underlying resources."""
+        ...
--- a/merino/configs/development.toml
+++ b/merino/configs/development.toml
@@ -17,7 +17,7 @@ dev_logger = true
 level = "DEBUG"
 format = "pretty"
 
-[development.providers.accuweather]
+[development.accuweather]
 api_key = "test"
 
 [development.jobs.wikipedia_indexer]
--- a/merino/providers/weather/backends/accuweather.py
+++ b/merino/providers/weather/backends/accuweather.py
@@ -77,6 +77,15 @@ class AccuweatherBackend:
         self.url_current_conditions_path = url_current_conditions_path
         self.url_forecasts_path = url_forecasts_path
 
+    def cache_inputs_for_weather_report(self, geolocation: Location) -> Optional[bytes]:
+        """Return the inputs used to form the cache key for looking up and storing the current
+        conditions and forecast for a location.
+        """
+        if geolocation.country is None or geolocation.postal_code is None:
+            return None
+
+        return (geolocation.country + geolocation.postal_code).encode("utf-8")
+
     async def get_weather_report(
         self, geolocation: Location
     ) -> Optional[WeatherReport]:
--- a/merino/providers/weather/provider.py
+++ b/merino/providers/weather/provider.py
@@ -1,8 +1,17 @@
 """Weather integration."""
+import hashlib
+import json
 import logging
+from datetime import timedelta
 from typing import Any, Optional
 
-from merino.exceptions import BackendError
+from merino.cache.protocol import CacheAdapter
+from merino.exceptions import (
+    BackendError,
+    CacheAdapterError,
+    CacheEntryError,
+    CacheMissError,
+)
 from merino.middleware.geolocation import Location
 from merino.providers.base import BaseProvider, BaseSuggestion, SuggestionRequest
 from merino.providers.weather.backends.protocol import (
@@ -27,21 +36,27 @@ class Provider(BaseProvider):
     """Suggestion provider for weather."""
 
     backend: WeatherBackend
+    cache: CacheAdapter
     score: float
+    cached_report_ttl_sec: int
 
     def __init__(
         self,
         backend: WeatherBackend,
+        cache: CacheAdapter,
         score: float,
         name: str,
         query_timeout_sec: float,
+        cached_report_ttl_sec: int,
         enabled_by_default: bool = False,
         **kwargs: Any,
     ) -> None:
         self.backend = backend
+        self.cache = cache
         self.score = score
         self._name = name
         self._query_timeout_sec = query_timeout_sec
+        self.cached_report_ttl_sec = cached_report_ttl_sec
         self._enabled_by_default = enabled_by_default
         super().__init__(**kwargs)
 
@@ -52,14 +67,78 @@ class Provider(BaseProvider):
     def hidden(self) -> bool:  # noqa: D102
         return False
 
+    def cache_key_for_weather_report(self, geolocation: Location) -> Optional[str]:
+        """Compute a Redis key used to look up and store cached weather reports for a location."""
+        cache_inputs: Optional[bytes] = self.backend.cache_inputs_for_weather_report(
+            geolocation
+        )
+        if not cache_inputs:
+            return None
+
+        return f"{self.name}:v1:report:{hashlib.blake2s(cache_inputs).hexdigest()}"
+
+    async def fetch_cached_weather_report(
+        self, geolocation: Location
+    ) -> Optional[WeatherReport]:
+        """Fetch a cached weather report, if available, for a location.
+
+        Raises:
+            - `CacheMissError` if there's no entry in the cache for this location.
+            - `CacheEntryError` if the cached weather report can't be deserialized.
+        """
+        cache_key: Optional[str] = self.cache_key_for_weather_report(geolocation)
+        if not cache_key:
+            raise CacheMissError
+
+        cache_value: Optional[bytes] = await self.cache.get(cache_key)
+        if not cache_value:
+            raise CacheMissError
+
+        try:
+            weather_report_dict = json.loads(cache_value)
+            if not weather_report_dict:
+                return None
+            return WeatherReport.parse_obj(weather_report_dict)
+        except ValueError as exc:
+            # `ValueError` is the common superclass of `json.JSONDecodeError` and
+            # `pydantic.ValidationError`.
+            raise CacheEntryError("Failed to parse cache entry") from exc
+
+    async def store_cached_weather_report(
+        self, geolocation: Location, weather_report: Optional[WeatherReport]
+    ):
+        """Store a cached weather report, or the absence of one, for a location."""
+        cache_key: Optional[str] = self.cache_key_for_weather_report(geolocation)
+        if not cache_key:
+            return
+
+        # If the request to the backend succeeded, but didn't return a report, we want to
+        # negatively cache an empty value, so that subsequent requests for that location won't make
+        # additional backend calls every time. This case is separate from a transient backend
+        # error, which isn't negatively cached.
+        cache_value = weather_report.json().encode("utf-8") if weather_report else b"{}"
+        await self.cache.set(
+            cache_key, cache_value, ttl=timedelta(seconds=self.cached_report_ttl_sec)
+        )
+
     async def query(self, srequest: SuggestionRequest) -> list[BaseSuggestion]:
         """Provide weather suggestions."""
         geolocation: Location = srequest.geolocation
         weather_report: Optional[WeatherReport] = None
+
         try:
-            weather_report = await self.backend.get_weather_report(geolocation)
-        except BackendError as backend_error:
-            logger.warning(backend_error)
+            weather_report = await self.fetch_cached_weather_report(geolocation)
+        except (CacheAdapterError, CacheEntryError, CacheMissError) as exc:
+            if not isinstance(exc, CacheMissError):
+                logger.warning(f"Failed to load cached weather report: {exc}")
+            try:
+                weather_report = await self.backend.get_weather_report(geolocation)
+                try:
+                    await self.store_cached_weather_report(geolocation, weather_report)
+                except CacheAdapterError as exc:
+                    logger.warning(f"Failed to store cached weather report: {exc}")
+            except BackendError as backend_error:
+                logger.warning(backend_error)
 
         if weather_report is None:
             return []
@@ -76,3 +155,7 @@ class Provider(BaseProvider):
                 forecast=weather_report.forecast,
             )
         ]
+
+    async def shutdown(self) -> None:
+        """Shut down the provider."""
+        await self.cache.close()
--- a/tests/integration/api/v1/suggest/test_suggest_weather.py
+++ b/tests/integration/api/v1/suggest/test_suggest_weather.py
@@ -13,6 +13,7 @@ from fastapi.testclient import TestClient
 from pytest import LogCaptureFixture
 from pytest_mock import MockerFixture
 
+from merino.cache.none import NoCacheAdapter
 from merino.exceptions import BackendError
 from merino.providers.weather.backends.protocol import (
     CurrentConditions,
@@ -29,7 +30,9 @@ from tests.types import FilterCaplogFixture
 @pytest.fixture(name="backend_mock")
 def fixture_backend_mock(mocker: MockerFixture) -> Any:
     """Create a WeatherBackend mock object for test."""
-    return mocker.AsyncMock(spec=WeatherBackend)
+    backend_mock = mocker.AsyncMock(spec=WeatherBackend)
+    backend_mock.cache_inputs_for_weather_report.return_value = None
+    yield backend_mock
 
 
 @pytest.fixture(name="providers")
@@ -38,9 +41,11 @@ def fixture_providers(backend_mock: Any) -> Providers:
     return {
         "weather": Provider(
             backend=backend_mock,
+            cache=NoCacheAdapter(),
             score=0.3,
             name="weather",
             query_timeout_sec=0.2,
+            cached_report_ttl_sec=10,
             enabled_by_default=True,
         )
     }
--- a/tests/unit/providers/weather/backends/test_accuweather.py
+++ b/tests/unit/providers/weather/backends/test_accuweather.py
@@ -787,3 +787,37 @@ async def test_get_forecast_error(
         await accuweather.get_forecast(mock_client, location_key)
 
     assert str(accuweather_error.value) == expected_error_value
+
+
+@pytest.mark.parametrize(
+    "cache_inputs_by_location",
+    [
+        (Location(country="US", region="CA", city="San Francisco", dma=807), None),
+        (
+            Location(region="CA", city="San Francisco", dma=807, postal_code="94105"),
+            None,
+        ),
+        (
+            Location(
+                country="US",
+                region="CA",
+                city="San Francisco",
+                dma=807,
+                postal_code="94105",
+            ),
+            b"US94105",
+        ),
+    ],
+)
+@pytest.mark.asyncio
+async def test_cache_inputs_for_weather_report(
+    accuweather: AccuweatherBackend,
+    cache_inputs_by_location: tuple[Location, Optional[bytes]],
+) -> None:
+    """Test that `cache_inputs_for_weather_report` computes the correct cache inputs when
+    given locations with various missing fields.
+    """
+    cache_inputs: Optional[bytes] = accuweather.cache_inputs_for_weather_report(
+        cache_inputs_by_location[0]
+    )
+    assert cache_inputs == cache_inputs_by_location[1]
--- a/tests/unit/providers/weather/test_provider.py
+++ b/tests/unit/providers/weather/test_provider.py
@@ -9,7 +9,9 @@ from typing import Any
 import pytest
 from pytest import LogCaptureFixture
 from pytest_mock import MockerFixture
+from redis.asyncio import Redis
 
+from merino.cache.redis import RedisAdapter
 from merino.config import settings
 from merino.exceptions import BackendError
 from merino.middleware.geolocation import Location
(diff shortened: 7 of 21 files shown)


merino/configs/development.toml
```diff
@@ -17,7 +17,7 @@ dev_logger = true
 level = "DEBUG"
 format = "pretty"
 
-[development.providers.accuweather]
+[development.accuweather]
```
tiftran, line 20: oops tyty

merino/providers/__init__.py
```diff
@@ -63,9 +65,11 @@ async def init_providers() -> None:
                         if setting.backend == "accuweather"
                         else FakeWeatherBackend()
                     ),
+                    redis=Redis.from_url(settings.redis.server),
```
ncloudioj, line 68: Per [API doc](https://redis-py.readthedocs.io/en/stable/examples/asyncio_examples.html#Connecting-and-Disconnecting), we should close the connection pool when the app is ready to shut down. Each Merino provider comes with a `shutdown()` function, which will be automatically called at the app showdown, we can do the closing over there :)

There is an example in the [Wikipedia provider](https://github.com/mozilla-services/merino-py/blob/main/merino/providers/wikipedia/provider.py#L86-L88).

merino/configs/development.toml
ncloudioj: oops 😬

merino/providers/weather/backends/accuweather.py
```diff
@@ -76,6 +76,17 @@ def __init__(
...
+    def cache_inputs_for_weather_report(self, geolocation: Location) -> Optional[bytes]:
+        """Return the inputs used to form the cache key for looking up and storing the current
+        conditions and forecast for a location.
+        """
+        if geolocation.country is None or geolocation.postal_code is None:
+            return None
+
+        return geolocation.country.encode("utf-8") + geolocation.postal_code.encode(
+            "utf-8"
+        )
```
ncloudioj, line 88: Perhaps overkill, but python also has [pattern matching](https://peps.python.org/pep-0622/)!
```suggestion
        match geolocation:
            case Location(country, postal_code) | if country is None or postal_code is None:
                 return None
             case Location(country, postal_code):
                 return f"{country}{postal_code}".encode("utf-8")
```

merino/providers/weather/provider.py
```diff
@@ -52,14 +74,71 @@ async def initialize(self) -> None:
...
+
+    async def store_cached_weather_report(
+        self, geolocation: Location, weather_report: Optional[WeatherReport]
+    ):
+        """Store a cached weather report, or the absence of one, for a location."""
+        cache_key: Optional[str] = self.cache_key_for_weather_report(geolocation)
+        if not cache_key:
+            return
+
+        cache_value = json.dumps(weather_report.dict() if weather_report else {})
```
ncloudioj, line 118:
```suggestion
        cache_value = weather_report.json() if weather_report else "{}"
```
Also, shall we not cache the empty value so that Merino can try fetching from the upstream soon?

merino/providers/weather/provider.py
```diff
@@ -52,14 +74,71 @@ async def initialize(self) -> None:
...
         """Provide weather suggestions."""
         geolocation: Location = srequest.geolocation
         weather_report: Optional[WeatherReport] = None
+
         try:
-            weather_report = await self.backend.get_weather_report(geolocation)
-        except BackendError as backend_error:
-            logger.warning(backend_error)
+            weather_report = await self.fetch_cached_weather_report(geolocation)
+        except (RedisError, BadCacheEntry, CacheMissError) as exc:
```
ncloudioj, line 128: A little unfortunate that we had to use exceptions to represent those responses. Python's enum can't do :/

```
enum CacheResponse:
    Miss
    BadEntry
    RedisError
    Hit(Entry)
```

merino/providers/weather/provider.py
```diff
@@ -52,14 +74,71 @@ async def initialize(self) -> None:
...
+            if not isinstance(exc, CacheMissError):
+                logger.warning(f"Failed to load cached weather report: {exc}")
+            try:
+                weather_report = await self.backend.get_weather_report(geolocation)
+                try:
+                    # The call to `get_weather_report` can succeed, but return `None`. In that
+                    # case, we want to cache the absence of the report, so that subsequent requests
+                    # for that location aren't forwarded to the backend.
+                    await self.store_cached_weather_report(geolocation, weather_report)
```
ncloudioj, line 137: Non-issue; just noting: the provider doesn't need to wait for cache writes to proceed, so we can dispatch this to a background cache writer task to reduce the query latency in the future.

merino/providers/weather/provider.py
```diff
@@ -52,14 +74,71 @@ async def initialize(self) -> None:
...
+                except Exception as exc:
+                    logger.warning(f"Failed to store cached weather report: {exc}")
+            except BackendError as backend_error:
```
ncloudioj, line 140: Looks like this `backendError` should only apply to the `get_weather_report()` call, we can reduce its scope to alleviate the nested try-except here.
```
    try:
        weather_report = await self.backend.get_weather_report(geolocation)
    except BackendError as backend_error:
        logger.warning(backend_error)

    if weather_report:
       try:
           await self.store_cached_weather_report(geolocation, weather_report)
       except Exception as exc:
           logger.warning(f"Failed to store cached weather report: {exc}")
```

merino/providers/weather/provider.py
```diff
@@ -52,14 +74,71 @@ async def initialize(self) -> None:
...
+
+        cache_value: Optional[bytes] = await self.redis.get(cache_key)
+        if not cache_value:
+            raise CacheMissError
+
+        try:
+            weather_report_dict = json.loads(cache_value)
+            if not weather_report_dict:
+                return None
+            return WeatherReport.parse_obj(weather_report_dict)
```
ncloudioj, line 106: Pydantic also supports [`parse_raw()`](https://docs.pydantic.dev/usage/models/#helper-functions) which may save that json step for us.

ncloudioj:
Looks good. \o/

A couple of follow-up questions:
- This adds Redis as a hard dependency, we will have to run a Redis server even for local development. Shall we make it optional? For example, we can disable this cache layer via configuration, or use an in-memory cache as the alternative backend
- @Trinaa poke poke 😁, would love to hear your thoughts about testing ergonomics. For simplicity, the cache logic is baked directly inside the weather provider, and we mock Redis calls for testing.

merino/providers/weather/provider.py
linabutler: > shall we not cache the empty value so that Merino can try fetching from the upstream soon?

Great question—I wanted to ask you about that, actually! 😅 `get_weather_report`'s signature suggests that it can succeed but return `None`:

```py
async def get_weather_report(self, geolocation: Location) -> Optional[WeatherReport]
```

And I noticed we have a [test](https://github.com/mozilla-services/merino-py/blob/9a6e0ec40c7d9610a60061b98a17cf42db64355e/tests/unit/providers/weather/test_provider.py#L118-L132) for that behavior. I was wondering if that's intentional—like if AccuWeather doesn't know the location we're requesting, or doesn't have enough data for it?

I opted to explicitly cache the empty value in case that's intentional, because it didn't seem useful to continue sending requests to the upstream if there's no way it can return a report for it. But, if returning `None` means a transient failure, I wonder if we should also change the signature to:

```py
async def get_weather_report(self, geolocation: Location) -> WeatherReport
```

...And have `get_weather_report` raise an error if it can't return a complete report (and not cache that result). WDYT? 😄 (/cc @Trinaa, too)

merino/providers/weather/provider.py
linabutler: Right? 💔 I asked about that in [this `#backend` thread](https://mozilla.slack.com/archives/C02N82L2JMN/p1676417266522259), and our Python Brain Trust® agreed that an exception was probably the way to go.

merino/providers/weather/provider.py
linabutler: Good catch! I had a TODO for that in the first take (the generic `RedisCachingProvider`), but forgot to carry it over.

merino/providers/weather/provider.py
linabutler: Check out my [note here](https://github.com/mozilla-services/merino-py/pull/202#discussion_r1110097305)! 😄 If we decide not to cache empty results, we can for sure reduce the scope of the `except`.

But, if we keep it, I think we need something like this, to tell the difference between "`weather_report` is `None` because `get_weather_report` threw a `BackendError`" (so don't cache it) and "`weather_report` is `None` because `get_weather_report` returned `None`" (cache it).

merino/providers/weather/provider.py
linabutler: Yep! I liked the symmetry between `json.dumps(model.dict())` and `json.loads(Model.parse_obj(dict))`, so I spelled it out explicitly—I didn't see a `model.raw()` method to match `Model.parse_raw`. I don't feel strongly about it at all, though—do you have a preference?

merino/providers/weather/backends/accuweather.py
linabutler: Oooh, fancy—I like how that reads! 😄 Thanks!

linabutler:
Thanks for the review, @ncloudioj! 🚀

> Shall we make it optional? For example, we can disable this cache layer via configuration, or use an in-memory cache as the alternative backend.

That sounds good to me! We can have a `CacheBackend` protocol with a Redis implementation and a fake implementation.

merino/providers/weather/provider.py
ncloudioj: Oh, speaking of symmetry, see my other [comment](https://github.com/mozilla-services/merino-py/pull/202/files#r1110000833). So we can do `weather_report.json()` and `WeatherReport.parse_raw()` to save two intermediate steps. Not a strong preference though, it's up to you :)

merino/providers/weather/provider.py
ncloudioj: No worries, we can leave it as is.

merino/providers/weather/provider.py
ncloudioj: Through another discussion offline, we decided to stick to the current way of caching empty resutls.

merino/providers/weather/provider.py
ncloudioj: Yup, since we will cache results, let's leave it as is.

merino/providers/weather/provider.py
linabutler: Oooh, I completely missed that you used `weather_report.json()` there, thank you very much!

> I didn't see a `model.raw()` method to match `Model.parse_raw`.

...Didn't look hard enough! 🙈

merino/providers/weather/provider.py
linabutler: Sounds good! I'll tweak the comment and docstrings to explicitly mention "negative caching" also.

merino/providers/weather/backends/accuweather.py
linabutler: I tried writing this a few different ways—in addition to yours, this was one of the others:

```py
match (geolocation.country, geolocation.postal_code):
    case (None, _) | (_, None):
        return None
    case (country, postal_code):
        return f"{country}{postal_code}".encode("utf-8")
```

...But I couldn't figure out how to teach Mypy that `country` and `postal_code` are non-optional in the second arm! Even with various checks for `None` in the first arm, it still thinks both are optionals in the second arm, and also complains about a missing return statement—it doesn't see that the match is exhaustive.

linabutler: I found a few similar-but-not-quite-the-same reports in the Mypy issue tracker, and filed python/mypy#14731 for completeness.

ncloudioj: Of course, mypy ;)

We've seen this kind of false alarm with pattern matching before. Had to add a [dummy wildcard](https://github.com/mozilla-services/merino-py/blob/main/merino/utils/user_agent_parsing.py#L33) to appease mypy, yet also had to exclude that from coverage checking.

I will let you decide which route to take, feel free to stick to your original version. Sorry for taking you on this detour 🙇‍♂️

merino/providers/weather/provider.py
```diff
@@ -52,14 +74,71 @@ async def initialize(self) -> None:
...
+        except ValueError as exc:
```
ncloudioj, line 107: `Model.parse_*` will throw `ValidationError` if the validation fails. Let's catch that as well. 
```suggestion
        # from pydantic import ValidationError
        except (ValidationError, ValueError) as exc:
```

merino/providers/weather/provider.py
```diff
@@ -52,14 +74,71 @@ async def initialize(self) -> None:
...
 
+    def cache_key_for_weather_report(self, geolocation: Location) -> Optional[str]:
+        """Compute a Redis key used to look up and store cached weather reports for a location."""
+        cache_inputs = self.backend.cache_inputs_for_weather_report(geolocation)
+        if not cache_inputs:
+            return None
+
+        return f"{self.name}:v1:report:{hashlib.blake2s(cache_inputs).hexdigest()}"
+
+    async def fetch_cached_weather_report(
```
quiiver, line 85: It would be nice to have metrics reported around cache miss and hit rates here

tests/integration/api/v1/suggest/test_suggest_weather.py
```diff
@@ -90,6 +99,7 @@ def test_suggest_with_weather_report(client: TestClient, backend_mock: Any) -> N
             forecast=weather_report.forecast,
         )
     ]
+    backend_mock.cache_inputs_for_weather_report.return_value = None
```
Trinaa, line 102: This `return_value` assignment is consistent across tests, an option could be to assign it once in the fixture_backend_mock.

Alternatively, were you thinking we could have an integration test where we simulate the cache workflow?

merino/providers/weather/provider.py
linabutler: `ValueError` is a superclass of both `pydantic.ValidationError` and `json.JSONDecodeError`! 😄

merino/providers/weather/provider.py
linabutler: Great call-out, thank you!

merino/providers/weather/backends/accuweather.py
linabutler: That's okay! I'll go ahead and leave it as it is; with all the changes to make Mypy happy, this function ends up being twice as long. 😅

tests/unit/providers/weather/test_provider.py
```diff
@@ -161,3 +173,285 @@ async def test_query_error(
...
+@pytest.mark.asyncio
+async def test_query_redis_unavailable(
+    redis_mock: Any,
+    backend_mock: Any,
+    provider: Provider,
+    geolocation: Location,
+) -> None:
+    """Test that Redis errors don't prevent requests to the backend."""
+    redis_mock.get.side_effect = RedisError("mercury in retrograde")
+    redis_mock.set.side_effect = RedisError("synergies not aligned")
```
Trinaa, line 405: ROFL! 😹  Love these error messages!!!!

Trinaa (commented):
So sorry for my tardiness in my comments. Similarly I will be OOF the next couple days and so will trust you to take or leave anything I've said.

Thank you for this work, especially the thoughtful tests!

merino/providers/weather/provider.py
ncloudioj: To minimize the scope of this PR, we're going to add metrics in [DISCO-2029](https://mozilla-hub.atlassian.net/browse/DISCO-2029).

merino/providers/weather/provider.py
ncloudioj: Ah, I see. Could you add a quick comment to note that for future reference?

merino/providers/weather/provider.py
linabutler: Oooh, right, I forgot we had DISCO-2029 on file for a follow-up—thanks!

merino/providers/weather/provider.py
linabutler: Done!

merino/providers/weather/provider.py
linabutler: I switched to using `weather_report.json()` in `store_cached_weather_report`, but didn't switch to `parse_raw` here, because `parse_raw` will raise a validation error for an empty dict, while we want to return `None`.

We could check for `cache_value == b"{}"` before calling `parse_raw`, but that feels a little too clever compared to `json.loads` followed by `parse_obj`. 😅

merino/providers/weather/provider.py
linabutler: Sure thing! I filed DISCO-2252 for post-MVP.

merino/providers/weather/backends/accuweather.py
```diff
@@ -76,6 +76,17 @@ def __init__(
         self.url_current_conditions_path = url_current_conditions_path
         self.url_forecasts_path = url_forecasts_path
 
+    def cache_inputs_for_weather_report(self, geolocation: Location) -> Optional[bytes]:
```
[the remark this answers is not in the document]
linabutler: > Would it make sense for this value to be part of the `Location` object and assigned in the `GeolocationMiddleware`?

Great suggestion! I thought about that, and also about making this be a computed `@property` or method on `Location`. In the end, I decided against it, because the way we compute the cache key inputs seems specific to a weather backend, and I wanted for that logic to live in the same place.

If we ever add another backend that returns results at a different granularity, or if we want to add caching to another provider, they might have different ways of computing the cache key inputs, using sources in addition to (or instead of) the location.

Thinking about it from that perspective, adding methods like `cache_inputs_for_accuweather`, etc. to `Location` felt clunkier, even though the only cache inputs we use in this case are from the `Location` object.

> This method may need some tests...

Yep, I think it needs an addition to `test_accuweather.py`! 😊

ncloudioj:
@linabutler Doh, sorry, the PR I merged yesterday caused lots of conflicts with yours, some rebasing is needed :/

This looks good to me. Did you want to add the `CacheBackend` and a simple in-memory cache to this PR as well? If you wanted to do that for a followup, that's also fine, but we will need to wire the provisioned Redis server with Merino in the cloudops-infra repo.

linabutler:
@ncloudioj Yep, I just pushed a sketch of a `CacheAdapter` protocol, with Redis and in-memory implementations. WDYT? I think I'm happy with the general shape of it; if you are, too, I'll go ahead and take care of the remaining nits, TODOs, and tests.

merino/cache/no.py
```diff
@@ -0,0 +1,24 @@
+"""..."""
+from datetime import timedelta
+from typing import Optional, Union
+
+
+class NoCacheAdapter:
```
ncloudioj, line 6: Imagine `from merino.cache import no; no.NoCacheAdapter`. What a cold module 🤣

ncloudioj:
> Yep, I just pushed a sketch of a CacheAdapter protocol, with Redis and in-memory implementations. WDYT? I think I'm happy with the general shape of it; if you are, too, I'll go ahead and take care of the remaining nits, TODOs, and tests.

LGTM 👍

merino/cache/no.py
linabutler: 😂😂😂

merino/providers/weather/backends/accuweather.py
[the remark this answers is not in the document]
Trinaa: Oh I see your point! The added contextual value of having the method in AccuWeather may come in handy!

tests/integration/api/v1/suggest/test_suggest_weather.py
linabutler: I went ahead and assigned it in the `fixture_backend_mock`!

Hmm, we'd still need to mock out Redis in our integration tests, right? My gut feeling is that simulating the cache workflow wouldn't tell us more than the unit tests we already have, or the contract tests with a real Redis container that we can add later—WDYT?

linabutler:
@ncloudioj Ready for re-review! I'll fold the `fixup!` commits into the earlier commits before merging, but I kept them separate for now to make it easier to see the interdiff.

merino/cache/protocol.py
```diff
@@ -0,0 +1,25 @@
...
+
+from datetime import timedelta
+from typing import Optional, Protocol, Union
+
+
+class CacheAdapter(Protocol):
+    """A protocol describing a cache backend."""
+
+    async def get(self, key: str) -> Optional[bytes]:  # pragma: no cover
+        """Get the value associated with the key. Returns `None` if the key isn't in the cache."""
```
ncloudioj, line 11: Let's also note in the docstring that `CacheAdatperError` could be raised from `get/set` by protocol implementations.

tests/unit/providers/weather/backends/test_accuweather.py
```diff
@@ -787,3 +787,37 @@ async def test_get_forecast_error(
...
+                postal_code="94105",
+            ),
+            b"US94105",
+        ),
+    ],
+)
+@pytest.mark.asyncio
+async def test_cache_inputs_for_weather_report(
+    accuweather: AccuweatherBackend,
+    cache_inputs_by_location: Tuple[Location, Optional[bytes]],
```
ncloudioj, line 815: Nit: Tuple types can be annotated directly as `tuple[Location, Optional[bytes]]`. Likewise, `dict`, list`, `set` can be used for typing.

tests/unit/providers/weather/test_provider.py
```diff
@@ -161,3 +174,285 @@ async def test_query_error(
...
+
+@pytest.mark.asyncio
+async def test_query_cached_weather_report(
+    redis_mock: Any,
+    backend_mock: Any,
+    provider: Provider,
+    geolocation: Location,
+) -> None:
+    """Test that weather reports are cached in Redis after the first request to the backend."""
+    cache_keys: dict[str, bytes] = {}
```
ncloudioj, line 187: Alternatively, we can make a simple `DictCacheAdapter` for testing.

ncloudioj (approved):
LGTM, just a few nits for your consideration.

Thanks! :shipit:

tests/unit/providers/weather/test_provider.py
ncloudioj: For extra context, `Protocols` are widely used in this repo, we tend to create simple fake protocol implementations for testing rather than mocking. See examples [here](https://github.com/mozilla-services/merino-py/blob/main/merino/providers/wikipedia/backends/fake_backends.py).

linabutler: That makes sense, thank you for the extra context! 🚀 I considered adding an `InMemoryAdapter` while I was working on this, but didn't for a few reasons.

1. We're specifically interested in the sequence of invocations to the cache adapter and the backend, so the mock's `assert_called_*` methods are very helpful here.
2. We want to make sure that errors are propagated and handled correctly—specifically that adapter errors on `get()` force a request, and errors on `set()` are caught. `side_effect` comes in handy here.
3. The `CacheAdapter` interface lets you set a per-key `ttl`, which Redis will handle for us, and which we aren't really interested in testing here, beyond "is the TTL passed to the adapter correctly".

An `InMemoryAdapter` could still handle (1) and (2) for us with `AsyncMock(wrap=)`, and we could easily add fake expiration to it for (3) by deleting expired entries on `get`.

(Aside: I noticed we're already pulling in `cachetools` as a transitive dependency, which has a [`TTLCache`](https://cachetools.readthedocs.io/en/latest/#cachetools.TTLCache) implementation. But it's at odds with our `CacheAdapter` protocol, which passes the `ttl` to `set()` for each key, while `TTLCache` needs it in the constructor. We can alter  `CacheAdapter`, but then we have the problem of using different times for different keys—if we want to introduce caching for AccuWeather location keys in the future, for example).

Rolling our own `InMemoryAdapter` isn't terribly complicated, but it is extra code, and something that we wouldn't use in production. And I think we'd still want use the mock with it, for (1) and (2). With those considerations in mind, `InMemoryAdapter` didn't pass the sniff test for me—it's an extra implementation with extra logic that we'll still want to augment in tests, in a way similar to what we do now.

I'll definitely keep the `Protocol` vs. mocking in mind for the future, but I think using a mock might actually help us in this specific case!

ncloudioj: Thanks for sharing that context! It sounds reasonable to me. Using fakes is more a convention instead of enforcement, though that doesn't mean we shouldn't use mocks for Merino. I also use it quite a lot ;)

PR #204, opened 2023-02-16
[Disco-2215] feat dockerflow version check at merino launch

### LINKED ISSUE #203: Implement checks for version file at app launch
Given Dockerflow enforces a specification (see [here](https://github.com/mozilla-services/Dockerflow#containerized-app-requirements)) that a version.json file (see  here) must reside in the app/ directory of a deployed application and exist with defined values of ["source", "commit", "version", "build"], we want to enforce that at Merino startup, there is a check to ensure this file exists.  

The best location in main.py will be [here](https://github.com/mozilla-services/merino-py/blob/main/merino/main.py#L20). This makes error handling simpler and fails the application outright should there be a failure.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/DISCO-2247)

taddes:
Feature to have a check at Merino startup. This ensures `version.json` is present, which ensures:
- Dockerflow requirement of `version.json` is met.
- Prevents unexpected server error when `__version__` endpoint hit.
- That merino crashes with the particular exception being raised. 
Since the intended behavior is for Merino to crash, the exceptions are not being handled. I had considered an alternative method to catch the exceptions and raise a new type of `VersionError` (see example below) but the provided stack trace from the crashes provide sufficient clarity as to what is going on.

```python
    try:
        version_file_content: dict = json.loads(version_file.read_text())
    except FileNotFoundError as fnf_error:
        raise VersionError(
            f"version.json file not found in directory: {merino_root_path}. \
            Check that version.json file exists."
        ) from fnf_error
    except JSONDecodeError as jd_error:
        raise VersionError(
            f"Error decoding JSON file at directory: {merino_root_path}. \
            Check for malformed data in JSON file."
        ) from jd_error
    except ValidationError as v_error:
        raise VersionError(
            "Validation of Version Model failed. Check version.json file for \
            invalid fields."
        ) from v_error
    else:
        return Version(**version_file_content)
```

This check implements a call to the `fetch_app_version_from_file`, which is the origin to the exceptions. It seemed sensible that if the call to this function succeeds, a boolean of True is returned but it isn't used. Happy to consider another option, but given the other init functions don't return anything in `main.py`

Closes #203

### DIFF
--- a/merino/main.py
+++ b/merino/main.py
@@ -11,6 +11,7 @@ from merino.config_logging import configure_logging
 from merino.config_sentry import configure_sentry
 from merino.metrics import configure_metrics, get_metrics_client
 from merino.middleware import featureflags, geolocation, logging, metrics, user_agent
+from merino.utils.version import check_version
 from merino.web import api_v1, dockerflow
 
 app = FastAPI()
@@ -19,6 +20,7 @@ app = FastAPI()
 @app.on_event("startup")
 async def startup_configuration() -> None:
     """Set up various configurations such as logging."""
+    check_version()
     configure_logging()
     configure_sentry()
     await configure_metrics()
--- a/merino/utils/version.py
+++ b/merino/utils/version.py
@@ -1,11 +1,8 @@
 """Versioning utility module"""
 import json
-import logging
 import pathlib
 
 from pydantic import BaseModel, Extra, HttpUrl
-
-logger = logging.getLogger(__name__)
 
 
 class Version(BaseModel, extra=Extra.forbid):
@@ -24,6 +21,13 @@ def fetch_app_version_from_file(
     commit value, repo source url, version, and CI build values .
     During deployment, this file is written and values are populated for
     the current version of Merino in production and staging.
+
+    Errors are not handled here as the desired behavior is for merino to crash.
+    The following exceptions can possibly be raised.
+    Raises:
+        FileNotFoundError if file cannot be found.
+        JSONDecodeError if the file cannot be processed.
+        ValidationError if Pydantic model validation for Version fails.
     """
     version_file: pathlib.Path = merino_root_path / "version.json"
 
@@ -31,3 +35,15 @@ def fetch_app_version_from_file(
     # and closes the file like a context manager. Uses built-in open() function.
     version_file_content: dict = json.loads(version_file.read_text())
     return Version(**version_file_content)
+
+
+def check_version(
+    merino_root_path: pathlib.Path = pathlib.Path.cwd(),
+) -> bool:
+    """Check the instance of version.json file at merino startup.
+    If an exception is raised, merino will not initialize.
+
+    Calls on fetch_app_version_from_file() for check. Possible exceptions
+    defined in that function definition.
+    """
+    return True if fetch_app_version_from_file(merino_root_path) else False
--- a/tests/contract/client/tests/models.py
+++ b/tests/contract/client/tests/models.py
@@ -8,7 +8,7 @@ from enum import Enum
 from typing import Any, Literal, Optional
 from uuid import UUID
 
-from pydantic import BaseModel, Extra, Field
+from pydantic import BaseModel, Extra, Field, HttpUrl
 
 
 class Service(Enum):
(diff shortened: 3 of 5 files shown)


hackebrot (changes requested):
Hi @taddes! 👋🏻 

Thank you for working on a solution for this!

I think it's a good idea to crash the app on startup when the version information cannot be retrieved. There are a few things we need to consider:

Does raising an exception in the FastAPI `startup` event handler bring down the entire FastAPI application or just this one worker? Does `uvicorn` try to spawn new workers after that? This is something we need to answer and verify through tests. We have a contract test for something similar to this for Contile (see [Contile CircleCI config](https://github.com/mozilla-services/contile/blob/main/.circleci/config.yml#L266-L281)).

The `check_version()` function is a thin wrapper for `fetch_app_version_from_file()`. The `fetch_app_version_from_file()` function always returns a True-ish value or it raises an Exception. The `check_version()` function will never return `False`. So it doesn't really add any value for us. I would recommend calling `fetch_app_version_from_file()` directly in the `startup` handler (if that's what crashes the entire app, see comment above). 

We can cache the result, so the filesystem is checked only once: [functools.cache](https://docs.python.org/3/library/functools.html#functools.cache)

It's great that you've added a unit test for `check_version`! When designing new tests, I always think about what's the worst thing that could happen with a particular piece of code and do we have test coverage for that? I would argue that the aim of this pull request is to verify that the merino app checks for the version information at startup. The worst thing that could happen is that we merge a change that deletes the check from the `startup` handler and we miss that in code review.

We don't have test coverage for that scenario. Let's add that! 🚀 

For this concrete case:

* **unit tests** for `version.py` ensure that the functionality of reading version of information from disk and creating a model with the data is working as expected.
* **integration tests** for `main.py` ensure that the version call is performed on app startup and errors are not silenced (you can use the `client_with_events` fixture from `tests/integration/api/conftest.py` for that).
* **contract tests** for Merino ensure that starting the Merino Docker service without version information results in a crash of the entire service with an appropriate exit code.

It's OK if we want to add the contract test later (please create a new Jira ticket), but we need unit tests and integration tests for this functionality.

cc @Trinaa 😸

ncloudioj:
I'd like us to take a step back and think about do we really need those checks on "version.json".

My take is that we don't need to be super defensive on this for the following reasons:
- For dev/testing, there is a dummy "version.json" at the root directory. Unless it gets deleted somehow, it should be always present with some valid dummy values
- For CI/production, we have this ["write-version"](https://github.com/mozilla-services/merino-py/blob/main/.circleci/config.yml#L157) step while building the docker image. The chances of having this file missing or with invalid data are really slim. If it fails, the Merino image won't be successfully built anyway.

I'd recommend leaving it as is, no need to check it again at runtime as it's already enforced in CI at the build time.

@taddes What do you think?

taddes:
Thanks for the thorough comments @hackebrot , you raise a lot of valuable points.  I knew the bool return from the function was perfunctory, but it seemed sensible even if it's not used since we want the exceptions to raise loud and clear. I can add this context to another ticket should we wish, but I want to point out @ncloudioj 's comments on this in asessing if we should even worry about this, as it is a very paranoid case that would likely have other causes, and even larger consequences. Given the 'write-version' step, I think that can reasonably verify we will have this version file. 

So, we can either go further and add what @hackebrot recommended, or just leave things as is and cancel this, since @ncloudioj does make a solid point for this being covered indirectly. I am okay either way.

taddes:
Just seeing if you had any more thoughts @ncloudioj on if we should can this one. Thx!

ncloudioj:
Hey @taddes, sorry for the delay. I think we can pass this one as `version.json` can be treated similarly to those TOML files. If they somehow go missing or malformed, something must be very broken elsewhere that would need our immediate attention.

Thank you for exploring this anyway!

taddes:
Agreed 100%  @ncloudioj . We don't need to be extra paranoid with the checks here, especially any related problem will be loud and clear. Thx for the feedback and thoughts.

taddes:
Won't continue with merge as current app state is sufficient as an indicator were something to go wrong at app launch. 

Closes #203

PR #209, opened 2023-02-21
Add a lifecycle policy for `enwiki` index

mtbottle:
This lifecycle policy will expire the index after 90 days (around 3 months). We need to add the lifecycle policy to the cluster before we can merge this code change.

### DIFF
--- a/docs/ops.md
+++ b/docs/ops.md
@@ -288,3 +288,27 @@ The provider that is backed by the locally indexed Wikipedia through Elasticsear
   as a floating point number. Defaults to 0.23.
 
 [log]:../merino/config_logging.py
+
+
+### Elasticsearch Index Policy
+
+We want to ensure that the index expire after 3 months,
+so we need to add a lifecycle policy for this deletion to happen.
+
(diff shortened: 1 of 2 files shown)


docs/ops.md
```diff
@@ -288,3 +288,27 @@ The provider that is backed by the locally indexed Wikipedia through Elasticsear
   as a floating point number. Defaults to 0.23.
 
 [log]:../merino/config_logging.py
+
+
+### Elasticsearch Index Policy
```
ncloudioj, line 293: Could you add a new section for "Merino Jobs", and put this under that?

ncloudioj (approved):
Looks good. I guess the plan is to bake the policy directly into TF configurations so we don't need to apply it separately, correct?

docs/ops.md
mtbottle: I have a task to add some [documentation to Merino Jobs](https://mozilla-hub.atlassian.net/browse/DISCO-2253), so I think it might be worth rethinking about where to put this documentation along with that ticket.

Issue #208, opened 2023-02-17
Re-use Merino Remote Settings backend module in test


data-sync-user:
In-line with the team's single repository direction (see [PR](https://github.com/mozilla-services/merino-py/pull/186)), remove duplicated Remote Settings query logic from the load test framework and re-use the existing Remote Settings backend module in Merino. This will make the test framework more resilient to change and reduce maintenance.

**Acceptance Criteria**

* The `RemoteSettingsBackend` module is refactored into methods that can be re-used in testing
* The load test framework re-uses the `RemoteSettingsBackend` module

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/DISCO-2248)
┆Epic: Expand load testing to report stats in CD and include all providers

data-sync-user:
➤ Katrina Anderson commented:

Update (2023/02/27)

  * Including merino in the locust Dockerfile necessitates making changes in the setup_k8s.sh
    * The setup_k8s.sh must now be run from root
    * Line 19 must be changed to MERINO_DIRECTORY=$DIRECTORY/tests/load/kubernetes-config
    * The command on line 49 must be updated to point to the load test Dockerfile
      * While a path to the dockerfile can be specified with gcloud builds submit it will mean the context is relative, which will not work in this case where we need the root context (ref: stack-overflow ( https://stackoverflow.com/questions/58327157/specify-dockerfile-for-gcloud-build-submit ))
      * We will need to add a cloudbuild.yaml (ref: https://cloud.google.com/build/docs/configuring-builds/create-basic-configuration ( https://cloud.google.com/build/docs/configuring-builds/create-basic-configuration|smart-link ) )
        * We will need to substitute variable values (ref: https://cloud.google.com/build/docs/configuring-builds/substitute-variable-values ( https://cloud.google.com/build/docs/configuring-builds/substitute-variable-values|smart-link ) )

PR #214, opened 2023-02-27
[DISCO-2248] Re-use Merino Remote Settings backend module in test

Trinaa:
# Description
* Refactor RemoteSettingsBackend for re-use in test frameworks
    * Sort methods by execution order
    * Replace deprecated asyncio `as_completed` with TaskGroup 
    * Add exception handling
* Refactor/Add unit tests
* Replace kinto_client module with the remotesettings module in the load test framework
* Update load test Docker files to include merino modules
* Update setup_k8s
* Add cloudbuild.yaml and update setup_k8s.sh to accommodate new Docker context in distributed tests

# Issue(s)
#208 | [DISCO-2248](https://mozilla-hub.atlassian.net/browse/DISCO-2248)

[DISCO-2248]: https://mozilla-hub.atlassian.net/browse/DISCO-2248?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


Trinaa:
@dlactin This change would require the inclusion of merino directory modules in the docker image. This has proven tricky because the load test framework Dockerfile is not located in the project root.
1. Do you have any feedback regarding my changes to the Dockerfile or docker-compose.yaml? Can this be done better?
2. Will this impact the load tests in CD?

dlactin:
> @dlactin This change would require the inclusion of merino directory modules in the load test docker image. This has proven tricky because the load test framework Dockerfile is not located in the project root.
> 
>     1. Do you have any feedback regarding my changes to the Dockerfile or docker-compose.yaml? Can this be done better?
> 
>     2. Will this impact the load tests in CD?

Your dockerfile changes look good to me. 

No impact on the load tests in CD, though as discussed over slack we might be better served with a circleci job to build and publish the image to dockerhub.

Trinaa:
Re-based and fixed some stuff I found in DISCO-2231

merino/providers/adm/backends/remotesettings.py
```diff
@@ -34,91 +54,151 @@ def __init__(self, server: str, collection: str, bucket: str) -> None:
...
+        try:
+            async with asyncio.TaskGroup() as task_group:
+                for record in data_records:
+                    task_group.create_task(
+                        extend_suggestions_lst(
+                            url=urljoin(
+                                base=attachment_host,
+                                url=record["attachment"]["location"],
+                            ),
+                            lst=suggestions,
```
ncloudioj, line 168: Mutating this suggestion list across all the tasks feels a little odd to me even though Merino is single-threaded. Alternatively, you can return suggestions from each task and then await them and join the results (or use a generator).

```py
         async def extend_suggestions_lst(url: str, lst: list[KintoSuggestion]):
             return await self.get_attachment(url)

         tasks = []
         async with asyncio.TaskGroup() as task_group:
                for record in data_records:
                    tasks.append(
                        task_group.create_task(
                            extend_suggestions_lst(
                                url=urljoin(
                                    base=attachment_host,
                                    url=record["attachment"]["location"],
                             )
                          )
                      )

         for task in tasks:
            yield await task
         # Or just join all the results for simplicity
```

tests/load/locust_tests/locustfile.py
```diff
@@ -43,7 +50,8 @@
 KINTO__COLLECTION = os.environ["KINTO__COLLECTION"]
 
 # This will be populated on each worker and
```
ncloudioj, line 52: Looks like this comment gets truncated.

ncloudioj (approved):
r+ wc.

Nice reuse!

merino/providers/adm/backends/remotesettings.py
Trinaa: I can iterate over tasks to extend suggestions, this is clearer, ty for the suggestion!

tests/load/locust_tests/locustfile.py
Trinaa: removed the 'and', looking through the history the addition is a mystery

PR #215, opened 2023-02-27
[DISCO-2029] Collect weather provider cache metrics

linabutler:
This commit adds:

* Timings for reading from and writing to Redis.
* Counters for cache hits, misses, and errors.

### DIFF
--- a/merino/providers/weather/provider.py
+++ b/merino/providers/weather/provider.py
@@ -5,6 +5,8 @@ import logging
 from datetime import timedelta
 from typing import Any, Optional
 
+import aiodogstatsd
+
 from merino.cache.protocol import CacheAdapter
 from merino.exceptions import (
     BackendError,
@@ -37,6 +39,7 @@ class Provider(BaseProvider):
 
     backend: WeatherBackend
     cache: CacheAdapter
+    metrics_client: aiodogstatsd.Client
     score: float
     cached_report_ttl_sec: int
 
@@ -44,6 +47,7 @@ class Provider(BaseProvider):
         self,
         backend: WeatherBackend,
         cache: CacheAdapter,
+        metrics_client: aiodogstatsd.Client,
         score: float,
         name: str,
         query_timeout_sec: float,
@@ -53,6 +57,7 @@ class Provider(BaseProvider):
     ) -> None:
         self.backend = backend
         self.cache = cache
+        self.metrics_client = metrics_client
         self.score = score
         self._name = name
         self._query_timeout_sec = query_timeout_sec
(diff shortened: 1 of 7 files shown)


merino/providers/weather/provider.py
```diff
@@ -127,15 +132,30 @@ async def query(self, srequest: SuggestionRequest) -> list[BaseSuggestion]:
...
-            if not isinstance(exc, CacheMissError):
+            if isinstance(exc, CacheMissError):
+                self.metrics_client.increment(f"providers.{self.name}.query.cache.miss")
+            else:
+                self.metrics_client.increment(
+                    f"providers.{self.name}.query.cache.error"
+                )
                 logger.warning(f"Failed to load cached weather report: {exc}")
             try:
                 weather_report = await self.backend.get_weather_report(geolocation)
```
linabutler, line 149: question: Is it worth recording the backend latency separately, or do we think the overall provider query timing is enough?

merino/providers/weather/provider.py
```diff
@@ -127,15 +132,30 @@ async def query(self, srequest: SuggestionRequest) -> list[BaseSuggestion]:
         weather_report: Optional[WeatherReport] = None
 
         try:
-            weather_report = await self.fetch_cached_weather_report(geolocation)
+            weather_report = await self.metrics_client.timeit_task(
+                self.fetch_cached_weather_report(geolocation),
+                f"providers.{self.name}.query.cache.get",
```
linabutler, line 137: issue: Update `docs/data.md` before landing, assuming we're happy with this set of metrics!

merino/providers/weather/provider.py
```diff
@@ -127,15 +132,30 @@ async def query(self, srequest: SuggestionRequest) -> list[BaseSuggestion]:
...
+            )
+            self.metrics_client.increment(f"providers.{self.name}.query.cache.hit")
         except (CacheAdapterError, CacheEntryError, CacheMissError) as exc:
-            if not isinstance(exc, CacheMissError):
+            if isinstance(exc, CacheMissError):
+                self.metrics_client.increment(f"providers.{self.name}.query.cache.miss")
+            else:
```
ncloudioj, line 143: For some extra context, I think we should either instrument `CacheEntryError` and `CacheAdapterError` separately such as `"query.cache.error.adapter"` and `"query.cache.error.entry"` or add a tag for the error type for `"query.cache.error"`.

merino/providers/weather/provider.py
```diff
@@ -127,15 +132,30 @@ async def query(self, srequest: SuggestionRequest) -> list[BaseSuggestion]:
...
+                self.metrics_client.increment(
+                    f"providers.{self.name}.query.cache.error"
+                )
                 logger.warning(f"Failed to load cached weather report: {exc}")
             try:
                 weather_report = await self.backend.get_weather_report(geolocation)
                 try:
-                    await self.store_cached_weather_report(geolocation, weather_report)
+                    await self.metrics_client.timeit_task(
```
ncloudioj, line 151: `timeit_task()` will create an `asyncio.Task` for the wrapped coroutine which is probably an overkill in this case. I think we can use the `[timeit](https://gr1n.github.io/aiodogstatsd/usage/#timeit)` context manager instead.

merino/providers/weather/provider.py
```diff
@@ -127,15 +132,30 @@ async def query(self, srequest: SuggestionRequest) -> list[BaseSuggestion]:
...
+                        self.store_cached_weather_report(geolocation, weather_report),
+                        f"providers.{self.name}.query.cache.set",
+                    )
                 except CacheAdapterError as exc:
+                    self.metrics_client.increment(
+                        f"providers.{self.name}.query.cache.error"
```
ncloudioj, line 157: `providers.{set.name}.query.cache.error` will capture both fetch and store errors, should we split it into two? Or maybe add a tag to differentiate fetch/store errors.

merino/providers/weather/provider.py
ncloudioj: This also applies to the other `timeit_task()` use case above.

linabutler: [`timeit` is a regular `@contextmanager`](https://github.com/Gr1N/aiodogstatsd/blob/4c363d795df04d1cc4c137307b7f91592224ed32/aiodogstatsd/client.py#L255-L263),not an `@asynccontextmanager`, though? `store_cached_weather_report` and `get_cached_weather_report` are async functions; I don't think I can `await` then inside of a non-async context manager?

merino/providers/weather/provider.py
linabutler: Check out my comment below about `query.cache.error`—I think it depends on how granular we want to be with metrics, and how we want to handle them when they come up! 😄

merino/providers/weather/provider.py
linabutler: We can, for sure!

I captured both in the same metric intentionally—my thinking was, once we see any kind of cache error, we'd want to look at the logs for actionable insights. For the same reason, I didn't emit separate metrics for the different kinds of cache errors: failing to deserialize a cache entry is a different kind of error than a Redis exception, but the way we'd handle both is by looking at the logs to see what happened. At a glance, it didn't feel like "this specific kind of cache error happened" would be more actionable than "some cache error happened"—both are things that we'd want to poke at right away, with the full error message.

WDYT?

tests/unit/providers/weather/test_provider.py
```diff
@@ -51,12 +52,21 @@ def fixture_redis_mock(mocker: MockerFixture) -> Any:
     return mocker.AsyncMock(spec=Redis)
 
 
+@pytest.fixture(name="statsd_mock")
```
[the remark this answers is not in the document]
linabutler: Oooh, nice!

merino/providers/weather/provider.py
linabutler: Sounds good, I'll add it!

merino/providers/weather/provider.py
ncloudioj: Right, bummer.

Maybe we can have a quick decorator to time those async questions. The good news is that we already use [wrapt](https://github.com/GrahamDumpleton/wrapt) in Merino. WDYT?

```py
import asyncio

import wrapt


def timeit(name: str, tags: Optional[dict[str, str]] = None):
    """A decorator that can time the async methods of `Provider`"""
    @wrapt.decorator
    async def wrapper(wrapped, instance, args, kwargs):
        loop = asyncio.get_event_loop()
        begin = loop.time()
        try:
            return await wrapped(*args, **kwargs)
        finally:
            elapsed = loop.time() - begin
            # `instance` here will be the provider instance
            instance.metrics_client.timing(name, value=int(elapsed), tags=tags)
    return wrapper


class Provider(BaseProvider):

    @timeit("providers.accuweather.query.cache.set")
    async def store_cached_weather_report(self, ...):
        ...

    @timeit("providers.accuweather.query.cache.get")
    async def fetch_cached_weather_report(self, ...):
        ...
```

Note that this requires the hardcoded name as it doesn't have access to `{self.name}`

merino/providers/weather/provider.py
ncloudioj: Ah, I see, makes sense.

Can you add a quick comment (anywhere that makes sense to you) to note your rationale?

merino/providers/weather/provider.py
linabutler: Maybe an `@asynccontextmanager`, so that we can use f-strings to form the metric name, and make the `self.metrics_client` reference explicit? 😊 (I just learned those could also work as decorators now!)

```py
import asyncio
import contextlib

@contextlib.asynccontextmanager
def timing(client: aiodogstatsd.Client, name: str, tags: Optional[dict[str, str]] = None) -> None:
    loop = asyncio.get_event_loop()
    begin = loop.time()
    try:
        yield
    finally:
        elapsed = loop.time() - begin
        client.timing(name, value=int(elapsed), tags=tags)

async def store_cached_weather_report(self, ...):
    async with timing(self.metrics_client, f"providers.{self.name}.query.cache.set"):
        ...
```

It adds an extra level of indentation, but it's not too bad!

linabutler: > I don't think I can `await` them inside of a non-async context manager?

I'm so sorry, I was confused! We need `async with` if we wanted to `await` anything _before_ entering the context, but there's nothing like that here—we're timing the calls themselves, and it's completely fine to use `await` inside a non-async context manager—and `init_providers` has an example.

Your original suggestion of using `timeit` works perfectly! ❤️

merino/providers/weather/provider.py
linabutler: Done!

ncloudioj (approved):
Looks great, thanks!

PR #216, opened 2023-03-01
[No bug] docs: add ADR template

linabutler:
This commit adds a template for our Architecture Decision Records, using a similar format to Firefox Accounts, and already used for DISCO-2045.

It's more structured than Nygard's status-context-decision-consequences template, but explicitly enumerates the alternatives considered and the pros and cons of each, which has been helpful in other projects (FxA, Application Services, UniFFI) that use this format.

Factored out from #152—thanks for the reminder, @Trinaa! 😊

### DIFF
--- a/docs/adr/template.md
+++ b/docs/adr/template.md
@@ -1,0 +1,94 @@
+# [short title of solved problem and solution]
+
+* Status: [proposed | rejected | accepted | deprecated | … | superseded by [ADR-0005](0005-example.md)] <!-- optional -->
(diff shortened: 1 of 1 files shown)


Trinaa (approved):

ncloudioj (approved):
Great template, thanks!

ncloudioj:
@linabutler Just a kind reminder, for changes like this you can add `[do not deploy]` to the commit message to opt out of the deployment.

PR #205, opened 2023-02-17
[DISCO-2178]  initial contract tests for version endpoint 

taddes:

PR #217, opened 2023-03-01
[DISCO-2178] Version Contract Tests for all environments

### LINKED ISSUE #170: __version__ endpoint contract tests
Add a contract test for the `/__version__` endpoint and verify that the endpoint does *not* return `TBD` for `build` etc. The `version.json` file is overwritten by a CircleCI step, so the contract test will only pass on CircleCI. 

**Resources**

* [https://circleci.com/docs/variables/#built-in-environment-variables](https://circleci.com/docs/variables/#built-in-environment-variables|smart-link) (see CIRCLECI)

**Acceptance Criteria**

* The contract test framework is updated to support skipping contract tests based on environment (CircleCI vs local) 
* The contract test suite includes a test that verifies a proper response for the `__version__` endpoint. 
* The contract test README is updated accordingly

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/DISCO-2178)
┆Epic: Implementation of Versioning for Merino

taddes:
# Description
* Version endpoint contract tests. 
* Took alternative path to cover all environments and to not simply skip in local. Achieved through:
- regex validation of sha-1 hash
- type check for HttpUrl to verify type check variance
- empty version string once in stage/prod

# Issue(s)
#170 | [DISCO-2178](https://mozilla-hub.atlassian.net/browse/DISCO-2178)

Replaces [old branch](https://github.com/mozilla-services/merino-py/pull/205) that was created when dealing with blocking cache bug that was resolved [DISCO-2264](https://mozilla-hub.atlassian.net/browse/DISCO-2264)

[DISCO-2178]: https://mozilla-hub.atlassian.net/browse/DISCO-2178?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[DISCO-2264]: https://mozilla-hub.atlassian.net/browse/DISCO-2264?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/tests/contract/client/tests/test_merino.py
+++ b/tests/contract/client/tests/test_merino.py
@@ -10,7 +10,6 @@ import time
 from typing import Any, Callable
 
 import pytest
-from pydantic import HttpUrl
 import requests
 from kinto import (
     KintoEnvironment,
@@ -29,6 +28,7 @@ from models import (
     Suggestion,
     VersionResponseContent,
 )
+from pydantic import HttpUrl
 from requests import Response as RequestsResponse
 
 # We need to exclude the following fields on the response level:
--- a/tests/contract/client/tests/test_merino.py
+++ b/tests/contract/client/tests/test_merino.py
@@ -218,9 +218,9 @@ def assert_200_version_endpoint_response(
     """
     expected_content_dict = step_content
     merino_content_dict = merino_version_content
-    assert expected_content_dict.source == merino_content_dict.source
+    assert expected_content_dict.source != merino_content_dict.source
 
-    if os.environ.get("MERINO_ENV") == "ci":
+    if os.environ.get("TEST_ENV") == "ci":
         # The version data is constructed from the build of merino, including
         # the commit hash, a build url to circleci and then an empty value for version.
         # Source is identitical between local and production.
--- a/tests/contract/client/tests/test_merino.py
+++ b/tests/contract/client/tests/test_merino.py
@@ -24,6 +24,7 @@ from models import (
     Service,
     Step,
     Suggestion,
+    VersionResponseContent,
 )
 from requests import Response as RequestsResponse
 
@@ -114,12 +115,22 @@ def fixture_merino_step(
             # instance for checking the field types and comparing a dict
             # representation of the model instance with the expected response
             # content for this step in the test scenario.
-            assert_200_response(
-                step_content=step.response.content,
-                merino_content=ResponseContent(**response.json()),
-                fetch_kinto_icon_url=fetch_kinto_icon_url,
-            )
-            return
+
+            if step.request.path == "/__version__":
+                assert_200_version_endpoint_response(
+                    # type ignored to appease mypy, does not infer 2 possible types.
+                    step_content=step.response.content,  # type: ignore
+                    merino_version_content=VersionResponseContent(**response.json()),
+                )
+                return
+            else:
+                assert_200_response(
+                    # type ignored to appease mypy, does not infer 2 possible types.
+                    step_content=step.response.content,  # type: ignore
+                    merino_content=ResponseContent(**response.json()),
+                    fetch_kinto_icon_url=fetch_kinto_icon_url,
+                )
+                return
 
         if response.status_code == 204:
             # If the response status code is 204 No Content, load the response content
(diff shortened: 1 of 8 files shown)


tests/contract/client/tests/test_merino.py
```diff
@@ -114,12 +118,22 @@ def merino_step(step: Step) -> None:
...
-            return
+
+            if step.request.path == "/__version__":
+                assert_200_version_endpoint_response(
+                    # type ignored to appease mypy, does not infer 2 possible types.
+                    step_content=step.response.content,  # type: ignore
+                    merino_version_content=VersionResponseContent(**response.json()),
+                )
+                return
+            else:
```
Trinaa, line 129: Returning in the if statement makes the else statement redundant. You could either eliminate the else, or extract the return from both if and else clauses.

tests/contract/client/tests/test_merino.py
```diff
@@ -114,12 +118,22 @@ def merino_step(step: Step) -> None:
...
             # representation of the model instance with the expected response
             # content for this step in the test scenario.
-            assert_200_response(
-                step_content=step.response.content,
-                merino_content=ResponseContent(**response.json()),
-                fetch_kinto_icon_url=fetch_kinto_icon_url,
-            )
-            return
+
+            if step.request.path == "/__version__":
```
Trinaa, line 122: I suggest to change this condition to use the expected response type instead of the request path

```suggestion
            if type(step.response.content) == VersionResponseContent:
```

Trinaa (changes requested):
Hi @taddes, thank you for contributing this test!!!

I will leave the final implementation up to you, but my 2 cents is I would have preferred a 'skip' feature to this implementation, which is very specific to the needs of the version endpoint tests. I believe the 'skip' would have future benefit and don't think there is added benefit to running the __version__ test locally.

tests/contract/client/tests/test_merino.py
taddes: The intended behavior is to make the else statement redundant, since there are 2 types of Response content, one for all the other Merino responses and this one for the version endpoint.  If a 200 returns, the first check triggers whether or not it's the case of being for the `version` endpoint. After the version assertion runs, we don't want there to be another check with the wrong model. The original code for the base 200 assertion function had a return right after it like before.

`python

        if response.status_code == 200:
            # If the response status code is 200 OK, use the
            # assert_200_response() helper function to validate the content of
            # the response from Merino. This includes creating a pydantic model
            # instance for checking the field types and comparing a dict
            # representation of the model instance with the expected response
            # content for this step in the test scenario.
            assert_200_response(
                step_content=step.response.content,
                merino_content=ResponseContent(**response.json()),
                fetch_kinto_icon_url=fetch_kinto_icon_url,
            )
            return`

taddes:
Hey @Trinaa ! Thanks for the review comments. As requested, I've done the following

- Added docker command to Makefile, great idea! 
- Modified docs for clarity and put Make-related stuff in the upper level as you thought.
- Removed that unnecessary `return` stmt
- Re: skip logic, as we discussed in Slack, I'm gonna leave things be so we can at least verify local results since I wasn't a fan of putting misleading data in the scenarios file. Also, I like the gut check provided by us verifying the shape of data in stage/prod all while also having a local verification. We can definitely implement skip logic in the future should we wish, since this revealed for us how to scope and extract environment variables that were previously not accessible to us. Therefore gonna leave the scenario data as-is, since it conforms to this pattern. 
- Re: `  if type(step.response.content) == VersionResponseContent:` for now, I added this as an and to the initial `__version__` path because just doing that single check makes Pydantic mad when it comes to the `wiki_fruit` tests and can't grok them, so wasting time fighting with those just isn't worth the time. I like the premise of checking the type. This way, we do that and also verify the path, further scoping the version test. Hope that's fine with you. 
```python
            if (
                step.request.path == "/__version__"
                and type(step.response.content) == VersionResponseContent
            ):```

Otherwise, all resolved and good to go if you are!

Trinaa (approved):
Okay. Lets merge. 🚀

PR #218, opened 2023-03-02
[DISCO-2264] changing docker contract test caching due to client errors 

taddes:
This bug was raised when working on contract tests for version endpoint. See [JIRA issue](https://mozilla-hub.atlassian.net/browse/DISCO-2264) for specifics. Changes in test client locally were not being reflected and new scenarios were causing test failures in CI. After digging, it was discovered that caching the docker image for contract tests was the issue. 

 Verified by @Trinaa

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -153,13 +153,13 @@ jobs:
     steps:
       - checkout
       - setup_remote_docker:
-            docker_layer_caching: true
+          docker_layer_caching: true
       - write-version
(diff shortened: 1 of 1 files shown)


ncloudioj (approved):
Looks good, nice catch!

taddes:
> Looks good, nice catch!

Thanks @ncloudioj ! Should save us future headaches. Glad I wasn't going crazy!

PR #220, opened 2023-03-02
Add category content filtering to Wikipedia Indexing

mtbottle:
We want to filter out all the articles in certain categories. The categories list currently live in GitHub.

### DIFF
--- a/merino/configs/default.toml
+++ b/merino/configs/default.toml
@@ -158,3 +158,5 @@ gcs_path = ""
 gcp_project = ""
 # Wikipedia export base URL
 export_base_url = "https://dumps.wikimedia.org/other/cirrussearch/current/"
+# Blocklist file as CSV. Contains a list of the categories for articles that we want to block.
+blocklist_file_url = "https://raw.githubusercontent.com/mozilla/search-terms-sanitization/a021d80defb40f3afb98b86e46e86b3e3017ec73/non_sensitive/wikipedia-content-moderation/blocklist_cats.csv"
--- a/merino/jobs/wikipedia_indexer/__init__.py
+++ b/merino/jobs/wikipedia_indexer/__init__.py
@@ -7,6 +7,7 @@ from elasticsearch import Elasticsearch
 from merino.config import settings as config
 from merino.jobs.wikipedia_indexer.filemanager import FileManager
 from merino.jobs.wikipedia_indexer.indexer import Indexer
+from merino.jobs.wikipedia_indexer.util import create_blocklist
 
 logger = logging.getLogger(__name__)
 
@@ -43,12 +44,13 @@ def index(
     elasticsearch_cloud_id: str = job_settings.es_cloud_id,
     elasticsearch_api_key: str = job_settings.es_api_key,
     elasticsearch_alias: str = job_settings.es_alias,
+    blocklist_file_url: str = job_settings.blocklist_file_url,
     index_version: str = version_option,
     total_docs: int = job_settings.total_docs,
     gcs_path: str = gcs_path_option,
     gcp_project: str = gcp_project_option,
 ):
-    """Index file from gcs to elasticsearch"""
+    """Index file from GCS to Elasticsearch"""
     es_client = Elasticsearch(
         cloud_id=elasticsearch_cloud_id,
         api_key=elasticsearch_api_key,
@@ -57,7 +59,14 @@ def index(
 
     file_manager = FileManager(gcs_path, gcp_project, "")
 
-    indexer = Indexer(index_version, file_manager, es_client)
+    blocklist = create_blocklist(blocklist_file_url)
+
+    indexer = Indexer(
+        index_version,
+        blocklist,
+        file_manager,
+        es_client,
+    )
     indexer.index_from_export(total_docs, elasticsearch_alias)
 
 
@@ -67,7 +76,7 @@ def copy_export(
     gcs_path: str = gcs_path_option,
     gcp_project: str = gcp_project_option,
 ):
-    """Copy file from wikimedia to gcs"""
+    """Copy file from Wikimedia to GCS"""
     file_manager = FileManager(gcs_path, gcp_project, export_base_url)
 
     logger.info(
--- a/merino/jobs/wikipedia_indexer/filemanager.py
+++ b/merino/jobs/wikipedia_indexer/filemanager.py
@@ -104,10 +104,6 @@ class FileManager:
 
         return latest_gcs
 
-    def stream_latest_from_gcs(self) -> Generator:
-        """Stream latest file from GCS"""
-        yield from self._stream_from_gcs(self.get_latest_gcs())
-
     def _stream_dump_to_gcs(self, dump_url: str):
         """Write latest to GCS without storing locally"""
         # 40 MB chunk_size. This is the default Blob chunk size.
@@ -127,7 +123,7 @@ class FileManager:
                     completed = writer.write(chunk)
                     reporter.report(completed)
 
-    def _stream_from_gcs(self, blob: Blob) -> Generator:
+    def stream_from_gcs(self, blob: Blob) -> Generator:
         """Streaming reader from GCS"""
         reader: BlobReader
         with blob.open("rb") as reader:
--- a/merino/jobs/wikipedia_indexer/indexer.py
+++ b/merino/jobs/wikipedia_indexer/indexer.py
@@ -2,7 +2,7 @@
 import json
 import logging
 import time
-from typing import Any, Mapping, Optional
+from typing import Any, Generator, Mapping
 
 from elasticsearch import Elasticsearch
 from google.cloud.storage import Blob
@@ -13,6 +13,26 @@ from merino.jobs.wikipedia_indexer.suggestion import Builder
 from merino.jobs.wikipedia_indexer.util import ProgressReporter
 
 logger = logging.getLogger(__name__)
+
+
+def stream_bulk(
+    stream: Generator[str, None, None]
+) -> Generator[tuple[str, str], None, None]:
+    """Aggregate each bulk component into a single yield. Each bulk component consists of:
+    1. First line as the operator (ie. `index`)
+    2. Second line as the document data for the operator.
+
+    Since we're only expecting index lines, each bulk component will always consist of 2 lines.
+    """
+    operation = None
+    for row in stream:
+        if operation is None:
+            operation = row
+        else:
+            # NOTE: Mypy thinks that the following line is unreachable,
+            # even though they actually are reachable.
+            yield operation, row  # type: ignore
+            operation = None
 
 
 class Indexer:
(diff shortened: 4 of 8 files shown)


merino/configs/default.toml
```diff
@@ -152,3 +152,5 @@ gcs_path = ""
 gcp_project = ""
 # Wikipedia export base URL
 export_base_url = "https://dumps.wikimedia.org/other/cirrussearch/current/"
+# Blocklist file as CSV. Contains a list of the categories for articles that we want to block.
+blocklist_file_url = "https://raw.githubusercontent.com/mozilla/search-terms-sanitization/a021d80defb40f3afb98b86e46e86b3e3017ec73/non_sensitive/wikipedia-content-moderation/blocklist_cats.csv"
```
mtbottle, line 156: @quiiver I think the category names in this CSV is the _exact_ format of the ones that show up in the WIkimedia dump, correct? I just want to make sure that using the raw string for the category name will match the ones that come from the document object.

tests/integration/jobs/wikipedia_indexer/test_indexer.py
```diff
@@ -180,10 +190,68 @@ def test_index_from_export(file_manager, es_client):
...
+        "foo/enwiki-20220101-cirrussearch-content.json.gz", "bar"
+    )
+
+    def check_bulk_side_effect(operations):
+        """Use a side effect to test that we send exactly 2 lines to bulk operation
+        (ie. one index line, one document line).
+        The operations list is mutable, so we need to check that the contents
+        are in place at the time of call.
+        """
+        assert len(operations) == 2
```
mtbottle, line 214: @Trinaa I want to check that a mutable list was called in a mock with exactly 2 items (it later gets cleared, so I can't just grab it with `call_args`). Currently I'm checking this in a side effect, because `assert_called_with` will require _exact_ data matching (will be brittle and unnecessary, I just need to check a stub). Do you have some advice on how to make this test better?

merino/configs/default.toml
quiiver: I believe that's correct

merino/jobs/wikipedia_indexer/indexer.py
```diff
@@ -50,20 +56,28 @@ def index_from_export(self, total_docs: int, elasticsearch_alias: str):
...
                 doc = json.loads(line)
-                if prior and (i + 1) % 2 == 0:
-                    self._enqueue(index_name, (prior, doc))
+                categories: set[str] = set(doc.get("category", []))
+
+                if index_line and categories & self.blocklist:
+                    # Takes the intersection of the categories and blocklist.
+                    # If there exists some shared categories with the blocklist,
+                    # then skip processing this document.
+                    index_line = None
```
quiiver, line 73: I think it's probably ideal to pull this out into its own function so we can more easily add individual page blocking as well as category blocking

tests/integration/jobs/wikipedia_indexer/test_indexer.py
Trinaa: I'm not sure if you saw the [coping with mutable arguments](https://docs.python.org/3/library/unittest.mock-examples.html#coping-with-mutable-arguments) documentation? If not perhaps give it a look over. Your `side_effect` solution is in-line with the recommendation imo.

When you say '_exact_ data matching' for the `assert_called_with` method, are you referring to the assigned values or data structure? There is this [ANY](https://docs.python.org/3/library/unittest.mock.html#any) function that can be useful when we want to obfuscate data values.
 
If you have a model for the contents of operations elements, you can strengthen the assert by checking type, but I'm not sure if that is also subject to change in future and would cause brittleness.

merino/jobs/wikipedia_indexer/__init__.py
```diff
@@ -43,12 +44,13 @@ def index(
...
     elasticsearch_api_key: str = job_settings.es_api_key,
     elasticsearch_alias: str = job_settings.es_alias,
+    blocklist_file_url: str = job_settings.blocklist_file_url,
     index_version: str = version_option,
     total_docs: int = job_settings.total_docs,
     gcs_path: str = gcs_path_option,
     gcp_project: str = gcp_project_option,
 ):
-    """Index file from gcs to elasticsearch"""
+    """Index file from GCS to Elasticsearch"""
```
quiiver, line 53: I swear these get changed with every PR wtf lol

tests/integration/jobs/wikipedia_indexer/test_indexer.py
mtbottle: > If you have a model for the contents of operations elements, you can strengthen the assert by checking type

Ah, interesting idea! I don't think this is easily applicable because it's mocking an Elasticsearch function (so I don't think we can modify that easily). It might be worth thinking about wrapping this, if we do need a better check.

FWIW, I did base this solution on the recommendation in the documentation. I just also thought that it's a bit hard to read. It doesn't sound like there's a better suggestion, so I'll just leave it.

merino/jobs/wikipedia_indexer/indexer.py
```diff
@@ -50,20 +56,28 @@ def index_from_export(self, total_docs: int, elasticsearch_alias: str):
         logger.info("Ensuring index exists", extra={"index": index_name})
 
         if self._create_index(index_name):
-            prior: Optional[Mapping[str, Any]] = None
+            index_line: Optional[Mapping[str, Any]] = None
```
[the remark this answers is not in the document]
mtbottle: Good idea! Added this and code is clearer now!

merino/jobs/wikipedia_indexer/indexer.py
mtbottle: I added a new method `_should_filter` which returns a boolean. We can add the other content moderation into that method.

merino/jobs/wikipedia_indexer/filemanager.py
```diff
@@ -104,10 +104,6 @@ def stream_latest_dump_to_gcs(self, latest_gcs: Optional[Blob] = None) -> Blob:
 
         return latest_gcs
 
-    def stream_latest_from_gcs(self) -> Generator:
-        """Stream latest file from GCS"""
-        yield from self._stream_from_gcs(self.get_latest_gcs())
```
mtbottle, line 109: I removed this function as it looked unused. Also exposed the private method below (L130) because that was what we were using in Indexer. Let me know if we should keep it!

merino/jobs/wikipedia_indexer/indexer.py
```diff
@@ -88,6 +114,14 @@ def index_from_export(self, total_docs: int, elasticsearch_alias: str):
         else:
             raise Exception("Could not create the index")
 
+    def _should_filter(self, categories: set[str]) -> bool:
+        """Return True if we do not want to index this document."""
+        # Takes the intersection of the categories and blocklist.
+        # If there exists some shared categories with the blocklist,
+        # then skip processing this document.
+        overlapping_categories = categories & self.blocklist
+        return overlapping_categories != set()
```
linabutler, line 123: suggestion (non-blocking):

```suggestion
        return not self.blocklist.isdisjoint(categories)
```

...Or maybe renaming this to `_should_index`, and inverting the conditions?

merino/jobs/wikipedia_indexer/indexer.py
```diff
@@ -15,6 +15,26 @@
 logger = logging.getLogger(__name__)
 
 
+def stream_bulk(
```
linabutler, line 18: praise: Nice refactoring—this is a lot easier to follow factored out from `index_from_export`!

How would you feel about naming this something like `chunk_bulk_stream` or `split_bulk_stream`? It took a couple of reads before it clicked that you're splitting the stream into tuples of two, but then it made perfect sense.

linabutler (approved):
LGTM, thanks @mtbottle!

merino/jobs/wikipedia_indexer/indexer.py
mtbottle: Good suggestion for `chunk_bulk_stream`. I'll rename it that!

merino/jobs/wikipedia_indexer/indexer.py
mtbottle: Good idea!

merino/jobs/wikipedia_indexer/indexer.py
```diff
@@ -15,6 +15,26 @@
 logger = logging.getLogger(__name__)
 
 
+def chunk_bulk_stream(
```
quiiver, line 18: Not a blocker at all but i'm curious what the rational is for having this be an independent function and not part of the `Indexer` class?

merino/jobs/wikipedia_indexer/indexer.py
```diff
@@ -50,20 +76,20 @@ def index_from_export(self, total_docs: int, elasticsearch_alias: str):
...
-                doc = json.loads(line)
-                if prior and (i + 1) % 2 == 0:
-                    self._enqueue(index_name, (prior, doc))
+            gcs_stream = self.file_manager.stream_from_gcs(latest)
+            for (operator, document) in chunk_bulk_stream(gcs_stream):
+                op = json.loads(operator)
+                doc = json.loads(document)
+                categories: set[str] = set(doc.get("category", []))
+
+                if self._should_index(categories):
```
quiiver, line 90: I think `_should_index` should take the whole `doc` here so it can make decisions based on the doc. For example in a page based blocking scenario we would want to consider the `title` field in the doc.

quiiver (approved):
LGTM! I have a question and a follow up, both non-blocking.

merino/jobs/wikipedia_indexer/indexer.py
mtbottle: Ah, that was because this function doesn't rely on any other fields in the class (so don't need `self`). So, I could've either left it in as a static method, or extracted it out as a separate function (I think function is a bit clearer).

merino/jobs/wikipedia_indexer/indexer.py
mtbottle: Good idea! I updated to pass in the whole doc instead.

merino/jobs/wikipedia_indexer/indexer.py
```diff
@@ -15,6 +15,26 @@
...
+def chunk_bulk_stream(
+    stream: Generator[str, None, None]
+) -> Generator[tuple[str, str], None, None]:
+    """Aggregate each bulk component into a single yield. Each bulk component consists of:
+    1. First line as the operator (i.e. `index`)
+    2. Second line as the document data for the operator.
+
+    Since we're only expecting index lines, each bulk component will always consist of 2 lines.
+    """
+    operation = None
```
ncloudioj, line 27: If you don't mind using itertools, the following should do the same:
```py
from itertools import islice

while (chunk := tuple(islice(stream, 2))):
    yield chunk
```

merino/configs/default.toml
ncloudioj: We might want to confirm with SecOps if it's okay to pull stuff from here. Just be safe :)

merino/jobs/wikipedia_indexer/util.py
```diff
@@ -37,3 +41,15 @@ def report(self, completed: int):
...
+def create_blocklist(blocklist_file_url: str) -> set[str]:
+    """Create blocklist from a file url."""
+    categories = set()
+    block_list = requests.get(blocklist_file_url).text
+    file_like_io = StringIO(block_list)
+    csv_reader = csv.DictReader(file_like_io, delimiter=",")
+    for row in csv_reader:
+        categories.add(row["name"])
+
+    return categories
```
ncloudioj, line 55: Nit:
```suggestion
    return set(row["name"] for row in csv_reader)
```

ncloudioj (approved):
Looks good 👍

merino/jobs/wikipedia_indexer/indexer.py
```diff
@@ -50,20 +76,19 @@ def index_from_export(self, total_docs: int, elasticsearch_alias: str):
...
-            for i, line in enumerate(self.file_manager._stream_from_gcs(latest)):
-                doc = json.loads(line)
-                if prior and (i + 1) % 2 == 0:
-                    self._enqueue(index_name, (prior, doc))
+            gcs_stream = self.file_manager.stream_from_gcs(latest)
+            for (operator, document) in chunk_bulk_stream(gcs_stream):
+                op = json.loads(operator)
+                doc = json.loads(document)
+
+                if self._should_index(doc):
```
ncloudioj, line 89: Shall we keep a counter so that we know how many docs get blocked in the job summary?

merino/configs/default.toml
ncloudioj: Just circling back, SecOps said Okay to this, so we're good.

merino/jobs/wikipedia_indexer/indexer.py
mtbottle: Oh, this doesn't work the same way. `islice` will only ever return one item in the stream. It doesn't actually chunk the stream.

ncloudioj: It does, `islice(stream, 2)` the second argument `2` tells it to chunk the stream into pairs :)

mtbottle: Ah ya! It does work with `iterator`, which isn't well documented.

PR #225, opened 2023-03-07
Initialize the Elasticsearch client with URL

mtbottle:
I'm saving the cloud id path just in case we want to still have it used for staging. It should be removed when we have some local development set up that involves a docker bootstrapped version of Elasticsearch.

### DIFF
--- a/docs/ops.md
+++ b/docs/ops.md
@@ -293,6 +293,9 @@ These are production providers that generate suggestions.
     Either `elasticsearch` or `test`.
   - `es_cloud_id` (`MERINO_PROVIDERS__WIKIPEDIA__ES_CLOUD_ID`) - The Cloud ID of the
     Elasticsearch cluster.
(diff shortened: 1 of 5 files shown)


docs/ops.md
```diff
@@ -293,6 +293,8 @@ These are production providers that generate suggestions.
     Either `elasticsearch` or `test`.
   - `es_cloud_id` (`MERINO_PROVIDERS__WIKIPEDIA__ES_CLOUD_ID`) - The Cloud ID of the
     Elasticsearch cluster.
+  - `es_url` (`MERINO_PROVIDERS__WIKIPEDIA__ES_URL`) - The URL of the cluster that we
+    want to connect to.
```
ncloudioj, line 297: Looks like `es_url` will be used if both `es_url` and `es_cloud_id` are present. Let's note that here.

ncloudioj (approved):

PR #227, opened 2023-03-08
[DISCO-2261] Wikipedia Provider Content Filtering

taddes:
# Description
Creation of logic to filter out blocked suggestions based on article titles at the Wikipedia provider level.

This list can probably live in Merino as a start. This list is currently a stub, as we won’t have anything in there initially. List is not expected to ever be particular large, hence this implementation would not add significant latency to the return of suggestions. 

Requirements:
- A list that can have article titles added to it
- Instructions on adding to said list (see content-moderation.md)
- Processing the list for use by the provider (a function that reads the file, given the environment at init time for the provider). @mtbottle had the great idea of impl dependency injection here for passing in call through util mod.
-  Filtering results from the Wikipedia Provider (remove suggestions that have a title match with items on the list)
- Testing to verify functionality. 

# Issue(s)
#228 | [DISCO-2261](https://mozilla-hub.atlassian.net/browse/DISCO-2261)

[DISCO-2261]: https://mozilla-hub.atlassian.net/browse/DISCO-2261?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


merino/configs/ci.toml
```diff
@@ -48,3 +48,5 @@ top_picks_file_path = "dev/top_picks_for_ci.json"
 [ci.providers.wikipedia]
 # The backend of the provider. Either "elasticsearch" or "test".
 backend = "test"
+# Manual content moderation block list
+block_list_path = "dev/wiki_provider_block_list.txt"
```
ncloudioj, line 52: shouldn't this be `dev/wiki_provider_block_list_ci.txt`?

merino/providers/wikipedia/provider.py
```diff
@@ -37,10 +37,12 @@ class Provider(BaseProvider):
 
     backend: WikipediaBackend
     score: float
+    title_block_list: list[str]
```
ncloudioj, line 40: Let's use a `set` or a `fronzonset` to store these items.

dev/wiki_provider_block_list_ci.txt
```diff
@@ -0,0 +1,2 @@
+Unsafe_Content
```
ncloudioj, line 1: I'd recommend using a more structured file (e.g. JSON) rather than a plain txt.

merino/providers/wikipedia/provider.py
```diff
@@ -81,8 +87,15 @@ async def query(self, srequest: SuggestionRequest) -> list[BaseSuggestion]:
                 **suggestion,
             )
             for suggestion in suggestions
+            # Ensures titles that are in the block list are not returned as suggestions.
+            if suggestion["title"] not in self.title_block_list
```
ncloudioj, line 88: Wonder if we can do this check before issuing the request to ES, though that means we need to build a blocklist index to support prefix matching. Perhaps not worth it, post-checks should be fine for now.

dev/wiki_provider_block_list_ci.txt
ncloudioj: Alternatively, I think we can also consider hardcoding those blocked items in the source (e.g. in the provider) assuming the list will be quite limited for now. That should simplify this as we don't need to maintain that many blocklist files. We can move it to a dedicated data file if it gets too big and unwieldy to use in the future.

merino/providers/wikipedia/provider.py
taddes: Nah, plus we are going to do a check in this regard during the index job (have a ticket for that after this one is merged). This step in this PR covers the provider, then there's another one during indexing, during which we are also checking the full list of topics/titles provided in the sensitive content file.

taddes:
Thanks for the comments @ncloudioj !!! I made the following modifications per your comments, but also in context of our Slack convo, the result of the most recent changes are summarized:
- Converted files to json files that contain json arrays.
- Converted the block list to a set for faster membership checks as it grows.
- File IO remains as we discussed the need to maintain checks in both indexer and provider for now, so we won't maintain a static list, though that would be a nice option down the road.  If we find the IO is a problem as provider scales, we can address it. 
- Explained that we have a ticket out after this one to handle the checking of these static titles during the index. Since the index job is 1x per week, we want at the provider level to block titles, that then later can be blocked in the index job when it runs again.

mtbottle:
> File IO remains as we discussed the need to maintain checks in both indexer and provider for now, so we won't maintain a static list, though that would be a nice option down the road. 

We can technically still use the list for both indexer and provider (just need to put the list in a separate module that can be imported for both). I don't have a strong preference of putting it in a module, but my main reason for suggesting this was that it would be easier to use (since we have a small list) and we don't have to deal with IO (which adds complexity and small amount of latency).

docs/dev/content-moderation.md
```diff
@@ -0,0 +1,16 @@
+# Content Moderation
+Merino has many features that enable the filtering and blocking of sensitive content.
+
+## Dynamic Wikipedia Provider
+
+### Manual Block List
+Located in `/dev/wiki_provider_block_list.txt`
+
+This block list is used during the indexing job for dynamic wiki as well as at the provider
```
mtbottle, line 9:
```suggestion
This block list is used during the indexing job for Dynamic Wikipedia as well as at the provider 
```

docs/dev/content-moderation.md
```diff
@@ -0,0 +1,16 @@
...
+Located in `/dev/wiki_provider_block_list.txt`
+
+This block list is used during the indexing job for dynamic wiki as well as at the provider 
+level for ensuring suggestions of a given title are not returned to the client.
+This allows for granular control or ad-hoc updates of titles we wish to ignore.
+
+*NOTE:* In adding to the block list, ensure the title field is added as it appears
+in the wiki url, respecting case sensitivity. Typically, the titles have the
+first character capitalized, are separated by underscores with subsequent
+words capitalized.
```
mtbottle, line 16: >  In adding to the block list, ensure the title field is added as it appears in the wiki url

I don't think this is correct. The titles are actually with spaces, then it replaces the spaces with `_` for the URL: https://github.com/mozilla-services/merino-py/blob/main/merino/providers/wikipedia/backends/elastic.py#L77-L85 

Hmm... from a usability standpoint, being super precise about format feels kind of hard to use and a potential place for bugs. How do you feel about making the checks case insensitive? Wikipedia titles also don't offer consistency with casing (ie. `American Civil War` vs `American frontier`).

taddes:
Thanks @mtbottle . I think what I'd like to recommend is that we do this: Keep this implementation as-is and move forward with the next task so we have it covered at the provider and indexer level. Then lets see and possibly some data on if we're seeing any real latency that has a tangible effect. I just don't want to get caught in premature optimization, though if we do see any detrimental performance issues, let's definitely change things up! 

I don't think this really increases any complexity given it's a pretty simple function. It's especially not going to add any more overhead at the indexer level that would have an impact, given the nature of the indexing process!

tests/integration/api/v1/suggest/test_suggest_wikipedia.py
```diff
@@ -28,19 +29,47 @@
...
         expected_title=None,
         expected_logs={"A backend failure"},
     ),
+    "Case-III: Block list filter": Scenario(
+        providers={
+            "wikipedia": Provider(
+                backend=FakeEchoWikipediaBackend(),
+                title_block_list=block_list.read_block_list(
+                    settings.providers.wikipedia.block_list_path
+                ),
```
mtbottle, line 65: We're doing the read of the blocklist 3 times in this test file. Let's extract this as a variable.

tests/integration/api/v1/suggest/test_suggest_wikipedia.py
```diff
@@ -28,19 +29,47 @@
...
+            )
+        },
+        query="Unsafe_Content",
```
mtbottle, line 68: I don't think people usually type in search entries with capitalization and underscores 😛

tests/unit/providers/wikipedia/test_provider.py
```diff
@@ -13,23 +14,43 @@
...
 
 def test_hidden(wikipedia: Provider) -> None:
     """Test for the hidden method."""
-    assert not wikipedia.hidden()
+    assert wikipedia.hidden() is False
+
+
+@pytest.mark.asyncio
+@pytest.mark.parametrize("query_keyword", ["test_fail"])
+async def test_query_failure(
```
mtbottle, line 45: Thanks for testing this!

docs/dev/content-moderation.md
```diff
@@ -0,0 +1,16 @@
+# Content Moderation
+Merino has many features that enable the filtering and blocking of sensitive content.
```
mtbottle, line 2: nit: Not sure that we have `many`? I would avoid making this claim in the documentation.

mtbottle (commented):
I have a few comments about usability of the file. I think we should consider case insensitive matching. We can do that by matching the strings when we lowercase both of them.

merino/providers/wikipedia/provider.py
```diff
@@ -71,6 +74,8 @@ async def query(self, srequest: SuggestionRequest) -> list[BaseSuggestion]:
             logger.warning(f"{e}")
             return []
 
+        self.title_block_list = {entry.lower() for entry in self.title_block_list}
```
mtbottle, line 77: Let's move this conversion to `__init__` so we don't have to do it again every time we `query`.

merino/utils/block_list.py
```diff
@@ -0,0 +1,3 @@
+"""Util module containing block list for Dynamic Wikipedia provider."""
+
+block_list: set[str] = set()
```
mtbottle, line 3: Thoughts on using `frozenset`? This list will be immutable, so can restrict any changes by making it a `frozenset`. 

Also stylistic nit: this set will essentially be used like a constant, we should probably capitalize it to 
```suggestion
BLOCK_LIST: set[str] = set()
```

docs/dev/content-moderation.md
```diff
@@ -0,0 +1,16 @@
+# Content Moderation
+Merino is capable of filtering and blocking of sensitive content.
+
+## Dynamic Wikipedia Provider
+
+### Manual Block List
```
mtbottle, line 6: nit: Let's be specific that this is content moderation for Wikipedia titles.
```suggestion
### Wikipedia Title Block List
```

mtbottle (commented):
Just a few more comments! We're almost there!

taddes:
Thanks for the review @mtbottle . Here are my updates:
- Changed docs as requested
- updated block list to be a const in its appearance
- __init__ call iteration over the block list
- As for changing to a `frozenset`, I'm fine with that from an immutability perspective, but given we iterate over the block list to modify entries to lower case for more dependable matching, we're technically kind of breaking the immutability rule and papering over it with the `frozenset` constructor in that case. Since the constructor accepts an iterator, the comprehension modifies the initial values, we are just reassigning the same variable name to a new frozenset. I think a set will be fine, but if you feel strongly it should be modified, I can change it.

mtbottle (approved):

taddes:
Closes #228

PR #231, opened 2023-03-09
Allow Indexer job to configure URL Elasticsearch connection

mtbottle:
This change will allow us to use the Elasticsearch URL if it is configured. If it doesn't exist, then fallback to using the Cloud ID.

@quiiver You should be able to configure the URL either with command line: `--elasticsearch_url` or setting the environment variable `MERINO__JOBS__WIKIPEDIA_INDEXER__ES_URL`.


quiiver (approved):
🚢

Issue #160, opened 2022-12-30
Expand Merino-py load test to include Dynamic Wikipedia


data-sync-user:
Expand the load test solution query set to include Dynamic Wikipedia.

**Acceptance Criteria**

* Dynamic Wikipedia is included in the load test query set
  * 10%{color:#ff991f} {color}of request load should be allocated to this provider
  * The p95 latency threshold should be **200ms**
* Reporting is setup to isolate performance metrics for Weather queries
* Documentation is updated accordingly

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/DISCO-2167)
┆Epic: Expand Load Test Scope to Target Dynamic Wikipedia, Top Picks and Weather Providers

data-sync-user:
➤ Michelle Tran commented:

Some code to get started for getting some words to use in Elasticsearch:

from elasticsearch import Elasticsearch

es_client = Elasticsearch(cloud_id=CLOUD_ID, api_key=API_KEY)
results = es_client.search(index="enwiki-v1", size=200)

docs = results["hits"]["hits"]

keywords = [] # List of useful query words to use
for doc in docs:
    keyword = doc["_source"]["title"]
    keywords.append[keyword]Kind of like Topicks, you should also be able to get results by using “almost” the complete word. Not sure if the load test requires this complexity, but you can try removing some characters (ie. if you’re looking for “animal”, can try “anim” or something).

data-sync-user:
➤ Katrina Anderson commented:

Draft PR ( https://github.com/mozilla-services/merino-py/pull/233 ) is up.

Blocked since we are unable to connect to stage ES cluster ATM. Waiting on outcome (ref: slack ( https://mozilla.slack.com/archives/G01M6ALKVL4/p1678475398926519 ))

PR #233, opened 2023-03-10
[DISCO-2167] Expand Merino-py load test to include Dynamic Wikipedia [load test: warn]

Trinaa:
## References

JIRA: [DISCO-2167](https://mozilla-hub.atlassian.net/browse/DISCO-2167)
GitHub: #160 

## Description

- Replace source for Wikipedia queries with Elastic Search
- Update remote execution script and configurations
- Update README documentation

## PR Review Checklist

[DISCO-2167]: https://mozilla-hub.atlassian.net/browse/DISCO-2167?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


Trinaa:
Note: I am unable to test this code atm, will wait for the ES stage connection outcome (ref: [slack](https://mozilla.slack.com/archives/G01M6ALKVL4/p1678475398926519))
@mtbottle if you have a chance perhaps you can take an early look and see if there is anything you want to flag?

tests/load/locust_tests/locustfile.py
```diff
@@ -173,6 +194,36 @@ def add_queries(index: dict[str, list[int]], queries: dict[int, list[str]]):
...
 
+def get_wikipedia_queries(
+    cloud_id: str | None, api_key: str | None, index: str | None
+) -> list[str]:
+    """Get query strings for use in testing the Wikipedia Provider.
+
+    Args:
+        cloud_id: The Cloud ID of the Elasticsearch cluster
+        api_key: The base64 key used to authenticate on the Elasticsearch cluster
+        index: A comma-separated list of index names to search; use `_all` or empty
```
mtbottle, line 205: Hmm... Is there a reason to allow this to run on `_all` indices? We are actually going to be keeping old indices as backup, so they might be queried if you add `_all` (not to mention that if we add other indices, then they will potentially be queried as well).

mtbottle (commented):
Generally LGTM! I don't see anything major that needs to be addressed!

tests/load/locust_tests/locustfile.py
Trinaa: It's more about maintaining the same configuration flexibility as Merino. Do we filter/check for '_all' in Merino?

I don't think there is high risk that this value will deviate from "enwiki-v1" without intention, but if the value should change, it's assignment environment variable will be cohesive.

tests/load/locust_tests/locustfile.py
```diff
@@ -66,10 +67,23 @@
     os.getenv("MERINO_PROVIDERS__TOP_PICKS__FIREFOX_CHAR_LIMIT", 0)
 )
 
+# See Ops - Wikipedia Dynamic Match Provider
+# https://github.com/mozilla-services/merino-py/blob/main/docs/ops.md#wikipedia-dynamic-match-provider
```
mtbottle, line 71: I moved the configs in the documentation, so I think we'll need the updated link or maybe just linking the config file itself. The other option (to make the documentation more future proof) is to not include a link, but have a note that these are configurations for Dynamic Wikipedia and to look at the config file for more information. This will protect us from potentially moving the configurations or documentation and having a dead link in the code.

mtbottle (approved):
Generally LGTM! Just a comment about a potentially dead link.

tests/load/locust_tests/locustfile.py
Trinaa: I like the future proof option! I will fix that thank you.

PR #236, opened 2023-03-15
refactor: Replace chunk_bulk_stream with islice

ncloudioj:
`islice` is slightly shorter without annoying mypy. Also modified the tests to use generators as that's what the indexer expects (not sure why mypy didn't capture that).


mtbottle (approved):
LGTM! 

Although I think this solution is very clever, the initial reason why I opted to not add the `islice` was because the documentation for `islice` does not reflect this behaviour (ie. https://docs.python.org/3/library/itertools.html?highlight=islice#itertools.islice talks about how it behaves on an _iterable_, but doesn't mention the behaviour when it's passed a _generator/iterator_).

Issue #234, opened 2023-03-13
Content Moderation Filtering Titles in Indexer


taddes:
This [task](https://mozilla-hub.atlassian.net/browse/DISCO-2271) is the second part of total content moderation

Rather than having this just in Merino, now we also want to do the title filtering in the Indexing. 

Remember to update any documentation to explain the process for adding manual entries to content moderation.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/DISCO-2289)

data-sync-user:
➤ Taddes Korris commented:

Already complete https://mozilla-hub.atlassian.net/browse/DISCO-2271 ( https://mozilla-hub.atlassian.net/browse/DISCO-2271|smart-link )

PR #237, opened 2023-03-15
[DISCO-2271] Indexer Content Moderation

taddes:
# Description
Creation of logic to filter out manually created blocked suggestions during the Wikipedia provider indexer job. This is relevant to the [previously merged task](https://github.com/mozilla-services/merino-py/pull/227) for provider level filtering. 

The creation of a module to store blocked titles was added in the last ticket. This utility, which is a set of strings containing the block list tiles, is merged via union with the dynamically read block list to filter all blocked titles during the index job

Changes :
- Updated `create_blocklist` function to consume the utility containing the title block list content.
-  Added logic in tests to capture addition of new block list content.
- Renamed `util.py` to `utils.py` to be more semantic. 

# Issue(s)
#234  | [DISCO-2271](https://mozilla-hub.atlassian.net/browse/DISCO-2271)

[DISCO-2271]: https://mozilla-hub.atlassian.net/browse/DISCO-2271?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


merino/jobs/wikipedia_indexer/indexer.py
```diff
@@ -118,7 +121,10 @@ def index_from_export(self, total_docs: int, elasticsearch_alias: str):
     def _should_index(self, doc: Dict[str, Any]) -> bool:
         """Return True if we want to index this document."""
         categories: set[str] = set(doc.get("category", []))
-        return self.blocklist.isdisjoint(categories)
+        titles: set[str] = set(doc.get("title", []))
```
quiiver, line 124: I'm not sure this is what you want since `title` should be a string in the `doc` dictionary

```
>>> doc = {"title": "Foo_Bar"}
>>> set(doc.get("title", []))
{'a', 'B', '_', 'o', 'F', 'r'}
```

quiiver: oh sorry, didn't notice this is still in draft. i'll 🤫

mtbottle (commented):
Some comments below!

merino/jobs/wikipedia_indexer/indexer.py
```diff
@@ -59,7 +61,8 @@ def __init__(
         self.file_manager = file_manager
         self.es_client = client
         self.suggestion_builder = Builder(index_version)
-        self.blocklist = blocklist
+        self.category_blocklist = {entry.lower() for entry in category_blocklist}
```
mtbottle, line 64: Let's not modify the fields for the categories. Since the list is automatically generated, it should match the category exactly. So let's remove the `.lower()` to allow for exact matching.

The difference with the title is that it's added manually, so there's more room for errors when entering the title. For lists and fields that are _not_ manually created, we can leave them without case sensitivity.

tests/integration/jobs/wikipedia_indexer/test_indexer.py
```diff
@@ -15,21 +15,27 @@
...
-    return {"meme", "nsfw", "anger"}
+def category_blocklist() -> set:
+    """Return category blocklist."""
+    return {"meme", "nsfw", "Anger"}
+
+
+@pytest.fixture
+def title_blocklist() -> set:
+    """Return title blocklist."""
+    return {"Bad Things"}
```
mtbottle, line 26: The case sensitivity matches the title below. It would be useful to check that the case sensitivity _doesn't_ matter (so change it so that it's not the same as the title in line 348.

merino/jobs/wikipedia_indexer/indexer.py
taddes: Good call, makes sense! 👍

merino/jobs/wikipedia_indexer/indexer.py
```diff
@@ -116,9 +119,14 @@ def index_from_export(self, total_docs: int, elasticsearch_alias: str):
             raise Exception("Could not create the index")
 
     def _should_index(self, doc: Dict[str, Any]) -> bool:
-        """Return True if we want to index this document."""
+        """Return True if we want to index this document.
+        Checks for existence of matching categories or title in both title and category blocklists.
+        """
         categories: set[str] = set(doc.get("category", []))
-        return self.blocklist.isdisjoint(categories)
+        title: str | None = doc.get("title").lower() if doc.get("title") else None  # type: ignore
```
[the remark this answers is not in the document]
taddes: Unfortunately not, MyPy isn't smart enough to evaluate one liners like that `Item "None" of "Optional[Any]" has no attribute "lower"` is the result.  It's very clean and clear this way, but no other way for MyPy to ignore. I'd add the [union-attr] to the comment, but that takes the line length over the max, so there's just no winning with the linter here 😿

quiiver, line 126: is there a reason you need to use `None` vs a zero value string? 

```py
should_filter_title = doc.get("title", "").lower() in self.title_blocklist
```

merino/jobs/wikipedia_indexer/__init__.py
```diff
@@ -6,10 +6,11 @@
 from merino.config import settings as config
 from merino.jobs.wikipedia_indexer.filemanager import FileManager
 from merino.jobs.wikipedia_indexer.indexer import Indexer
-from merino.jobs.wikipedia_indexer.util import (
+from merino.jobs.wikipedia_indexer.utils import (
     create_blocklist,
     create_elasticsearch_client,
 )
+from merino.utils.block_list import TITLE_BLOCKLIST
```
linabutler, line 13: nitpick: How would you feel about naming this module `merino.utils.blocklist`, since you're using `blocklist` and `BLOCKLIST` in other places?

merino/jobs/wikipedia_indexer/indexer.py
mtbottle: Ah, I think I suggested `None` so that we can differentiate missing titles and titles that _happen_ to be empty when we do a conditional:

```
if title is None:
    return False   # don't index missing titles
```

Looking at this now, I think we can still do that match with an empty string, because if the title is actually empty, we probably don't want to index it anyways.

taddes:
Any further thoughts @mtbottle on why the first call to `es_client.bulk.assert_called_once()` in `test_index_from_export` results in `E       AssertionError: Expected 'bulk' to have been called once. Called 0 times.1` Still working on debugging that. It doesn't happen in any of the subsequent `bulk` calls. Very weird.

merino/utils/blocklist.py
```diff
@@ -1,3 +1,3 @@
 """Util module containing block list for Dynamic Wikipedia provider."""
 
-BLOCK_LIST: set[str] = set()
+TITLE_BLOCKLIST: set[str] = set()
```
quiiver, line 3: What do you think about initializing this with an empty string and then change the logic to treat missing titles as an empty string that would provide filtering of empty or missing titles and clean up some mypy weirdness

merino/jobs/wikipedia_indexer/indexer.py
quiiver: Added a possible resolution in https://github.com/mozilla-services/merino-py/pull/237/files#r1146490754

linabutler: I really like @quiiver's suggestion, too, but, if you'd like to keep the `None` (or for the future), another way to write this and avoid the double `.get()` is:

```py
title: str | None = title.lower() if (title := doc.get("title")) else None
```

taddes: Thanks for the suggestions @quiiver  and @linabutler ! @mtbottle and I discussed and landed with the recently pushed change, which is kind of a compromise hybrid solution, though it would also have been ok to split it out in some `if` `else` checks. Were looking for simplicity, but agreed there was no need in the end to keep `None`
```python
        title: str = doc.get("title", "")
        should_filter_title: bool = title.lower() in self.title_blocklist if title != "" else True
```

We also renamed the function for clarity and altered logic so that the method is now `_should_filter`, which makes the boolean logic easier to understand in context.

merino/utils/blocklist.py
taddes: That is indeed a good idea. Lmk with the recent changes Michelle and I made to that logic if you then think it's ok, it ameliorates that issue with mypy now.

quiiver (approved):

merino/jobs/wikipedia_indexer/indexer.py
quiiver: that's fine yeah, i think we're at the splitting hairs point with this one.

taddes:
Closes #237

PR #239, opened 2023-03-16
[DISCO-2292] Add query string normalization to Merino

taddes:
# Description
## Add query string normalization to Merino
The navigational top picks and AdM suggestion lookups are only triggered if the keyword it's written in all lowercase. Upper case or mixed case keywords do not currently result in a lookup.

# Acceptance Criteria:

- Top Picks and AdM return suggestions irrespective of case, regardless of what character may or may not be capitalized in the query string.
- As an added precaution, ensure domain property is lowercased and stripped during indexing (top picks)
- Create a default implementation of` normalize_query()` in the Provider abstract class so providers can implement their own normalization logic, based on their unique needs.
- Add relevant testing

# Changes :
- Added `normalize_query()` class for `Provider`, is extended for AdM and Top Picks
- Added logic to top picks backend to strip and lowercase domains as an added layer of backup, should the json data have inaccuracies. 
- Added unit tests
- Added integration tests
- Added `# no cover` annotations in `base.py` for abstract classes

# Issue(s)
#240  | [DISCO-2292](https://mozilla-hub.atlassian.net/browse/DISCO-2292) | [Bugzilla-1806970](https://bugzilla.mozilla.org/show_bug.cgi?id=1806970)

[DISCO-2292]: https://mozilla-hub.atlassian.net/browse/DISCO-2292?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/merino/providers/adm/provider.py
+++ b/merino/providers/adm/provider.py
@@ -125,7 +125,7 @@ class Provider(BaseProvider):
 
     def _normalize_query(self, query: str) -> str:
         """Convert a query string to lowercase and remove trailing spaces."""
-        return query.rstrip().lower()
+        return query.strip().lower()
 
     async def query(self, srequest: SuggestionRequest) -> list[BaseSuggestion]:
         """Provide suggestion for a given query."""
--- a/merino/providers/adm/provider.py
+++ b/merino/providers/adm/provider.py
@@ -123,16 +123,13 @@ class Provider(BaseProvider):
     def hidden(self) -> bool:  # noqa: D102
         return False
 
-    def normalize_query(self, query: str) -> str:
-        """Normalize the query string when passed to the provider.
-        For AdM, queries should be case-insensitive and trailing space should
-        be removed.
-        """
+    def _normalize_query(self, query: str) -> str:
+        """Convert a query string to lowercase and remove trailing spaces."""
         return query.rstrip().lower()
 
     async def query(self, srequest: SuggestionRequest) -> list[BaseSuggestion]:
         """Provide suggestion for a given query."""
-        q: str = self.normalize_query(srequest.query)
+        q: str = self._normalize_query(srequest.query)
         if (suggest_look_ups := self.suggestion_content.suggestions.get(q)) is not None:
             results_id, fkw_id = suggest_look_ups
             res = self.suggestion_content.results[results_id]
--- a/merino/providers/adm/provider.py
+++ b/merino/providers/adm/provider.py
@@ -129,7 +129,7 @@ class Provider(BaseProvider):
 
     async def query(self, srequest: SuggestionRequest) -> list[BaseSuggestion]:
         """Provide suggestion for a given query."""
-        q: str = self.normalize_query(srequest.query)
+        q: str = srequest.query
         if (suggest_look_ups := self.suggestion_content.suggestions.get(q)) is not None:
             results_id, fkw_id = suggest_look_ups
             res = self.suggestion_content.results[results_id]
--- a/merino/providers/adm/provider.py
+++ b/merino/providers/adm/provider.py
@@ -125,7 +125,7 @@ class Provider(BaseProvider):
 
     def normalize_query(self, query: str) -> str:
         """Normalize the query string when passed to the provider.
-        For AdM, queries should be case-insensitive and tailing space should
+        For AdM, queries should be case-insensitive and trailing space should
         be removed.
         """
         return query.rstrip().lower()
--- a/merino/providers/adm/provider.py
+++ b/merino/providers/adm/provider.py
@@ -123,9 +123,16 @@ class Provider(BaseProvider):
     def hidden(self) -> bool:  # noqa: D102
         return False
 
+    def normalize_query(self, query: str) -> str:
+        """Normalize the query string when passed to the provider.
+        For AdM, queries should be case-insensitive and tailing space should
+        be removed.
+        """
+        return query.rstrip().lower()
+
     async def query(self, srequest: SuggestionRequest) -> list[BaseSuggestion]:
         """Provide suggestion for a given query."""
-        q = srequest.query
+        q: str = self.normalize_query(srequest.query)
         if (suggest_look_ups := self.suggestion_content.suggestions.get(q)) is not None:
             results_id, fkw_id = suggest_look_ups
             res = self.suggestion_content.results[results_id]
--- a/merino/providers/adm/provider.py
+++ b/merino/providers/adm/provider.py
@@ -123,13 +123,13 @@ class Provider(BaseProvider):
     def hidden(self) -> bool:  # noqa: D102
         return False
 
-    def _normalize_query(self, query: str) -> str:
+    def normalize_query(self, query: str) -> str:
         """Convert a query string to lowercase and remove trailing spaces."""
         return query.strip().lower()
 
     async def query(self, srequest: SuggestionRequest) -> list[BaseSuggestion]:
         """Provide suggestion for a given query."""
-        q: str = self._normalize_query(srequest.query)
+        q: str = self.normalize_query(srequest.query)
         if (suggest_look_ups := self.suggestion_content.suggestions.get(q)) is not None:
             results_id, fkw_id = suggest_look_ups
             res = self.suggestion_content.results[results_id]
--- a/merino/providers/base.py
+++ b/merino/providers/base.py
@@ -57,16 +57,16 @@ class BaseProvider(ABC):
         """
         ...
 
-    def normalize_query(self, query: str) -> str:  # pragma: no cover
+    def _normalize_query(self, query: str) -> str:  # pragma: no cover
         """Normalize the query string when passed to the provider.
         Each provider can extend this class given its requirements. Can be used to
-        strip whitespace, handle case sensitivity, etc. Default is no-op.
+        strip whitespace, handle case sensitivity, etc. Default is to return query unchanged.
 
         Trailing spaces are not stripped in Firefox, so stripping trailing spaces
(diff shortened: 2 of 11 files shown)


linabutler (changes requested):
Nice refactoring, @taddes! 🚀 I left a few suggestions to think about.

Trinaa (changes requested):

taddes:
Thx all for your ideas and comments:

@linabutler 
- great suggestion on the comments, made the change. 
- made the change to return the query string unchanged in default impl. 👍 
- Since we're meeting shortly, I'll explain in more detail, but when @ncloudioj laid out the spec, he stated want each provider to handle its own normalization in cases where the use cases are different, not calling it at all in some cases since it isn't needed.  He cited that Firefox doesn't pass in leading space and hence we only strip trailing space (though, calling strip might totally be fine too!).  Wiki and weather don't require the normalization at all, hence not calling it needlessly for those two providers, meanwhile leaving the possibility for them to extend this in the future. Totally see the value and simplicity of that approach, so we can discuss it!

@quiiver made the revision to have it as an internal to the provider 👍 

@Trinaa 
- added coverage for `query` method in adm provider by parametrizing input which is inclusive of queries with spaces and upper case
- removed the test `test_query_case_insensitive` case and just added parameterized queries with expected outputs to the query tests, as you suggested (good call!)
- For`test_query_filtered_input`, regarding the empty result sets: when the query method for Top Picks is called, there are 3 situations that result in an empty result set: given a query <= to firefox min (2 char) the `https://` prefix and a term longer than our index char max, those are three separate cases of filtering the provider execute. Each paramaterized query value tests those 3 discrete options. I think we want to keep all 3 because it efficiently tests each case once, verifying the 3 possibilities. Good question, though.
- Moved the `test_initialize` unit tests under the init test, the no-change sanity case, and incorporated the ids in the `parametrize` call, I really liked that suggestion, it does indeed help clarify, so I may add it to some other places in the future.

However with leading spaces @ncloudioj had cited that Firefox does not send any leading whitespace, hence none of the logic filters out leading whitespace. Adding leading space would break the test as we don't strip it away, since our logic is only concerned with trailing space. See [comment in Jira](https://mozilla-hub.atlassian.net/browse/DISCO-2292?focusedCommentId=660526). I'm not opposed to us just using the `strip()` method and don't see harm in it. We'd just need the 👍 from @ncloudioj as to whether this is necessary or not.

ncloudioj:
This is overkill, don't use it for this PR ;). I shared this idea with @Trinaa earlier, she thought it was cool, so I'd like to share it here as well.

```py
# merino/utils/normalization.py

import string
from functools import reduce


def identity(s: str) -> str:
    return s


def trim(s: str) -> str:
    return s.rstrip()


def lower(s: str) -> str:
    return s.lower()


# For depunctuation
PUNCT_MAPPING = str.maketrans("", "", string.punctuation)
def depunctuation(s: str) -> str:
    return s.translate(PUNCT_MAPPING)


def compose(*funcs):
    """A composer to compose query normalizers"""
    def inner(f, g):
        return lambda s: f(g(s))
    return reduce(inner, funcs)


"""
class SomeProvider:
    def __init__(self):
        self.normalizer = compose(lower, trim, depunctuation)

    def normalize_query(self, query: str) -> str:
        return self.normalizer(query)
"""
```

ncloudioj:
> However with leading spaces @ncloudioj had cited that Firefox does not send any leading whitespace, hence none of the logic filters out leading whitespace. Adding leading space would break the test as we don't strip it away, since our logic is only concerned with trailing space.

I think it's okay to do `strip` here even if the leading spaces are already trimmed in Firefox :)

merino/providers/adm/provider.py
```diff
@@ -123,9 +123,16 @@ async def _fetch(self) -> None:
...
+        """Normalize the query string when passed to the provider.
+        For AdM, queries should be case-insensitive and trailing space should
+        be removed.
+        """
+        return query.rstrip().lower()
+
     async def query(self, srequest: SuggestionRequest) -> list[BaseSuggestion]:
         """Provide suggestion for a given query."""
-        q = srequest.query
+        q: str = self.normalize_query(srequest.query)
```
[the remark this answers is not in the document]
ncloudioj: Good idea! Since now we have the identity normalizer as the default, we can do this even if the provider doesn't override it.

merino/providers/base.py
```diff
@@ -35,26 +35,39 @@ class BaseProvider(ABC):
...
+        self, srequest: SuggestionRequest
+    ) -> list[BaseSuggestion]:  # pragma: no cover
         """Query against this provider.
 
         Args:
           - `srequest`: the suggestion request.
         """
         ...
 
+    def _normalize_query(self, query: str) -> str:  # pragma: no cover
```
linabutler, line 60: suggestion (blocking): It sounds like @ncloudioj is also on board with the suggestion in https://github.com/mozilla-services/merino-py/pull/239#discussion_r1139144951! Would you mind adding it in your next revision, please? ☺️ 

Responding to your feedback with a bit more detail:

> each provider to handle its own normalization in cases where the use cases are different, not calling it at all in some cases since it isn't needed. [...]  Wiki and weather don't require the normalization at all, hence not calling it needlessly for those two providers, meanwhile leaving the possibility for them to extend this in the future.

Thanks for the additional context—that helps me see where you're coming from!

I'm not sure that making `normalize_query` would be at odds with that spec—it would actually let you take advantage of making `normalize_query` part of the `BaseProvider`:

* If the provider wants to normalize inputs, then handling it in the `api_v1` layer means that the provider's `query` method will receive the normalized input, instead of having to remember to call `normalize_query` first.
* If the provider doesn't want normalized inputs, then the default implementation you just added will pass the query through as-is.

linabutler: nit: If you implement that suggestion, please also rename `_normalize_query` back to `normalize_query` (sorry! 😅)—because we'll be using it as part of the public `BaseProvider` interface.

merino/providers/top_picks/provider.py
```diff
@@ -52,14 +52,18 @@ async def initialize(self) -> None:
     def hidden(self) -> bool:  # noqa: D102
         return False
 
+    def _normalize_query(self, query: str) -> str:
+        """Convert a query string to lowercase and remove trailing spaces."""
+        return query.rstrip().lower()
```
linabutler, line 57: issue: You're using `record["domain"].strip()` above, but `query.rstrip()` here—for simplicity, let's use `strip()` everywhere.

```suggestion
        return query.strip().lower()
```

merino/providers/adm/provider.py
```diff
@@ -123,9 +123,13 @@ async def _fetch(self) -> None:
     def hidden(self) -> bool:  # noqa: D102
         return False
 
+    def _normalize_query(self, query: str) -> str:
+        """Convert a query string to lowercase and remove trailing spaces."""
+        return query.rstrip().lower()
```
linabutler, line 128:
```suggestion
        return query.strip().lower()
```

linabutler (changes requested):
Thanks @taddes! I wouldn't mind seeing this one more time before approving.

merino/providers/base.py
Trinaa: Given the above, the two `test_normalize_query` methods in `adm/test_provider.py` and `top_picks/test_provider.py` can become one `test_normalize_query` method in a new module `test_base.py`. A nice removal of duplication.

Trinaa (commented):
Removing the block, since the test I required has been added. TY Taddes!

merino/providers/adm/provider.py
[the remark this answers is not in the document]
taddes: Yes, terrific idea @linabutler , thanks for this suggestion and discussing the rationale! I've got it in the most recent push.

merino/providers/adm/provider.py
```diff
@@ -123,9 +123,13 @@ async def _fetch(self) -> None:
...
         return False
 
+    def normalize_query(self, query: str) -> str:
+        """Convert a query string to lowercase and remove trailing spaces."""
+        return query.strip().lower()
+
     async def query(self, srequest: SuggestionRequest) -> list[BaseSuggestion]:
         """Provide suggestion for a given query."""
-        q = srequest.query
+        q: str = self.normalize_query(srequest.query)
```
linabutler, line 132: nitpick: Since you're normalizing the query in `api_v1` now, you can revert this change!

```suggestion
        q = srequest.query
```

merino/providers/top_picks/provider.py
```diff
@@ -52,14 +52,18 @@ async def initialize(self) -> None:
...
+
     async def query(self, srequest: SuggestionRequest) -> list[BaseSuggestion]:
         """Query Top Pick data and return suggestions."""
         # Ignore https:// and http://
         if srequest.query.startswith("http"):
             return []
 
         qlen: int = len(srequest.query)
-        query: str = srequest.query
+        query: str = self.normalize_query(srequest.query)
```
linabutler, line 66: nitpick: Ditto here.

```suggestion
        query: str = srequest.query
```

linabutler (approved):
Fantastic job, @taddes! Just a couple of nits, but this looks great overall! Great job on folding in all the feedback, and the test coverage!

merino/providers/adm/provider.py
taddes: doh! Thank you!

Trinaa (approved):

taddes:
Closes #240

Issue #193, opened 2023-02-09
AdM Provider Internal Server Error During Initialization


data-sync-user:
A set of 500 - Internal Server Errors were found in Merino-py Stage during load test execution. Further investigation indicates the error is originating from the AdM provider during initialization, while establishing suggestion data from Remote Settings. The act of fetching data is conducted on an interval, so that Merino is able to recover. This error is considered a race-condition.

**Sentry Error Message**

!image-20230217-213622.png|width=1294,height=334!

**Code screenshots**

/merino-py/merino/providers/adm/provider.py

!image-20230217-213000.png|width=609,height=237!

**References**

* [Sentry](https://sentry.io/organizations/mozilla/issues/3927373152/?query=is%3Aunresolved&referrer=issue-stream)
* [Slack thread](https://mozilla.slack.com/archives/G01M6ALKVL4/p1675960973712519)

┆Issue is synchronized with this [Jira Bug](https://mozilla-hub.atlassian.net/browse/DISCO-2236)
┆Attachments: <a href="https://mozilla-hub.atlassian.net/rest/api/2/attachment/content/34134">image-20230217-213000.png</a> | <a href="https://mozilla-hub.atlassian.net/rest/api/2/attachment/content/34137">image-20230217-213622.png</a>

data-sync-user:
➤ Taddes Korris commented:

Hey Nan Jiang ! Just wanted to tag your thoughts on this one as Katrina Anderson and I were discussing this task. Wanted to know what you think the appropriate approach for this would be.

We could log that the error is happening in a metric to measure its frequency of occurrence. From there, we could ignore it if it isn’t causing problems or take some action. This way we’re at least taking note of the issue but not having to expend significant effort to solve it, especially if it’s not actually causing negative effects.

Alternatively, the trickier solution to this that might involve somehow tracking stat would be to not take in queries until all the providers are initialized and the indexes are in place.

data-sync-user:
➤ Nan Jiang commented:

[...]

PR #248, opened 2023-03-23
[DISCO-2236] AdM Provider Internal Server Error During Initialization

taddes:
## References

JIRA: [DISCO-2236](https://mozilla-hub.atlassian.net/browse/DISCO-2236)
GitHub: [#193 ](https://github.com/mozilla-services/merino-py/issues/193)

## Description

This solution will ensure the provider is initialized with empty data and that queries will instead provide no result, opposed to raising a server error. An empty `suggestion_content` object will be used until the cron job gets a chance to retry the lookup of suggestion data. The retry cron has a fairly low latency so having no data to serve during that timeframe should be fine. This error should be rare in practice and since there is a retry mechanism in place, we do not need to take any other actions. 

The solution is to implement the initialization of an empty SuggestionContent object at the initialization of the AdM provider in the rare case this issue crops up.

See Jira ticket for more relevant information.

## PR Review Checklist

[DISCO-2236]: https://mozilla-hub.atlassian.net/browse/DISCO-2236?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/tests/unit/providers/adm/conftest.py
+++ b/tests/unit/providers/adm/conftest.py
@@ -12,6 +12,17 @@ from merino.providers.adm.backends.protocol import AdmBackend, SuggestionContent
 from merino.providers.adm.provider import Provider
 
 
+@pytest.fixture(name="adm_uninitialized_suggestion_content")
+def fixture_adm_uninitialized_suggestion_content() -> SuggestionContent:
+    """Define empty suggestion content prior to provider initialization call."""
+    return SuggestionContent(
+        suggestions={},
+        full_keywords=[],
+        results=[],
+        icons={},
+    )
+
+
 @pytest.fixture(name="adm_suggestion_content")
 def fixture_adm_suggestion_content() -> SuggestionContent:
     """Define backend suggestion content for test."""
--- a/tests/unit/providers/adm/conftest.py
+++ b/tests/unit/providers/adm/conftest.py
@@ -12,17 +12,6 @@ from merino.providers.adm.backends.protocol import AdmBackend, SuggestionContent
 from merino.providers.adm.provider import Provider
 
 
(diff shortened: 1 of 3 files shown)


tests/unit/providers/adm/test_provider.py
```diff
@@ -89,6 +99,8 @@ async def test_initialize_remote_settings_failure(
     assert len(records) == 1
     assert records[0].__dict__["error message"] == error_message
     assert adm.last_fetch_at == 0
+    # SuggestionContent should be empty as initialize was unsuccessful.
+    assert adm.suggestion_content == adm_uninitialized_suggestion_content
```
ncloudioj, line 103: You can call `adm.query()` right here to ensure it returns an empty list without throwing anything.

tests/unit/providers/adm/conftest.py
```diff
@@ -12,6 +12,17 @@
 from merino.providers.adm.provider import Provider
 
 
+@pytest.fixture(name="adm_uninitialized_suggestion_content")
```
ncloudioj, line 15: I think we can drop this fixture as long as we do the check as mentioned in https://github.com/mozilla-services/merino-py/pull/248/files#r1147899323.

taddes:
Great comments @ncloudioj  👍 , I liked those suggestions for testing brevity. Made the changes to verify query returns empty result, removed the fixture, though I did leave a single test to clearly show an empty `SuggestionContent` object as I thought it compliments the logic. Lmk if there's anything else you'd like changed!

ncloudioj (approved):
LGTM, thanks!

taddes:
Closes #248

PR #260, opened 2023-04-04
Merino job for creating navigational suggestion specific artifacts

abhi-agg:
## References

JIRA: https://mozilla-hub.atlassian.net/browse/DENG-524

## Description
This PR adds the capability to generate navigation suggestion specific artifacts via merino jobs. These artifacts will be consumed by merino.
- Uploads domain icons to mozilla's internal cdn (GCS)
- Uploads top picks file to mozilla's internal cdn (GCS)

Tested this change by running the command locally:
`merino-jobs navigational-suggestions prepare-domain-metadata --dst-gcp-project "wstuckey-sandbox" --dst-gcs-bucket "merino-images-sandbox"`

I couldn't use the config specified in `development.toml` file as I don't have write access to the corresponding GCS project

## PR Review Checklist

## ToDo
Integration and unit tests will follow soon. I pushed the PR to seek an early feedback.


merino/configs/development.toml
```diff
@@ -25,3 +25,11 @@ api_key = "test"
 gcs_path = "merino-jobs-dev/wikipedia-exports"
 # GCP project name where the GCS bucket lives.
 gcp_project = "contextual-services-dev"
+
+[development.jobs.navigational_suggestions]
+# GCP project name that contains domain data tables
+source_gcp_project = "mozdata"
```
ncloudioj, line 31: After a second thought, perhaps we should use a dev GCP project & buckets for development rather than `merino-nonprod` since the latter is only accessible to most of us.

Somewhat related, you can also override those configurations via local configurations, such as putting overridding settings into `default.local.toml`, those are not tracked by source control, so you can put your custom settings/secrets there.

merino/jobs/navigational_suggestions/__init__.py
```diff
@@ -0,0 +1,101 @@
...
+    help="Command for preparing top domain metadata for navigational suggestions",
+)
+
+
+def _construct_top_picks(
+    domain_data: list[dict],
+    favicons: list[str],
+    urls_and_titles: list[dict],
+    second_level_domains: list[str],
+):
```
ncloudioj, line 52: FYI, we encourage folks to adopt type annotation for Merino so that static type analyzers such as Mypy can help us capture typing errors sooner. 
```suggestion
) -> str:
```

merino/jobs/navigational_suggestions/__init__.py
```diff
@@ -0,0 +1,101 @@
...
+    result = []
+    for index in range(len(domain_data)):
```
ncloudioj, line 54:
```suggestion
    for index, domain in enumerate(domain_data):
```

merino/jobs/navigational_suggestions/domain_metadata_extractor.py
```diff
@@ -0,0 +1,215 @@
...
+    """Extract domain metadata from domain data"""
+
+    LINK_SELECTOR = (
+        "link[rel=apple-touch-icon], link[rel=apple-touch-icon-precomposed],"
+        'link[rel="icon shortcut"], link[rel="shortcut icon"], link[rel="icon"],'
+        'link[rel="SHORTCUT ICON"], link[rel="fluid-icon"]'
+    )
+    META_SELECTOR = "meta[name=apple-touch-icon]"
+    FIREFOX_UA = (
+        "Mozilla/5.0 (Macintosh; Intel Mac OS X 10.12; rv:58.0)"
```
ncloudioj, line 24: Let's update UA version to the latest, e.g. `112`. Didn't realize that we've gone through that MANY releases of Firefox over the past years.

ncloudioj:
Hi @abhi-agg, I've skimmed over your patch, looks pretty good. Could you try adding some tests for it? For context, we've enforced the minimal test coverage (95%) for Merino, including all the Merino jobs, so tests are required for all new code additions.

You can reference Merino's [testing guidance](https://mozilla-services.github.io/merino-py/dev/testing.html) for more details.

merino/jobs/navigational_suggestions/__init__.py
abhi-agg: Hey @ncloudioj This is great. I had added type annotation everywhere but missed this one 👍🏾

merino/jobs/navigational_suggestions/__init__.py
abhi-agg: Done.

merino/jobs/navigational_suggestions/domain_metadata_extractor.py
abhi-agg: Done. Setting it to 111 as that is the latest public release.

merino/configs/development.toml
abhi-agg: Do you mean that we can leave these entries empty here and let every developer override them via `default.local.toml` file for local development?

ncloudioj: I mean if we find it handy to have a dev environment, we can set up a dev environment and store those settings here. But we probably won't use the merino staging environment for that, so let's not save those staging settings here.

abhi-agg: @ncloudioj Got it. For now, I will remove gcs staging settings and leave them empty here just like in `default.toml`.

I am wondering what should be done for `source_gcp_project`. Considering that `mozdata is accessible to Mozilla staff by default as per [this](https://docs.telemetry.mozilla.org/cookbooks/bigquery/access.html#bigquery-access-request), I think we can keep it. If you feel otherwise then I can remove it as well. What do you think?

taddes (changes requested):
Great work @abhi-agg , there's a lot of great stuff here and its nice the job resides here in the repo! I have some suggestions on adding a backup lookup function, expanding the invalid titles list and a few minor nits like adding context to exception handling.

I spoke with @ncloudioj and he suggested we can look at working together on getting unit test coverage up, doing some end to end test setup, etc.  That or I can work off this branch and add the testing, as we'll have to get some mocking in place and also getting some dev GCP data setup so we can test the process.

merino/jobs/navigational_suggestions/domain_metadata_extractor.py
```diff
@@ -0,0 +1,216 @@
...
+        "Access to this page has been denied",
+        "Loading…",
+        "Page loading",
+        "Just a moment...",
+    ]
+
+    browser: RoboBrowser
+
+    def __init__(self) -> None:
+        self.browser = RoboBrowser(user_agent=self.FIREFOX_UA, parser="html.parser")
```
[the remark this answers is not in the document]
quiiver: what does `mechanize` provide here that `RoboBrowser` does not?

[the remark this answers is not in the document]
taddes: Hey @quiiver ! It's not so much a question of what one provides over the other. 

In my investigation of invalid titles as I shared in [Jira here](https://mozilla-hub.atlassian.net/browse/DISCO-2300), `mechanize` did the best job of re-checking the invalid titles after this process and had a high success rate. Web App Firewalls, anti-bot logic, scraping countermeasures all limit our ability to capture titles. By implementing another secondary means to lookup and verify titles, we get a significant benefit for little cost, in my opinion. Far better to have a full title opposed to just a domain name. If `RoboBrowser` was the initial library used to capture titles for the first `top_picks.json` file, there were a lot of empty and invalid titles, hence the need to use another library after the `RoboBrowser` lookup. Mechanize and its settings have a different fingerprint and configured headers, so it may not trigger the same WAF/bot blocks.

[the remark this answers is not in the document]
quiiver: Thanks! AFAICT they are both web scraping libraries. meaning that they are just sending HTTP requests and not running a headless browser. Given that they should be sending roughly the same HTTP requests or can be configured to send the same request via header configuration. I'm not quite sure what else you mean by 'fingerprinting' 

RoboBrowser was not the library i used initially, i just used requests and beautiful soup for html parsing. very bare bones. I'm fairly certain, though this hasn't been well tested, that the reason for the high block rate was not the scraping library used rather that the requests came from within the GCP network. meaning the exit IP would correlate to a GCP owned IP which is a pretty common thing to block. Therefore, i'm not certain that adding the complexity of another library would necessarily solve the issue since this will be run from within GCP via Airflow.

merino/jobs/navigational_suggestions/domain_metadata_extractor.py
```diff
@@ -0,0 +1,216 @@
...
+        'link[rel="SHORTCUT ICON"], link[rel="fluid-icon"]'
+    )
+    META_SELECTOR = "meta[name=apple-touch-icon]"
+    FIREFOX_UA = (
+        "Mozilla/5.0 (Macintosh; Intel Mac OS X 13.3; rv:111.0) Gecko/20100101 "
+        "Firefox/111.0"
+    )
+    TIMEOUT = 60
+    # A non-exhaustive list of substrings of invalid titles
+    INVALID_TITLES = [
```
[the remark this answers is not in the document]
abhi-agg: @taddes We are running `Robobrowser` to extract titles and the reason that the blocked title list is so small because I didn't encounter the rest of the ones that you mentioned here after running this code on top 1k domains.

I would just like to mention that the invalid title list contains the invalid title substrings and not the complete invalid strings. The most common error that you mentioned is already covered by `Attention Required` substring which is already present in the list. Doing this will also keep our blocked list short. e.g. `Access to this page has been denied` in the list also covers for `Access to this page has been denied.`. So we don't need to add `Access to this page has been denied.` separately.

However, I am fine with adding the rest of the invalid titles that you mentioned to the blocked list just as a safe measure.

merino/jobs/navigational_suggestions/domain_metadata_extractor.py
```diff
@@ -0,0 +1,216 @@
...
+            domain = domain_data["domain"]
+            url = f"https://{domain}"
+            title = self._extract_title(url)
+            if title is None and "www." not in domain:
+                # Retry with www. in the domain as some domains require it explicitly.
+                url = f"https://www.{domain}"
+                title = self._extract_title(url)
+
+            # if no valid title is present then fallback to use the second level domain as title
+            if title is None:
```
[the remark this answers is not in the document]
abhi-agg: This is already covered in `_extract_title` function.

merino/jobs/navigational_suggestions/domain_metadata_extractor.py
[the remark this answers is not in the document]
abhi-agg: @taddes Thanks a lot for running the experiment with `mechanize`. As per your Jira ticket:

1. There were 131 domains with invalid titles (due to web scraping issues, resulting in titles like ‘Access Denied or 'Attention Required! | Cloudflare')
2. 16 have no title at all

Could you share that for how many of the 131 invalid titles was mechanize unsuccessful to return a valid title?

abhi-agg:
@taddes Thanks for reviewing the PR. I am currently writing the tests for this feature. Let me get back to you if I have  questions regarding that. I might need your help in setting up a dev GCP for running tests 🙂

merino/configs/development.toml
abhi-agg: @ncloudioj I removed the merino specific staging environment settings from here. Does it look good now?

merino/jobs/navigational_suggestions/domain_metadata_extractor.py
```diff
@@ -0,0 +1,216 @@
...
+        return result
+
+    def _extract_title(self, url: str) -> Optional[str]:
+        logger.info(f"Extracting titles for {url}")
+        titles = []
+        try:
+            self.browser.open(url, timeout=self.TIMEOUT)
+            titles = [title.text for title in self.browser.select("title")]
+        except Exception as e:
+            logger.info(f"Exception: {str(e)} while extracting titles from document")
```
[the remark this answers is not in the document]
abhi-agg: > I would also recommend supplying the url to the exception so that logging will indicate the problematic url. It may be useful for possible metrics and error reporting.

Hey @taddes Line 172 already has url information.

merino/configs/development.toml
ncloudioj: I think you can just drop `destination_gcp_project` and `destination_gcs_bucket` from here, it will use those from `default.toml`.

Other than that, this looks good. Thanks!

abhi-agg: Great. I will do it. Thanks.

abhi-agg:
Hey @ncloudioj @quiiver @taddes , I have addressed all the comments on this PR and pushed changes to resolve all of them. I added unit and integration tests and CI is green now. Please feel free to review the PR.

merino/jobs/navigational_suggestions/domain_metadata_extractor.py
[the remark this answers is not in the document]
taddes: Great, thanks for explaining that. From Wil's comment below it makes sense now given another library was used before RoboBrowser. Again, it all seems to be what headers are defined, some of which are set by the library unless overridden, which define how successful a lookup is.  Thanks for adding them though, knowing we have them just in case, as well as the substring check is a good idea.

merino/jobs/navigational_suggestions/domain_metadata_extractor.py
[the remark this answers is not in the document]
taddes: Hey @quiiver , yup, they are scraping libraries, but depending on what headers are set - some of which are defaulted in the libraries - there are sometimes pretty different result. I noticed significant performance differences between BeautifulSoup and Mechanize, and it's just based on what header options are defined or overridden. Good to know that RoboBrowser wasn't the original, I experimented with the simple HTTP requests and BeautifulSoup before and found a significant performance difference, given whatever options and headers are set.  That's what I meant by fingerprinting, a lot of server logic, to my knowledge, is enabled to detect scraping libraries and this can be bypassed depending on how a scraper is configured. I'm no expert in this field per-se, but toying around with libraries and configs have shown me they don't all behave the same, nor deliver the same results.  Fine by me to not add another if you think that's best, maybe we can see how much of the titles the current job will capture and review. Certainly this will be perfectly enough for the experiment.

merino/jobs/navigational_suggestions/domain_metadata_extractor.py
```diff
@@ -0,0 +1,224 @@
...
+
+            # Some domains have a default "favicon.ico" in their root without explicitly
+            # specifying them via rel attribute of link tag.
+            default_favicon_url = self._get_default_favicon(url)
+            if default_favicon_url is not None:
+                favicons.append({"href": default_favicon_url})
+
+        except Exception as e:
+            logger.info(f"Exception {str(e)} while extracting favicons for {url}")
+            pass
```
ncloudioj, line 100:
```suggestion
            logger.warning(f"Exception {e} while extracting favicons for {url}")
```

merino/jobs/navigational_suggestions/domain_metadata_extractor.py
```diff
@@ -0,0 +1,224 @@
...
+                    with Image.open(BytesIO(response.content)) as img:
+                        width, height = img.size
+                        if width != height:
+                            logger.info(
+                                f'favicon {favicon} shape "{width}*{height}" is not square'
+                            )
+                            width = min(width, height)
+                except Exception as e:
+                    logger.info(f'Exception "{str(e)}" for favicon {favicon}')
+                    pass
```
ncloudioj, line 149:
```suggestion
                    logger.warning(f'Exception "{e}" for favicon {favicon}')
```

merino/jobs/navigational_suggestions/domain_metadata_extractor.py
```diff
@@ -0,0 +1,224 @@
...
+
+    def _extract_title(self, url: str) -> Optional[str]:
+        logger.info(f"Extracting titles for {url}")
+        titles = []
+        try:
+            self.browser.open(url, timeout=self.TIMEOUT)
+            titles = [title.text for title in self.browser.select("title")]
+        except Exception as e:
+            logger.info(f"Exception: {str(e)} while extracting titles from document")
+            pass
```
ncloudioj, line 185:
```suggestion
            logger.warning(f"Exception: {e} while extracting titles from document")
```

merino/jobs/navigational_suggestions/domain_metadata_extractor.py
```diff
@@ -0,0 +1,224 @@
...
+        'link[rel="SHORTCUT ICON"], link[rel="fluid-icon"]'
+    )
+    META_SELECTOR = "meta[name=apple-touch-icon]"
+    FIREFOX_UA = (
+        "Mozilla/5.0 (Macintosh; Intel Mac OS X 13.3; rv:111.0) Gecko/20100101 "
+        "Firefox/111.0"
+    )
+    TIMEOUT = 60
+    # A non-exhaustive list of substrings of invalid titles
+    INVALID_TITLES = [
```
ncloudioj, line 29: We can use a set for this list.

merino/jobs/navigational_suggestions/domain_metadata_extractor.py
```diff
@@ -0,0 +1,224 @@
...
+        except Exception as e:
+            logger.info(f"Exception: {str(e)} while extracting titles from document")
+            pass
+
+        return (
+            titles[0]
+            if len(titles) > 0
+            and not any(map(titles[0].__contains__, self.INVALID_TITLES))
+            else None
+        )
```
ncloudioj, line 192: This will only check the first title of the titles list, the following returns the _first_ title that is not in INVALID_TITLES. Alternatively, we can get all the valid titles and then use the longest one assuming that's more informative than other shorter ones. 

```py
from itertools import islice, takewhile

valid_titles: list[str] = list(
    islice(takewhile(lambda t: t not in self.INVALID_TITLES, titles), 1)
)
return valid_titles[0] if valid_titles else None
```

ncloudioj (approved):
LGTM, r+wc.

@abhi-agg Thanks for this great work!

ncloudioj:
Would also like to note that I don't think we should go super deep in polishing the title quality in this PR, image hosting is the key for the moment, we can improve the former as follow-up later.

merino/jobs/navigational_suggestions/domain_metadata_extractor.py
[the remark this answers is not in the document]
taddes: Hi @abhi-agg . I don't recall exactly the number, but it was close to 30 that it was still unsuccessful for, just like in your test. I recall lowes was a domain it really had trouble with. If it's a similar count then that sounds good to me!

merino/jobs/navigational_suggestions/domain_metadata_extractor.py
taddes, line 29: Agreed with @ncloudioj , set is a good call

tests/unit/jobs/navigational_suggestions/test___init__.py
```diff
@@ -0,0 +1,63 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
```
ncloudioj, line 1: Let's rename this to `test_navigational_suggestions.py`.

merino/jobs/navigational_suggestions/domain_metadata_extractor.py
```diff
@@ -0,0 +1,224 @@
...
+
+
+class DomainMetadataExtractor:
+    """Extract domain metadata from domain data"""
+
+    LINK_SELECTOR = (
+        "link[rel=apple-touch-icon], link[rel=apple-touch-icon-precomposed],"
+        'link[rel="icon shortcut"], link[rel="shortcut icon"], link[rel="icon"],'
+        'link[rel="SHORTCUT ICON"], link[rel="fluid-icon"]'
+    )
```
taddes, line 21: Just a minor nit here, but if the linter permits, perhaps we want a single space after each constant? Also, while it may not be enforced by mypy and implicit, perhaps adding an explicit type declaration to each one is the best way to go for consistency. Ex. `META_SELECTOR: str`. Totally minor, though.

merino/jobs/navigational_suggestions/domain_metadata_extractor.py
```diff
@@ -0,0 +1,224 @@
...
+            if default_favicon_url is not None:
+                favicons.append({"href": default_favicon_url})
+
+        except Exception as e:
+            logger.info(f"Exception {str(e)} while extracting favicons for {url}")
+            pass
+
+        return favicons
+
+    def _get_best_favicon(self, favicons: list[dict]) -> str:
```
taddes, line 104: Missing docstring?

merino/jobs/navigational_suggestions/domain_metadata_extractor.py
```diff
@@ -0,0 +1,224 @@
...
+                favicons = self._extract_favicons(url)
+
+            logger.info(
+                f"{len(favicons)} favicons extracted for {url}. Favicons are: {favicons}"
+            )
+            best_favicon = self._get_best_favicon(favicons)
+            result.append(best_favicon)
+        return result
+
+    def _extract_title(self, url: str) -> Optional[str]:
```
taddes, line 177: Missing docstring?

merino/jobs/navigational_suggestions/domain_metadata_extractor.py
```diff
@@ -0,0 +1,224 @@
...
+
+    def __init__(self) -> None:
+        self.browser = RoboBrowser(user_agent=self.FIREFOX_UA, parser="html.parser")
+
+    def _fix_url(self, url: str) -> str:
+        if not url.startswith("http"):
+            return f"https:{url}"
+        return url
+
+    def _get_default_favicon(self, url: str) -> Optional[str]:
```
taddes, line 55: Missing docstring.

merino/jobs/navigational_suggestions/domain_metadata_extractor.py
```diff
@@ -0,0 +1,224 @@
...
+        default_favicon_url = f"{url}/favicon.ico"
+        response = requests.get(
+            url, headers={"User-agent": self.FIREFOX_UA}, timeout=self.TIMEOUT
+        )
+        return default_favicon_url if response.status_code == 200 else None
+
+    def _extract_favicons(self, url: str) -> list[dict]:
```
taddes, line 62: Missing docstring?

merino/jobs/navigational_suggestions/domain_metadata_extractor.py
taddes: +1

taddes (approved):
Great stuff @abhi-agg . I have a few comments, but will unblock for approval as things look good. Just a few suggestions along with @ncloudioj . There are a few docstrings missing and a few minor nits regarding type definitions for constants.  I know for private methods, you don't always need to define a docstring, but I think it doesn't hurt.

Also, and this is a broader recommendation I'd like us to consider as an enhancement (or if you wanted to add some quick logic here), but I think having some metrics to look at, or at least a dictionary that captures the invalid titles or those that we have to provide a basic domain for would be useful. This way, we can have some insights into what fails on the job when it comes to title scraping. Just a thought, not at all a blocker for you, but I wanted us to consider it. It's a small addition that I think could give us good insights!

Thanks for the great work! Let me know if you want any help with the merging process or have any other questions.

merino/jobs/navigational_suggestions/domain_metadata_extractor.py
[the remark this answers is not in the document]
abhi-agg: Resolving this one as well.

merino/jobs/navigational_suggestions/domain_metadata_extractor.py
abhi-agg: @ncloudioj Is there a specific advantage of using set here?

I would like to add that this list contains the invalid title "substrings" and not the complete invalid title strings. So, after extracting a title, I would be checking whether the title contains any one of these substrings.

quiiver (approved):
looks good. thanks @abhi-agg

merino/jobs/navigational_suggestions/domain_metadata_extractor.py
abhi-agg: @ncloudioj Ideally a page shouldn't have multiple `title` tags. But if it does (e.g. https://www.microsoft.com, https://live.com) then I am selecting the first title tag encountered while parsing the document to be the title of the website. I actually ran this for top 1k websites and filtered the websites that had multiple titles. I found that for all these websites the first title was the relevant one because `<head>` element that contains the `title` tag of a page is specified in the beginning of a document.

Now one can argue that there might be exceptions to this. Therefore, to make the current logic more robust, I can parse `title` tag of the `head` element specifically (because that is our actual title) rather than parsing all the title tags of a document. Submitting this change right away.

I believe, considering other title tags to be the contenders of a website's title might give us weird results.

Please let me know if I missed something.

merino/jobs/navigational_suggestions/domain_metadata_extractor.py
ncloudioj: Oh, I see. You wanted to check if the title is a substring of any of INVALID_TITLES rather than exact matching. In that case, using the list here makes sense. Two extra notes:

merino/jobs/navigational_suggestions/domain_metadata_extractor.py
ncloudioj: If we're sure the first one is likely the best one, then I'd recommend dropping the rest of the candidates at the page parsing time (as none of them would be considered anyway) to avoid confusion.

merino/jobs/navigational_suggestions/domain_metadata_extractor.py
abhi-agg: Adopting both of your suggestions 👍🏾

merino/jobs/navigational_suggestions/domain_metadata_extractor.py
abhi-agg: @ncloudioj This is exactly what I did in my last commit. I am just parsing the `title` tag of the `head` element now instead of parsing the entire document for all `title` tags followed by choosing the first one 👍🏾

Issue #261, opened 2023-04-06
Exclude sensitive data from Sentry reports


taddes:
Ensure sensitive data is not sent to Sentry Reporting. 

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/DISCO-2235)

PR #262, opened 2023-04-06
[DISCO-2235] Exclude Sensitive Data from Sentry 

taddes:
## References

JIRA: [DISCO-2235](https://mozilla-hub.atlassian.net/browse/DISCO-2235)
GitHub: [#261 ](https://github.com/mozilla-services/merino-py/issues/261)

## Description
It is essential to filter out potentially sensitive data when emitting events to Sentry. Currently, query strings and search terms are the values to be filtered and removed.

Accessing this through sentry requires defining a callback function that is associated with the `before_send` parameter in sentry_init(). See [Sentry documentation here](https://sentry-docs-o2paie5ivq-uc.a.run.app/error-reporting/configuration/filtering/?platform=python). This function grants access to the `event` object, in which all context of stack frames exit.  

It was found that at the API request level, the `query_string` parameter passed to Sentry is found at `event["request"]["query_string"]`. 

What is more complicated is finding other instances of the query parameter for each provider, matching the key `q` or `query`. This requires iterating through `event['exception']['values'][0]['stacktrace']['frames']` where each frame is contained. This contained 3 possible locations where the `q` value occurred:
- `event["exception"]["values"][0]["stacktrace"]["frames"][<idx>]["vars"]["q"]`
-  `event["exception"]["values"][0]["stacktrace"]["frames"][<idx>]["vars"]["values"]["q"]`
- `event["exception"]["values"][0]["stacktrace"]["frames"][<idx>]["vars"]["solved_result"][0]["q"]`

Given the nested nature of these values, a simple `try/except` block with a `continue` statement when a `KeyError` is raised was used to minimize complexity.

To explore the event object, do the following:
1. Define `pdb.set_trace()` at the end of the `strip_sensitive_data` function, before the return of the event.
2.  Raise a `RuntimeException()` in `api_v1.py` [here](https://github.com/mozilla-services/merino-py/blob/dae6c05fe694fffa344706fd963781966cd6bb23/merino/web/api_v1.py#L99) 
3. Interactively explore the event object context. 

Note that currently, I have the values set to empty strings. Alternatively, the dictionary keys can be deleted. For testing sake, I had originally used discrete strings to make sure each case modified the actual query term. If there's a good reason to instead just delete all possible query keys, we could do that.
## PR Review Checklist

[DISCO-2235]: https://mozilla-hub.atlassian.net/browse/DISCO-2235?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/merino/config_sentry.py
+++ b/merino/config_sentry.py
@@ -1,5 +1,7 @@
 """Sentry Configuration"""
 
+from typing import Any
+
 import sentry_sdk
 from sentry_sdk.integrations.fastapi import FastApiIntegration
 from sentry_sdk.integrations.starlette import StarletteIntegration
@@ -22,9 +24,16 @@ def configure_sentry() -> None:  # pragma: no cover
         ],
         release=version_sha,
         debug="debug" == settings.sentry.mode,
+        before_send=strip_sensitive_data,  # type: ignore [arg-type]
         environment=settings.sentry.env,
         # Set traces_sample_rate to 1.0 to capture 100%
         # of transactions for performance monitoring.
         # We recommend adjusting this value in production,
         traces_sample_rate=settings.sentry.traces_sample_rate,
     )
+
+
+def strip_sensitive_data(event) -> Any:
+    """Filter out sensitive data from Sentry events."""
+    #  See: https://docs.sentry.io/platforms/python/configuration/filtering/
+    return event
--- a/merino/config_sentry.py
+++ b/merino/config_sentry.py
@@ -41,7 +41,7 @@ def strip_sensitive_data(event: dict, hint: dict) -> Any:
     """Filter out sensitive data from Sentry events."""
     #  See: https://docs.sentry.io/platforms/python/configuration/filtering/
     if event["request"]["query_string"]:
-        event["request"]["query_string"] = "query_str_foo"
+        event["request"]["query_string"] = ""
 
     if "exc_info" in hint:
         exc_type, exc_value, tb = hint["exc_info"]
@@ -49,20 +49,15 @@ def strip_sensitive_data(event: dict, hint: dict) -> Any:
             for entry in event["exception"]["values"][0]["stacktrace"]["frames"]:
                 try:
                     if entry["vars"].get("q"):
-                        entry["vars"]["q"] = "vars_foo"
+                        entry["vars"]["q"] = ""
                     if entry["vars"].get("query"):
-                        entry["vars"]["query"] = "picked_repl"
+                        entry["vars"]["query"] = ""
                     if entry["vars"].get("srequest"):
-                        entry["vars"]["srequest"] = "sreq_repl"
+                        entry["vars"]["srequest"] = ""
                     if entry["vars"]["values"].get("q"):
-                        entry["vars"]["values"]["q"] = "vars_values_foo"
+                        entry["vars"]["values"]["q"] = ""
                     if entry["vars"]["solved_result"][0].get("q"):
-                        entry["vars"]["solved_result"][0]["q"] = "solved_foo"
-
-                except KeyError as e:
-                    logger.debug(
-                        f"Sentry filtering RuntimeError query value not found: {e}"
-                    )
+                        entry["vars"]["solved_result"][0]["q"] = ""
+                except KeyError:
                     continue
-
     return event
--- a/merino/config_sentry.py
+++ b/merino/config_sentry.py
@@ -1,5 +1,6 @@
 """Sentry Configuration"""
 
+import logging
 from typing import Any
 
 import sentry_sdk
@@ -8,6 +9,8 @@ from sentry_sdk.integrations.starlette import StarletteIntegration
 
 from merino.config import settings
 from merino.utils.version import fetch_app_version_from_file
+
+logger = logging.getLogger(__name__)
 
 
 def configure_sentry() -> None:  # pragma: no cover
@@ -24,7 +27,7 @@ def configure_sentry() -> None:  # pragma: no cover
         ],
         release=version_sha,
         debug="debug" == settings.sentry.mode,
-        before_send=strip_sensitive_data,  # type: ignore [arg-type]
+        before_send=strip_sensitive_data,
         environment=settings.sentry.env,
         # Set traces_sample_rate to 1.0 to capture 100%
         # of transactions for performance monitoring.
@@ -33,11 +36,27 @@ def configure_sentry() -> None:  # pragma: no cover
     )
 
 
-def strip_sensitive_data(event) -> Any:
+def strip_sensitive_data(event: dict, hint: dict) -> Any:
     """Filter out sensitive data from Sentry events."""
     #  See: https://docs.sentry.io/platforms/python/configuration/filtering/
-    if event.q:
-        delattr(event, "q")
-    if event.query:
-        delattr(event, "query")
+    if event["request"]["query_string"]:
+        event["request"]["query_string"] = "query_str_foo"
+
+    if "exc_info" in hint:
+        exc_type, exc_value, tb = hint["exc_info"]
+        if isinstance(exc_value, RuntimeError):
+            for entry in event["exception"]["values"][0]["stacktrace"]["frames"]:
+                try:
+                    if entry["vars"].get("q"):
+                        entry["vars"]["q"] = "vars_foo"
+                    if entry["vars"]["values"].get("q"):
+                        entry["vars"]["values"]["q"] = "vars_values_foo"
+                    if entry["vars"]["solved_result"][0].get("q"):
+                        entry["vars"]["solved_result"][0]["q"] = "solved_foo"
+                except KeyError as e:
+                    logger.debug(
+                        f"Sentry filtering RuntimeError query value not found: {e}"
+                    )
+                    continue
+
     return event
--- a/merino/config_sentry.py
+++ b/merino/config_sentry.py
@@ -58,7 +58,9 @@ def strip_sensitive_data(event: dict, hint: dict) -> dict:
                 if entry["vars"]["solved_result"][0].get("q"):
                     entry["vars"]["solved_result"][0]["q"] = ""
 
-    except KeyError as e:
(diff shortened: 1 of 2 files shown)


merino/config_sentry.py
```diff
@@ -22,9 +28,36 @@ def configure_sentry() -> None:  # pragma: no cover
...
         # of transactions for performance monitoring.
         # We recommend adjusting this value in production,
         traces_sample_rate=settings.sentry.traces_sample_rate,
     )
+
+
+def strip_sensitive_data(event: dict, hint: dict) -> Any:
+    """Filter out sensitive data from Sentry events."""
+    #  See: https://docs.sentry.io/platforms/python/configuration/filtering/
+    if event["request"]["query_string"]:
```
ncloudioj, line 43: An exception raised from outside of the handler paths (e.g. from the cron jobs) might blow off this test. I guess you can use pattern matching here:

```py
match event:
    case {
        "request": {
            "query_string": _
        }
    }:
        event["request"]["query_string"] = ""
    case _:
        pass

```

merino/config_sentry.py
```diff
@@ -22,9 +28,36 @@ def configure_sentry() -> None:  # pragma: no cover
...
+        event["request"]["query_string"] = ""
+
+    if "exc_info" in hint:
+        exc_type, exc_value, tb = hint["exc_info"]
+        if isinstance(exc_value, RuntimeError) or isinstance(exc_value, BackendError):
```
ncloudioj, line 48:
```suggestion
        if isinstance(exc_value, (RuntimeError, BackendError)):
```
BTW, I don't think we have explicitly raised any `RuntimeError` in Merino, shall we just do this for all exceptions?

merino/config_sentry.py
```diff
@@ -22,9 +28,36 @@ def configure_sentry() -> None:  # pragma: no cover
...
+            for entry in event["exception"]["values"][0]["stacktrace"]["frames"]:
```
ncloudioj, line 49: Q: do we just need to handle the first entry of the `values` list? Are `stacktrace` and `frames` always present in the entry? If we're not sure about that, let's wrap this inside a "try/except" just in case.

merino/config_sentry.py
```diff
@@ -22,9 +28,36 @@ def configure_sentry() -> None:  # pragma: no cover
...
+                    if entry["vars"].get("q"):
+                        entry["vars"]["q"] = ""
+                    if entry["vars"].get("query"):
+                        entry["vars"]["query"] = ""
+                    if entry["vars"].get("srequest"):
+                        entry["vars"]["srequest"] = ""
+                    if entry["vars"]["values"].get("q"):
+                        entry["vars"]["values"]["q"] = ""
+                    if entry["vars"]["solved_result"][0].get("q"):
+                        entry["vars"]["solved_result"][0]["q"] = ""
```
ncloudioj, line 60: Maybe do this instead? It won't raise so try/except is not needed here.

```py
vars = entry.get("var", {})
match vars:
    case { "q": _ }:
        vars["q"] = ""
    case { "query": _ }:
        vars["query"] = ""
    case { "srequest": { "query": _ } }:
        vars["srequest"]["query"] = ""
    case { "values": { "q": _ } }:
        vars["values"]["q"] = ""
    case { "solved_result": [ { "q": _ }, *_ ] }:  # `*_` matches a requence of any length including 0
        vars["solved_result"][0]["q"] = ""
    case _:
        pass
```

merino/config_sentry.py
```diff
@@ -1,12 +1,18 @@
...
+
 import sentry_sdk
 from sentry_sdk.integrations.fastapi import FastApiIntegration
 from sentry_sdk.integrations.starlette import StarletteIntegration
 
 from merino.config import settings
+from merino.exceptions import BackendError
 from merino.utils.version import fetch_app_version_from_file
 
+logger = logging.getLogger(__name__)
```
ncloudioj, line 14: Unused, forgot to drop?

merino/config_sentry.py
taddes: Love it! Much cleaner, thanks for the suggestion. Only hiccup I'm having is :
```python
case { "solved_result": [ { "q": _ }, *_ ] }:  # `*_` matches a requence of any length including 0
    vars["solved_result"][0]["q"] = ""```

I'm working on tweaking it, but mypy throws the below error and that particular result is not being sanitized.  Lmk if you have any thoughts, meanwhile I'm trying to work out an equally flexible tweak that works in the match case.
```python
merino/config_sentry.py:76: error: Statement is unreachable  [unreachable]
                            vars["solved_result"][0]["q"] = ""
                            ```

File: merino/config_sentry.py (hunk 2 above)
Comment (taddes) [reply merino/config_sentry.py:48]: Correct, therefore I think that's sensible!  Many child exceptions fall under `RuntimeError`, but capturing all under `Exception` is a good idea.

**Comment by taddes** [comment]
Hey @ncloudioj! Thanks again for the great suggestions. So, I experimented with the match case only to discover that there is a single frame in which `"values" `and `"solved_result"` occur, meaning that the conditional check for one would bypass the other, as if all the checks were chains of elif clauses. 

@linabutler was also wonderfully helpful in recommending some additional `.get()` calls in this context so there are no cases of Key errors when specifically accessing the "`solved_result"`. It was set at the end as it's one of the last most frames we see in the `event` payload, however it was throwing a KeyError in some instances where `"values"` is in the same frame. You'll see new checks added for that. 

I know it's less pretty to have the match case, but following the debugger, I ensured that there wouldn't be key errors in the internal iteration. I did also add a `try/except` block to capture the entirety of the check in case the `event` doesn't resemble what we expect, which is very unlikely, but your comment made sense.

Lina also suggested option that we clear everything in vars. That's a simple `entry.pop("vars", None) call within the loop for each entry. However, that clears out everything and no local vars are then in the stack trace, so wanted to run that by you if it happened to be an option you'd want to consider.

File: merino/config_sentry.py [hunk 6]
```diff
@@ -22,9 +26,51 @@ def configure_sentry() -> None:  # pragma: no cover
...
+    #  See: https://docs.sentry.io/platforms/python/configuration/filtering/
+    match event:
+        case {"request": {"query_string": _}}:
+            event["request"]["query_string"] = ""
+        case _:
+            pass
+
+    if "exc_info" in hint:
+        exc_type, exc_value, tb = hint["exc_info"]
+        if isinstance(exc_value, Exception):
```
ncloudioj, line 49: Is this still necessary? I guess `exc_value` will always be some kind of Exception, so this check won't do anything.

merino/config_sentry.py
```diff
@@ -22,9 +26,51 @@ def configure_sentry() -> None:  # pragma: no cover
...
+            try:
+                for entry in event["exception"]["values"][0]["stacktrace"]["frames"]:
+                    try:
```
ncloudioj, line 52: This try/catch can be dropped as well as the following shouldn't raise any KeyError.

merino/config_sentry.py
taddes: Good point @ncloudioj , maybe then we should just capture the few we're sure of to be specific opposed to general? Ideally, we'd not always want this to fire, depending on the log event.  Happy to remove, though I could change it to something like what you had before:
```python
        if isinstance(exc_value, RuntimeError, BackendError):
```

ncloudioj: I think we can remove this line since maintaining a list of exceptions can be an ordeal over time.

merino/config_sentry.py
```diff
@@ -22,9 +22,41 @@ def configure_sentry() -> None:  # pragma: no cover
...
+                        entry["vars"]["srequest"] = ""
+                    if entry["vars"].get("values", {}).get("q"):
+                        entry["vars"]["values"]["q"] = ""
+                    if (
+                        entry["vars"].get("solved_result", [])
+                        and len(entry["vars"].get("solved_result", [])) > 0
+                    ):
+                        if entry["vars"]["solved_result"][0].get("q"):
+                            entry["vars"]["solved_result"][0]["q"] = ""
+                except KeyError:
```
ncloudioj, line 59: Sorry for not being clear. I think we should keep the outer try/except for `for entry in event["exception"]["values"][0]["stacktrace"]["frames"]:` and drop this inner try/except because all those if statements are infallible.

taddes: Ok makes sense, sorry for my confusion on that!!

merino/config_sentry.py
taddes: Great point! 👍

merino/config_sentry.py
taddes: I added it back and emit a log event for which we can determine the key that raised the issue, should the KeyError occur, and then we `pass` as not to crash the whole Sentry operation. I wrote a test to capture the log for this as well.

taddes:
Thanks again @ncloudioj ! I removed that check for `"exc_info`", however I left the tests intact that check it, since the hint parameter must be required for the invocation of the callback. When testing without, it broke in debugger. Agreed, we don't need that check anymore, but I wanted to conform to the boilerplate definition. Thx for the review.

merino/config_sentry.py
```diff
@@ -22,9 +25,40 @@ def configure_sentry() -> None:  # pragma: no cover
...
+            if entry["vars"].get("values", {}).get("q"):
+                entry["vars"]["values"]["q"] = ""
+            if (
+                entry["vars"].get("solved_result", [])
+                and len(entry["vars"].get("solved_result", [])) > 0
+            ):
+                if entry["vars"]["solved_result"][0].get("q"):
+                    entry["vars"]["solved_result"][0]["q"] = ""
+
+    except KeyError as e:
```
ncloudioj, line 61: Let's make this to `except (KeyError, Index) as e` for `event[..][..][0]`

ncloudioj (approved):

taddes:
Closes #261

PR #266, opened 2023-04-20
Add ADR for Merino Suggest API Response

mtbottle:
## Description

Opening up this ADR for comments on content (decision has not been made yet). 

This ADR proposes a new format for the Suggest API Endpoint to manage provider specific data.

## PR Review Checklist

### DIFF
--- a/docs/adr/0002-merino-general-response.md
+++ b/docs/adr/0002-merino-general-response.md
@@ -1,7 +1,8 @@
 # Merino Suggest API Response Structure
 
-* Status: proposed
-* Deciders: [list everyone involved in the decision] <!-- optional -->
+* Status: accepted
+* Deciders: Michelle Tran, Lina Butler, Nan Jiang, Wil Stuckey,
+  Drew Willcoxon, Taddes Korris, Tiffany Tran
 * Date: 2023-04-20
 
 ## Context and Problem Statement
@@ -42,15 +43,49 @@ In rough order of importance:
 
 ## Decision Outcome
 
-Chosen option: ???
+Chosen option: B
 
-### Positive Consequences <!-- optional -->
+We will also *not* increase the version number of the API for this ADR.
+So, going forward, we will encode option B into the response design
+without changing the existing providers.
+This means that the following providers will *not* have
+their bespoke fields removed from top level:
+* AdM Provider
+* Top Picks Provider
+* Weather Provider
+* Wikipedia Provider
+* WikiFruit Provider
 
-* ???
+However, this does not preclude these providers from duplicating the fields
+to `custom_details` in the v1 API.
 
-### Negative Consequences <!-- optional -->
+### Positive Consequences of Option B
 
-* ???
+* Clear isolation of fields that belong together (i.e. grouped by provider).
+* Clear ownership of fields through the structure.
+* Simpler validation logic than other options due to less need for conditionals.
+
+### Negative Consequences of Option B
+
+* Potentially some redundancy caused by extra nesting.
+* Might not be as flexible with a provider that returns different fields based on
+  what type of suggestion it is.
+
+### Positive Consequences of not Increasing API Version
+
+* We do not have to worry about migrating Firefox (and other clients) into the new format.
+  The migration is going to be quite a lot of extra work that adds little benefits
+  (other than consistency of design, it doesn't add more features nor improve
+  any known time sinks with development).
+* Do not have to support 2 versions of the API.
+
+### Negative Consequences of not Increasing API Version
+
+* Some inconsistencies with how providers add fields to the response.
+  We will likely want to resolve this as we migrate to v2,
+  but it's a known issue at the moment.
+* Might be missing an opportune time to migrate, as features are currently not out yet
+  which means the flexibility for change is higher.
 
 ## Pros and Cons of the Options 
 
--- a/docs/adr/0002-merino-general-response.md
+++ b/docs/adr/0002-merino-general-response.md
@@ -1,0 +1,275 @@
+# Merino Suggest API Response Structure
+
+* Status: proposed
+* Deciders: [list everyone involved in the decision] <!-- optional -->
+* Date: 2023-04-20
+
+## Context and Problem Statement
+
+As Merino continues to add more suggestions, 
+suggestion providers are going to have to return all sorts of
+data to the clients that are bespoke to the particular suggestion.
+For instance, weather suggestion returns a `temperature`.
+Currently, we do not have a strategy to manage these bespoke pieces of data
+which results in them returned at the top level of the suggestion object.
+However, this will pose a problem when
+
+1. names of fields are shared between providers, but have different semantics 
+  (i.e. `rating` may be a decimal value between 0-1 in one type, 
+   and a "star" integer rating between 1-5 in another)
+1. the API is unclear about what will _necessarily_ exist, and what is _optional_,
+   which leads to client confusion about the contract
+
+So, this ADR is to make a decision on how we want to handle provider specific fields
+going forward.
+
+## Decision Drivers
+
+In rough order of importance:
+
+1. Explicitness of Ownership - i.e. the `rating` field belongs to the `addons` provider
+1. Compatibility with [JSON] Schema Validation
+1. Adherence to the Fx Suggest Design Framework
+1. Backwards Compatibility with Current Schema
+
+## Considered Options
+
+* A. Continue to add to Top Level with Optional Fields
+* B. Custom Details Field for Bespoke Provider Fields
+* C. Custom Details Field for a "Type"
+* D. Component Driven `custom_details`
+
+## Decision Outcome
+
+Chosen option: ???
+
+### Positive Consequences <!-- optional -->
+
+* ???
+
+### Negative Consequences <!-- optional -->
+
+* ???
+
+## Pros and Cons of the Options <!-- optional -->
+
+### A. Continue to add to Top Level with Optional Fields
+
+This is the status quo option.
+We will continue to append bespoke values to the top level suggestion,
+and ensure that they're optional.
+Resolving type differences will just require us to be more specific
+with the fieldnames, as we continue to grow.
+
+Example, to differentiate Addons rating and Pocket Collections rating,
+we will call one `addons_rating` and the other `pocket_collection_rating`.
+This will look like:
+
+```json
+{
+  "suggestions": [
+    {
+      ...
+      "provider": "addons",
+      "addons_rating": "4.123",
+      ...
+    },
+    {
+      ...
+      "provider": "pocket_collections",
+      "pocket_collection_rating": 0.123,
+      ...
+    },
+    ...
+  ],
+  ...
+}
+```
+
+#### Pros
+
+* Merino is still kind of immature, so it still might be too early to think about design.
+* Less nesting in the models (resulting in less complexity).
+* Currently, backwards compatible as we don't have to do anything 
+  to existing providers, as this follows the existing patterns.
+
+#### Cons
+
+* Can't effectively do schema validation because 
+  we don't truly know what is or isn't required. i.e. `rating` might not be _optional_ for `addons` type suggestions,
+  but it has to remain _optional_ for the schema to work for other providers.
+* Not clear what is shared between _all_ suggestions, vs. what is bespoke to specific provider.
+* Lack of isolation for bespoke fields; poorly named fields will cause chaos across multiple providers
+  i.e. using `rating` as a field for `addons` suggestions, then needing a `movie_rating` field for `movie` suggestions
+ then wonder why the client is confused about why they are missing a `rating` for their `movie` suggestion.
+
+### B. Custom Details Field for Bespoke Provider Fields
+
+We introduce a `custom_details` field that uses a provider name as key
+to an object with the bespoke values to that provider.
+
+Example:
+
+```json
+{
+  "suggestions": [
+    {
+      ...
+      "provider": "addons",
+      "custom_details": {
+        "addons": {
+          "rating": "4.7459"
+        }
+      }
+    },
+    ...
+  ],
+  ...
+}
+```
+
+The specific fields in `custom_details` will all be optional (i.e. `addons` will be an optional key)
+but the _shape_ of what goes in `addons` can be more strict (i.e. `addons` require a `rating` field).
+
+A partial schema specification for the above might look like[^jsonschema]:
+
+```json
+{
+  "$schema": "https://json-schema.org/draft/2020-12/schema",
+  "title": "Suggest API Response v1",
+  "description": "Response for /api/v1/suggest",
+  "type": "object",
+  "properties": {
+    "provider": {
+      "description": "id for the provider type",
+      "type": "string"
+    },
+    "custom_details": {
+      "type": "object",
+      "properties": {
+        "addons": {
+          "type": "object",
+          "description": "Custom Addon Fields",
+          "properties": {
+            "rating": {
+              "type": "number"
+            }
+          },
+          "required": ["rating"]
+        }
+      }
+    }
+  },
+  "required": ["provider"]
+}
+```
+
+[^jsonschema]: Can play with JSON schema in https://www.jsonschemavalidator.net/
+
+#### Pros
+
+* Can specify specific validation per provider.
+* Clear ownership of `rating` to `addons` via structure.
+* Fields outside of `custom_details` can be fields that are more universal across suggestions.
+  These fields can potentially be correlated directly to the Fx Suggest Design Framework
+  (i.e. `context_label`, `url`, `title`, `description`, etc.).
+* Having a clear distinction for Fx Suggest Design Framework fields vs. bespoke fields makes this more
+  backwards compatible, as the fields in the Design Framework can render the _default_ suggestion case
+  for clients who haven't upgraded their clients.
+
+#### Cons
+
+* No guarantee that `provider` field will match the `custom_detail` object.
+* We'll likely need to migrate existing providers at some point. But in the meantime,
+  some fields will not follow convention to maintain backwards compatibility.
+
+### C. Custom Details Field for a "Type"
+
+This is similar to option B, except that we want to introduce a new `type` field
+to differentiate it from the provider.
+The `custom_details` will be keyed by this type, rather than the `provider` name.
+These `types` are kind of analogous to a _rendering component_,
+as they will likely be used to specify a specific rendering path in the client.
+
+Example:
+
+```json
+{
+  "suggestions": [
+    {
+      ...
+      "provider": "addons",
+      "type": "addons_type",
+      "custom_details": {
+        "addons_type": {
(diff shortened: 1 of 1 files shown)


0c0w3:
Thanks @mtbottle. We face a similar problem on desktop Firefox (and I presume mobile does as well), where we have various types of urlbar results from different providers that ultimately must be collected and shown together in the UI: history, bookmarks, search suggestions, etc. Roughly, result objects look like this:

```js
{
  // The name of the provider that created this result
  providerName: "SearchSuggestions",

  // The type of result [1]: URL, search, tab, etc.
  type: SEARCH,

  // Where the result comes from [2]: network, user profile, bookmarks, etc.
  source: NETWORK,

  title: "chocolate chip cookies",

  // Properties specific to the result's `type` [3]
  payload: {
    suggestion: "chocolate chip cookies",
    engine: "Google",
  },
}
```

[1] [Source](https://searchfox.org/mozilla-central/rev/26790fecfcda622dab234b28859da721b80f3a35/browser/components/urlbar/UrlbarUtils.sys.mjs#84-106)
[2] [Source](https://searchfox.org/mozilla-central/rev/26790fecfcda622dab234b28859da721b80f3a35/browser/components/urlbar/UrlbarUtils.sys.mjs#108-123)
[3] [Source](https://searchfox.org/mozilla-central/rev/26790fecfcda622dab234b28859da721b80f3a35/browser/components/urlbar/UrlbarUtils.sys.mjs#1458-1852)

It's similar to your option C. We use both `providerName` and `type` throughout urlbar when deciding what UI to create for a result, how it responds to clicks, etc. We also use a third technique, which is to use common payload property names across result types, a big example being `url`. If a result has a `payload.url`, it's typically the URL that should load when the user clicks the result. Payload properties are typically in service of the UI -- they include info that ultimately needs to be shown or loaded -- but we try to keep them loosely coupled, and it's not too hard.

Other urlbar engineers might disagree, but I've found over time that `type` becomes less important and `providerName` becomes more important. As we add more highly specific kinds of suggestions, it's usually not the case that multiple providers return the same type. Types become pretty tightly coupled to their providers. In fact we have a newer special `type` called the dynamic result type that circumvents the `type` concept entirely. These results are very tightly coupled to their providers and their payloads are schema-less. (I'm biased here since these were my idea.)

One more thing: I've been refactoring the client code recently in preparation for newer Suggest types (addons, Pocket, etc.), and I've also been helping Wil with exposure telemetry. In both cases, it's been convenient to treat a suggestion's Merino provider name as its type, essentially. For example, we're going to use the Merino provider name directly in telemetry to identify some Merino suggestions. We don't have to do it that way but it makes sense right now. I'm also working on [a patch](https://phabricator.services.mozilla.com/D176111) that makes it easy to add a urlbar provider that corresponds directly to a Merino provider, based on the Merino provider name. Here too the coupling doesn't have to be based on the provider name, it could be based on a formalized "type," but using the provider makes sense right now.

docs/adr/0002-merino-general-response.md
```diff
@@ -0,0 +1,275 @@
...
+
+* Merino is still kind of immature, so it still might be too early to think about design.
+* Less nesting in the models (resulting in less complexity).
+* Currently, backwards compatible as we don't have to do anything 
+  to existing providers, as this follows the existing patterns.
+
+#### Cons
+
+* Can't effectively do schema validation because 
+  we don't truly know what is or isn't required. i.e. `rating` might not be _optional_ for `addons` type suggestions,
```
linabutler, line 99: I think we can do validation, and express the current response schema in a type-safe way, if we treat the `provider` as the discriminant.

The prototype Merino Rust client that @tiftran and I hacked on does this: because it's [deserializing each Merino suggestion into a statically-typed enum](https://github.com/mozilla/application-services/blob/8d1c699c81882d9931e649667a86bc9b108aebcc/components/merino_client/src/client.rs#L271-L299), it needs to know what fields to expect for each suggestion.

If we use the `provider` as the discriminant:

* Clients can expect to find all `BaseSuggestion` model fields on every suggestion.
* For suggestions where `"provider": "accuweather"`, the client can additionally expect `current_conditions` and `forecast` object fields—because these are required by the `WeatherSuggestion` model.
* For suggestions where `"provider": "adm"`, it can expect an integer `block_id`, a string `full_keyword`, etc., because these are all required by the `SponsoredSuggestion` and `NonsponsoredSuggestion` models.

Extending that approach to your example:

* The client will only use the `rating` field for suggestions where `"provider": "addons"`.
* If we ever introduce another kind of suggestions, like `"movie"`, with a different `rating` field, that's still okay—because a `"provider": "addons"` suggestion is different than a `"provider": "movie"` suggestion, so they can have fields with the same name but different meanings.

docs/adr/0002-merino-general-response.md
```diff
@@ -0,0 +1,275 @@
...
+  but it has to remain _optional_ for the schema to work for other providers.
+* Not clear what is shared between _all_ suggestions, vs. what is bespoke to specific provider.
+* Lack of isolation for bespoke fields; poorly named fields will cause chaos across multiple providers
+  i.e. using `rating` as a field for `addons` suggestions, then needing a `movie_rating` field for `movie` suggestions
+ then wonder why the client is confused about why they are missing a `rating` for their `movie` suggestion.
```
linabutler, line 104: If we use the `provider` as the discriminant, I think we still get isolation for bespoke fields. Sticking with the example above, we could have something like this in Merino:

```py
class AddonSuggestion(BaseSuggestion)
    rating: string

class MovieSuggestion(BaseSuggestion)
    rating: int
```

Both of those models define a `rating` field—with different types, even—but there's no conflict, because they're two different suggestion models returned from two different providers. The `provider` is how we know we're looking at an add-on rating or a movie rating.

docs/adr/0002-merino-general-response.md
```diff
@@ -0,0 +1,275 @@
...
+```
+
+#### Pros
+
+* All the pros for B applies here
+* Can decouple the `custom_details` from `provider`. This will be helpful for potentially 
+  sharing the `type` with other suggestions produced by different providers. For instance,
+  we may want this to specify different _rendering_ paths in the client 
+  (i.e. a "top picks" type to be shared between `addons` and `top_picks` providers,
+  as there's many shared fields because they're rendered similarly).
```
linabutler, line 222: This is a great point! I agree with your second con below, though: the `provider` feels like a good-enough discriminant that we have today, and adding `type` when `provider` can serve the same purpose for our use cases feels a little premature.

docs/adr/0002-merino-general-response.md
linabutler: [Here's this approach expressed in a partial JSON Schema](https://jsonschema.dev/s/n12xP), using an ["if-then-else" conditional](https://json-schema.org/understanding-json-schema/reference/conditionals.html#if-then-else). It has different definitions of `rating` for `"provider": "addons"` and `"provider": "movie"`, and different sets of required fields for each provider, in addition to all the required base fields.

linabutler (commented):
Thanks for writing this, @mtbottle!

I'm still mulling this over, and in particular the advantages of using `custom_details` over checking the `provider` to distinguish between different suggestions (and validating fields accordingly), but I wanted to get some early comments up!

docs/adr/0002-merino-general-response.md
mtbottle: Oh nice! I didn't know you can do such complex matching with JSON schema! Ya I will update Option A to use this example. The con also will be updated, as it's possible to do this type of schema validation.

ncloudioj (commented):
@mtbottle Thanks for kicking this off with great insights! I've seen a lot of great ideas here and also added my thoughts to rally the discussion :)

docs/adr/0002-merino-general-response.md
mtbottle: Hmm... I think from a client perspective, there is a dependency of the `provider` to the `rating` that isn't super clear via the structure of the response. I feel like the dependency on the `provider` field is non-obvious unless you look at the validation rules or the Pydantic objects used to create the JSON. So in that aspect, I feel like there is not good isolation of these fields, and incorrect assumptions on the API will make it easy to use it incorrectly.

docs/adr/0002-merino-general-response.md
```diff
@@ -0,0 +1,275 @@
...
+  as there's many shared fields because they're rendered similarly). 
+
+#### Cons
+
+* All the cons for B applies here
+* Potentially over-engineering for `type`, as it's use is currently hypothetical.
+
+### D. Component Driven `custom_details`
+
+This solution will model distinct UI components in the `custom_details` section.
```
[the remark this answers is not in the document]
linabutler: @ncloudioj Hehe, yes, exactly! You might be interested in [this adjacently-tagged definition](https://github.com/mozilla/application-services/blob/8d1c699c81882d9931e649667a86bc9b108aebcc/components/merino_client/src/client.rs#L271-L299) of the current Merino suggestion format! 😁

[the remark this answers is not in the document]
ncloudioj: Nice, tagged enums in action 🙌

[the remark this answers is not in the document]
quiiver: > Now we still need to keep the rest of the fields at the top level (i.e. BaseSuggestion) for backward compatibility.
Possibly orthogonal to this discussion but what about defining a v2 api endpoint to get around backwards compat issues?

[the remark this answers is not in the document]
ncloudioj: I like that idea. Let's build a clean slate for v2. @mtbottle @linabutler @0c0w3 What do you think about dropping the backwards compatibility from the requirement list?

docs/adr/0002-merino-general-response.md
taddes: Great suggestions @linabutler ! Yes, the JSON schema I just had to make for Shepherd, for [example here](https://github.com/mozilla-services/consvc-shepherd/blob/main/schema/allocation.schema.json), showed me the additional abilities to run regex patterns against values, so enforcement of specific fields would be possible in this option as @linabutler pointed out. 

Additionally, since FastAPI does such a good job of handling our JSON serialization/de-serialization, its verification of models in terms of validation still provides us with pretty robust checking, though of course the pointed out issues with values returned at the top level. With Pydantic baked into FastAPI, it is possible that we also implement more specific validators, beyond the simple ones that verify types and certain constraints. Enforcing a lot more scoped validation at the provider level could provide us with more of what we need, see docs [here](https://fastapi.tiangolo.com/tutorial/query-params-str-validations/).

docs/adr/0002-merino-general-response.md
taddes: I do like that you point out for this option @mtbottle that we can be choosy about field names to avoid confusion and enforce this though good docs and validations.  I like what @linabutler has to say here about `provider` being the discriminant, given the `provider` context in the returned response.

docs/adr/0002-merino-general-response.md
```diff
@@ -0,0 +1,275 @@
...
+    }
+  },
+  "required": ["provider"]
+}
+```
+
+[^jsonschema]: Can play with JSON schema in https://www.jsonschemavalidator.net/
+
+#### Pros
+
```
taddes, line 170: Yeah, I think this option has a lot of merit @mtbottle . Given the schema spec and scoped values for each provider, this gives enough separation from the parent response being returned at every instance and gives us a lot of options in tailoring logic to handle provider-specific fields. I guess the issues is maintaining the custom fields for each provider, I can see how that could get tricky in having to customize `custom_detail` for each provider, maintain the logic around validation for each provider, etc.

docs/adr/0002-merino-general-response.md
```diff
@@ -0,0 +1,275 @@
...
+* Can decouple the `custom_details` from `provider`. This will be helpful for potentially 
+  sharing the `type` with other suggestions produced by different providers. For instance,
+  we may want this to specify different _rendering_ paths in the client 
+  (i.e. a "top picks" type to be shared between `addons` and `top_picks` providers,
+  as there's many shared fields because they're rendered similarly). 
+
+#### Cons
+
+* All the cons for B applies here
+* Potentially over-engineering for `type`, as it's use is currently hypothetical.
```
taddes, line 227: I agree with this con for sure in this potentially being over-engineered, though its merits are certainly evident.

docs/adr/0002-merino-general-response.md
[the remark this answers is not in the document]
taddes: Yeah, like the idea of a v2 api in whatever decision we make in the end. I do like structure of option D here and provides a lot of ancillary benefits over the other options. I guess my only question would be how much do we want to tailor the response specifically to the design framework, versus considering how the design framework ingests the API data? The point you raise @mtbottle on tight coupling between the UI and API I think warrants a bit of exploration I'd be curious about.

Trinaa:
Pulling the latest should fix that flaky test blocking this PR 😬😢

docs/adr/0002-merino-general-response.md
Trinaa: +1

docs/adr/0002-merino-general-response.md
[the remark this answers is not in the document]
Trinaa: I am also partial to option `B.5 Custom Details Field without the Provider Nesting`.

Something to consider if v2 is on the table is that it will have a significant impact on the tests, both functional and performance. Worst case we could see a double-up, like in the contract tests, which would also possibly mean looking at optimization in the CI so it doesn't take twice as long 😵‍💫

docs/adr/0002-merino-general-response.md
linabutler: > I feel like the dependency on the `provider` field is non-obvious unless you look at the validation rules or the Pydantic objects used to create the JSON.

I'm not sure that this dependency is avoidable for any of the other options, either, though. Even with a `custom_details` object, the dependency on the `provider` (Options B, B.5) or `type` (Option C) is still there, because that tells us which custom details to look for, right? In other words, you still need to know what type a suggestion is in order to understand its custom fields.

docs/adr/0002-merino-general-response.md
```diff
@@ -0,0 +1,416 @@
...
+            "rating": {
+              "type": "number"
+            }
+          },
+          "required": ["rating"]
+        }
+      }
+    }
+  },
+  "required": ["provider"]
```
linabutler, line 219: This schema will validate for a suggestion like this:

```json
{
  "provider": "addons",
  "custom_details": {
    "addons": {
      "rating": 4
    },
    "movies": {
      "rating": "4.5"
    }
  }
}
```

Is that intentional, or do we expect that `custom_details.addons` will only be present if the `provider` is `addons`, `custom_details.movies` only if `provider` is `movies`, etc.?

docs/adr/0002-merino-general-response.md
mtbottle: Ah, actually B and C encode the dependency in the structure, so there's not an implicit dependency at all on `provider` or `type`. Those fields are more for convenience in these scenarios. For example, if we wanted to render `addons` without looking at the `provider` field, it is possible to just check `custom_details` for the `addon` object. 

This is partly why I also have a preference for option B over B.5, as B.5 removes this structural dependency that makes it clear how the fields relate to the provider without needing a non-intuitive reference.

docs/adr/0002-merino-general-response.md
```diff
@@ -0,0 +1,416 @@
...
+* Less nesting in the models (resulting in less complexity).
+* Currently, backwards compatible as we don't have to do anything 
+  to existing providers, as this follows the existing patterns.
+
+#### Cons
+
+* Lack of isolation for bespoke fields; `ratings` is coupled with 2 specific providers,
+  and by just looking at the response, it's not clear that they are related.
+* Not clear what is shared between _all_ suggestions, vs. what is bespoke to specific provider.
+* It is not obvious that the `provider` field should signal how you should perform validation.
```
linabutler, line 158: I think this is true of all the options?

* For B, you need to know the `provider` to know which sub-object in `custom_fields` to expect—unless we explicitly want to allow something like https://github.com/mozilla-services/merino-py/pull/266/files#r1179586540, which would be great to call out explicitly if so!
* For B.5, you need to know the `provider` to understand how to interpret `custom_fields`.
* For C, you need to know the `type` to know which sub-object in `custom_fields` to expect.
* For D, you need to know the `provider` to understand how to interpret some of the `custom_fields` (in your example, `ratings`).

docs/adr/0002-merino-general-response.md
mtbottle: Not intentional :p This schema validation was a short proof of concept that we can write some sort of validation. I can see if I can add some conditional checks on this if it would make it clearer.

docs/adr/0002-merino-general-response.md
linabutler: Oh, I see, the _presence_ of `custom_details.addons` is what tells you that it's an add-on suggestion. Thanks for clarifying that, I was missing this point!

This is actually my first time seeing a tagged union represented like that in the wild—cool! Serde calls it the ["externally tagged"](https://serde.rs/enum-representations.html#externally-tagged) representation.

docs/adr/0002-merino-general-response.md
linabutler: Oops, no, you clarified in https://github.com/mozilla-services/merino-py/pull/266#discussion_r1179588442 that it's not true for B. Thanks!

docs/adr/0002-merino-general-response.md
linabutler: Got it. I think I was also missing the very important point that `custom_details` in this proposal will only ever have one field, whose name is the type of the suggestion (`addons`), and the value is a sub-object (`{ "rating": 4 }`).

docs/adr/0002-merino-general-response.md
linabutler: > This is partly why I also have a preference for option B over B.5, as B.5 removes this structural dependency that makes it clear how the fields relate to the provider without needing a non-intuitive reference.

Yep, I see where you're going with this now, thank you! The discriminant (tag) is `provider` for both A and B.5, but, in A, the provider-specific fields are included in the top-level object, at the same level as `provider` and the other common fields; and in B, the provider-specific fields are in the `custom_details` object.

In Serde terms (hi @ncloudioj! 😁):

* A is internally tagged.
* B is externally tagged, using the `provider` as the tag / discriminant.
* B.5 is adjacently tagged.
* C is also externally tagged, using a new `type` field as the tag.
* D is...not a tagged union at all, I don't think! It's a struct ("product type"), where its fields are attributes that describe what the suggestion _has_ ("this suggestion has ratings and a highlight context label"), not what it _is_ ("this suggestion is an add-on").

linabutler (approved):
:shipit: Thank you for the great discussion, for folding in the suggestions, and for driving us to the decision!

ncloudioj (approved):
Thank you so much for driving this to the conclusion! I really enjoyed those discussions and insights from everyone!

PR #268, opened 2023-04-20
Fix the display for Dynamic Wikipedia

mtbottle:
## Description

This adds "Wikipedia -" to the title and return the matching keyword (rather than the whole title) so that it looks like the clientside version of Wikipedia suggestions.

This is how the version on the client looks like: 
Screenshot 2023-04-20 at 16 40 50

This is how this it looks with the changes:

Screenshot 2023-04-21 at 10 39 25

## PR Review Checklist

### DIFF
--- a/merino/providers/wikipedia/backends/elastic.py
+++ b/merino/providers/wikipedia/backends/elastic.py
@@ -19,6 +19,23 @@ class ElasticBackendError(BackendError):
     """Error with Elastic Backend"""
 
 
+def get_best_keyword(q: str, title: str):
+    """Try to get the best autocomplete keyword match from the title. If there are no matches,
+    then return the full title as a match. Lowercase everything.
+    """
+    title = title.lower()
+    q = q.strip().lower()
+    start_index = title.find(q)
+    if start_index < 0:
+        return title
+
+    end_index = title.find(" ", start_index + len(q) - 1)
+    if end_index < start_index:
+        return title[start_index:]
+
+    return title[start_index:end_index]
+
+
 class ElasticBackend:
     """The client that works with the Elasticsearch backend."""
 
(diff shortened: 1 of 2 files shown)


merino/providers/wikipedia/backends/elastic.py
```diff
@@ -19,6 +19,18 @@ class ElasticBackendError(BackendError):
...
 
 
+def get_best_keyword(q: str, title: str):
+    """Try to get the best autocomplete keyword match from the title. If there are no matches,
+    then return the full title as a match. Lowercase everything.
+    """
+    title_words = title.split()
+    for w in title_words:
+        if w.lower().startswith(q):
+            return w.lower()
```
quiiver, line 29: I think you want and accumulator of words here so that you return `taylor` for `tay` and `taylor switft` for `taylor sw`

mtbottle: I've changed the strategy to get the substring via matching (then finding the next "space" char). This does a more specific match.

merino/providers/wikipedia/backends/elastic.py
```diff
@@ -19,6 +19,23 @@ class ElasticBackendError(BackendError):
     """Error with Elastic Backend"""
 
 
+def get_best_keyword(q: str, title: str):
```
quiiver, line 22: do we always get a lowercase query string from the client?

quiiver (approved):
lgtm, thanks

merino/providers/wikipedia/backends/elastic.py
mtbottle: Oh good point! I think it's worth being defensive here (even if it is always lowercase). I'll quickly add this change, thanks!

ncloudioj: @mtbottle Sorry for the late comment. I think we should do this via `Provider.normalize_query()`, which will be automatically called for each provider [here](https://github.com/mozilla-services/merino-py/blob/main/merino/web/api_v1.py#L91).

quiiver: @ncloudioj query normalization should be for preparing queries not formatting for output

ncloudioj: Yup, I was referring to the handling of `q`, which was passed from the provider.

quiiver: oh duh i didn't realize what comment thread we were on 🤦🏻‍♂️

ncloudioj: No worries. It's a weird case, we don't actually need query normalization for this provider because ES will do that for us anyway, but we still need normalization here for output for each hit. So we can now let the query normalizer do this for us.

PR #273, opened 2023-04-25
Add the Addons Provider to Merino

mtbottle:
## References

JIRA: [DISCO-2357](https://mozilla-hub.atlassian.net/browse/DISCO-2357)

## Description
This adds the Addons Provider to Merino. It currently has the `is_top_pick` flag to render it in the "top picks" format. However, we'll likely want to revert that once the _real_ UX is complete.

Here is how a sample Addon Suggestion look in the browser: 

Screenshot 2023-04-24 at 19 24 01

## PR Review Checklist

[DISCO-2357]: https://mozilla-hub.atlassian.net/browse/DISCO-2357?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


merino/providers/addons/provider.py
```diff
@@ -0,0 +1,89 @@
...
+                url=addon.url,
+                score=self.score,
+                is_sponsored=True,
+                provider=self.name,
+                icon=addon.icon,
+                custom_details=CustomDetails(addons=AddonsDetails(rating=addon.rating)),
+            )
+        ]
+
+    def _reverse_and_expand_index_keywords(
```
linabutler, line 75: suggestion: How would you feel about making this a `@staticmethod`, or a free function?

merino/providers/addons/provider.py
```diff
@@ -0,0 +1,89 @@
...
+        q: str = srequest.query
+        if len(q) < self.min_chars:
+            return []
+
+        matched_addon = self.addon_keywords.get(q)
+
+        if matched_addon is None:
+            return []
+
+        addon: Addon = await self.backend.get_addon(matched_addon)
```
linabutler, line 60: suggestion: Catch and log `Addon(API)BackendException` here?

merino/providers/addons/backends/api.py
```diff
@@ -0,0 +1,51 @@
...
+from merino.providers.addons.backends.protocol import Addon
+
+
+class AddonAPIBackendException(BackendError):
+    """Addon API Exception"""
+
+    pass
+
+
+class AddonAPIBackend:
```
linabutler, line 15: suggestion: Maybe `DynamicBackend` would make a nicer name than `APIBackend`? Or `AMOBackend`?

merino/providers/custom_details.py
```diff
@@ -0,0 +1,14 @@
...
+from pydantic import BaseModel
+
+
+class AddonsDetails(BaseModel):
+    """Addon specific data to be used in Custom Details"""
+
+    rating: str
+
+
+class CustomDetails(BaseModel):
```
linabutler, line 11: Oooh, nice use of your custom details proposal! 🚀

merino/providers/base.py
```diff
@@ -24,7 +25,9 @@ class BaseSuggestion(BaseModel):
     provider: str
     is_sponsored: bool
     score: float
+    description: str | None = None
```
linabutler, line 28: suggestion: Would this be good to put as a non-optional field in the `(Temporary)AddonSuggestion` model, instead of an optional field in the `BaseSuggestion` model?

tests/contract/volumes/client/scenarios.yml
```diff
@@ -32,6 +32,8 @@ scenarios:
                 is_sponsored: false
                 icon: "https://en.wikipedia.org/favicon.ico"
                 score: 0.0
+                custom_details: null
```
linabutler, line 35: suggestion: Let's set the `response_model_exclude_none=True` option (or at least `response_model_exclude_unset`; the [FastAPI docs](https://fastapi.tiangolo.com/tutorial/response-model/#use-the-response_model_exclude_unset-parameter) say it supports both of those options) on the `/suggest` endpoint, [here](https://github.com/mozilla-services/merino-py/blob/a70962823db7b62ae56067f500e5af8dbb212ee6/merino/web/api_v1.py#L43-L48), so that we're not including null fields in all our responses.

linabutler (commented):
This is tremendous work, @mtbottle, thank you so much for doing this! 🎉 🚀

I left some initial review comments, but they're nits and suggestions in the grand scheme of things; the overall shape of this look great!

Holding off on officially approving in case you had more changes planned, or wanted to get this up as a work-in-progress patch—but if you're aiming to get this landed directly, I'm happy to r+wc next time!

linabutler:
> Here is how a sample Addon Suggestion look in the browser:
> 
> Screenshot 2023-04-24 at 19 24 01

Terrific! 🤩

merino/providers/addons/provider.py
```diff
@@ -0,0 +1,89 @@
...
+        min_chars=settings.providers.addons.min_chars,
+        score=settings.providers.addons.score,
+        **kwargs: Any
+    ):
+        """Initialize Addon Provider"""
+        self._name = name
+        self.score = score
+        self.backend = backend
+        self.min_chars = min_chars
+        self.addon_keywords = self._reverse_and_expand_index_keywords(keywords)
```
ncloudioj, line 41: Let's move this call to `initialize()`, which is made exactly for tasks like this.

merino/providers/addons/provider.py
ncloudioj: For the API backend, what do you think about pre-fetching all the addons from the AMO endpoint during initialization? We only have a small number of addons and I guess the addon data won't change that often. We can cache the results in memory so that Merino doesn't need to make network calls for every single keyword hit. If addon data get volatile in the future, we can spin a cron job to refresh them in the background just like what the `adm` provider does.

merino/providers/addons/provider.py
```diff
@@ -0,0 +1,89 @@
...
+
+        addon: Addon = await self.backend.get_addon(matched_addon)
+
+        return [
+            TemporaryAddonSuggestion(
+                title=addon.name,
+                description=addon.description,
+                url=addon.url,
+                score=self.score,
+                is_sponsored=True,
```
ncloudioj, line 68: We might want to confirm with Product regarding whether this belongs to sponsored or not.

merino/providers/addons/provider.py
```diff
@@ -0,0 +1,89 @@
...
+    def _reverse_and_expand_index_keywords(
+        self, keywords: dict[str, set[str]]
+    ) -> dict[str, str]:
+        """Reverse the keywords index.
+        param keywords: mapping of addon key -> keywords
+        returns: mapping of keyword -> addon key
+        """
+        reverse_index = {}
+        for (addon_name, kws) in keywords.items():
+            for word in list(kws):
```
ncloudioj, line 84:
```suggestion
            for word in kws:
```

merino/providers/addons/provider.py
```diff
@@ -0,0 +1,89 @@
...
+        ]
+
+    def _reverse_and_expand_index_keywords(
+        self, keywords: dict[str, set[str]]
+    ) -> dict[str, str]:
+        """Reverse the keywords index.
+        param keywords: mapping of addon key -> keywords
+        returns: mapping of keyword -> addon key
+        """
+        reverse_index = {}
```
ncloudioj, line 82: nit: I think we usually use "inverted index" for structures like this.

merino/providers/addons/addons_data.py
```diff
@@ -0,0 +1,168 @@
...
+    "privacy-badger17": {
+        "Privacy",
+        "Privacy Badger",
+        "Anti Tracking",
+        "Cybersecurity",
+        "Tracking",
+        "Trackers",
+        "Spyware",
+        "Malware",
+        "Ad tracking",
```
ncloudioj, line 145: Looks like "Ad tracking" is used by both this and `ublock-origin`, since the keyword index follows the "LIFO" semantics, shall we drop one to avoid confusion?

merino/providers/addons/addons_data.py
```diff
@@ -0,0 +1,168 @@
...
+        "Invisible trackers",
+        "Security",
+    },
+    "ublock-origin": {
+        "Adblock",
+        "ad block",
+        "ad blocker",
```
ncloudioj, line 152:
```suggestion
        "Ad block",
        "Ad blocker",
```

merino/providers/addons/backends/api.py
ncloudioj: Good point. Shall we also call this provider "amo" instead of `addon`?

merino/providers/addons/backends/api.py
```diff
@@ -0,0 +1,51 @@
...
+    async def get_addon(self, addon_key: str) -> Addon:
+        """Get Addons given the list of keys"""
+        static_info = STATIC_DATA[addon_key]
+        return self._get_dynamic_addon_data(static_info, addon_key)
+
+    def _get_dynamic_addon_data(
+        self, static_data: dict[str, str], addon_key: str
+    ) -> Addon:
+        """Get the Addon information via the Addons API."""
+        res = requests.get(f"{self.api_url}/{addon_key}")
```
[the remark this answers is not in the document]
ncloudioj: Nice catch, Lina. Yup, we prefer async clients over the blocking counterparts for the Merino web app since it runs as a single-threaded app, whereas it's fine to use the latter for Merino jobs.

tests/unit/providers/addons/backends/test_static.py
```diff
@@ -0,0 +1,42 @@
+"""Test StaticAddonsBackend."""
+import pytest
+
+from merino.providers.addons.addons_data import STATIC_DATA
```
ncloudioj, line 4: Can you add a test to enforce those addon keys are consistent across `STATIC_DATA`, `KEYWORDS`, and `STATIC_RATING_AND_ICONS`? Because we're calling `STATIC_DATA[addon_key]` and `STATIC_RATING_AND_ICONS[addon_key]` throughout the code base, any typos or missing keys would hit the `KeyError`s.

merino/configs/default.toml
```diff
@@ -110,6 +110,16 @@ resync_interval_sec = 10800
 score = 0.3
 score_wikipedia = 0.2
 
+[default.providers.addons]
```
ncloudioj, line 113: It's always a good idea to add config validations to `merino/config.py`. Also, let's add some docs to `docs/operations/configs.md` as it's still referenced by some folks.

ncloudioj (changes requested):
Great stuff 🙌! Thanks for adding this new provider to Merino!

There are some review comments for further tweaking, otherwise, it looks great.

merino/providers/addons/provider.py
```diff
@@ -0,0 +1,89 @@
...
+        for (addon_name, kws) in keywords.items():
```
ncloudioj, line 83:
```suggestion
        for addon_name, kws in keywords.items():
```

merino/providers/addons/addons_data.py
```diff
@@ -0,0 +1,168 @@
+"""Static Mapping for the Addon data, along with the associated keywords
```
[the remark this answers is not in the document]
mtbottle: Yes! I intentionally chose the Python module so we didn't have to read it! For small datasets that isn't shared with any other processing (i.e. aggregation job etc.), this is probably the most easily maintainable. Having it in code allows us to: 

1. Have format checking during runtime (ie. missing commas, brackets, or wrong indentation will not be an issue)
2. Be more efficient by avoiding extra IO processing steps
3. Skip the complexity of ensuring that we have the right filepaths to be used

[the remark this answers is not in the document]
quiiver: i'm into it 😎

merino/providers/addons/provider.py
mtbottle: > what do you think about pre-fetching all the addons from the AMO endpoint during initialization?

Ya! I think this might actually be a better idea than caching (especially since some calls take longer than 200ms). I'm going to try to initialize and store the Addons data in an object for now.

merino/providers/addons/backends/api.py
```diff
@@ -0,0 +1,51 @@
...
+    """
+
+    api_url: str
+
+    def __init__(self, api_url: str):
+        """Initialize Backend"""
+        self.api_url = api_url
+
+    async def get_addon(self, addon_key: str) -> Addon:
+        """Get Addons given the list of keys"""
```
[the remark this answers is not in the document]
mtbottle: Whoops! Thanks for the catch!

> But also, addon_key almost feels like it wants to be an enum or a [StrEnum](https://docs.python.org/3/library/enum.html#enum.StrEnum), with STATIC_DATA and KEYWORDS both using it as the key. WDYT?

This is a good idea! This will also ensure that the keys passed between provider and backend are valid.

merino/providers/addons/provider.py
mtbottle: I've confirmed with product that we don't want this to be sponsored. Will update this.

tests/unit/providers/addons/backends/test_static.py
mtbottle: I've switched to using an enum for the keys, so we shouldn't have KeyErrors anymore.

merino/providers/addons/backends/api.py
mtbottle: > Shall we also call this provider "amo" instead of addon?

I'm sticking with `addons` as I feel like it's clearer to me :p 

But I've renamed this backend `DynamicAddonsBackend`. It's actually _less_ dynamic right now (I load it once, and never update). I think we can use the next ticket to update the object in a cron so that it's _fresher_.

merino/providers/base.py
linabutler: Gotcha, thanks!

linabutler (approved):
r+wc, thanks @mtbottle! This is exciting! 🎊

merino/providers/addons/backends/dynamic.py
```diff
@@ -0,0 +1,65 @@
...
+        for addon_key in SupportedAddons:
+            async with AsyncClient() as client:
+                res = await client.get(
+                    f"{self.api_url}/{addon_key}", follow_redirects=True
+                )
+
+            if res.status_code != 200:
+                raise DynamicAddonsBackendException(
+                    f"Addons API could not find key: {addon_key}"
+                )
```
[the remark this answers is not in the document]
mtbottle: > If any error happens, this will terminate the entire initialization with a partial self.dynamic_data, do we want to fall back to the static data in that case?

No, I don't think we want to fall back on static, but have the partial data. I'm intending the cron job follow-up, so we can just leave this for now.

merino/configs/default.toml
```diff
@@ -110,6 +110,20 @@ resync_interval_sec = 10800
 score = 0.3
 score_wikipedia = 0.2
 
+[default.addons.dynamic]
+# This is the URL for the Addons API to get more information for particular addons
+api_url = "https://addons.mozilla.org/api/v5/addons/addon/"
+
+[default.providers.addons]
```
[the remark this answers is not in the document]
mtbottle: > we usually use the name of the suggestion source as the provider ID instead of the type of the suggestion

OK! I'll change it to `amo` instead of `addons`.

merino/providers/addons/backends/dynamic.py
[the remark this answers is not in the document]
ncloudioj: Sounds good, please disregard all my previous comments on fallback-on-static then.

merino/providers/addons/backends/dynamic.py
```diff
@@ -0,0 +1,65 @@
...
+
+    def __init__(self, api_url: str):
+        """Initialize Backend"""
+        self.api_url = api_url
+        self.dynamic_data = {}
+
+    async def initialize_addons(self) -> None:
+        """Initialize the dynamic Addon information via the Addons API."""
+        self.dynamic_data = {}  # ensure that it's empty
+        for addon_key in SupportedAddons:
```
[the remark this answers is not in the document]
mtbottle: > This might be a good place to use a TaskGroup, to issue all the requests to the API in parallel?

As mentioned in a different comment, I'm not going to parallelize this for now because there's not a lot of calls made and it's going to be done async.

tests/contract/volumes/client/scenarios.yml
```diff
@@ -65,6 +67,8 @@ scenarios:
...
```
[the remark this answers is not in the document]
mtbottle: It looks like `response_model_exclude_unset` needs to work with the `response_model` (i.e. `BaseSuggestion`) but what we're currently returning is a `JSONResponse` casting of the Suggestion. However, once I remove extra `JSONResponse` casting and return the Suggestion directly, it looks like the extra fields in the extended `BaseSuggestion` models (i.e. like `SponsoredSuggestions` specific fields) are not recognized. So the fix for this looks to be non-trivial, so maybe it's worth exploring later (I can create a ticket for this)? What do you think @linabutler ?  There's also a part of me that feels like returning `null` is _ok_ because it's part of the API contract and returning them (even though they're null) gives more explicitness to the contract. However, we'll probably want to _not_ show all of the `custom_details` fields... so we'll likely want to fix this at some point.

Currently I've removed the configuration, as it's not working as expected.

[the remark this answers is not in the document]
linabutler: > what we're currently returning is a `JSONResponse` casting of the Suggestion.

Oh, I'm so sorry I missed that! It looks like you can do `JSONResponse(content=jsonable_encoder(response, exclude_unset=True))`, but I'd be okay with doing this in a follow-up in case that's also messy!

> Currently I've removed the configuration, as it's not working as expected. 

Sounds good, thank you!

PR #290, opened 2023-05-12
[DISCO-2359] Refactor navigational suggestions tests

abhi-agg:
## References

JIRA: https://mozilla-hub.atlassian.net/browse/DISCO-2359
GitHub:

## Description
Refactored `test_domain_metadata_extractor.py` test:
1. Remove reliance on external network request
2. Re-categorized the test as unit
3. Assert the collection as opposed to a for-loop with individual asserts, for simplicity and better error messages
4. Use the ids field with pytest.mark.parametrize to communicate test cases at report level
5. Apply type annotations to facilitate the static type checker
6. `test_domain_metadata_extractor.py` was also calling the network for testing a scenario. Hence refactored that as well.

Credits to @Trinaa for providing an initial example to mocking using dependency injection

## PR Review Checklist


abhi-agg:
Checked the reason of `docs-build` failure:
```
error: failed to compile `mdbook v0.4.29`, intermediate artifacts can be found at `/tmp/cargo-installVMgOwc`

Caused by:
  package `mdbook v0.4.29` cannot be built because it requires rustc 1.65 or newer, while the currently active rustc version is 1.64.0
```

I guess we just need to use the 1.65.0 version of rust image.

ncloudioj (approved):
r=wc.

Looks good, thanks for fixing those flaky tests!

Q: there are various favicon types in [here](https://github.com/mozilla-services/merino-py/blob/main/merino/jobs/navigational_suggestions/domain_metadata_uploader.py#L113-L132), shall we drop some of those (e.g. `avif, gif, tiff`) to silence the coverage warnings? My understanding is that most of those types are not that commonly used in the wild.

ncloudioj:
This shows the icon type distribution for the inflight domain file. The last five account for 99%, I think we can narrow down that list here.

```sh
$ jq -r '.domains[].icon' dev/top_picks.json | grep https | awk '{print substr($1, length($1)-2)}' | sort | uniq -c

2 ebp => webp
1 gif
56 ico
35 oct
4 peg => jpeg
399 png
31 svg
```

abhi-agg:
Great. Will submit a PR for this.

PR #293, opened 2023-05-16
Improve URL for navigational suggestions

abhi-agg:
## References
~~Fixes https://bugzilla.mozilla.org/show_bug.cgi?id=1832555~~

JIRA:
GitHub:

## Description
 - Using scraping library with `allow_redirects=True`
 - Use final redirected base url returned by scraping library for top picks file
 - Skip domains for which the corresponding final redirected base url doesn't have domain itself as a substring
 - Skip domains that are unreachable
 - Updated unit test cases

~~Tested with `minecraft` and `facebook` domain to confirm that the final redirected base urls are being stored in top picks file.~~

## PR Review Checklist

### DIFF
--- a/merino/jobs/navigational_suggestions/domain_metadata_extractor.py
+++ b/merino/jobs/navigational_suggestions/domain_metadata_extractor.py
@@ -2,7 +2,7 @@
 import logging
 from io import BytesIO
 from typing import Any, Optional
-from urllib.parse import urljoin
+from urllib.parse import urljoin, urlparse
 
 import requests
 from PIL import Image
@@ -40,7 +40,9 @@ class Scraper:
     browser: RoboBrowser
 
     def __init__(self) -> None:
-        self.browser = RoboBrowser(user_agent=FIREFOX_UA, parser="html.parser")
+        self.browser = RoboBrowser(
+            user_agent=FIREFOX_UA, parser="html.parser", allow_redirects=True
+        )
 
     def scrape_favicon_data(self, url: str) -> FaviconData:
         """Scrape the favicon data from the given url.
@@ -73,16 +75,20 @@ class Scraper:
         )
         return default_favicon_url if response.status_code == 200 else None
 
-    def scrape_title(self, url: str) -> str:
-        """Scrape the title from the header of the given url.
+    def scrape_url_and_title(self, url: str) -> tuple[str, str]:
+        """Scrape the title from the header of the given url and return it along with the
+        final redirected url.
 
         Args:
             url: URL to open and scrape
         Returns:
-            str: Title from header of the given URL
+            tuple[str,str]: The final url and the title extracted from header of the given url
         """
         self.browser.open(url, timeout=TIMEOUT)
-        return str(self.browser.find("head").find("title").string)
+        return (
+            str(self.browser.url),
+            str(self.browser.find("head").find("title").string),
+        )
 
 
 class DomainMetadataExtractor:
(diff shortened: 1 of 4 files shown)


merino/jobs/navigational_suggestions/domain_metadata_extractor.py
```diff
@@ -73,16 +75,19 @@ def get_default_favicon(self, url: str) -> Optional[str]:
         )
         return default_favicon_url if response.status_code == 200 else None
 
-    def scrape_title(self, url: str) -> str:
+    def scrape_title(self, url: str) -> tuple[str, str]:
```
ncloudioj, line 78: Let's rename this method (and the doc) to reflect what it's really doing.

ncloudioj:
> Tested with minecraft and facebook domain to confirm that the final redirected urls are being stored in top picks file.

What do those URLs look like? One concern about using those redirected URLs is that they might look confusing and weird to Firefox users. Also some redirected URLs could be more volatile (or dynamically generated) than their pre-redirected counterparts.

As an aside, I don't see any big concern with the issue. Having multiple history entries sounds inevitable to me, if the user visits those URLs by typing in the URL bar, the issue should remain anyway.

quiiver:
> What do those URLs look like? One concern about using those redirected URLs is that they might look confusing and weird to Firefox users. Also some redirected URLs could be more volatile (or dynamically generated) than their pre-redirected counterparts.

A concern I have with using the full url returned is that it will include the `path` and `query` parts of the URL as well. I think what we want with redirects is to just get the final `scheme` and `host` parts.

abhi-agg:
> > Tested with minecraft and facebook domain to confirm that the final redirected urls are being stored in top picks file.
> 
> What do those URLs look like? One concern about using those redirected URLs is that they might look confusing and weird to Firefox users. Also some redirected URLs could be more volatile (or dynamically generated) than their pre-redirected counterparts.
> 

Hey @ncloudioj For minecraft, it is: `https://www.minecraft.net/en-us` and for facebook it is: `https://www.facebook.com/`. Without redirected URLs, these were `https://minecraft.net` and `https://facebook.com` respectively.

> As an aside, I don't see any big concern with the issue. Having multiple history entries sounds inevitable to me, if the user visits those URLs by typing in the URL bar, the issue should remain anyway.

You are right about this. If users visit these urls, they will see multiple histories.

> > What do those URLs look like? One concern about using those redirected URLs is that they might look confusing and weird to Firefox users. Also some redirected URLs could be more volatile (or dynamically generated) than their pre-redirected counterparts.
> 
> A concern I have with using the full url returned is that it will include the `path` and `query` parts of the URL as well. I think what we want with redirects is to just get the final `scheme` and `host` parts.

This is fixable. I can parse the `scheme` and `hostname` from the redirected url and use only these components to generate the final url for top picks.

Although, I see a point in what Nan is saying. This bug doesn't sound like a bug.

merino/jobs/navigational_suggestions/domain_metadata_extractor.py
abhi-agg: done.

merino/jobs/navigational_suggestions/domain_metadata_extractor.py
```diff
@@ -245,45 +251,49 @@ def get_favicons(
...
+                final_redirected_base_url, title = self._extract_url_and_title(url)
 
             # if no valid title is present then fallback to use the second level domain as title
             if title is None:
                 title = self._get_second_level_domain(domain_data)
                 title = title.capitalize()
 
-            logger.info(f"url {url} and title {title}")
-            result.append({"url": url, "title": title})
+            logger.info(f"url {final_redirected_base_url} and title {title}")
```
ncloudioj, line 295:
```suggestion
            logger.info(f"URL: {final_redirected_base_url}; Title: {title}")
```

ncloudioj (approved):
LGTM!

0c0w3:
Thanks for linking me to this PR @abhi-agg. I know it was my idea to follow redirects, but on second thought I have a few concerns:

The final URLs may be more volatile, as @ncloudioj pointed out. But if we have a process to refresh this data periodically, that's much less of a concern IMO.

If we discard the path and query parts of the final URL, as @quiiver suggested, then it's not the final URL anymore, so the user will still get redirected. That doesn't solve the problem we were trying to solve here, and IMO there's no point in doing any of this if we do that.

Typical users will have some browsing history built up. Firefox gives redirects a very low frecency, so it's likely the user will have other history that matches their search string that Firefox will show instead of low-frecency redirects.

For these reasons, I'm leaning toward wont-fixing the bug, personally. However, I don't think any of this is critical, so I'm fine with going either way.

abhi-agg:
> If we discard the path and query parts of the final URL, as @quiiver suggested, then it's not the final URL anymore, so the user will still get redirected. 

I agree on this one.

However, I just want to throw light on a different issue which will require the exact same change as done in this PR.

We would like to skip all the domains from the final top pick list for which the corresponding urls don't contain the domain itself. i.e. for a hypothetical domain `domain.com`, if the corresponding url comes out to be anything except `https://www.domain.com` then we want to skip it.

e.g. 4th most visited domain in the current list `amazonaws.com` can resolve to the url `https://aws.amazon.com`. If we don't exclude this domain from the final list then it would mean that `https://aws.amazon.com` is the 4th most visited website which is not true since our dataset shows that it is `s3.amazonaws.com` which is 4th most visited website and not `aws.amazon.com`.

The change required to accommodate this logic is same as done in this PR:
1. Get the url returned by scraping library
2. Extract the base url from it
3. Check if the base url contains the domain as a substring. If it does then include it in the final list else skip it.

Therefor, I will change the heading of this PR and we all agree that https://bugzilla.mozilla.org/show_bug.cgi?id=1832555 is a won't fix bug.

quiiver (approved):

PR #311, opened 2023-05-31
Return defaultly set values and make AMO case insensitive

mtbottle:
## Description
Some providers hard-code some return fields in the model itself because these fields are not changeable for the provider. So, it looks like the fields are not "set" because they are default. We don't want to exclude these default values from being returned.

Also, AMO provider needs to be case insensitive. 

## PR Review Checklist


ncloudioj (approved):
Looks good, thanks! Yup, that `exclude_unset` seems unnecessary here.

PR #320, opened 2023-06-09
Cache Accuweather Responses with Expiry TTL

mtbottle:
## References

JIRA: [DISCO-2418](https://mozilla-hub.atlassian.net/browse/DISCO-2418)

## Description
This adds some caching for the exact Accuweather request, rather than aggregating the response into a `WeatherReport` to put into the cache. This provides a few possible flexibilities:

1. We can be agnostic about the response type coming from Accuweather. So, this can open up the possibilities to also cache 4xx errors (currently not doing that in this PR), as retrying them in short successions will be a waste of API credit. (Redis mem graph [here](https://earthangel-b40313e5.influxcloud.net/explore?left=%7B%22datasource%22:%22Stackdriver%20-%20CloudOps%22,%22queries%22:%5B%7B%22datasource%22:%7B%22type%22:%22stackdriver%22,%22uid%22:%22o0Z_lCHMz%22%7D,%22refId%22:%22A%22,%22intervalMs%22:0,%22type%22:%22timeSeriesQuery%22,%22queryType%22:%22metrics%22,%22metricQuery%22:%7B%22aliasBy%22:%22Used%20Memory%22,%22alignmentPeriod%22:%22%2B1800s%22,%22crossSeriesReducer%22:%22REDUCE_MAX%22,%22editorMode%22:%22visual%22,%22filters%22:%5B%5D,%22groupBys%22:%5B%22resource.label.pod_name%22%5D,%22metricKind%22:%22GAUGE%22,%22metricType%22:%22redis.googleapis.com%2Fstats%2Fmemory%2Fusage%22,%22perSeriesAligner%22:%22ALIGN_MEAN%22,%22preprocessor%22:%22none%22,%22projectName%22:%22moz-fx-merino-metrics-d487%22,%22query%22:%22%22,%22valueType%22:%22INT64%22,%22view%22:%22FULL%22%7D%7D,%7B%22datasource%22:%7B%22type%22:%22stackdriver%22,%22uid%22:%22o0Z_lCHMz%22%7D,%22refId%22:%22B%22,%22intervalMs%22:0,%22type%22:%22timeSeriesQuery%22,%22queryType%22:%22metrics%22,%22metricQuery%22:%7B%22aliasBy%22:%22Maximum%20Memory%22,%22alignmentPeriod%22:%22cloud-monitoring-auto%22,%22crossSeriesReducer%22:%22REDUCE_MEAN%22,%22editorMode%22:%22visual%22,%22filters%22:%5B%5D,%22groupBys%22:%5B%5D,%22metricKind%22:%22GAUGE%22,%22metricType%22:%22redis.googleapis.com%2Fstats%2Fmemory%2Fmaxmemory%22,%22perSeriesAligner%22:%22ALIGN_MEAN%22,%22preprocessor%22:%22none%22,%22projectName%22:%22moz-fx-merino-metrics-d487%22,%22query%22:%22%22,%22valueType%22:%22INT64%22,%22view%22:%22FULL%22%7D%7D%5D,%22range%22:%7B%22from%22:%22now-2d%22,%22to%22:%22now%22%7D%7D&orgId=1))
2. We can treat each API differently based on our expected expiry for that data.

The potential area of concerns are:

1. We are storing a lot more data than we're using. However, checking the Redis disk usage, it _doesn't_ seem to be a concern. However, we should monitor this in case we use up too much disk space.
2. Data migration risks. 

To derisk some of these things, I've opted to open up this change to _add_ caching at the request level, while preserving the existing caching. The amount of space we have on our Redis instance makes me feel like this is an _ok_ strategy, as we're currently using about 12% of our maximum capacity (if my graph is correct). Once this is out and it looks like we're populating the cache properly, then I would remove the old caching code.

I am also considering compressing the responses when storing it in Redis. However, I want to have it run without compression to see how much space it actually takes up first. Adding compression might also affect the latency, so it would be useful to have some sort of comparison data.

## PR Review Checklist

[DISCO-2418]: https://mozilla-hub.atlassian.net/browse/DISCO-2418?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


mtbottle:
I've tested this locally with the local Redis docker container and it stores and retrieves values from Redis.

merino/providers/manager.py
```diff
@@ -53,6 +53,13 @@ def _create_provider(provider_id: str, setting: Settings) -> BaseProvider:
             return WeatherProvider(
                 backend=AccuweatherBackend(
                     api_key=settings.accuweather.api_key,
+                    cache=RedisAdapter(
```
linabutler, line 56: suggestion: Can we create one `RedisAdapter` instance, above line 53, then pass it here and on line 76 below?

merino/providers/weather/backends/accuweather.py
```diff
@@ -92,6 +110,87 @@ def cache_inputs_for_weather_report(self, geolocation: Location) -> Optional[byt
...
 
+    def cache_key_for_accuweather_request(
+        self, url: str, query_params: dict[str, str] = {}
+    ) -> str:
+        """Get the cache key for the accuweather request.
+        Also ensure that the API key is stripped out.
+        """
+        params = query_params.copy()
+        if "apikey" in params:
+            del params["apikey"]
```
linabutler, line 121:
```suggestion
        params.pop(settings.accuweather.url_param_api_key, None)
```

nitpick: If it makes sense, let's use `settings.accuweather.url_param_api_key` here, instead of hard-coding `"apikey"`?

merino/providers/weather/backends/accuweather.py
```diff
@@ -92,6 +110,87 @@ def cache_inputs_for_weather_report(self, geolocation: Location) -> Optional[byt
...
+
+        if params:
+            extra_identifiers = hashlib.blake2s(str(params).encode("utf-8")).hexdigest()
```
linabutler, line 124:
```suggestion
            hasher = hashlib.blake2s()
            for key, value in sorted(params.items()):
                hasher.update(key.encode("utf-8") + value.encode("utf-8"))
            extra_identifiers = hasher.hexdigest()
```

suggestion: I'm not sure if the dictionary string representation—or dictionary enumeration order—is guaranteed to be stable, so we might end up with spurious cache misses for the same set of `params`, depending on how exactly the dictionary is populated.

It's a bit more code, but for cases like this, I like to explicitly canonicalize the representation.

linabutler (changes requested):
This looks lovely; I really like your approach of pushing the caching down into the backend, and totally agree that it gives us more flexibility, compared to caching the entire report at the provider level!

I'll do another pass once the jet lag wears off, but here's a first round of feedback 😄

merino/providers/weather/backends/accuweather.py
```diff
@@ -92,6 +110,87 @@ def cache_inputs_for_weather_report(self, geolocation: Location) -> Optional[byt
...
+        self, cache_key: str, response_dict: dict[str, Any], response_expiry: str
+    ):
+        """Store the request into cache. Also ensures that the cache ttl is
+        at least `cached_report_ttl_sec`.
+        """
+        with self.metrics_client.timeit("accuweather.cache.store"):
+            expiry_delta: timedelta = (
+                datetime.datetime.strptime(
+                    response_expiry, ACCUWEATHER_CACHE_EXPIRY_DATE_FORMAT
+                )
```
linabutler, line 172: suggestion (non-blocking): The HTTP date format is a [variation of RFC 2822](https://datatracker.ietf.org/doc/html/rfc7231#section-7.1.1.1), so I _think_ using [`email.utils.parsedate_to_datetime`](https://python.readthedocs.io/en/latest/library/email.util.html#email.utils.parsedate_to_datetime) here instead of `strptime` will also work!

tests/unit/providers/weather/backends/test_accuweather.py
```diff
@@ -40,20 +65,37 @@ def fixture_accuweather_parameters() -> dict[str, str]:
...
 @pytest.fixture(name="accuweather_with_partner_code")
 def fixture_accuweather_with_partner_code(
-    accuweather_parameters: dict[str, str]
+    redis_mock_cache_miss: AsyncMock, accuweather_parameters: dict[str, Any]
 ) -> AccuweatherBackend:
     """Create an Accuweather object with a partner code for test."""
     return AccuweatherBackend(
         url_param_partner_code="partner",
         partner_code="acme",
+        cache=redis_mock_cache_miss,
```
linabutler, line 98:
```suggestion
        cache=RedisAdapter(redis_mock_cache_miss),
```

nitpick: The test will still work, but I think this will match how the "real" code does it—IIUC, `cache` is a `CacheAdapter` (`from merino.cache.redis import RedisAdapter`), not the Redis client directly.

tests/unit/providers/weather/backends/test_accuweather.py
```diff
@@ -251,13 +293,15 @@ def fixture_accuweather_forecast_response_fahrenheit(
...
+def test_init_api_key_value_error(
+    mocker: MockerFixture, accuweather_parameters: dict[str, Any]
+) -> None:
     """Test that a ValueError is raised if initializing with an empty API key."""
     expected_error_value: str = "AccuWeather API key not specified"
     accuweather_parameters["api_key"] = ""
 
     with pytest.raises(ValueError) as accuweather_error:
-        AccuweatherBackend(**accuweather_parameters)
+        AccuweatherBackend(cache=mocker.AsyncMock(spec=Redis), **accuweather_parameters)
```
linabutler, line 304:
```suggestion
        AccuweatherBackend(cache=RedisAdapter(mocker.AsyncMock(spec=Redis)), **accuweather_parameters)
```

tests/unit/providers/weather/backends/test_accuweather.py
```diff
@@ -40,20 +65,37 @@ def fixture_accuweather_parameters() -> dict[str, str]:
...
-    """Create an Accuweather object for test."""
-    return AccuweatherBackend(**accuweather_parameters)
+def fixture_accuweather(
+    redis_mock_cache_miss: AsyncMock,
+    accuweather_parameters: dict[str, Any],
+    statsd_mock: Any,
+) -> AccuweatherBackend:
+    """Create an Accuweather object for test. This object always have cache miss."""
+    return AccuweatherBackend(
+        cache=redis_mock_cache_miss,
```
linabutler, line 85:
```suggestion
        cache=RedisAdapter(redis_mock_cache_miss),
```

tests/unit/providers/weather/backends/test_accuweather.py
```diff
@@ -283,7 +327,7 @@ def test_init_url_value_error(
     accuweather_parameters[url_value] = ""
 
     with pytest.raises(ValueError) as accuweather_error:
-        AccuweatherBackend(**accuweather_parameters)
+        AccuweatherBackend(cache=mocker.AsyncMock(spec=Redis), **accuweather_parameters)
```
linabutler, line 330:
```suggestion
        AccuweatherBackend(cache=RedisAdapter(mocker.AsyncMock(spec=Redis)), **accuweather_parameters)
```

tests/unit/providers/weather/backends/test_accuweather.py
```diff
@@ -661,9 +763,49 @@ async def test_get_current_conditions(
...
+    expected_conditions: CurrentConditions = CurrentConditions(
+        url=current_conditions_url,
+        summary="Mostly cloudy",
+        icon_id=6,
+        temperature=Temperature(c=15.5, f=60),
+    )
+    location_key: str = "39376_PC"
+    mock_client: Any = mocker.AsyncMock(spec=AsyncClient)
+
+    accuweather = AccuweatherBackend(cache=redis_mock, **accuweather_parameters)
```
linabutler, line 793:
```suggestion
    accuweather = AccuweatherBackend(cache=RedisAdapter(redis_mock), **accuweather_parameters)
```

tests/unit/providers/weather/backends/test_accuweather.py
```diff
@@ -776,9 +924,51 @@ async def test_get_forecast(
...
+        url=expected_forecast_url,
+        summary="Pleasant Saturday",
+        high=Temperature(f=70),
+        low=Temperature(f=57),
+    )
+
+    location_key: str = "39376_PC"
+    mock_client: Any = mocker.AsyncMock(spec=AsyncClient)
+
+    accuweather = AccuweatherBackend(cache=redis_mock, **accuweather_parameters)
```
linabutler, line 956:
```suggestion
    accuweather = AccuweatherBackend(cache=RedisAdapter(redis_mock), **accuweather_parameters)
```

tests/unit/providers/weather/backends/test_accuweather.py
```diff
@@ -864,3 +1055,169 @@ async def test_cache_inputs_for_weather_report(
...
+    url = "/forecasts/v1/daily/1day/39376_PC.json"
+
+    async def mock_get(key):
+        return json.dumps({"key": key}).encode("utf-8")
+
+    redis_mock.get.side_effect = mock_get
+
+    mock_client: Any = mocker.AsyncMock(spec=AsyncClient)
+
+    accuweather = AccuweatherBackend(cache=redis_mock, **accuweather_parameters)
```
linabutler, line 1108:
```suggestion
    accuweather = AccuweatherBackend(cache=RedisAdapter(redis_mock), **accuweather_parameters)
```

tests/unit/providers/weather/backends/test_accuweather.py
```diff
@@ -864,3 +1055,169 @@ async def test_cache_inputs_for_weather_report(
...
+        status_code=200,
+        headers={"Expires": expiry_date.strftime(ACCUWEATHER_CACHE_EXPIRY_DATE_FORMAT)},
+        content=json.dumps(expected_client_response).encode("utf-8"),
+        request=Request(
+            method="GET",
+            url=f"test://test/{url}?apikey=test",
+        ),
+    )
+
+    accuweather = AccuweatherBackend(cache=redis_mock, **accuweather_parameters)
```
linabutler, line 1177:
```suggestion
    accuweather = AccuweatherBackend(cache=RedisAdapter(redis_mock), **accuweather_parameters)
```

tests/unit/providers/weather/backends/test_accuweather.py
```diff
@@ -864,3 +1055,169 @@ async def test_cache_inputs_for_weather_report(
...
+):
+    """Test that an error is raised for cache miss."""
+    redis_mock = mocker.AsyncMock(spec=Redis)
+
+    async def mock_get(key):
+        return mock_cache_entry
+
+    redis_mock.get.side_effect = mock_get
+
+    accuweather = AccuweatherBackend(cache=redis_mock, **accuweather_parameters)
```
linabutler, line 1220:
```suggestion
    accuweather = AccuweatherBackend(cache=RedisAdapter(redis_mock), **accuweather_parameters)
```

tests/unit/providers/weather/backends/test_accuweather.py
```diff
@@ -26,11 +33,29 @@
...
+
+    mock = mocker.AsyncMock(spec=Redis)
+    mock.get.side_effect = mock_get
+    mock.set.side_effect = mock_set
+    return mock
+
+
 @pytest.fixture(name="accuweather_parameters")
-def fixture_accuweather_parameters() -> dict[str, str]:
+def fixture_accuweather_parameter(statsd_mock: Any) -> dict[str, Any]:
```
linabutler, line 53: Did you mean to drop the plural (`accuweather_parameters` => `accuweather_parameter`) here? It's okay either way; I was just curious, since the fixture name still has the plural! 👀

tests/unit/providers/weather/backends/test_accuweather.py
```diff
@@ -553,9 +615,46 @@ async def test_get_location(
...
+    location: Optional[AccuweatherLocation] = await accuweather.get_location(
+        mock_client, country, postal_code
+    )
+
+    assert location == expected_location
+    redis_mock.get.assert_called_once_with(
+        f"AccuweatherBackend:v1:/locations/v1/postalcodes/{country}/search.json:"
+        f"{hashlib.blake2s(str({'q': postal_code}).encode('utf-8')).hexdigest()}"
+    )
+    assert not mock_client.get.called
```
linabutler, line 650:
```suggestion
    mock_client.get.assert_not_called()
```

suggestion: `assert_not_called` has a nicer error message with details (mock name, number of times it was called) if the assertion fails.

tests/unit/providers/weather/backends/test_accuweather.py
```diff
@@ -661,9 +763,49 @@ async def test_get_current_conditions(
...
+
+    accuweather = AccuweatherBackend(cache=redis_mock, **accuweather_parameters)
+    conditions: Optional[CurrentConditions] = await accuweather.get_current_conditions(
+        mock_client, location_key
+    )
+    assert conditions == expected_conditions
+    redis_mock.get.assert_called_once_with(
+        "AccuweatherBackend:v1:/currentconditions/v1/39376_PC.json"
+    )
+    assert not mock_client.get.called
```
linabutler, line 801:
```suggestion
    mock_client.get.assert_not_called()
```

tests/unit/providers/weather/backends/test_accuweather.py
```diff
@@ -776,9 +924,51 @@ async def test_get_forecast(
...
+
+    accuweather = AccuweatherBackend(cache=redis_mock, **accuweather_parameters)
+    forecast: Optional[Forecast] = await accuweather.get_forecast(
+        mock_client, location_key
+    )
+    assert forecast == expected_forecast
+    redis_mock.get.assert_called_once_with(
+        "AccuweatherBackend:v1:/forecasts/v1/daily/1day/39376_PC.json"
+    )
+    assert not mock_client.get.called
```
linabutler, line 964:
```suggestion
    mock_client.get.assert_not_called()
```

tests/unit/providers/weather/backends/test_accuweather.py
```diff
@@ -864,3 +1055,169 @@ async def test_cache_inputs_for_weather_report(
...
+    accuweather = AccuweatherBackend(cache=redis_mock, **accuweather_parameters)
+
+    results: dict[str, Any] = await accuweather.get_request(mock_client, url)
+    assert results == {
+        "key": "AccuweatherBackend:v1:/forecasts/v1/daily/1day/39376_PC.json"
+    }
+
+    statsd_mock.timeit.assert_called_once_with("accuweather.cache.fetch")
+    statsd_mock.increment.assert_called_once_with("accuweather.cache.hit.forecasts")
+    assert not mock_client.get.called
```
linabutler, line 1117:
```suggestion
    mock_client.get.assert_not_called()
```

tests/unit/providers/weather/backends/test_accuweather.py
```diff
@@ -864,3 +1055,169 @@ async def test_cache_inputs_for_weather_report(
         cache_inputs_by_location[0]
     )
     assert cache_inputs == cache_inputs_by_location[1]
+
+
+@pytest.mark.parametrize(
```
linabutler, line 1060: Fantastic new tests! 🏆 I'm always inspired to be more thorough in writing tests when I review your PRs!

merino/providers/weather/backends/accuweather.py
mtbottle: Oh good point! I think Python 3.7+ [has stable `dict` enumeration order](https://mail.python.org/pipermail/python-dev/2017-December/151283.html) (which makes me presume that the string will be stable too). However, `dicts` should _not_ be assumed to be ordered, as that's not the intent of this type of data structure. So let's make this explicit!

tests/unit/providers/weather/backends/test_accuweather.py
mtbottle: Nope, this is a typo! Thanks for the catch!

merino/providers/weather/backends/accuweather.py
ncloudioj: Just a note that accessing a deeply nested property of the Dynaconf's settings object can be slow, in particular on the hot code paths as it will do multiple lookups. We've been bitten by that before. As a workaround, we often use a module constant or a class member to store its value for future reference.

```
URL_PARAM_API_KEY: str = settings.accuweather.url_param_api_key
```

merino/providers/weather/backends/accuweather.py
```diff
@@ -92,6 +110,87 @@ def cache_inputs_for_weather_report(self, geolocation: Location) -> Optional[byt
...
+        if "apikey" in params:
+            del params["apikey"]
+
+        if params:
+            extra_identifiers = hashlib.blake2s(str(params).encode("utf-8")).hexdigest()
+            return f"{self.__class__.__name__}:v1:{url}:{extra_identifiers}"
+        return f"{self.__class__.__name__}:v1:{url}"
+
+    async def get_request(
+        self, client: AsyncClient, url: str, params: dict[str, str] = {}
```
ncloudioj, line 129: Calling it `url` seems a bit misleading here, it's actually a path as we have set the base URL upon the creation of the client. I was confused when I saw `request_type: str = url.strip("/").split("/")[0]` down below cuz given a full URL that'll return `https:`.

merino/providers/weather/backends/accuweather.py
```diff
@@ -92,6 +110,87 @@ def cache_inputs_for_weather_report(self, geolocation: Location) -> Optional[byt
...
+    ) -> dict[str, Any]:
+        """Get API response. Attempt to get it from cache first,
+        then actually make the call if there's a cache miss.
+        """
+        cache_key = self.cache_key_for_accuweather_request(url, params)
+        response_dict: dict[str, str]
+
+        # The top level path in the URL gives us a good enough idea of what type of request
+        # we are calling from here.
+        request_type: str = url.strip("/").split("/")[0]
```
ncloudioj, line 139: Since we only care about the 1st entry, no need to split the rest.
```suggestion
        request_type: str = url.strip("/").split("/", 1)[0]
```

merino/providers/weather/backends/accuweather.py
ncloudioj: On top of Lina's suggestion, we can do this to save a dict copy and a `dict.pop()` for all current condition and forecast requests.
```suggestion
           if len(params) > 1:  // "apikey" is included for all requests
                hasher = hashlib.blake2s()
                for key, value in sorted(params.items()):
                    if key != self.url_param_api_key:
                        hasher.update(key.encode("utf-8") + value.encode("utf-8"))
                extra_identifiers = hasher.hexdigest()
```

ncloudioj:
This looks great, r=nanj. Feel free to merge once Lina's ✅ is granted.

merino/providers/weather/backends/accuweather.py
mtbottle: > we can do this to save a dict copy and a dict.pop() for all current condition and forecast requests.

Oh! This is actually a much better idea. I'll do this instead, thanks!

merino/providers/weather/backends/accuweather.py
mtbottle: Ah ya, `url_path` is more accurate. I'll change it to that.

linabutler (approved):
Looks great, thanks! r+wc about the exceptions.

PR #332, opened 2023-06-17
feat: add score to AMO RS suggestions and fallback scores to chunked RS uploader

0c0w3:
## Description

Modify the AMO RS uploader job by adding scores to AMO RS suggestions. The purpose of this is to give AMO RS suggestions a high score to ensure the client will prefer them to adM suggestions when there are keyword collisions. (There are currently 22 collisions total.)

This is intended as a quick fix for the upcoming AMO suggestions experiment. The client already looks for a `score` property in AMO suggestions, but currently none is defined, so the default 0.2 score is used. If/when AMO suggestions are enabled by default instead of being enabled only in experiments, we will want to revisit this.

To implement this in a way that can be used for all RS suggestion types, add a param to the chunked RS uploader that lets the consumer specify a fallback score for suggestions. If a suggestion doesn't include a score and a fallback is defined, the fallback value will be added to the suggestion as its score.

## PR Review Checklist

### DIFF
--- a/merino/jobs/amo_rs_uploader/chunked_rs_uploader.py
+++ b/merino/jobs/amo_rs_uploader/chunked_rs_uploader.py
@@ -82,6 +82,7 @@ class ChunkedRemoteSettingsUploader:
     dry_run: bool
     kinto: kinto_http.Client
     record_type: str
+    suggestion_score_fallback: float | None
 
     def __init__(
         self,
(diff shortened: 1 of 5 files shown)


merino/jobs/amo_rs_uploader/chunked_rs_uploader.py
```diff
@@ -106,6 +109,8 @@ def add_suggestion(self, suggestion: dict[str, Any]) -> None:
         """Add a suggestion. If the current chunk becomes full as a result, it
         will be uploaded before this method returns.
         """
+        if self.suggestion_score_fallback and "score" not in suggestion:
+            suggestion = suggestion | {"score": self.suggestion_score_fallback}
```
ncloudioj, line 113: Nit:
```suggestion
            suggestion |= {"score": self.suggestion_score_fallback}
```

ncloudioj (approved):
LGTM.

Note that this ranking logic makes sense for experimentation. For production use, we might double check with Prod as we currently enforce this "Sponsored suggestions take precedency over other suggestions" for Firefox Suggest.

0c0w3:
> For production use, we might double check with Prod as we currently enforce this "Sponsored suggestions take precedency over other suggestions" for Firefox Suggest.

Right -- for the experiment, I confirmed with Nive that addon suggestions should take precedence over sponsored suggestions, but we'll probably want to revisit if/when addon suggestions are enabled by default.

PR #349, opened 2023-06-23
fix: score AMO suggestions below adM suggestions

0c0w3:
## Description

Change the score of AMO suggestions so it's less than the score of adM suggestions. Use the new score for both Merino and remote settings AMO suggestions. This ensures the client will show sponsored suggestions instead of AMO suggestions when keyword collisions occur, which according to Product is the desired behavior for both the addons-suggestion experiment and general release.

## PR Review Checklist

### DIFF
--- a/merino/configs/default.toml
+++ b/merino/configs/default.toml
@@ -129,7 +129,7 @@ api_url = "https://addons.mozilla.org/api/v5/addons/addon/"
 [default.providers.amo]
 type = "amo"
 enabled_by_default = false
-score = 0.3
+score = 0.25
 # Specifies which backend to use. Currently defaulting to dynamic backend.
 backend = "dynamic"
 # The minimum number of characters to be considered for matching.
@@ -209,6 +209,3 @@ min_favicon_width = 48
 [default.jobs.amo_rs_uploader]
 # The "type" of each remote settings record
 record_type = "amo-suggestions"
-# While the addon experiment is active, addon suggestions should be preferred
-# over all other types.
-score = 1.0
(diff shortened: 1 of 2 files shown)


mtbottle (approved):

merino/configs/default.toml
```diff
@@ -129,7 +129,7 @@ api_url = "https://addons.mozilla.org/api/v5/addons/addon/"
 [default.providers.amo]
 type = "amo"
 enabled_by_default = false
-score = 0.3
+score = 0.25
```
ncloudioj, line 132: Note that `top_picks` (for Navigational Queries) also has the same score, if we end up enabling it by default post-experiment, it might introduce some more collisions with `amo`. Do we want to bump the score up a bit to avoid that?

0c0w3: Yeah, it's not clear how amo should be ranked vs. top picks IMO, or any other type vs. any other, except we know we probably want sponsored to have the highest score, and maybe Wikipedia should have the lowest. OTOH giving two types the same score will result in nondeterminism on the client, which isn't great. Wdyt if we leave this at 0.25 for now and revisit if/when amo (and other types) are enabled for everyone?

ncloudioj: Yup, revisting later sounds good to me.

quiiver: I just want to highlight that changes to the ranking post-experiment could mean that the observed effects change when the score changes.

0c0w3: The only reason we'd change the amo ranking post-experiment is because we're enabling some other type(s) of suggestion alongside amo, whether in a new experiment or by default. In that case, the observed effects would change anyway due to the presence of the other suggestion types, right?

Based on discussion with Nive over the past few days, I think we're pretty sure we want the ranking to remain (1) adm, (2) amo, (3) wikipedia, as long as we're talking about only those three being enabled together.

Do you think we need to decide the relative ranking of other suggestion types before making this change?

PR #353, opened 2023-06-28
feat: add Pocket RS suggestions uploader

0c0w3:
## Description

Add a job that uploads Pocket suggestions to remote settings. The source data is expected to be in a CSV file because we're collaborating with the Pocket team to build a list of suggestions and keywords using [this Google sheet](https://docs.google.com/spreadsheets/d/14mjaEYWnHD3VuTT5vq1HCzCz9e98cm8Z5a9fwjtKzGw/edit?usp=sharing). Running this job requires an initial manual step of exporting the sheet as CSV. I didn't add the CSV to this commit because I don't think we want to keep it in the repo.

Move chunked_rs_uploader.py to a new jobs/utils directory so it can be shared between the AMO and Pocket uploader and any future uploader jobs. Modify it to zero-pad the chunk start and end indexes in record IDs similar to how indexes are zero-padded in adm records.

## PR Review Checklist

### DIFF
--- a/merino/jobs/pocket_rs_uploader/__init__.py
+++ b/merino/jobs/pocket_rs_uploader/__init__.py
@@ -1,22 +1,37 @@
 """CLI commands for the pocket_rs_uploader module"""
 import asyncio
 import csv
+import io
 from typing import Any
 
 import typer
+from pydantic import BaseModel, HttpUrl, validator
 
 from merino.config import settings as config
 from merino.jobs.utils.chunked_rs_uploader import ChunkedRemoteSettingsUploader
 
-# Maps from CSV field name in the input data to the corresponding suggestion
-# property in the output.
-SUGGESTION_PROPERTIES_BY_CSV_FIELD = {
-    "Collection Title": "title",
-    "Collection Description": "description",
-    "Collection URL": "url",
-    "High-Confidence Keywords": "highConfidenceKeywords",
-    "Low-Confidence Keywords": "lowConfidenceKeywords",
-}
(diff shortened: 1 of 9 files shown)


merino/jobs/pocket_rs_uploader/__init__.py
```diff
@@ -0,0 +1,171 @@
...
+        # Generate the full list of suggestions before creating the chunked
+        # uploader so we can detect errors in the source data before deleting
+        # existing records and starting the upload.
+        suggestions: list[dict[str, Any]] = []
+        for row in csv_reader:
+            suggestion: dict[str, Any] = {}
+            for field, prop in SUGGESTION_PROPERTIES_BY_CSV_FIELD.items():
+                if not row[field]:
+                    raise Exception(f"Empty value for '{field}' in row: {row}")
+                suggestion[prop] = row[field]
```
ncloudioj, line 145: Do we want to validate the CSV data? For example, certain fields should be required and valid (e.g. URL). If so, we can define a simple Pydantic model (already available in Merino) for `Suggestion` and let it do validation for us.

ncloudioj (approved):
Looks good, just one suggestion for input data validation.

merino/jobs/pocket_rs_uploader/__init__.py
0c0w3: Good point, I added some validation. The changes aren't trivial so could you review again please?

ncloudioj (approved):
r+, thanks!

PR #354, opened 2023-07-04
[DISCO-2473] refactor: Reuse HTTP client for accuweather backend

ncloudioj:
## References

JIRA: [DISCO-2473](https://mozilla-hub.atlassian.net/browse/DISCO-2473)
GitHub: N/A

## Description

Profiled Merino via Scalene and the result suggested the `AsyncClient` creation in the AccuWeather backend really requires a lot of CPU cycles and Merino created an AsyncClient for each request!

The benchmark results show a 5-6x boost in throughput and latency with the patch.

**Before: 225 QPS / 522 ms (p95 latency)
After: 1160 QPS / 115 ms (p95 latency)** 

## PR Review Checklist

[DISCO-2473]: https://mozilla-hub.atlassian.net/browse/DISCO-2473?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/merino/providers/weather/backends/accuweather.py
+++ b/merino/providers/weather/backends/accuweather.py
@@ -61,6 +61,7 @@ class AccuweatherBackend:
     url_forecasts_path: str
     url_param_partner_code: Optional[str]
     partner_code: Optional[str]
+    http_client: AsyncClient
 
     def __init__(
         self,
@@ -107,6 +108,7 @@ class AccuweatherBackend:
         self.url_forecasts_path = url_forecasts_path
         self.url_param_partner_code = url_param_partner_code
         self.partner_code = partner_code
+        self.http_client = AsyncClient(base_url=self.url_base)
 
     def cache_key_for_accuweather_request(
         self, url: str, query_params: dict[str, str] = {}
@@ -127,7 +129,7 @@ class AccuweatherBackend:
         return f"{self.__class__.__name__}:v1:{url}"
 
     async def get_request(
-        self, client: AsyncClient, url_path: str, params: dict[str, str] = {}
+        self, url_path: str, params: dict[str, str] = {}
     ) -> dict[str, Any]:
         """Get API response. Attempt to get it from cache first,
         then actually make the call if there's a cache miss.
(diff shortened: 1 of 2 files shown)


mtbottle (approved):
Nice catch!

merino/providers/weather/backends/accuweather.py
```diff
@@ -107,6 +108,7 @@ def __init__(
         self.url_forecasts_path = url_forecasts_path
         self.url_param_partner_code = url_param_partner_code
         self.partner_code = partner_code
+        self.http_client = AsyncClient(base_url=self.url_base)
```
Trinaa, line 111: In order to maintain the same level of test-ability as before, I recommend injecting the AsyncClient into the backend, instead of using patch with side_effect.

In the signature we could do
```
http_client: AsyncClient = None 
```

And then instantiate like bellow
```
self.http_client = (
    AsyncClient(base_url=self.url_base) if http_client is None else http_client
)
```

Then in the tests we could inject the mock_client into the Accuweather backend, setting return values like:
```
accuweather.http_client.get.return_value = Response...
```

merino/providers/weather/backends/accuweather.py
```diff
@@ -61,6 +61,7 @@ class AccuweatherBackend:
     url_forecasts_path: str
     url_param_partner_code: Optional[str]
     partner_code: Optional[str]
+    http_client: AsyncClient
```
Trinaa, line 64: Do we need to `close` the client?

merino/providers/weather/backends/accuweather.py
ncloudioj: Good point, let's do it for better testability.

merino/providers/weather/backends/accuweather.py
ncloudioj: Yes, it'll be closed [here](https://github.com/mozilla-services/merino-py/pull/354/files#diff-920fdf1cb932b5beb9b8d1494db8b081550d945a58280bf8e626fb991634a44bR403)

Trinaa: oops! missed that. Thank you @ncloudioj

merino/providers/weather/backends/accuweather.py
Trinaa: FYI: https://github.com/mozilla-services/merino-py/pull/356

Issue #366, opened 2023-07-11
Merino: Titles in Wikipedia Suggestions Have Trailing Punctuation


taddes:
Reported on the #contextual-services-tech Slack channel on June 28th 2023:

Some titles on wikipedia seem to have a comma at the end (this one is likely because the title is “Split, Croatia”).

Fix so that the comma does not return in the result suggestion.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/DISCO-2502)

PR #367, opened 2023-07-11
[DISCO-2474] Merino: Titles in Wikipedia Suggestions Have Trailing Punctuation

taddes:
## References

JIRA: [DISCO-2474](https://mozilla-hub.atlassian.net/browse/DISCO-2474)
GitHub: [#366 ](https://github.com/mozilla-services/merino-py/issues/366)

## Description
Reported on the #contextual-services-tech Slack channel on June 28th 2023

Some titles on wikipedia seem to have a comma at the end (this one is likely because the title is “Split, Croatia”). It’s low priority but we should fix it at some point if we can if there are multiple occurrences.

Multiple occurrences found, mostly with geographic locations.
Screen Shot 2023-07-11 at 6 52 27 PM
Screen Shot 2023-07-11 at 7 00 27 PM

As discussed in ticket, we could just again return the title, but upon investigation this method required the removal of the comma from passed in title, as `get_best_keyword` method splits on the space, which is inclusive of the comma.  Have added logic to remove any and all punctuation that could also cause this issue. Just returning the title and then the text to the immediate right being the same (just possibly lowercase) seems like it would be cluttered and redundant. 

## PR Review Checklist

[DISCO-2474]: https://mozilla-hub.atlassian.net/browse/DISCO-2474?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/merino/providers/wikipedia/backends/elastic.py
+++ b/merino/providers/wikipedia/backends/elastic.py
@@ -1,5 +1,6 @@
 """The Elasticsearch backend for Dynamic Wikipedia."""
 import logging
+import string
 from typing import Any, Final
 from urllib.parse import quote
 
@@ -23,7 +24,7 @@ def get_best_keyword(q: str, title: str):
     """Try to get the best autocomplete keyword match from the title. If there are no matches,
     then return the full title as a match. Lowercase everything.
     """
-    title = title.lower().replace(",", "")
+    title = title.lower().translate(str.maketrans("", "", string.punctuation))
     q = q.strip().lower()
     start_index = title.find(q)
     if start_index < 0:
--- a/merino/providers/wikipedia/backends/elastic.py
+++ b/merino/providers/wikipedia/backends/elastic.py
@@ -24,7 +24,7 @@ def get_best_keyword(q: str, title: str):
     """Try to get the best autocomplete keyword match from the title. If there are no matches,
     then return the full title as a match. Lowercase everything.
     """
-    title = title.lower().translate(str.maketrans("", "", string.punctuation))
+    title = title.lower()
     q = q.strip().lower()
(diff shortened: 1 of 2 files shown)


merino/providers/wikipedia/backends/elastic.py
```diff
@@ -23,7 +24,7 @@ def get_best_keyword(q: str, title: str):
     """Try to get the best autocomplete keyword match from the title. If there are no matches,
     then return the full title as a match. Lowercase everything.
     """
-    title = title.lower()
+    title = title.lower().translate(str.maketrans("", "", string.punctuation))
```
ncloudioj, line 27: Translating the entire string might be surprising, e.g. `spider-man` will be translated into "spiderman" which is vastly different from the original keyword, searching the latter won't lead to the result of the former. How about we just strip the trailing punctuation of the full keyword?

```py
    return title[start_index:end_index].rstrip(string.punctuation)
```

ncloudioj (changes requested):
Could you also add a test for this?

merino/providers/wikipedia/backends/elastic.py
taddes: Wfm, makes sense. Good thought @ncloudioj 🕷🦸‍♂️

ncloudioj: Can you add a quick test for this, please?

ncloudioj (approved):
Looks good, having a test case for this would be nice.

Thanks!

taddes:
Closes #366

PR #372, opened 2023-07-18
[DISCO-2530] fix: Use common configurations for async http client

ncloudioj:
## References

JIRA: [DISCO-2530](https://mozilla-hub.atlassian.net/browse/DISCO-2530)
GitHub: N/A

## Description
This adds common configurations for timeout and connection pool to the HTTP client that Merino uses.

Per client connection pool max size: 100 => 1024
Timeouts:
- Socket connection timeout: 5 secs => 1  sec
- HTTP request timeout: 5 secs unchanged
- Pool connection acquisition: 5 secs => 1 sec 

Currently, all the async clients in Merino use the same configurations, we can customize those configs for individual cases if needed in the future.

## PR Review Checklist

[DISCO-2530]: https://mozilla-hub.atlassian.net/browse/DISCO-2530?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/merino/utils/http_client.py
+++ b/merino/utils/http_client.py
@@ -7,7 +7,7 @@ from httpx import AsyncClient, Limits, Timeout
 
 def create_http_client(
     base_url: str = "",
-    max_connection: int = 1024,
+    max_connections: int = 1024,
     connect_timeout: float = 1.0,
     request_timeout: float = 5.0,
     pool_timeout: float = 1.0,
@@ -25,6 +25,6 @@ def create_http_client(
     """
     return AsyncClient(
         base_url=base_url,
-        limits=Limits(max_connections=max_connection),
+        limits=Limits(max_connections=max_connections),
         timeout=Timeout(request_timeout, connect=connect_timeout, pool=pool_timeout),
     )


mtbottle (commented):
Changes LGTM. 

Did we want to try testing our theory with a smaller connection pool and see if we can replicate the issue?

mtbottle (approved):

ncloudioj:
> Did we want to try testing our theory with a smaller connection pool and see if we can replicate the issue?

Yup, I've tested that locally with a small pool with the following findings:
- HTTP connections from cancelled tasks were correctly released to the pool, which disproves my guess about connection leaking by task cancelling 
- A fully occupied connection pool did block the subsequent requests, the default connection acquisition timeout is 5 seconds, now we lower that down to 1 sec with a larger pool (100 => 1024), so Merino will not block for that long if the new connection can not be acquired. Assuming the average request handling duration to be 200ms, this setup allows Merino to perform up to 1024 * (1 / 200) = 5K requests per second per client, which should be sufficient compared to the existing setup: 100 * (1 / 200) = 500
- Since we didn't see massive traffic spikes during the degradation, it's still unclear to me what exactly cause the issue (i.e. massive connection acquisition timeouts). Unfortunately, we use a generic exception handler [here](https://github.com/mozilla-services/merino-py/blob/main/merino/providers/weather/backends/accuweather.py#L261), will add more logging/metrics to collect more details about the request error(s)

PR #373, opened 2023-07-19
Update Service Architecture and Move API Documentation

mtbottle:
## References

JIRA: [DISCO-2486](https://mozilla-hub.atlassian.net/browse/DISCO-2486)

## Description
This change does 2 things:
1. Update the Service Architecture Diagram to add new dependencies (i.e. more Airflow jobs, MaxmindDB, GCS dependencies etc.)
2. Move our API documentation to the auto-generated Swagger UI. 

I think moving the API documentation has a few benefits:

- documentation lives closer to the code, so will be easier to update
- the FastAPI/Swagger generated API page shows user how to _actually_ call the API and the shape of the data (and you can play around with the API on the page!)
- documentation is generated directly from the model objects that we use, which means they will be more accurate and less prone to being outdated
- error codes are more precise (i.e. explicit 422, rather than a broad 4xx in the original docs)

The biggest concern about the auto-generated documentation is around the extended models from `BaseSuggestion`. Since they're not referenced directly from any of the APIs, their models don't get auto-generated either (another reason in favour of [our chosen ADR solution](https://github.com/mozilla-services/merino-py/blob/main/docs/adr/0002-merino-general-response.md)). However, since there are existing models that extend from `BaseSuggestion` currently, I've just added the extra fields to the `BaseSuggestion` documentation for completeness. I would like some opinions on this approach. 

## PR Review Checklist

[DISCO-2486]: https://mozilla-hub.atlassian.net/browse/DISCO-2486?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


docs/api.md
```diff
@@ -164,31 +23,3 @@ the user opts out the telemetry collection of Firefox.
 
 The required behavior for interaction pings is TBD.
```
mtbottle, line 24: Is this still TBD? (and does it actually belong in the API section of the documentation?)

docs/intro.md
```diff
@@ -88,43 +105,47 @@ graph TD
...
     redis[(Redis Cache)]
     accuweather ..-> |tries to query cache first| redis
 
     kinto[(Remote Settings)]
-    offline ..-> |fetches adMarketplace and  <br/> static Wikipedia suggestions only.  <br/> Offline mode is fallback if Merino times out.| kinto
-    rsb --> kinto
+    rsb --- kinto
+    addons_sync ..- |Add Addons Suggestions to Remote Settings| kinto
 
+    toppicks_data[(GCS Top Picks Data,<br/>metadata and icons to be <br/>displayed on browser)]
```
ncloudioj, line 144:
```suggestion
    toppicks_data[(GCS Top Picks Data,<br/>a list of Mozilla curated popular sites and metadata to be <br/>displayed on browser)]
```

ncloudioj (approved):
This looks great, thanks for porting this over to OpenAPI docs.

merino/providers/base.py
```diff
@@ -16,8 +16,76 @@ class SuggestionRequest(BaseModel):
...
+      If this is true, the UI must indicate to the user that the suggestion is
+      sponsored.
+    - `is_top_pick` - A boolean indicating if the suggestion requires a "top pick"
+      UI treatment.
+    - `city_name` - A string representing the city name. To be used with weather suggestions.
+    - `current_conditions` - An object that contains the current weather conditions for a location.
+      To be used with weather suggestions. Please reference `CurrentCondition` model
+      for more information.
+    - `forecast` - An object that contains the one-day weather forecast information.
+      To be used with weather suggestions. Please reference `Forecast` model for more information.
```
taddes, line 88: I have a general thought about this docstring, given it's a long one: why don't we annotate the attribute with the type right after its definition, separating with a colon? This saves space, reduces verbosity and makes it easier to immediately know the type. This also eliminates any ambiguity in the case a special class is being used say over a string (ex. a url could be a string or HttpUrl).

The added bonus is that it can be parsed by Sphinx and will show up in an IDE as part of the signature when used as we have now. This would make it a [NumPy/SciPy ](https://numpydoc.readthedocs.io/en/latest/format.html) compliant docstrings, which is officially supported by Sphinx and a formal specification. It's also the middle ground between the Google standard and reStructuredText, which I think is a bit ugly and harder to maintain.

Just an example: 
```suggestion
    """Base model for suggestions.

    Fields:
    -------
    title : str 
         The full title of the suggestion resulting from the query. Using the
        example of apples above, this might be "Types of Apples in the Pacific
        Northwest". This field should be treated as plain text.
    url : HttpUrl 
        The URL of the page that should be navigated to if the user selects
        this suggestion. This will be a resource with the title specified in the
        `title` field.
    provider : str 
         Identifies the provider of this suggestion, such as
        "adM". In general, this field is not intended to be directly displayed to the user.
```
etc.

merino/web/api_v1.py
```diff
@@ -61,14 +61,80 @@ async def suggest(
...
 
-    Args:
-    - `q`: The query string
-    - `client_variants`: [Optional] A comma separated string indicating the client variants
-    - `providers`: [Optional] A comma separated string indicating the suggestion providers for
-      this query
-
-    Returns:
-    A list of suggestions or an empty list if nothing was found.
+    This is the primary endpoint that consumes user input and suggests
```
taddes, line 64: Same here, converting the format to something officially supported and easier to read.

taddes (approved):
Great work @mtbottle . I had some thoughts regarding converting the long docstrings into Sphinx-compliant, NumPy/SciPy docstrings to eliminate some verbosity and increase readability, plus it allows us to conform to an official formal specification while also reaping the benefit of types being right there next to the field. More details in the comments.

merino/providers/base.py
mtbottle: Interesting thought! I tried out the new format and it doesn't play nicely with the markdown formatter in OpenJson:

Screenshot 2023-07-20 at 09 39 31
 
I did do some investigation and found that you can actually add a description to the fields themselves, so I added the documentation there! The documentation should loo more like this now:

Screenshot 2023-07-20 at 09 35 53

There's a fair bit more clicking, but the documentation is now closer to the field, which is good I think...

taddes: Yeah, that looks really good @mtbottle !Wonder if it's just the indent level that controls whether that formatting kicks in. Either way, the description added to the fields with the type is nice!

mtbottle:
@ncloudioj  and @taddes I did some more changes, including moving documentation to the actual field itself. Since this is a large change, I'm re-requesting reviews!

ncloudioj (approved):
:shipit:

taddes (approved):
lgtm! 🚀

PR #375, opened 2023-07-19
[DISCO-2533] Upgrade & Migration: FastAPI, Pydantic and Uvicorn Libraries in Merino [load test: abort]

taddes:
## References

JIRA: [DISCO-2533](https://mozilla-hub.atlassian.net/browse/DISCO-2533)
GitHub: [#390 ](https://github.com/mozilla-services/merino-py/issues/390)

## Description
Upgrades core dependencies and ensure Merino runs without error due to the upgrade. 

The latest FastAPI (0.100.0) comes with some exciting features/improvements such as Pydantic 2.0. This supports the high priority H2 objective: Project Maintenance & Enhancements (Priority: High), as part of our Q3 maintenance. 

Pydantic underwent a significant upgrade and numerous models and methods are depricated. We don’t want to support an out of date library. Furthermore, there are significant performance, stability and usage improvements. See [release notes ](https://fastapi.tiangolo.com/release-notes/#01000). 

Benefits:
Improved correctness in corner cases.
Clearer method names
Safer types.
Better performance and less energy consumption.
Better extensibility.

PR Notes:
For simplicity, not every `Optional` has been taken out. The following PR that will port all these remaining types over and upgrades mypy will do this. Essential methods have been replaced and dependent v1 methods are removed. 

A major issue in the migration was data loss. The subclass attributes for a Suggestion would be emitted from each individual provider, however something during serialization was making them vanish, only returning `BaseSuggestion` attributes. For instance, `block_id` and` is_top_picks` would not be emitted when querying merino.  Pydantic didn’t include this in the migration guide (:cat-evil: ). Because of how weird this seemed, I started wondering if it was a [bug](https://github.com/pydantic/pydantic/issues/6895) or some odd change that wasn’t front and center.  Many folks who were migrating were having the same issue... When I found this filed bug on their issues, it led to [here](https://docs.pydantic.dev/latest/usage/serialization/#subclass-instances-for-fields-of-basemodel-dataclasses-typeddict), where they really changed how subclassing works when serializing model data. Needed to wrap with SerializeAny and voila! Now the child attributes show in all the providers as expected. This is why unit tests would pass, but the integration tests would fail. Just some context to explain this change

Also note that the new `HttpUrl` model by default appends a trailing / to a url. This was clarified by the contract tests failing, as the urls in suggestions don’t have them. Checking the api-as is, it returns https://google.com/, and with V2, it will be https://google.com/ . This is only added after the TLD, Anything beyond the index does not get an appended. Ex: `pydantic.HttpUrl("https://google.com")` and you get `Url('https://google.com/')`

## PR Review Checklist

[DISCO-2533]: https://mozilla-hub.atlassian.net/browse/DISCO-2533?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


ncloudioj:
Nice, good to see everything is green now.

I just did some test runs, looks like there are a few warnings (will comment in the patch).

Also the benchmark shows this branch regressed the performance by ~25% (against `main`) for both throughput and latency, which is interesting. Will take a closer look into it.

merino/utils/log_data_creators.py
```diff
@@ -22,14 +22,15 @@ class LogDataModel(BaseModel):
...
+    @field_validator("time", mode="before")
+    @classmethod
     def validate_time(cls, v, values, **kwargs):
         """Attempt to parse the provided value for `time` into a datetime
         model which will check that the value is a valid datetime representation.
         Then ensure that we output the datetime in a consistent isoformat,
         if it wasn't already in that format.
         """
-        dt: datetime = parse_obj_as(datetime, v)
+        dt: datetime = TypeAdapter(datetime).validate_python(v)
```
ncloudioj, line 33: The profile result suggests this validation really requires lots of CPU cycles, dropping it yields a huge delta in performance gain. @Trinaa can you remind me why did we want to do this validation instead of passing in the ISO-formatted string directly [here](https://github.com/mozilla-services/merino-py/blob/main/merino/middleware/logging.py#L51C1-L56)? The validation seems unnecessary as the `datatime` object is created by Merino rather than by external producers.

pyproject.toml
```diff
@@ -62,8 +62,8 @@ merino-jobs = "merino.jobs.cli:cli"
 
 [tool.poetry.dependencies]
 python = "^3.11"
-fastapi = "^0.79.0"
-uvicorn = {extras = ["standard"], version = "^0.18.2"}
+fastapi = "^0.100.0"
```
ncloudioj, line 65: Can you do a rebase against the main branch? Dependabot updated various things which might get downgraded by this patch.

ncloudioj (commented):
Looks good, a few warning fixes and a performance fix are needed.

merino/cache/none.py
```diff
@@ -1,20 +1,19 @@
...
+    async def get(self, key: str) -> bytes | None:  # noqa: D102
         return None
 
     async def set(
         self,
         key: str,
-        value: Union[bytes, str],
-        ttl: Optional[timedelta] = None,
+        value: bytes | str,
+        ttl: timedelta | None = None,
```
ncloudioj, line 16: Is it really necessary? Neither `Union` nor `Optional` is deprecated in typing.py, I think we should allow folks to use either one instead of forcing them to use the `A | B`.

@taddes can you confirm?

taddes: Hey @ncloudioj  👋 The`|` operator for typing was added part of [PEP-604](https://peps.python.org/pep-0604/) and references `The new union syntax should be accepted for function, variable and parameter annotations.` This is intended for > Python 3.10, so while it is acceptable, the new modern standard is pointing to using the operator opposed to `Union` or `Optional`. I suppose we can let folks use either, but since it's part of an official PEP, it is gently recommended to abandon the old keywords.

merino/utils/log_data_creators.py
taddes: Good call, it's likely some of the other validations could yield improvement. However, it's a little confusing as Pydantic marketed their V2 methods to be _faster_... Strange that `parse_obj_as` should be somehow faster than using the `TypeAdapter`. I will do a little profiling as well, but can chat with @Trinaa about this too!

merino/providers/weather/backends/protocol.py
```diff
@@ -9,10 +9,10 @@
 class Temperature(BaseModel):
     """Model for temperature with C and F values."""
 
-    c: Optional[float] = None
-    f: Optional[float] = None
+    c: float | None = None
+    f: float | None = None
 
-    def __init__(self, c: Optional[float] = None, f: Optional[float] = None):
+    def __init__(self, c: float | None = None, f: float | None = None):
```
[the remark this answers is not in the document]
taddes: Makes sense! I did make a slightly different version that lets the `c` and `f` be a bit more permissive allowing `int | float | None`, which means we don't need `cc` and `ff`. Thoughts?

ncloudioj:
@taddes sorry, #392 introduced a couple of more things to this migration. Especially, I used `parse_obj_as()` in v1, looks like we can do `model_validate_json()` instead in v2 ([link](https://docs.pydantic.dev/latest/usage/models/#helper-functions)).

taddes:
Hey @ncloudioj , no worries, I can do the rebase and make those changes. If I have any questions, I'll ping you, but seems pretty straightforward.

merino/utils/log_data_creators.py
linabutler: It's interesting for sure! I tend to think about advertised performance gains like that as a ballpark guideline—because it's very tough to reduce performance down to a single number.

It could still be that the more common paths in Pydantic are faster overall, but this specific one isn't—maybe it was a shorter, pure-Python code path before, but now involves trampolining into Rust through more methods and types.

Investigating if this validation is needed at all is a great next step—especially since we're [already constructing the `datetime` object in our own code](https://github.com/mozilla-services/merino-py/blob/f3d8efed346e17008f95cc589b72bda413c73d39/merino/middleware/logging.py#L43), as @ncloudioj says. The fastest code is code that doesn't have to run at all! 😊

merino/web/api_v1.py
```diff
@@ -123,6 +123,10 @@ async def suggest(
 
     emit_suggestions_per_metrics(metrics_client, suggestions, search_from)
 
+    # NOTE SuggestResponse suggestions attr is BaseSuggestion wrapped in SerializeAsAny.
+    # Ensures that all subclass attributes are part of the suggestion returned to client.
+    # See Pydantic docs for context.
+    # https://docs.pydantic.dev/latest/usage/serialization/#serializing-subclasses
```
linabutler, line 129:
```suggestion
```

suggestion (non-blocking): Let's drop this comment, since you already explain it very well—and link to the same docs—in `SuggestResponse`.

merino/web/models_v1.py
```diff
@@ -14,7 +14,10 @@ class ProviderResponse(BaseModel):
 class SuggestResponse(BaseModel):
     """Model for the `suggest` API response."""
 
-    suggestions: list[BaseSuggestion]
+    # SerializeAsAny required to ensure subclassed attributes serialized and emitted as expected.
```
linabutler, line 17:
```suggestion
    # `SerializeAsAny` ensures that all fields of `BaseSuggestion` and its subclasses are included when converting this model to a dictionary. 
```

suggestion (non-blocking): This is an excellent comment—it calls attention to a subtle behavior! I wonder if we can be just a bit more precise in the wording, to something like this? (Please feel free to tweak as you like!)

Specifically, let's call out _which_ subclasses we're referring to, and let's not rely on Pydantic's terminology in our explanation. As their docs note, "serialization" generally refers to converting data structures to strings or bytes, but "serialization" in the Pydantic world means something slightly different, which might be confusing or ambiguous to someone just skimming this code.

linabutler (commented):
Thank you so much for powering through this, @taddes! You're almost there!

merino/web/api_v1.py
taddes: Good call @linabutler ! Thanks!

merino/web/models_v1.py
taddes: That's a good call @linabutler , I especially agree with the section about calling out the "serialization" terminology in particular. Thanks for adding clarity to the comment, it's easier to follow.  I added one small blurb in parenthesis to go off what you said about what subclasses, what do you think?

```suggestion
    # `SerializeAsAny` ensures that all fields of `BaseSuggestion` and its subclasses
    # (ex. provider-specific Suggestion classes) are included when converting this model
    # to a dictionary.  See Pydantic docs for context:
```

merino/providers/weather/backends/protocol.py
```diff
@@ -9,19 +9,21 @@
 class Temperature(BaseModel):
     """Model for temperature with C and F values."""
 
-    c: Optional[float] = None
-    f: Optional[float] = None
+    c: int | None = None
```
linabutler, line 12: nitpick: For extra credit (meaning, please feel free to push back 😊), I'd love to see the `Temperature` changes split out into their own commit, before this one.

Having this commit separate from the Pydantic migration one already makes your PR _so_ much easier to skim, to the point where I wouldn't mind leaving the `Temperature` changes here—but it does go a bit beyond the `Option` -> `| None` conversion, since `c` and `f` change types from floats to integers (even though the behavior stays the same).

linabutler (changes requested):
This is looking wonderful, @taddes; great job! I wouldn't mind seeing it one more time—you're gonna be a pro at `git rebase` after you're done with this PR—but all I have are nits 😊

Let's see how the load test goes, and then we can get this landed!

merino/featureflags.py
```diff
@@ -45,17 +38,20 @@ class BucketingScheme(str, Enum):
 class FeatureFlag(BaseModel):
     """Model representing a feature flag."""
 
-    enabled: Enabled
+    enabled: confloat(ge=0.0, le=1.0)  # type: ignore [valid-type]
```
quiiver, line 41: I would recommend using the 3.9+ `Annotated` type here. It is what `confloat` uses now and is recommended for describing type constraints. You should also be able to drop the mypy type ignore comment.

```suggestion
    enabled: Annotated[int, Field(gt=0.0, lte=1.0)]
```

https://docs.pydantic.dev/latest/usage/types/custom/

merino/providers/weather/backends/protocol.py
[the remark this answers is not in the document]
ncloudioj: Thanks for your comments, Lina!

Just wanted to add a bit of color here: This model is only used in the API response, and we have decided to show integer temperatures (rounded) in Firefox. That's why I recommended using `int | None` instead of `int | float | None` as the latter doesn't convey what exactly the type is and API consumers might presume that they will need to do rounding again to convert them into integers. Looks like Firefox assumes rounding is already done in Merino and uses temperature values as is (see source [here](https://searchfox.org/mozilla-central/source/browser/components/urlbar/UrlbarProviderWeather.sys.mjs#263-267)) from the API response.

As you mentioned, JSON doesn't distinguish the two. So alternatively, we can stick to `float | None` and convert all `c` and `f` into floats after rounding. Python will serialize them e.g. `23.0` in JSON, I guess JSON deserialization will drop the decimal bits automatically and we still get `23` eventually.

What do you think? @linabutler @taddes

merino/featureflags.py
taddes: Thanks for the suggestion @quiiver !

merino/jobs/csv_rs_uploader/__init__.py
```diff
@@ -5,6 +5,7 @@
 import io
 
 import typer
+from pydantic import BaseModel, HttpUrl, field_validator
```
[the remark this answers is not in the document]
taddes: Hmm. Very odd the linter didn't catch this... 🕵️

tests/integration/api/v1/suggest/test_suggest_top_picks.py
```diff
@@ -75,10 +75,10 @@ def test_top_picks_query(client: TestClient, query: str) -> None:
     assert response.status_code == 200
 
     result = response.json()
-    actual_suggestions: list[Suggestion] = [
-        Suggestion(**suggestion) for suggestion in result["suggestions"]
-    ]
-    assert actual_suggestions == expected_suggestion
```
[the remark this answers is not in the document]
taddes: Yup, you got it @linabutler ! It's to align with `test_top_picks_short_domains`, didn't seem to make sense to have 2 similar functions doing a different thing to get a relatively similar result. I wanted to do some double duty on these tests by instantiating a `Suggestion` from the result data, which I feel robustly ensures we're getting what we expect.  Good call on checking, though, I know what you mean by unexpected test changes.

merino/providers/weather/backends/protocol.py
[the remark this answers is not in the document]
linabutler: Oooh, I like your point about using `int | None` to communicate that the temperature is already rounded! Let's leave it the way it's written now (`int | None`, and keep the `cc` and `ff` logic).

I didn't think about how including decimals in the response, or using `int | float | None`, might actually be more confusing about the expected precision. In JS, `23` and `23.0` are the same, but if you look at the raw JSON response and see `23.0` or `57.0`, you might wonder if you'll ever see a fractional temperature.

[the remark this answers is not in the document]
taddes: Yeah, good points above. I like the delineation about the types based on the creation of the `Temperature` class and then what it passes back, makes for it to be very clear. Thank you both for the insights on this @ncloudioj & @linabutler .

merino/utils/log_data_creators.py
mtbottle: > IIUC, it looks like the (only?) reason LogDataModel.time is a string and not a datetime is to work around Pydantic 1.x's lack of custom per-field serializers. Does that sound right to you, @hackebrot and @mtbottle?

Yes! I believe that was the reason. The example looks good too!

taddes: Excellent! Glad Pydantic supports this feature now and we won't need workarounds in the future!

merino/featureflags.py
quiiver: oh yes that's right

taddes:
Thanks for the excellent coaching in aligning the commits with the work done @linabutler , you're a master at this, appreciate it!

linabutler (approved):
Perfect! Thanks again, @taddes! :shipit:

taddes:
Closes #390

PR #392, opened 2023-08-01
[DISCO-2482] refactor: Use bulk fetch for accuweather backend [load test: abort]

ncloudioj:
## References

JIRA: [DISCO-2482](https://mozilla-hub.atlassian.net/browse/DISCO-2482)
GitHub: N/A

## Description
This patch uses a Lua script to bulk fetch weather data from Redis. It cuts the Redis lookup calls from 3 to 1 per each request. Local benchmark suggests it has increased ~50% in throughput and ~30% in latency in a high cache-hit scenario (similar to production). It should perform better in other scenarios as well as it reduces the overall overhead for Redis cache lookups.

Main vs This branch
QPS: 1,150 vs 1,750 (52%)
Latency (p95): 110ms vs 77ms (30%)  

Need to add/fix tests

## PR Review Checklist

[DISCO-2482]: https://mozilla-hub.atlassian.net/browse/DISCO-2482?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/merino/cache/redis.py
+++ b/merino/cache/redis.py
@@ -1,9 +1,10 @@
 """Redis cache adapter."""
 
 from datetime import timedelta
-from typing import Optional
+from typing import Any, Optional
 
 from redis.asyncio import Redis, RedisError
+from redis.commands.core import AsyncScript
 
 from merino.exceptions import CacheAdapterError
 
@@ -12,6 +13,7 @@ class RedisAdapter:
     """A cache adapter that stores key-value pairs in Redis."""
 
     redis: Redis
+    scripts: dict[str, AsyncScript] = {}
 
     def __init__(self, redis: Redis):
         self.redis = redis
@@ -52,3 +54,37 @@ class RedisAdapter:
     async def close(self) -> None:
         """Close the Redis connection."""
         await self.redis.close()
+
+    def register_script(self, sid: str, script: str) -> None:
+        """Register a Lua script in Redis. Regist multiple scripts using the same `sid`
+        will overwrite the previous ones.
+
+        Note that script registration is lazy, no network call will be made for this.
+
+        Params:
+            - `sid` {str}, a script identifier
+            - `script` {str}, a Redis supported Lua script
+        """
+        self.scripts[sid] = self.redis.register_script(script)
+
+    async def run_script(self, sid: str, keys: list[str], args: list[str]) -> Any:
+        """Run a given script with keys and arguments.
+
+        Params:
+            - `sid` {str}, a script identifier
+            - `keys` list[str], a list of keys used as the global `KEYS` in Redis scripting
+            - `args` list[str], a list of arguments used as the global `ARGV` in Redis scripting
+        Returns:
+            A Reids value based on the return value of the specified script
+        Raises:
+            - `CacheAdapterError` if Redis returns an error
+            - `KeyError` if `sid` does not have a script associated
+        """
+        try:
+            res = await self.scripts[sid](keys, args)
+        except RedisError as exc:
+            raise CacheAdapterError(
+                f"Failed to run script {id} with error: `{exc}`"
+            ) from exc
+
+        return res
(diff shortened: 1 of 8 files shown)


merino/providers/weather/backends/accuweather.py
```diff
@@ -119,9 +154,24 @@ def cache_key_for_accuweather_request(
...
 
-        return f"{self.__class__.__name__}:v2:{url}"
+    @functools.cache
+    def cache_key_template(self, dt: WeatherDataType) -> str:
+        """Get the cache key template for weather data."""
+        match dt:
+            case WeatherDataType.CURRENT_CONDITIONS:
+                return self.cache_key_for_accuweather_request(
+                    self.url_current_conditions_path,
+                    query_params={self.url_param_api_key: self.api_key},
```
quiiver, line 168: Can we put the query params into a variable so they stay in sync across the two different keys?

merino/providers/weather/backends/accuweather.py
```diff
@@ -34,6 +30,36 @@
...
+#   - If the location key is missing, it will return an empty array
+LUA_SCRIPT_CACHE_BULK_FETCH: str = """
+    local location_key = redis.call("GET", KEYS[1])
+
+    if not location_key then
+        return {}
+    end
+
+    local key = cjson.decode(location_key)["key"]
+    local condition_key = string.gsub(ARGV[1], "{location_key}", key)
```
quiiver, line 53: What do you think about moving the string `{location_key}` to a script argument? It took a bit of digging to figure out where that string is defined in the configs. moving it out to the script call site would make it a little clearer imo. open to suggestions. maybe a comment is fine as well.

merino/cache/redis.py
```diff
@@ -52,3 +54,37 @@ async def set(
...
+
+        Note that script registration is lazy, no network call will be made for this.
+
+        Params:
+            - `sid` {str}, a script identifier
+            - `script` {str}, a Redis supported Lua script
+        """
+        self.scripts[sid] = self.redis.register_script(script)
+
+    async def run_script(self, sid: str, keys: list[str], args: list[str]) -> Any:
```
quiiver, line 70: Is there a constrained set of types that can be returned from this? if so would it make sense to use a TypeVar with constraints?

quiiver (approved):
Thanks for this, I don't have any blocking concerns. I do worry a little bit about the extra level of abstraction introduced by the script. however the performance benefits make this change feel worth that indirection.

merino/providers/weather/backends/accuweather.py
```diff
@@ -34,6 +30,36 @@
...
+#     if they are not present in the cache
+#   - If the location key is missing, it will return an empty array
+LUA_SCRIPT_CACHE_BULK_FETCH: str = """
+    local location_key = redis.call("GET", KEYS[1])
+
+    if not location_key then
+        return {}
+    end
+
+    local key = cjson.decode(location_key)["key"]
```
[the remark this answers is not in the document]
ncloudioj: Upon malformed data, it will raise a `RedisError` which will be captured by `get_weather_report()` with error logging and metrics. Let me add some comments to note this.

merino/providers/weather/backends/accuweather.py
ncloudioj: Good point. Will add a comment for this as well.

merino/cache/redis.py
ncloudioj: Unfortunately, the Redis client also uses `Any` as the return type, which means it returns a value of any type depending on what your script returns 🤷‍♂️.

merino/providers/weather/backends/accuweather.py
ncloudioj: Sending this placeholder as an argument.

PR #395, opened 2023-08-03
feat: factor out CSV remote settings uploader

0c0w3:
## Description
Factor out a generic CSV remote settings uploader job from the Pocket uploader. We need to upload MDN suggestions and instead of adding another job for that, let's make a generic uploader job.

We still want validation of the input data. To accomplish that, I added modules that implement Pydantic models for Pocket and MDN. Any time we add a new type of suggestion for remote settings, we can add a new model module.

I updated jobs.md with info on how to run it. Hopefully it also provides a good explanation of how this all works.

## PR Review Checklist


merino/jobs/csv_rs_uploader/base.py
```diff
@@ -0,0 +1,53 @@
...
+        equivalents that users are more likely to type, leading and trailing
+        whitespace is stripped, and all whitespace is replaced with spaces and
+        collapsed. Subclasses may call this method to validate keywords in their
+        models.
+        """
+        value = value.lower()
+        value = re.sub(r"\s+", " ", value)
+        value = re.sub(r"’", "'", value)
+        keywords = [
+            *filter(
```
ncloudioj, line 46:
```suggestion
            filter(
```

ncloudioj (approved):
Looks great!

merino/jobs/csv_rs_uploader/base.py
0c0w3: Thanks Nan, that ends up putting a filter object inside the array instead of the keywords themselves. So I'll leave this as is but let me know if I'm misunderstanding.

ncloudioj: Oops, sorry, I guess it should be `list(filter(...))`. Just a nit, feel free to disregard this.

PR #399, opened 2023-08-10
[DISCO-2575] refactor: Perform fine-grained cache refreshing for accuweather backend [load test: abort]

ncloudioj:
## References

JIRA: [DISCO-2575](https://mozilla-hub.atlassian.net/browse/DISCO-2575)

## Description
Merino currently refreshes "Current Conditions" and "Forecast" at the same time regardless of its cache validity, this patch implements fine-grained refreshing so that it only refreshes the cached data if they're expired. 

## PR Review Checklist

[DISCO-2575]: https://mozilla-hub.atlassian.net/browse/DISCO-2575?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/docs/operations/configs.md
+++ b/docs/operations/configs.md
@@ -212,7 +212,7 @@ Configuration for using AccuWeather as a weather backend
   - `partner_code` (`MERINO_ACCUWEATHER__PARTNER_CODE`) -
     The partner code to append to URLs in the current conditions and forecast responses.
 
-### AMO API 
+### AMO API
 
 - `api_url` (`MERINO_AMO__DYNAMIC__API_URL`) - the base URL of the Addons API.
 
@@ -260,10 +260,6 @@ These are production providers that generate suggestions.
   - `query_timeout_sec` (`MERINO_PROVIDERS__ACCUWEATHER__QUERY_TIMEOUT_SEC`) - A floating
     point (in seconds) indicating the maximum waiting period when Merino queries
     for weather forecasts. This will override the default query timeout.
-  - `cached_report_ttl_sec` (`MERINO_PROVIDERS__ACCUWEATHER__CACHED_REPORT_TTL_SEC`) - The
-    number of whole seconds (as an integer) indicating the time-to-live for a cached weather
-    report (current conditions and forecast) for a location. After the TTL expires, the provider
-    will fetch and cache the report again on the next suggestion request for that location.
 
 #### AMO Provider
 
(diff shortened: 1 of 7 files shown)


docs/operations/configs.md
```diff
@@ -260,10 +260,6 @@ These are production providers that generate suggestions.
   - `query_timeout_sec` (`MERINO_PROVIDERS__ACCUWEATHER__QUERY_TIMEOUT_SEC`) - A floating
     point (in seconds) indicating the maximum waiting period when Merino queries
     for weather forecasts. This will override the default query timeout.
-  - `cached_report_ttl_sec` (`MERINO_PROVIDERS__ACCUWEATHER__CACHED_REPORT_TTL_SEC`) - The
```
ncloudioj, line 263: Didn't add new configurations here as we're looking to drop this doc.

merino/configs/default.toml
```diff
@@ -95,7 +95,9 @@ cache = "none"
 enabled_by_default = false
 score = 0.3
 query_timeout_sec = 5.0
-cached_report_ttl_sec = 2700 # 45 mins.
+cached_location_key_ttl_sec = 604800 # 7 days
+cached_current_condition_ttl_sec = 1800 # 1/2 hr
+cached_forecast_ttl_sec = 3600 # 1 hr
```
ncloudioj, line 101: All the above are set empirically. We can see how they perform in practice and tune them as needed.

merino/providers/weather/backends/accuweather.py
```diff
@@ -375,38 +380,61 @@ async def get_weather_report(
...
-        # Fetch current conditions and forecast from AccuWeather.
-        # To simplify the implementation we always fetch both as they share the
-        # same cache TTL now.
-        try:
-            async with asyncio.TaskGroup() as tg:
-                task_current = tg.create_task(self.get_current_conditions(location.key))
-                task_forecast = tg.create_task(self.get_forecast(location.key))
-        except ExceptionGroup as e:
-            raise AccuweatherError(f"Failed to fetch weather report: {e.exceptions}")
+        if current_conditions is None and forecast is None:
```
linabutler, line 412: thought: It would be so nice if `asyncio` had a helper that could wrap a value in an awaitable, like Twisted's `defer.succeed` or JavaScript's `Promise.resolve`, so you could write this block as something like:

```py
async with asyncio.TaskGroup() as tg:
    awaitable_current = tg.create_task(...) if current_conditions is None else as_awaitable(current_conditions)
    awaitable_forecast = tg.create_task(...) if forecast is None else as_awaitable(forecast)

current_conditions = await awaitable_current
forecast = await awaitable_forecast
```

...But it looks like we'd have to write that helper ourselves:

```py
async def as_awaitable(value):
    return value
```

linabutler (approved):
Cool!

merino/providers/weather/backends/accuweather.py
ncloudioj: I like this. It makes the code much cleaner with perhaps negligible overhead on `TaskGroup` and the extra coroutine. Let's do this!

merino/cache/redis.py
```diff
@@ -45,7 +45,9 @@ async def set(
             - `CacheAdapterError` if Redis returns an error.
         """
         try:
-            await self.redis.set(key, value, ex=ttl.seconds if ttl else None)
+            await self.redis.set(
+                key, value, ex=ttl.days * 86400 + ttl.seconds if ttl else None
```
ncloudioj, line 49: This is needed to support more-than-24hr TTLs.

PR #414, opened 2023-09-01
[DISCO-2586] Navigational Suggestions Top Picks file generation job

taddes:
## References

JIRA: [DISCO-2586](https://mozilla-hub.atlassian.net/browse/DISCO-2586)

## Description
Much like the WIkipedia Indexing job, we have a Top Picks file CLI job that runs to create suggestions for merino. This file is then indexed to have suggestions to match against. We should document the CLI job so that other developers can run the job in the future, in case data eng (which currently runs the job) ends up not doing so, or in the scenario where we need to make a rapid update.

## PR Review Checklist

[DISCO-2586]: https://mozilla-hub.atlassian.net/browse/DISCO-2586?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


mtbottle (approved):
LGTM! The steps in the instructions are clear and each step is actionable. Normally I would be against having 2 sets of instructions to do the same thing, but given how un-intuitive the UI is, I think it's OK to leave in both for now. When we get a better sense of which one is easier to work with, we can keep just that one as the recommended instructions. Having the images annotated is an excellent touch! Well done :shipit:

PR #415, opened 2023-09-05
fix to resolve kinto key error for load tests [load test: abort]

taddes:
## References
JIRA: [DISCO-2534](https://mozilla-hub.atlassian.net/browse/DISCO-2534)

## Description
Related to DISCO-2534.

Fix for production env of locust where change could not be verified locally. `KeyError` occurred from `KINTO__SERVER_URL` not being found. 

Due to a change to enforce stricter typing of imported enviornment variables, `os.environ["KEY"]` replaced a call to `os.getenv("KEY")`. This was done as `getenv` can return `None`, which did not match the class signatures which these variables are passed into. 

Upon further investigation, it became clear that in `remotesettings.py`, a check is done at the creation of `RemoteSettingsBackend` (see [HERE](https://github.com/mozilla-services/merino-py/blob/28b3775ed9a0dc9ba446d1627149255693afd1d4/merino/providers/adm/backends/remotesettings.py#L37-L57)) to see if either server, collection, bucket or bucket are falsy. 

Therefore, it made sense to update the `RemoteSettingsBackend` signature to allow `None`, which resolved the mypy typing issue, as this will dependably raise a `ValueError` in any environment. Then, updating `get_adm_queries` to accept the same env variables now with possible `None` will handle error cases without causing a `KeyError` during load tests.

## PR Review Checklist

[DISCO-2534]: https://mozilla-hub.atlassian.net/browse/DISCO-2534?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/tests/load/locustfiles/locustfile.py
+++ b/tests/load/locustfiles/locustfile.py
@@ -82,14 +82,14 @@ MERINO_PROVIDERS__WIKIPEDIA__ES_API_KEY: str | None = os.getenv(
 MERINO_PROVIDERS__WIKIPEDIA__ES_INDEX: str | None = os.getenv(
     "MERINO_PROVIDERS__WIKIPEDIA__ES_INDEX"
 )
-MERINO_REMOTE_SETTINGS__SERVER: str = os.getenv(
-    "MERINO_REMOTE_SETTINGS__SERVER", default=os.environ["KINTO__SERVER_URL"]
+MERINO_REMOTE_SETTINGS__SERVER: str | None = os.getenv(
+    "MERINO_REMOTE_SETTINGS__SERVER", os.getenv("KINTO__SERVER_URL")
 )
-MERINO_REMOTE_SETTINGS__BUCKET: str = os.getenv(
-    "MERINO_REMOTE_SETTINGS__BUCKET", default=os.environ["KINTO__BUCKET"]
+MERINO_REMOTE_SETTINGS__BUCKET: str | None = os.getenv(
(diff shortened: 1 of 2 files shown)


linabutler (approved):
Oops 😬 Thanks!

tests/load/locustfiles/locustfile.py
```diff
@@ -82,14 +82,14 @@
 MERINO_PROVIDERS__WIKIPEDIA__ES_INDEX: str | None = os.getenv(
     "MERINO_PROVIDERS__WIKIPEDIA__ES_INDEX"
 )
-MERINO_REMOTE_SETTINGS__SERVER: str = os.getenv(
-    "MERINO_REMOTE_SETTINGS__SERVER", default=os.environ["KINTO__SERVER_URL"]
```
linabutler, line 86: Oops, I'm so sorry I didn't look closely enough at your original change—I see that this will still look up `KINTO__SERVER_URL` even if `MERINO_REMOTE_SETTINGS__SERVER` is set. Thanks for double-checking that these are already optional in `RemoteSettingsBackend`, and apologies for the runaround!

taddes: Not at all! This ends up being a cleaner solution all around anyways! Appreciate the quick review!

PR #418, opened 2023-09-08
[DISCO-2598] Add some extra logging to AMO request errors

mtbottle:
## References

JIRA: [DISCO-2598](https://mozilla-hub.atlassian.net/browse/DISCO-2598)

## Description
This is to help us hopefully get more information on what is causing the errors for AMO requests.

## PR Review Checklist

[DISCO-2598]: https://mozilla-hub.atlassian.net/browse/DISCO-2598?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/merino/providers/amo/backends/dynamic.py
+++ b/merino/providers/amo/backends/dynamic.py
@@ -46,6 +46,13 @@ class DynamicAmoBackend:
         """Fetch addon metadata via AMO. Return `None` on errors."""
         try:
             res = await client.get(f"{self.api_url}/{addon_key}", follow_redirects=True)
+        except httpx.HTTPError as e:
+            logger.error(
+                f"Addons API failed request to fetch addon: {addon_key}, {e}, {e.__class__}"
+            )
+            return None
+
+        try:
             res.raise_for_status()
 
             json_res = res.json()
@@ -55,7 +62,9 @@ class DynamicAmoBackend:
                 "number_of_ratings": json_res["ratings"]["count"],
             }
         except httpx.HTTPError as e:
-            logger.error(f"Addons API could not find key: {addon_key}, {e}")
+            logger.error(
+                f"Addons API could not find key: {addon_key}, {e}, {e.__class__}"
+            )
         except (KeyError, JSONDecodeError):
             logger.error(
                 "Problem with Addons API formatting. "
--- a/tests/unit/providers/amo/backends/test_dynamic.py
+++ b/tests/unit/providers/amo/backends/test_dynamic.py
@@ -1,6 +1,7 @@
 """Unit tests for the Addon API provider backend"""
 import json
 
+import httpx
 import pytest
 from _pytest.logging import LogCaptureFixture
 from httpx import AsyncClient, Request, Response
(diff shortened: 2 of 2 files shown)


merino/providers/amo/backends/dynamic.py
```diff
@@ -46,6 +46,13 @@ async def _fetch_addon(
         """Fetch addon metadata via AMO. Return `None` on errors."""
         try:
             res = await client.get(f"{self.api_url}/{addon_key}", follow_redirects=True)
+        except httpx.HTTPError as e:
+            logger.error(
+                f"Addons API failed request to fetch addon: {addon_key}, {e}, {e.__class__}"
+            )
+            return None
+
```
taddes, line 54: Great addition to determine whether it's the API request or the key!

In looking at this, I'm not sure I see the value in `e.__class__`.  We will get the class as a part of `e`, so in the case of the tests, we get:
 `TimeoutException('timedout!')` for `e`
 `<class 'httpx.TimeoutException'>` for `e.__class__`. 

Given it's kind of the same I'm not sure we need it, since the except statements will trigger from something that's a descendant of `httpx.HTTPError`, so I don't know if it would help all that much. What do you think? 

```suggestion
        except httpx.HTTPError as e:
            logger.error(
                f"Addons API failed request to fetch addon: {addon_key}, {e}"
            )
            return None

```

merino/providers/amo/backends/dynamic.py
```diff
@@ -55,7 +62,9 @@ async def _fetch_addon(
                 "number_of_ratings": json_res["ratings"]["count"],
             }
         except httpx.HTTPError as e:
-            logger.error(f"Addons API could not find key: {addon_key}, {e}")
+            logger.error(
+                f"Addons API could not find key: {addon_key}, {e}, {e.__class__}"
+            )
```
taddes, line 67: Same as above. 
`HTTPStatusError("Client error '400 Bad Request' for url 'https://addons.mozilla.org/api/v5/addons/addon/video-downloadhelper'\nFor more information check: https://httpstatuses.com/400")` for `e`

`<class 'httpx.HTTPStatusError'>` for `e.__class__`


```suggestion
            logger.error(
                f"Addons API could not find key: {addon_key}, {e}"
            )
```

taddes (changes requested):
Hey @mtbottle ! Great addition of the second try/except arm, I think that should help clarify exactly which error is at the root of thigns!  I just added two comments regarding `e.__class__` not necessarily being that useful given that context is provided in `e`. More in the comments.

This should help narrow things down in Sentry for sure, great stuff 👍

merino/providers/amo/backends/dynamic.py
mtbottle: Ah, so the current problem is that our errors are logging empty "e". So currently, the only thing we see in the logs are literally: `Addons API could not find key: read-aloud, ` etc. I was hoping that adding the class will let me know which subclass is causing the error to help with investigating the actual issue (as `HttpError` [is really broad](https://www.python-httpx.org/exceptions/)). The only thing that I'm pretty sure it's not is `HTTPStatusError` (not 100, but that _should_ log something... usually some sort of request code in the message).

merino/providers/amo/backends/dynamic.py
mtbottle: Same as above. This is mainly logging a sanity check, as this _should_ be `HttpStatusError`. It would be interesting to see if this catches any other types of errors as well.

taddes (approved):
Hey @mtbottle ! After our discussion on this, I poked around a bit to see if there might be some other useful things to throw at logging. Here might be some interesting possibilities to provide the desired sanity check, so long as we can get an error payload at all from HTTPX. I looked at their docs and there's pretty scant details on error handling in general, aside from all the inherited types. Happy to look at this with you more this week, too!

`e.__repr__()` -> TimeoutException('timedout!') 
Within the `except`, before the log emission, use `e.add_note("note here")` and you can access via `e.__dict__`, providing `{'_request': None, '__notes__': ['test']}`, if it helps provide context as to which error is triggering the issue.

PR #420, opened 2023-09-14
Increase the timeout for AMO suggestions

mtbottle:
## References

JIRA: [DISCO-2598](https://mozilla-hub.atlassian.net/browse/DISCO-2598)

## Description
Currently we're seeing a lot of `ConnectTimeout` for AMO requests. We're overriding the default timeout of 1s and increasing it to see if this will fix our timeout issues. Since this is a background job, the side effects of this is that start up might be a bit slower.

## PR Review Checklist

### DIFF
--- a/merino/providers/amo/backends/dynamic.py
+++ b/merino/providers/amo/backends/dynamic.py
@@ -11,6 +11,8 @@ from merino.providers.amo.addons_data import ADDON_DATA, SupportedAddon
 from merino.providers.amo.backends.protocol import Addon, AmoBackendError
 from merino.utils.http_client import create_http_client
 
+AMO_CONNECT_TIMEOUT = 5.0
+
 logger = logging.getLogger(__name__)
 
 
@@ -82,7 +84,10 @@ class DynamicAmoBackend:
         tasks: list[Task] = []
 
         try:
-            async with (create_http_client() as client, TaskGroup() as group):
+            async with (
+                create_http_client(connect_timeout=AMO_CONNECT_TIMEOUT) as client,
+                TaskGroup() as group,
+            ):
                 for addon_key in SupportedAddon:
                     tasks.append(
                         group.create_task(


merino/providers/amo/backends/dynamic.py
```diff
@@ -11,6 +11,8 @@
 from merino.providers.amo.backends.protocol import Addon, AmoBackendError
 from merino.utils.http_client import create_http_client
 
+AMO_CONNECT_TIMEOUT = 5.0
```
taddes, line 14: nit: let's perhaps add the `float` type annotation?

```suggestion
AMO_CONNECT_TIMEOUT: float = 5.0
```

ncloudioj (approved):
LGTM.

taddes (approved):
Hey @mtbottle ! Let's hope this resolves all the weird AMO issues. Good call!

Approving, but had one nit with a type annotation (see comment) and a thought: how about defining the value for `AMO_CONNECT_TIMEOUT` in the `default.toml` along with our other constants and config vars? Assigning that const from accessing it via `from merino.config import settings`.

This way it's in one place along with our other constants and we don't have to search for it if we need to change it. We can also ensure a little more type safety and place some constraints on it via Dynaconf. What do you think?

The other AMO settings are [here](https://github.com/mozilla-services/merino-py/blob/bdfe242809d253ffe43d2293af0782955fc48723/merino/configs/default.toml#L139) under `[default.providers.amo]`, and we'd just need to add the Dynaconf validator in `config.py` [here](https://github.com/mozilla-services/merino-py/blob/bdfe242809d253ffe43d2293af0782955fc48723/merino/config.py#L56). If you also want a different setting in the tests, just consider adding in `testing.toml` too.

mtbottle:
> how about defining the value for AMO_CONNECT_TIMEOUT in the default.toml along with our other constants and config vars? Then assigning that const from accessing it via from merino.config import settings.

This feels like an unnecessary added complexity. The reason for having configurations is usually due to a need to configure it based on environment (i.e. local env vs test vs prod) or have modules that we want to reuse with dependency injection. At the moment, there is no need for this as 1. this is experimental (I'm testing out whether this fixes the issue) and 2. we don't have a real reason for needing to configure this value based on environment or otherwise. Let's only add complexity when we have a real reason to do so.

ncloudioj:
Agreed, I think using a constant is reasonable enough in this case as we won't re-configure this often in the future, nor will we use different timeouts in different envs.

As an aside, I suspect the same timeout might have caused [DISCO-2585](https://mozilla-hub.atlassian.net/browse/DISCO-2585) as docker's network latency is notoriously bad on M1 macbooks.

PR #442, opened 2023-10-22
[DISCO-2634] Disable Provider Disaster Recovery

taddes:
## References

JIRA: [DISCO-2634](https://mozilla-hub.atlassian.net/browse/DISCO-2634)

## Description
Implementation of a circuit-breaker to disable certain providers from initializing that can't be bypassed using normal API request params like `enabled_by_default` and `hidden`

Documentation in disaster recovery runbook updated as well. 

## PR Review Checklist

[DISCO-2634]: https://mozilla-hub.atlassian.net/browse/DISCO-2634?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/merino/providers/manager.py
+++ b/merino/providers/manager.py
@@ -141,7 +141,7 @@ def _create_provider(provider_id: str, setting: Settings) -> BaseProvider:
             raise InvalidProviderError(f"Unknown provider type: {setting.type}")
 
 
-def load_providers() -> dict[str, BaseProvider]:
+def load_providers(disabled_providers_list: list[str]) -> dict[str, BaseProvider]:
     """Load providers from configurations.
 
     Exceptions:
@@ -150,10 +150,6 @@ def load_providers() -> dict[str, BaseProvider]:
     providers: dict[str, BaseProvider] = {}
     for provider_id, setting in settings.providers.items():
         # Do not initialize provider if disabled in config.
-        if (
-            provider_id.lower() in settings.runtime.disabled_providers
-            and provider_id.lower() in [pt.value for pt in ProviderType]
-        ):
-            continue
-        providers[provider_id] = _create_provider(provider_id, setting)
+        if provider_id.lower() not in disabled_providers_list:
+            providers[provider_id] = _create_provider(provider_id, setting)
     return providers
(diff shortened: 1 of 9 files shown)


merino/providers/manager.py
```diff
@@ -148,8 +148,12 @@ def load_providers() -> dict[str, BaseProvider]:
       - `InvalidProviderError` if the provider type is unknown.
     """
     providers: dict[str, BaseProvider] = {}
-
     for provider_id, setting in settings.providers.items():
+        # Do not initialize provider if disabled in config.
+        if (
+            provider_id.lower() in settings.runtime.disabled_providers
+            and provider_id.lower() in [pt.value for pt in ProviderType]
```
mtbottle, line 155: Do we need the check on line 155? The added benefit it provides is that it tries to initialize a badly configured provider that also happens to be "disabled". I'm not sure that's a common case that we should protect for.

To save on the "continue" statement, we can do:

```
if provider_id.lower() not in settings.runtime.disabled_providers:
    providers[provider_id] = _create_provider(provider_id, setting)
```

or something instead.

tests/unit/providers/test_init_providers.py
```diff
@@ -30,6 +35,29 @@ async def test_init_providers() -> None:
...
+    await init_providers()
+
+    providers = load_providers()
+    assert provider in providers.keys()
+
+    monkeypatch.setattr(
+        settings,
+        "runtime",
+        {"disabled_providers": [provider]},
+    )
```
mtbottle, line 55: I would suggest modifying the `load_providers` to pass the `disabled_providers` setting instead of loading the configuration in method. This can help us avoid monkey patching this configuration.

So, something like:

```
def load_providers(disabled_providers_list: list[str]) -> dict[str, BaseProvider]:
     ...
     for provider_id, setting in settings.providers.items():
          if provider_id.lower() not in disabled_providers_list:

                 ...
```

ncloudioj (approved):
LGTM, thanks!

tests/unit/providers/test_init_providers.py
taddes: Great suggestion! I'll implement that and then yes, monkey patching can be avoided as the param takes care of which elements are in the list, which is excellent for testing.

PR #443, opened 2023-10-25
[DISCO-2626] Update merino_job to generate file with _latest prefix, update old file

taddes:
## References

JIRA: [DISCO-2626](https://mozilla-hub.atlassian.net/browse/DISCO-2626)

## Description
The merino_job currently generates a new file with the Unix Epoch timestamp prepending the file. We want to add _latest for easily grabbing the file and to also have functionality to update the existing _latest each time the job runs.

Some improvements were made to the `DomainMetadataUploader` like converting the `_destination_top_pick_name` method to a staticmethod with a parameter for suffix to make it more testable, as well as resolving the rounding issue where a needless trailing `.0` was generated after each epoch timestamp. Also, the timestamp is converted from microseconds to a standard timestamp with a simpler method, being `int(current.timestamp())`, given it is called in the UTC context anyways, avoiding a call to `time`. 

The `update_latest_filename_suffix` method takes in the parameters within the context of a `DomainMetadataUploader` instance, gets the blobs from the bucket, searches for the existing file with a `_latest` suffix and removes that suffix, all before the newest file is uploaded. This ensures that one and only one file contains the `_latest` suffix which makes it easy for the provider backend to ingest. 

Various type annotations added and code test coverage increased. 

Source docs from GCS for this patch:

Copy/Rename/Delete objects in GCS: https://cloud.google.com/storage/docs/copying-renaming-moving-objects#client-libraries 
List objects in GCS: https://cloud.google.com/storage/docs/listing-objects#storage-list-objects-python

## PR Review Checklist

[DISCO-2626]: https://mozilla-hub.atlassian.net/browse/DISCO-2626?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/merino/jobs/navigational_suggestions/domain_metadata_uploader.py
+++ b/merino/jobs/navigational_suggestions/domain_metadata_uploader.py
@@ -17,7 +17,7 @@ class DomainMetadataUploader:
     """Upload the domain metadata to GCS"""
 
     DESTINATION_FAVICONS_ROOT = "favicons"
-    DESTINATION_TOP_PICK_FILE_NAME_SUFFIX = "top_picks.json"
+    DESTINATION_TOP_PICK_FILE_NAME_SUFFIX = "top_picks_latest.json"
 
     bucket_name: str
     storage_client: Client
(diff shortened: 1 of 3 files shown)


merino/jobs/navigational_suggestions/domain_metadata_uploader.py
```diff
@@ -41,31 +39,57 @@ def __init__(
...
+    @staticmethod
+    def _destination_top_pick_name(suffix: str) -> str:
         """Return the name of the top pick file to be used for uploading to GCS"""
         current = datetime.datetime.now()
-        return (
-            str(time.mktime(current.timetuple()) * 1000)
-            + "_"
-            + self.DESTINATION_TOP_PICK_FILE_NAME_SUFFIX
-        )
+        return f"{str(int(current.timestamp()))}_{suffix}"
```
mtbottle, line 59: The timestamp doesn't have to be string-casted:

```suggestion
        return f"{int(current.timestamp())}_{suffix}"
```

merino/jobs/navigational_suggestions/domain_metadata_uploader.py
```diff
@@ -41,31 +39,57 @@ def __init__(
...
+
+    def update_latest_filename_suffix(
```
mtbottle, line 61: nit: how about we rename this `remove_latest_from_all_toppicks_files`? This function is literally renaming all the files with the `latest` suffix to not contain the latest suffix.

mtbottle (approved):
Just a few minor points! Overall LGTM

taddes:
Good suggestions @mtbottle , thanks for looking at it 🚀 !

PR #446, opened 2023-11-03
[DISCO-2622] Top Picks Diff Logic

taddes:
## References

JIRA: [DISCO-2622](https://mozilla-hub.atlassian.net/browse/DISCO-2622)

## Description
We require a tool to verify the changes made between runs of the Navigational Suggestions/Top Picks File.  We will roll this into an automated process that outputs the result in the airflow job completion email. However, we require an ad-hoc means of comparing historical files for auditing and security purposes.

## PR Review Checklist

[DISCO-2622]: https://mozilla-hub.atlassian.net/browse/DISCO-2622?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


ncloudioj:
While it's handy to have the diff information at our disposal, the added complexity and boilerplate are a bit concerning.  I think we'd only care about diff for investigation or other content moderation needs, and those can be done on demand using other existing tools (and we can document it or even provide scripts rather than do it within the Merino job).

For example, we can easily get the diff for domains by downloading the most recent JSON files and doing the following:

```
diff <(jq -r '.domains.[].domain' latest.json | sort) <(jq -r '.domains.[].domain' 2nd-latest.json | sort)
```

What do you all think?

taddes:
Hey @ncloudioj , thanks for looking at it!  Here's my thought: as part of our RRA, it's on our doc to provide a diff summary in the email notification we get when the run jobs. I know that was something that was seen as a really good thing in our meeting. This then takes someone only a few seconds to glance at without additional steps, with the potential for being able to resole any problems with greater speed, without review from another team and without a developer needing to take any additional steps.

It's intention in the RRA is to remove any developer steps required to do quality assurance, especially when it comes to info overload. The previous diffs QA went though with the two JSON files were onerous and time consuming. We don't want to keep going forward having to depend on QA to do ad-hoc checks of our data to determine if we need changes to then react. This way, we have an easily readable file at our disposal with every run of the indexer job that can be accessed via URL, kind of  like the remote settings change summaries we generate and then post in Slack! This also leaves us an easy audit trail, which I also recall was discussed in our RRA as being valuable. Just had to do this recently when QA asked about a domain related to sports betting. 

I'm happy to move this into its own module if that would make you feel better regarding the boilerplate, as I totally get that! Are you ok to see how this works for us for a run or two and we can re-asses from there? I'd just like to see if it meets those needs we determined in the RRA. Nice thing is it's just all in the background in the job!

merino/jobs/navigational_suggestions/__init__.py
```diff
@@ -133,16 +134,47 @@ def prepare_domain_metadata(
...
+    old_top_picks: dict[
+        str, list[dict[str, str]]
+    ] = domain_metadata_uploader.get_latest_file_for_diff(
+        client=domain_metadata_uploader.storage_client
+    )
+    domain_diff = DomainDiff(
+        latest_domain_data=top_picks, old_domain_data=old_top_picks
+    )
+    (
+        categories,
```
ncloudioj, line 153: As we discussed before, let's drop categories here and down below as they're not relevant to the digest.

merino/jobs/navigational_suggestions/domain_metadata_diff.py
```diff
@@ -0,0 +1,129 @@
...
+        self, domain_data: dict[str, list[dict[str, str]]]
+    ) -> list[str]:
+        """Process the domain list and return a list of all second-level domains."""
+        return [entry["domain"] for entry in domain_data["domains"]]
+
+    def process_urls(self, domain_data: dict[str, list[dict[str, str]]]) -> list[str]:
+        """Process the domain list and return a list of all urls."""
+        return [entry["url"] for entry in domain_data["domains"]]
+
+    def process_categories(
```
ncloudioj, line 25: Let's drop this for now to minimize complexity.

taddes:
Hey @ncloudioj ! As we discussed today, I simplified the diff output and moved everything into its own module for ease. Works a lot more cleanly and neatly. Let me know how it looks to you. We now just have a summary of new domains and urls, as well as the subdomains. Categories and unchanged are simply enumerated for informational purposes at top and no longer have any distinct entries.

merino/jobs/navigational_suggestions/domain_metadata_diff.py
```diff
@@ -0,0 +1,129 @@
...
+
+        return distinct_categories
+
+    def check_url_for_subdomain(
+        self, domain_data: dict[str, list[dict[str, str]]]
+    ) -> list[dict[str, str]]:
+        """Compare the domain and url to check if a subdomain occurs immediately
+        after scheme.
+        """
+        url_pattern: str = r"^https?://(www\.)?"
```
ncloudioj, line 47: I'd also recommend dropping these subdomain bits from the diff to make it easier to read.

merino/jobs/navigational_suggestions/domain_metadata_diff.py
```diff
@@ -0,0 +1,129 @@
...
+        unchanged_domains: set[str] = set(new_domains).intersection(
+            set(previous_domains)
+        )
+        # added_domains returns the domains not in the previous `_latest` file
+        added_domains: set[str] = set(new_domains).difference(set(previous_domains))
+        added_urls: set[str] = set(new_urls).difference(set(previous_urls))
+
+        return (categories, unchanged_domains, added_domains, added_urls, subdomains)
+
+    def create_diff_file(
```
ncloudioj, line 86: Nit: since it won't create a file
```suggestion
    def create_diff(
```

merino/jobs/navigational_suggestions/domain_metadata_uploader.py
```diff
@@ -79,6 +80,35 @@ def remove_latest_from_all_top_picks_files(
...
+
+    def upload_diff_file(self, diff_file_content: str) -> Blob:  # pragma: no cover
+        """Upload the top pick diff file to gcs."""
+        bucket = self.storage_client.bucket(self.bucket_name)
+        dst_top_pick_name = DomainMetadataUploader._destination_top_pick_name(
+            suffix=DomainMetadataUploader.DESTINATION_TOP_PICK_FILE_NAME_SUFFIX
+        )
+        diff_file_name: str = f"{dst_top_pick_name}_diff_file.txt"
+        diff_blob = bucket.blob(diff_file_name)
+        diff_blob.upload_from_string(diff_file_content)
```
ncloudioj, line 109: Hmm, why do we want to persist diffs? I think we can attach those to the email without storing them.

taddes:
Thx for the good suggestions @ncloudioj ! Tightened up file and only have urls and subdomains listed, rest is just numerical summary. Took out categories, removed upload of file and passed the diff as Xcom push for airflow job to access.  I did leave the `_latest` rename and capture logic as I have a separate ticket (DISCO-2620) to make that switcheroo and write both the `_latest` file and `<timestamp>_top_picks` file as discussed. Lmk if this looks good. Got the airflow patch here https://github.com/mozilla/telemetry-airflow/pull/1865

merino/jobs/navigational_suggestions/domain_metadata_diff.py
```diff
@@ -0,0 +1,99 @@
...
+    """Process and prepare diff for domain files."""
+
+    latest_domain_data: dict[str, list[dict[str, str]]]
+    old_domain_data: dict[str, list[dict[str, str]]]
+
+    def __init__(self, latest_domain_data, old_domain_data) -> None:
+        self.latest_domain_data = latest_domain_data
+        self.old_domain_data = old_domain_data
+
+    def process_domains(
```
ncloudioj, line 15: nit: this can be a static method as it doesn't use self.

merino/jobs/navigational_suggestions/domain_metadata_diff.py
```diff
@@ -0,0 +1,99 @@
...
+        self, domain_data: dict[str, list[dict[str, str]]]
+    ) -> list[str]:
+        """Process the domain list and return a list of all second-level domains."""
+        return [entry["domain"] for entry in domain_data["domains"]]
+
+    def process_urls(self, domain_data: dict[str, list[dict[str, str]]]) -> list[str]:
```
ncloudioj, line 21: ditto.

merino/jobs/navigational_suggestions/domain_metadata_diff.py
```diff
@@ -0,0 +1,99 @@
...
+        subdomains: list[dict[str, str]] = self.check_url_for_subdomain(new_data)
+
+        unchanged_domains: set[str] = set(new_domains).intersection(
+            set(previous_domains)
+        )
+        # added_domains returns the domains not in the previous `_latest` file
+        added_domains: set[str] = set(new_domains).difference(set(previous_domains))
+        added_urls: set[str] = set(new_urls).difference(set(previous_urls))
+
+        return (unchanged_domains, added_domains, added_urls, subdomains)
```
ncloudioj, line 66: It's perhaps me but I think all those type annotations make this not super pleasant to read. Personally, I'd leave out type hints whenever possible and let the type checker infer that automatically and only explicitly specify them if needed. Such as the following, what do you think?
```suggestion
        old_urls = self.process_urls(old_top_picks)
        new_urls = self.process_urls(new_top_picks)
        old_domains = self.process_domains(old_top_picks)
        new_domains = self.process_domains(new_top_picks)

        unchanged_domains = set(new_domains).intersection(old_domains)
        added_domains = set(new_domains).difference(old_domains)
        added_urls = set(new_urls).difference(previous_urls)

        return (unchanged_domains, added_domains, added_urls)
```

merino/jobs/navigational_suggestions/domain_metadata_diff.py
```diff
@@ -0,0 +1,99 @@
...
+        domains: set[str],
+        urls: set[str],
+        subdomains: list[dict[str, str]],
+    ) -> str:
+        """Create string representation of diff file comaring domain data."""
+        title = f"Top Picks Diff File {file_name}"
+        sep = "=" * 20
+
+        unchanged_summary = f"Total domain suggestions unchanged: {len(unchanged)}"
+        domain_summary = f"""Newly added domains: {len(domains)}\n{sep}\n"""
```
ncloudioj, line 81:
```suggestion
        domain_summary = f"Newly added domains: {len(domains)}\n{sep}\n"
```

merino/jobs/navigational_suggestions/domain_metadata_diff.py
```diff
@@ -0,0 +1,99 @@
...
+        url_summary = f"Newly added urls: {len(urls)}\n{sep}\n"
+        for url in urls:
+            url_summary += f"{url}\n"
```
ncloudioj, line 84:
```suggestion
        url_summary += "\n".join(urls)
```

merino/jobs/navigational_suggestions/domain_metadata_diff.py
```diff
@@ -0,0 +1,99 @@
...
+
+        subdomains_summary = f"Domains containing subdomain: {len(subdomains)}\n{sep}\n"
+        for subdomain in subdomains:
+            subdomains_summary += f"{subdomain}\n"
```
ncloudioj, line 88: ditto.

merino/jobs/navigational_suggestions/domain_metadata_diff.py
```diff
@@ -0,0 +1,99 @@
...
+        file = f"""
+        {title}
+
+        {unchanged_summary}
+        {domain_summary}
+        {url_summary}
+        {subdomains_summary}
+        """.strip()
+
+        return file
```
ncloudioj, line 99:
```suggestion
        return f"""
{title}

{unchanged_summary}
{domain_summary}
{url_summary}
{subdomains_summary}
""".strip()
```

ncloudioj (commented):
We're almost there :)

taddes:
Thx a bunch @ncloudioj , good call on the type annotations, let it infer a little more and just have it in the method signatures. That should cover those revisions!

merino/jobs/navigational_suggestions/domain_metadata_uploader.py
```diff
@@ -79,6 +80,24 @@ def remove_latest_from_all_top_picks_files(
                 bucket.delete_blob(blob.name)
         return
 
+    def get_latest_file_for_diff(
```
ncloudioj, line 83: This can also be simplified as there will be a fixed file `top_picks_latest.json`. Do you want to rename the file in a separate patch?

taddes: Yeah! I've got a separate ticket to do it all at once in merino and here for easier testing and such! 👍 DISCO-2620

PR #448, opened 2023-11-06
[DISCO-2669] update all localhost instances to ipv4 127.0.0.1 [load test: abort]

taddes:
## References

JIRA: [DISCO-2669](https://mozilla-hub.atlassian.net/browse/DISCO-2669)

## Description
I came across an interesting recommendation from an [engineer at Sentry](https://www.youtube.com/watch?v=98SYTvNw1kw) that determined the lax use of localhost can have deleterious results on performance and can cause bugs. See [Sentry repo example here](https://github.com/getsentry/snuba/pull/3736/commits/6f4b041bc3949f16f3fc9f569fa388424b518373)

It turns out that if you default to localhost, there is not a guaranteed translation to the IPv4 127.0.0.1. More often than not, a IPv6 lookup is made first. For example:
```python
import socket
socket.getaddrinfo('localhost', 0)

# result from python 3.11 merino env
[(<AddressFamily.AF_INET6: 30>, <SocketKind.SOCK_DGRAM: 2>, 17, '', ('::1', 0, 0, 0)),
(<AddressFamily.AF_INET6: 30>, <SocketKind.SOCK_STREAM: 1>, 6, '', ('::1', 0, 0, 0)),
(<AddressFamily.AF_INET: 2>, <SocketKind.SOCK_DGRAM: 2>, 17, '', ('127.0.0.1', 0)),
(<AddressFamily.AF_INET: 2>, <SocketKind.SOCK_STREAM: 1>, 6, '', ('127.0.0.1', 0))]
```

In this example, you can see 2 IPv6 lookups were made first.

This ticket prevents this from being an issue on some operating systems, reduces confusion and may provide modest performance gains. Prevents needless system calls.

## PR Review Checklist

[DISCO-2669]: https://mozilla-hub.atlassian.net/browse/DISCO-2669?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/dev/kinto.ini
+++ b/dev/kinto.ini
@@ -18,7 +18,7 @@ kinto.changes.resources = /buckets/main
 kinto.attachment.base_path = /app/attachments
 kinto.attachment.base_url =
 # See uwsgi static-map setting
-kinto.attachment.extra.base_url = http://localhost:8888/attachments
+kinto.attachment.extra.base_url = http://127.0.0.1:8888/attachments
 kinto.attachment.folder = {bucket_id}/{collection_id}
 
 [uwsgi]
--- a/docs/dev/dependencies.md
+++ b/docs/dev/dependencies.md
@@ -59,7 +59,7 @@ $ make docker-compose-up-daemon
 $ make docker-compose-down
 ```
 
-Redis is listening on port 6397 and can be connected via `redis://localhost:6397`.
+Redis is listening on port 6397 and can be connected via `redis://127.0.0.1:6397`.
 
 This Dockerized set up is optional. Feel free to run the dependent services by
 any other means as well.
@@ -69,7 +69,7 @@ any other means as well.
 The docker-compose setup also includes some services that can help during
 development.
 
-- Redis Commander, http://localhost:8081 - Explore the Redis database started
+- Redis Commander, http://127.0.0.1:8081 - Explore the Redis database started
   above.
 
 
--- a/docs/testing/contract-tests/client.md
+++ b/docs/testing/contract-tests/client.md
@@ -127,11 +127,11 @@ tests within a Python virtual environment to prevent dependency cross contaminat
     The following environment variables are set in `docker-compose.yml`, but will
     require local setup via command line, pytest.ini file or IDE configuration:
     * `MERINO_URL`: The URL of the Merino service
-      * Example: `MERINO_URL=http://localhost:8000`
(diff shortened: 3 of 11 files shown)


ncloudioj:
First off, thanks for sharing your findings!

However, I am not convinced by the proposed benefits of this conversion. On the contrary, I think it perhaps would cause more confusion if we disallow the use of `localhost` in Merino. To me, it's crystal clear that when we use `localhost`, it resolves to the IP v4 loopback address `127.0.0.1` and to the IP v6 loopback address `::1`. Why would we force people to use IP v4 in Merino (or prohibit the use of `localhost`)?

When clarity is needed, we specify exactly what we want to use. For instance,
- We use `0.0.0.0` for the [Merino docker image](https://github.com/mozilla-services/merino-py/blob/main/Dockerfile#L60) because we want Merino to be reachable at all the IP addresses on its running host
- We use `127.0.0.1` for the [Nginx config](https://github.com/mozilla-services/cloudops-infra/blob/master/projects/merino/k8s/charts/merinopy/templates/nginx-configmap.yaml#L75) because we run Merino within an IP v4 environment in production.

Does that make sense to you?

taddes:
Hey @ncloudioj ! Thanks for your thoughts. I'm fine if we don't do this in the end, but it's less about disallowing someone using `localhost` and more-so optimizing the lookups on whatever os is running merino. 

It's not the understanding that `localhost` == `127.0.0.1`, but rather the background process of `::1` always resolving first, invoking several needless system calls and connection attempts that can be avoided with this process. The rationale that Andrew Sottile gives is improving clarity for one, removing needless lookups and sys calls for the translation, which in the snippet above shows the IPv6 addresses are called first. He stated at Lyft there was a 2% improvement on latency across the board and other similar improvements in latency at Sentry in their core HTTP client.

I'll leave it up to you, let me know!

ncloudioj:
We don't use `localhost` in production and I don't think that 2% performance gain in other environments even justifies this prohibition. So let's not do that.

taddes:
Ok @ncloudioj , sounds good. Shall I just close this out then or was there any place we wanted to use it in the patch?

ncloudioj:
Let's close it out. Thanks!

PR #449, opened 2023-11-13
[DISCO-2681] fake default backend top picks

taddes:
## References

JIRA: [DISCO-2681](https://mozilla-hub.atlassian.net/browse/DISCO-2681)

## Description
All other providers that present dynamic data have a Fake Provider for testing and in the case of initialization failure.

Now that the Top Picks provider will access data from a remote source, it requires a Fake Backend just like Adm, Wikipedia, and AMO. Want this patch in place so the remote GCS patch can make use of this logic. 

## PR Review Checklist

[DISCO-2681]: https://mozilla-hub.atlassian.net/browse/DISCO-2681?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/merino/providers/top_picks/backends/fake_backends.py
+++ b/merino/providers/top_picks/backends/fake_backends.py
@@ -1,0 +1,11 @@
+"""Test backend for the Top Picks provider."""
+
+from merino.providers.top_picks.backends.protocol import TopPicksData
+
+
+class FakeTopPicksBackend:  # pragma: no cover
+    """A fake test backend that always returns empty results."""
+
+    async def fetch(self) -> TopPicksData:
+        """Get fake Top Picks data."""
+        return TopPicksData()  # type: ignore [call-arg]
--- a/merino/providers/top_picks/provider.py
+++ b/merino/providers/top_picks/provider.py
@@ -7,8 +7,8 @@ import logging
 
 from merino.exceptions import BackendError
 from merino.providers.base import BaseProvider, BaseSuggestion, SuggestionRequest
-from merino.providers.top_picks.backends.protocol import TopPicksBackend, TopPicksData
 from merino.providers.top_picks.backends.fake_backends import FakeTopPicksBackend
+from merino.providers.top_picks.backends.protocol import TopPicksBackend, TopPicksData
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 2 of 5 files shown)


merino/providers/top_picks/provider.py
```diff
@@ -26,7 +27,7 @@ class Provider(BaseProvider):
 
     def __init__(
         self,
-        backend: TopPicksBackend,
+        backend: TopPicksBackend | FakeTopPicksBackend,
```
ncloudioj, line 30: `TopPickBackend` is the protocol for `FakeTopPicksBackend`, so you don't need `FakeTopPicksBackend` here.

merino/providers/top_picks/backends/fake_backends.py
```diff
@@ -0,0 +1,11 @@
+"""Test backend for the Top Picks provider."""
+
+from merino.providers.top_picks.backends.protocol import TopPicksData
+
+
+class FakeTopPicksBackend:  # pragma: no cover
```
ncloudioj, line 6: Did you just want to add a placeholder here? It feels odd to add a fake backend without using it anywhere. Since this seems to be a small change, I'd recommend adding this to the patch that actually uses it. What do you think?

merino/providers/top_picks/provider.py
taddes: Ah, was handling a mypy error hence adding, but I'll remove, I see the mypy ignore is called for the backend in `manager.py`

merino/providers/top_picks/backends/fake_backends.py
taddes: Heya! I was adding this as a seperate PR as the GCS remote one is getting kinda large and I thought it made sense to have it, as using the FakeBackend may be the way to go to allow testability of that feature.  I'm happy to add this in elsewhere then, I just wanted to break things into chunks to make turning on the feature later easier. Lmk!

ncloudioj: Ah, I see. Let's cherrypick this into the GCS patch then. No need for `#praga: no cover` :)

PR #455, opened 2023-11-17
[DISCO-2677] runbooks and images for wiki deletion runbook

taddes:
## References

JIRA: [DISCO-2677](https://mozilla-hub.atlassian.net/browse/DISCO-2677)

## Description
We’d like to have a runbook that explains how to properly delete data in the Wikipedia indexer in the case of an error.

## PR Review Checklist

[DISCO-2677]: https://mozilla-hub.atlassian.net/browse/DISCO-2677?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


docs/operations/jobs/delete-wiki-export-data.md
```diff
@@ -0,0 +1,20 @@
+# Merino Jobs Operations
+
+## Delete Dynamic Wikipedia Data
+
+Wikipedia data is copied to a GCS bucket prior to indexing in the `wikipedia_indexer_copy_export` job. 
+![merino_jobs Wikipedia Indexer Graph View](wiki_graph_view.png "merino_jobs UI Graph View")
+
+There are situations in which you may need to delete the Wikipedia export data for a fresh run. This can be due to a processing error in the Kubernetes cluster, for instance.
```
mtbottle, line 8: Edited for conciseness and removed unnecessary information (i.e. we don't need to know that a Kubernetes cluster created a corrupted file... the user only cares that we may want to delete the file if it's corrupted).

```suggestion
We may want to delete a Wikipedia export file if it is corrupted.
```

docs/operations/jobs/delete-wiki-export-data.md
```diff
@@ -0,0 +1,20 @@
...
+
+There are situations in which you may need to delete the Wikipedia export data for a fresh run. This can be due to a processing error in the Kubernetes cluster, for instance.  
+
+### Deletion Steps
+1. Visit the [Airflow dashboard for `merino_jobs`][merino_jobs-graph].
+2. From the Grid View Tab, Click on the `wikipedia_indexer_copy_export` task
+![merino_jobs Wikipedia Indexer Graph View](wiki-indexer-grid-view.png "merino_jobs UI Grid View")
+3. In the log tab, you can see the name of the file in question in the format `enwiki-<date>-cirrussearch-content.json.gz`. Make note of the file name. If the job failed, a red box will show for the particular job.
+![merino_jobs Wikipedia Indexer Graph View](wiki-log-view.png "merino_jobs UI log view")
+4. Visit the [Google Cloud Console][https://console.cloud.google.com] and search for the  `moz-fx-data-prod-external-data` bucket in `shared-prod`. It will be under `contextual-services/merino-jobs/wikipedia-exports/<filename>`. Find the matching  `enwiki-<date>-cirrussearch-content.json.gz` file, click on it for 'Object details' and then delete it.
```
mtbottle, line 16: It doesn't look like we have delete access to it right now, so I'm wondering if we should add this step if we can't technically run that action. 

Also, it would be useful to point to the actual bucket in the link. We don't have list permissions, so I don't think the bucket is technically "searchable". The bucket link, FWIW is [this](https://console.cloud.google.com/storage/browser/moz-fx-data-prod-external-data/contextual-services/merino-jobs/wikipedia-exports).

mtbottle (commented):
I have a suggestion and question about whether we should add an instruction that we can't technically run. A few options:

1. Not have this runbook at all if we can't technically _fix_ it. Skipping one run for Wikipedia is not that detrimental. We can technically wait for the next processing of that file because Merino will still return Wikipedia results (just one week out of date, on data that isn't really time sensitive)
2. Instead of getting us to delete the file (as we don't have permissions), we can instruct ourselves with creating a ticket to people who will have permissions to do so. Of course, our data might just refresh before someone with access can delete the file, so might not be worth going down this route.
3. Make a code change to have an option to delete the file in GCS if it exists on every job run. Downstream affect of this change is that this will affect _all_ runs, so we might want to investigate how to make this more configurable in Airflow.

docs/operations/jobs/delete-wiki-export-data.md
taddes: Yeah, I wasn't 100% sure if we wanted to put the link as I had thought the same thing, but yes, since it's not searchable at at minimum we need GCS access, totally fine to add it.

taddes:
Hey @mtbottle ! Thanks for the comments. I think for now I can alter the instructions to firstly make a ticket to data eng to delete the file or secondly, wait till the next run. The runs won't be weekly but monthly, so it's possible waiting it out until the next won't be the best idea, though again not a big deal. I can leave all the context for the bucket and file since we'd need to share that anyway with whoever deletes it, should we not be granted access in the end.

Then, when we add whatever logic in airflow to delete the selected data, I can then make the change to the docs to match. This way we at least have a process in place until that's done. As we discussed last week, Airflow isn't super easy to control when it comes to using code to just delete a job here or there. 

Let me know what you think, I've included your suggestions and altered the directions accordingly.

docs/operations/jobs/delete-wiki-export-data.md
```diff
@@ -5,16 +5,16 @@
...
+4. Visit the [Google Cloud Console][https://console.cloud.google.com] and search access the `shared-prod` `moz-fx-data-prod-external-data` bucket [here][https://console.cloud.google.com/storage/browser/moz-fx-data-prod-external-data/contextual-services/merino-jobs/wikipedia-exports;tab=objects?prefix=&forceOnObjectsSortingFiltering=false]. It will be under `contextual-services/merino-jobs/wikipedia-exports/<filename>`. Find the matching  `enwiki-<date>-cirrussearch-content.json.gz` file. You will need this file. 
 ![GCS Wiki Bucket](gcs-wiki-bucket.png "GCS Wiki Object Details")
-
-5. Return to the Airflow console for `merino_jobs` and select 'Clear'. The job will re-run and fresh data will be created.
+5. File a Data Eng ticket to ask someone to delete the file for you [here][https://mozilla-hub.atlassian.net/jira/software/c/projects/DENG/boards/465]. Letting someone know on the team could be beneficial so the issue is on their radar. Make sure to provide the link of the file you took note of above in the ticket. Alternatively, you could wait for the next run, though you may not want to leave the job un-run for a considerable time.
```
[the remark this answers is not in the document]
taddes: @mtbottle  Haha, I was trying to imply 'ask Wil to do it' since he did it last time 😸, as Data Eng staff have the ability to do that without SRE.  Beneficial as in we don't need to wait or ping people we don't know. Plus we don't interact with Data SRE all that often and not sure how they'd be getting requests from us. I can leave it for SRE to delete it, but I know our SRE team doesn't have access to this from when we had the issue. 

I do like the language changes you put above though. I can change that, just let me know how to proceed.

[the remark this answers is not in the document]
taddes: Sounds good, makes sense. We could also move this into a bucket we control, which makes some sense I think. Once we sort out what we're doing with possibly deleting the file at each re-run, we'll have a better idea. I'll move this info out into the internal runbook, link it on the Jira ticket and direct you there, then we can close this out here. Thanks @mtbottle

PR #466, opened 2023-12-11
[DISCO-2620] Update `_latest` logic to write core domain file and timestamp file

taddes:
## References

JIRA: [DISCO-2620](https://mozilla-hub.atlassian.net/browse/DISCO-2620)

## Description
Like the spec for Shepherd storage.py [here](https://github.com/mozilla-services/consvc-shepherd/blob/main/consvc_shepherd/storage.py), we want merino jobs to similarly create a latest file which is always overridden (`top_picks_latest.json`) and used as the single domain source while also writing a second file for our records with the time stamp(`<0123>_top_picks_latest.json`).

This improves reliability and decreases possibility of bottlenecks when future data grows. 

Note: The process will now write a`top_picks_latest.json` file from which data will be read and overwritten as well as a timestamped file `<0123>_top_picks_latest.json` for our records. The timestamped files at each generation will have the `_latest` suffix which makes it simple to capture the value when making a comparison and passing the dated file to the diff email header and logs for auditing. This is preferable to doing any calculations based on the timestamp alone, which would be necessary given we may change the interval at which the job runs. The process then just removes `_latest` from the timestamped file and each time the job runs, the `top_picks_latest.json` file is overridden

Since the `_latest` will now only be removed from the timestamped files via regex and it runs on the job, there isn't a performance concern for merino. 

## PR Review Checklist

[DISCO-2620]: https://mozilla-hub.atlassian.net/browse/DISCO-2620?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/merino/jobs/navigational_suggestions/domain_metadata_uploader.py
+++ b/merino/jobs/navigational_suggestions/domain_metadata_uploader.py
@@ -2,7 +2,6 @@
 import hashlib
 import json
 import logging
-import re
 from datetime import datetime
 from urllib.parse import urljoin
 
@@ -17,7 +16,7 @@ class DomainMetadataUploader:
     """Upload the domain metadata to GCS"""
 
     DESTINATION_FAVICONS_ROOT: str = "favicons"
-    DESTINATION_TOP_PICK_FILE_NAME: str = "top_picks_latest.json"
+    DESTINATION_TOP_PICK_FILE_NAME: str = "top_picks.json"
 
     bucket_name: str
     storage_client: Client
@@ -47,11 +46,6 @@ class DomainMetadataUploader:
         dst_top_pick_name = DomainMetadataUploader._destination_top_pick_name(
             suffix=DomainMetadataUploader.DESTINATION_TOP_PICK_FILE_NAME
         )
-        self.remove_latest_from_all_top_picks_files(
-            bucket_name=self.bucket_name,
-            bucket=bucket,
-            storage_client=self.storage_client,
-        )
         latest_blob = bucket.blob(self.DESTINATION_TOP_PICK_FILE_NAME)
         latest_blob.upload_from_string(top_picks)
         dated_blob = bucket.blob(dst_top_pick_name)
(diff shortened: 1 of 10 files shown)


merino/jobs/navigational_suggestions/domain_metadata_uploader.py
```diff
@@ -38,40 +39,45 @@ def __init__(
...
         bucket: Bucket,
         storage_client: Client,
     ) -> None:
-        """If an existing file with the `_latest` suffix exists, remove it so the
-        most recent file has the suffix.
+        """If an existing file with a timestamp and the `_latest` suffix exists,
+        remove `_latest` so only the most recent file has the suffix.
         """
         blobs = storage_client.list_blobs(bucket_name)
+        file_pattern = re.compile(r"\d+_top_picks_latest\.json")
```
ncloudioj, line 77: You don't need this suffix anymore as you can sort all the blobs by their names and that's the whole point of using creation timestamps as the blob ID.

```
blobs = [blob for blob in client.list_blobs(bucket_name) if blob.name != "top_picks_latest.json")]
most_recent = sorted(blobs, key=lambda blob: blob.name, reverse=True)[0]
```

Likewise, you can get the N most recent files by slicing the return list of `sorted()`

taddes: Indeed @ncloudioj , terrific suggestion. I've altered it to do the above pattern and no more `_latest` suffix is used.

merino/jobs/navigational_suggestions/domain_metadata_uploader.py
```diff
@@ -16,7 +16,7 @@ class DomainMetadataUploader:
     """Upload the domain metadata to GCS"""
 
     DESTINATION_FAVICONS_ROOT: str = "favicons"
-    DESTINATION_TOP_PICK_FILE_NAME_SUFFIX: str = "top_picks_latest.json"
+    DESTINATION_TOP_PICK_FILE_NAME: str = "top_picks.json"
```
ncloudioj, line 19: Let's make it self-explanatory. 
```suggestion
    DESTINATION_TOP_PICK_FILE_NAME: str = "top_picks_latest.json"
```

merino/jobs/navigational_suggestions/domain_metadata_uploader.py
```diff
@@ -38,64 +38,45 @@ def __init__(
...
-            if file.name.endswith("top_picks_latest.json"):
-                data = file.download_as_text()
-                file_contents: dict = json.loads(data)
-                logger.info(f"Domain file {file.name} acquired.")
-
-                return file_contents
+        blobs = [
+            blob
+            for blob in bucket.list_blobs(self.bucket_name)
+            if blob.name != "top_picks.json"
```
ncloudioj, line 71:
```suggestion
            if blob.name != DESTINATION_TOP_PICK_FILE_NAME
```

merino/jobs/navigational_suggestions/domain_metadata_uploader.py
```diff
@@ -38,64 +38,45 @@ def __init__(
...
+        ]
+        most_recent = sorted(blobs, key=lambda blob: blob.name, reverse=True)[0]
+
+        if most_recent:
```
ncloudioj, line 75: Unless `blobs == []`, this should be always true, and the last return statement will be unreachable. You can do this to handle the empty list case:

```
if not blobs:
    return {"domains": []}

most_recent = sorted(...)
data = most_recent.download_as_text()
...
```

ncloudioj (approved):
r+wc.

merino/jobs/navigational_suggestions/domain_metadata_uploader.py
taddes: Ok! I'll change it in the provider too to grab `"top_picks_latest.json"`

taddes: Ah, though we don't want every single timestamped file to have `_latest`, so I'll create a const for `_latest` and add it only to the latest file, since updating `DESTINATION_TOP_PICK_FILE_NAME` will mean all files have `_latest`.

merino/jobs/navigational_suggestions/domain_metadata_uploader.py
```diff
@@ -38,65 +38,46 @@ def __init__(
         self.favicon_downloader = favicon_downloader
 
     def upload_top_picks(self, top_picks: str) -> Blob:
-        """Upload the top pick contents to gcs."""
+        """Upload the top pick contents to GCS.
+        One file is prepended by a timestamp for record keeping,
+        the other file is the latest entry from which data is loaded.
+        """
         bucket = self.storage_client.bucket(self.bucket_name)
         dst_top_pick_name = DomainMetadataUploader._destination_top_pick_name(
```
ncloudioj, line 46: Since the naming convention is much simpler now, I'd recommend defining it inline without having this `_destination_top_pick_name()`. What do you think?

ncloudioj (approved):

merino/jobs/navigational_suggestions/domain_metadata_uploader.py
taddes: I like that 👍

PR #468, opened 2023-12-13
Update method to exclude favicons and match only json files

taddes:
add delimiter to just get files in root and match_glob to only capture json files

## PR Review Checklist


mtbottle (approved):
Giving an approval to get things rolling! The documented use of `delimiter` is kind of different from how we've used it in the past, so we might want to understand why that's the case. However, the risk is low if the airflow job fails again, so we can try out this solution 👍

PR #472, opened 2024-01-08
[DISCO-2704] ADR: Use Testcontainer for Merino

Trinaa:
## References

JIRA: [DISCO-2704](https://mozilla-hub.atlassian.net/browse/DISCO-2704)

## Description

An ADR proposing expanding the scope of integration tests with Testcontainers.

## PR Review Checklist

[DISCO-2704]: https://mozilla-hub.atlassian.net/browse/DISCO-2704?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


docs/adr/0003-test-containers.md
```diff
@@ -0,0 +1,186 @@
...
+
+* **Status:** In Review
+* **Deciders:** Nan Jiang & Katrina Anderson
+* **Date:** 2024-01-08
+
+## Context and Problem Statement
+
+In 2024, it is anticipated that Merino will expand to be consumed by a greater set of Firefox
+surfaces and to include more content providers. This will challenge the current feature test
+strategy, which has shown weakness in detecting incompatibility with third-party integrations.
```
mtbottle, line 11: When I read "third-party integration", my mind goes straight to API dependencies for services outside of Mozilla (like Upday or Accuweather). AFAICT, the proposal doesn't really solve this, as TestContainers really only help solve the dependency issues with _first-party infrastructure_ (I consider Redis that we provision, or GCP bucket that we provision as "first-party").

I'm wondering if we should be more specific about the scope we're covering here in terms of what is actually "third-party".

mtbottle (commented):
👍 I agree with chosen option A! I also have a suggested option which I would also like to put forth. I don't think we should adopt it, but I think it would be good to consider as an opposition to the "test heavy" options. 😄 

Also, I would like to highlight my comments about the context. I'm still kind of uncertain about how TestContainers can solve the issue to managing Third-Party APIs (like Accuweather and Upday... so maybe it's actually out of scope). So, I think it'll be useful to be clear about what _kind_ of integrations we're trying to test for here and what _can't_ be solved with our chosen option (and even with the limitations, why we should adopt the chosen option anyways).

docs/adr/0003-test-containers.md
```diff
@@ -0,0 +1,186 @@
...
+**3. Cost Efficiency**
+
+The worker hours and tooling expenditures associated with the implementation and execution of
+the test strategy should ensure the profitability of Merino.
+
+## Considered Options
+
+* A. Yes. Expand the Scope of Integration Tests Using Dependency Docker Containers (Testcontainers)
+* B. Yes. Reduce the Dependency Overhead in Tests Using Development and Stage Environments
+* C. No. Fulfill the Current Test Strategy with Contract Test Coverage (Status quo)
```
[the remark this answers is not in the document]
ncloudioj: By "third party Integrations", if you meant the 3rd-party API integrations, then I don't think your point is controversial at all. Rather, I think it warrants its own ADR perhaps in the future for us to think about what tests can be further dropped or replaced if we feel good about "testing/fixing" those in production :). I am quite sure that there are many of those cases belonged to that category.

As for other dependencies, I still see tremendous amount of values to test those via integration tests. Things such as testing distributed locks in Redis, storage IOs with GCS, queries in database, etc. I think it will be much easier to do those through testing rather than in production. Those are main concerns that we're trying to solve in this ADR.

docs/adr/0003-test-containers.md
```diff
@@ -0,0 +1,186 @@
...
+  many popular programming languages. There are also a large number of community maintained clients
+  available for popular services such as PostgreSQL, Redis, Elasticsearch, etc.
+* Testcontainers is lightweight and sandboxed, meaning service resources aren't shared and are
+  cleaned up automatically, promoting test isolation and parallelization
+* Docker-compose is also supported by Testcontainers, facilitating use of multiple dependency
+  containers for more complex test cases
+* Testcontainers supports both Python and Rust languages and works well with their respective test
+  frameworks [PyTest][6] and [Cargo-Test][7]
+
+#### Cons
```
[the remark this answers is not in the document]
ncloudioj: True, let's call this out as well.

docs/adr/0003-test-containers.md
Trinaa: Currently, in integration tests we inject mock backends to simulate third-party API dependencies. In theory, using a combination of Testcontainers and Wiremocks for example, we could simulate the API dependencies without the backend injection, this expands the test coverage to our backend layer and the networking communication between us and the third-party API.

Testcontainers wouldn't help to detect contract changes with APIs, but we would be able to test our code and connection logic for them more thoroughly.

I tweaked (see https://github.com/mozilla-services/merino-py/pull/472/commits/1934a707428977119fd740a5073a3e4efd88fd81) the description of Option A a little bit to clarify.

docs/adr/0003-test-containers.md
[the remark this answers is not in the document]
Trinaa: I am in agreement with Nan and think this topic would be the subject of an independent ADR.

For me including this option in this ADR would be disingenuous as it is not something I'd really consider at this time given our bug/incident history and current monitoring setup. We know third party integrations are a point of weakness and tests are an appropriate mitigation in this case.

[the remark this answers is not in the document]
mtbottle: OK, I'm fine with leaving this out.

mtbottle (approved):
Approving to just add some support for the chosen option.

docs/adr/0003-test-containers.md
```diff
@@ -0,0 +1,193 @@
...
+surfaces and to include more content providers. This will challenge the current feature test
+strategy, which has shown weakness in detecting incompatibilities with third-party integrations.
+Examples:
+
+1. [Weather Incident with Redis Integration][1]
+2. [Navigation Suggestions Error with GCP Integration][2]
+
+The current test approach uses a combination of unit, integration, and contract feature tests,
+where third-party integrations such as cloud services, data storage services, and external API
+integrations are [test doubled][3] in the unit and integration tests and emulated with Docker
```
taddes, line 19: I agree here wholeheartedly, there tends to end up being duplication of mocks and fixtures between unit and integration tests, which makes maintenance a bit tricky at times. The lines blur between unit and integration tests, especially when dealing with I/O and requests from APIs/third-parties.

docs/adr/0003-test-containers.md
```diff
@@ -0,0 +1,186 @@
...
+containers in contract tests. While test doubles might be easier to work with, they lack the
+accuracy of working with real dependencies in terms of matching the production environment and
+covering all the integration surfaces and concerns in tests.
+
+Despite the potential to test with third-party integrations in contract tests, developers have
+refrained due to their lack of familiarity with Docker and CI tooling, as well as their belief
+in a poor return on investment for the time and effort required to create and maintain contract
+tests for experimental features.
```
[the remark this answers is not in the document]
taddes: I think this highlights a definite positive outcome from considering this ADR: this could develop better understanding and familiarity with Docker and CI tooling across the team, which broadly speaking is a technological standard across Mozilla and the industry. This has the added benefit of providing value beyond the implementation of this feature with less dependency on SRE even.

docs/adr/0003-test-containers.md
```diff
@@ -0,0 +1,193 @@
...
+
+Given the Merino service context, which has a rapid development pace and a high risk tolerance,
+is there a way to streamline the test strategy while ensuring robustness against third-party
+integrations?
+
+## Decision Drivers
+
+**1. Usability & Skill Transferability**
+
+Ease-of-use is the key criterion when we assess a tool for testing. The test strategy should
```
taddes, line 37: Given some of my research on testcontainers, ease of use is a prime driver of this technology. You have to know a bit about Docker, but compared to spinning up and managing the lifecycle of a container and dealing with whether tests pass, testcontainers provide a lot of abstraction that simplifies maintenance and effort.

docs/adr/0003-test-containers.md
```diff
@@ -0,0 +1,193 @@
...
+* Testcontainers, which has [recently been acquired by Docker][5], is fairly mature and supports
+  many popular programming languages. There are also a large number of community maintained clients
+  available for popular services such as PostgreSQL, Redis, Elasticsearch, etc.
+* Testcontainers is lightweight and sandboxed, meaning service resources aren't shared and are
+  cleaned up automatically, promoting test isolation and parallelization
+* Docker-compose is also supported by Testcontainers, facilitating use of multiple dependency
+  containers for more complex test cases
+* Testcontainers supports both Python and Rust languages and works well with their respective test
+  frameworks [PyTest][6] and [Cargo-Test][7]
+
```
taddes, line 110: I think it may be worth adding the following points given the other pros it can offer:

* Testcontainers has numerous pre-built modules exist to simplify access to services like Redis and Elasticsearch.
* Testcontainers allows the running of complex integration tests locally. Verify your application works with the technologies that you will run in production by running tests against the exact database or external services. 
* Results in tests being more repeatable and reproducable on all environments, whether on local dev machine, other environments or in CI, irrespective of operating system. Helps verify updates to new production images and versions by testing prior to deploying.
* Testcontainers do not require developers to manage the lifecycle of containers. This means a lot less effort is needed to create and tear down containers or deal with possible failures that could leave containers running. Lifecycle management is cumbersome and a deterrent to development, which may explain some of the difficulty with maintaining the contract tests in the past.
* Previous discussions have involved the desire for smoke tests/chaos testing. This may be beyond the scope of this ADR, but smoke and chaos testing are exceedingly simple to implement using test containers to check application health and emulate complex network topologies.  Ex. Toxiproxy (Shopify Docker Image) can be used with Redis to do complex network operations like introducing latency, dropping requests, introduce jitter with minimal developer effort, without having to rely on the platform to introduce these service degradation. Can do it minimally on developer laptop before considerable dev hours.

docs/adr/0003-test-containers.md
```diff
@@ -0,0 +1,193 @@
...
+* Developers will need to understand more about how to configure and work with dependency
+  containers. The development community has many popular services out of the box, but developers
+  would still need to know and do more than what's required when using test doubles
+* It could be challenging to provision test fixtures for the underlying containers. Feeding the
+  fixture data into the containers could be complex
+* Developers need to ensure version consistency across the development, testing, and production
+  environments
+* For third-party API integrations, if the provider doesn't provide a Docker image for their API,
+  Testcontainers alone will not help us much. It's possible to use fake API container generators,
+  but it comes with its own complexities as well
```
taddes, line 125: Adding some cons:

* Would imply the removal of many mocks and fixtures, which would amount to a fair bit of development effort to patch the holes left by removing tests.
* Pushes the testing approach of the application more towards integration testing and away from unit testing. May require reappraising how we consider test coverage.

Possibly removing the con: 
* Developers need to ensure version consistency across the development, testing, and production
  environments
  
  Rationale: I don't necessarily think this is a con, since the versions would be inherently consistent in testcontainers. This means a closer match of development to production. Implementing testcontainers actually simplifies the ease of version consistency across all environments and doesn't reduce it so much.

taddes (approved):

docs/adr/0003-test-containers.md
Trinaa: So many pros! I've tweaked the existing pros to use some of the language and terms you've used in this list. Trying to merge ideas into a concise list.

While I'm always up for discussing smoke and chaos testing, I agree that maybe this ADR is not the place. Trying to keep the Pros and Cons relevant to the problem statement.

docs/adr/0003-test-containers.md
Trinaa: I've included your first con (slightly different wording). Transforming the existing integration tests will be some work indeed.

For the version Con, I interpret this as being specific to integration tests. Right now this is only a concern in contract tests, where we've had run-ins with the Redis version before for example.

Trinaa:
@taddes Thank you for all your ideas and feedback!

docs/adr/0003-test-containers.md
```diff
@@ -0,0 +1,197 @@
...
+  result, we can use real dependencies in Merino's tests as opposed to test doubles
+* Testcontainers allows developers to programmatically manage the lifecycle of containers in the
+  test code. This simplifies its usage for developers, who will not need to run any Docker commands
+  separately for testing
+* Testcontainers, which has [recently been acquired by Docker][5], is fairly mature and supports
+  many popular programming languages. There are also a large number of community maintained clients
+  available for popular services such as PostgreSQL, Redis, Elasticsearch, etc.
+* Testcontainers is lightweight and sandboxed, meaning service resources aren't shared and are
+  cleaned up automatically, promoting test isolation and parallelization
+* Docker-compose is also supported by Testcontainers, facilitating use of multiple dependency
```
quiiver, line 106: I would really avoid this if possible as i think it has the potential for quite complex testing setups.

quiiver (approved):
Generally 👍🏻 I'm a little confused though as the problem statement talks a lot about how 3rd party integration tests are weak and need to be addressed. However, the proposal only addresses a subset of what we're calling 3rd party integrations which are things like database and caching dependencies. That aside I think the proposal is very reasonable. 

**Other concerns but not blockers**
* It would probably help y'all understand the success of the work to define some success criteria before starting.
* Make sure we're measuring test and CI time to completion. I have a unsubstantiated concern that this will increase CI and testing time non-linearly as containers are added.

ncloudioj:
> I'm a little confused though as the problem statement talks a lot about how 3rd party integration tests are weak and need to be addressed. However, the proposal only addresses a subset of what we're calling 3rd party integrations which are things like database and caching dependencies. That aside I think the proposal is very reasonable.

Yup, that (how to test 3rd party API integrations) remains an open question and we should note it in the ADR for future discussions. Personally, I am open to suggestions. We could either try using some generic API mocks (e.g. [WireMock](https://wiremock.org/docs/)) or just keep a thin test layer and handle API errors through other means (like what Michelle suggested in another thread).

> It would probably help y'all understand the success of the work to define some success criteria before starting.

Good point. I think we can come up with a set of success criteria and asset them once we have gather enough learning in the future.

>  Make sure we're measuring test and CI time to completion. I have a unsubstantiated concern that this will increase CI and testing time non-linearly as containers are added.

Definitely, will do.

ncloudioj:
Thanks to all the reviewers for the inputs!

What's next is that DISCO will set some success criteria as well as a milestone for assessment. Let's give it a spin!

PR #492, opened 2024-02-29
refactor: Extract gcp upload logic into a separate module

Herraj:
## References

JIRA: [DISCO-2742](https://mozilla-hub.atlassian.net/browse/DISCO-2742)

## Description
Refactoring and decoupling the GCP logic for uploading files into a separate module.

## PR Review Checklist

[DISCO-2742]: https://mozilla-hub.atlassian.net/browse/DISCO-2742?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/merino/content_handler/models.py
+++ b/merino/content_handler/models.py
@@ -1,0 +1,30 @@
+from abc import ABC, abstractmethod
+
+from pydantic import BaseModel
+
+
+class Image(BaseModel):
+    """Data model for Image contents and associated metadata."""
+
+    content: bytes
+    content_type: str
+
+
+class BaseContentUploader(ABC):
+    """Abstract class for uploading content to GCS."""
+
+    @abstractmethod
+    def upload_content(
+        self, content: str, destination_name: str, content_type: str
+    ) -> str:
+        """Abstract method for uploading content to our GCS Bucket."""
+        ...
+
+    @abstractmethod
+    def upload_image(
+        self,
+        image: Image,
+        destination_name: str,
+    ) -> str:
+        """Abstract method for uploading an image to our GCS Bucket."""
+        ...
--- a/merino/content_handler/models.py
+++ b/merino/content_handler/models.py
@@ -10,7 +10,7 @@ class Image(BaseModel):
 
     content: bytes
     content_type: str = Field(
-        description="Content type of the Image. Can be 'image/png', 'image/jpeg', 'image"
+        description="Content type of the Image. Can be 'image/png', 'image/jpeg', 'image'"
     )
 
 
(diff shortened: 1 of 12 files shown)


mtbottle (commented):
This mostly looks good to me! There's a few things to fix below, but I think this is almost there! :)

mtbottle (approved):
Just a small nit. I'm going to approve it as it generally looks good. Please fix the lint errors before merging.

merino/content_handler/models.py
```diff
@@ -14,13 +16,21 @@ class Image(BaseModel):
         description="Content type of the Image. Can be 'image/png', 'image/jpeg', 'image'"
     )
 
+    @staticmethod
+    def open(bytes_io: BytesIO) -> PILImage:
+        return PILImage.open(bytes_io)
+
```
mtbottle: This doesn't seem like it's being used anywhere?

mtbottle (approved):
Just a note on potentially unused piece of code.

merino/content_handler/models.py
Herraj: This is being used in `domain_metadata_extractor` on line 198. 

I saw the original change where it was using the `PIL`'s `Image.open()`. I basically extracted that into our own `Image` model!

Herraj: Basically, didn't want two `Image` types in the same module i.e PIL Image and our own Image. 

Not sure if I'm breaking any python patterns here 😅

mtbottle: Ah I see! I think that maybe we can make a change to line 282 in `domain_metadata_extractor` to not just pass the content, but the image object itself. Then we can call `image.open()` return the PILImage instead. This makes better use of the `Image` object, than passing in the static content type to this static method I think.

merino/jobs/navigational_suggestions/domain_metadata_extractor.py
```diff
@@ -196,9 +194,8 @@ def _fix_url(self, url: str) -> str:
 
     def _get_favicon_smallest_dimension(self, image: Image) -> int:
         """Return the smallest of the favicon image width and height"""
-        with PILImage.open(BytesIO(image.content)) as img:
-            width, height = img.size
-            return int(min(width, height))
+        width, height = image.open()
```
mtbottle, line 197: The logic here is a bit different? Should this actually be `image.open().size`?

Herraj: yep, good catch! updated it

mtbottle (approved):

PR #498, opened 2024-04-09
[DISCO-2777] Add merino job to upload relevancy data to RS

tiftran:
## References

JIRA: [DISCO-2777](https://mozilla-hub.atlassian.net/browse/DISCO-2777)

## PR Review Checklist

[DISCO-2777]: https://mozilla-hub.atlassian.net/browse/DISCO-2777?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/merino/jobs/relevancy_uploader/__init__.py
+++ b/merino/jobs/relevancy_uploader/__init__.py
@@ -30,14 +30,16 @@ class Category(Enum):
     Finance = 8
     Food = 9
     Government = 10
-    Hobbies = 11
-    Home = 12
-    News = 13
-    RealEstate = 14
-    Society = 15
-    Sports = 16
-    Tech = 17
-    Travel = 18
+    # Disable this per policy consultation
+    # Health = 11
+    Hobbies = 12
+    Home = 13
+    News = 14
+    RealEstate = 15
+    Society = 16
+    Sports = 17
+    Tech = 18
+    Travel = 19
 
 
 RELEVANCY_RECORD_TYPE = "category_to_domains"
@@ -76,14 +78,18 @@ UPLOAD_CATEGORY_TO_R2D2_CATEGORY: dict[str, Category] = {
 }
 
 
-class RelevancyData(BaseModel):
-    """Pydantic base model for remote settings relevancy data."""
+class RelevancyData:
+    """Class to relate to conforming data to remote settings structure."""
 
     @classmethod
     def csv_to_relevancy_data(
         cls, csv_reader
     ) -> defaultdict[Category, list[dict[str, str]]]:
-        """Read CSV file and extract required data for relevancy."""
+        """Read CSV file and extract required data for relevancy in the structure
+        [
+            { "domain" : <base64 string> }
+        ]
+        """
         rows = sorted(csv_reader, key=lambda x: x["categories"])
         data: defaultdict[Category, list[dict[str, str]]] = defaultdict(list)
         for row in rows:
@@ -92,13 +98,10 @@ class RelevancyData(BaseModel):
                 category_mapped = UPLOAD_CATEGORY_TO_R2D2_CATEGORY.get(
                     category, Category.Inconclusive
                 )
-                if category_mapped != Category.Inconclusive:
-                    md5_hash = md5(
-                        row["domain"].encode(), usedforsecurity=False
-                    ).digest()
-                    data[category_mapped].append(
-                        {"domain": base64.b64encode(md5_hash).decode()}
-                    )
+                md5_hash = md5(row["domain"].encode(), usedforsecurity=False).digest()
+                data[category_mapped].append(
+                    {"domain": base64.b64encode(md5_hash).decode()}
+                )
         return data
 
 
@@ -234,7 +237,9 @@ async def _upload_file_object(
     # so we can validate the source data before deleting existing records and
     # starting the upload.
     data = RelevancyData.csv_to_relevancy_data(csv_reader)
-    for category in data:
+    for category, domains in data.items():
+        # TODO: We want to clean this up and no need suggestion_score_fallback
+        #  https://mozilla-hub.atlassian.net/browse/DISCO-2794
         with ChunkedRemoteSettingsRelevancyUploader(
             auth=auth,
             bucket=bucket,
@@ -251,6 +256,5 @@ async def _upload_file_object(
             if delete_existing_records:
                 uploader.delete_records()
 
-            domains = data[category]
             for domain in domains:
                 uploader.add_relevancy_data(domain)
--- a/merino/jobs/relevancy_uploader/__init__.py
+++ b/merino/jobs/relevancy_uploader/__init__.py
@@ -1,5 +1,6 @@
 """CLI commands for the relevancy_csv_rs_uploader module"""
 import asyncio
+import base64
 import csv
 import io
 from collections import defaultdict
@@ -18,17 +19,17 @@ from merino.jobs.relevancy_uploader.chunked_rs_uploader import (
 class Category(Enum):
     """Enum of possible interests for a domain."""
 
-    Animals = 0
-    Arts = 1
-    Autos = 2
-    Business = 3
-    Career = 4
-    Education = 5
-    Fashion = 6
-    Finance = 7
-    Food = 8
-    Government = 9
-    Health = 10
+    Inconclusive = 0
+    Animals = 1
+    Arts = 2
+    Autos = 3
+    Business = 4
+    Career = 5
+    Education = 6
+    Fashion = 7
+    Finance = 8
+    Food = 9
+    Government = 10
     Hobbies = 11
     Home = 12
     News = 13
@@ -37,13 +38,12 @@ class Category(Enum):
     Sports = 16
     Tech = 17
     Travel = 18
-    Inconclusive = 19
 
 
 RELEVANCY_RECORD_TYPE = "category_to_domains"
 
 # Mapping to unify categories across the sources
-MAPPING: dict[str, Category] = {
+UPLOAD_CATEGORY_TO_R2D2_CATEGORY: dict[str, Category] = {
     "Sports": Category.Sports,
     "Economy & Finance": Category.Finance,
     "Ecommerce": Category.Inconclusive,
@@ -55,7 +55,7 @@ MAPPING: dict[str, Category] = {
     "Social Networks": Category.Inconclusive,
     "Instant Messengers": Category.Inconclusive,
     "Business": Category.Business,
-    "Health & Fitness": Category.Health,
+    "Health & Fitness": Category.Inconclusive,
     "Music": Category.Hobbies,
     "Home & Garden": Category.Home,
     "Science": Category.Education,
@@ -89,14 +89,15 @@ class RelevancyData(BaseModel):
         for row in rows:
             categories = row["categories"].strip("[]").split(",")
             for category in categories:
-                category_mapped = MAPPING.get(category, Category.Inconclusive)
+                category_mapped = UPLOAD_CATEGORY_TO_R2D2_CATEGORY.get(
+                    category, Category.Inconclusive
+                )
                 if category_mapped != Category.Inconclusive:
+                    md5_hash = md5(
+                        row["domain"].encode(), usedforsecurity=False
+                    ).digest()
                     data[category_mapped].append(
-                        {
-                            "domain": md5(
-                                row["origin"].encode(), usedforsecurity=False
-                            ).hexdigest()
-                        }
+                        {"domain": base64.b64encode(md5_hash).decode()}
                     )
         return data
 
--- a/merino/jobs/relevancy_uploader/__init__.py
+++ b/merino/jobs/relevancy_uploader/__init__.py
@@ -1,0 +1,255 @@
+"""CLI commands for the relevancy_csv_rs_uploader module"""
+import asyncio
+import csv
+import io
+from collections import defaultdict
+from enum import Enum
+from hashlib import md5
+
+import typer
+from pydantic import BaseModel
+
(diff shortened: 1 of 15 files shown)


merino/jobs/relevancy_uploader/__init__.py
```diff
@@ -0,0 +1,255 @@
...
+    "Education": Category.Education,
+    "Lifestyle": Category.Inconclusive,
+    "Cartoons & Anime": Category.Inconclusive,
+    "Gaming": Category.Hobbies,
+    "Magazines": Category.Inconclusive,
+    "Forums": Category.Inconclusive,
+    "Entertainment": Category.Inconclusive,
+    "Clothing": Category.Fashion,
+    "Weather": Category.Inconclusive,
+    "Government": Category.Government,
```
tiftran, line 75: Took this list from the interest-classification repo and mapped it with the AS classification categories.
Thoughts on how I mapped them, we can def get rid of the inconclusive ones later after the mapping is finalized?

merino/jobs/relevancy_uploader/__init__.py
```diff
@@ -0,0 +1,255 @@
...
+    for category in data:
+        with ChunkedRemoteSettingsRelevancyUploader(
+            auth=auth,
+            bucket=bucket,
+            chunk_size=chunk_size,
+            collection=collection,
+            dry_run=dry_run,
+            record_type=RELEVANCY_RECORD_TYPE,
+            server=server,
+            suggestion_score_fallback=0,
```
tiftran, line 245: this is icky... will need to make changes to `ChunkedRemoteSettingsUploader` to remove this

merino/jobs/relevancy_uploader/__init__.py
```diff
@@ -0,0 +1,255 @@
...
+        for row in rows:
+            categories = row["categories"].strip("[]").split(",")
+            for category in categories:
+                category_mapped = MAPPING.get(category, Category.Inconclusive)
+                if category_mapped != Category.Inconclusive:
+                    data[category_mapped].append(
+                        {
+                            "domain": md5(
+                                row["origin"].encode(), usedforsecurity=False
+                            ).hexdigest()
```
ncloudioj, line 98: Per spec, I believe we want to base64 encode the MD5 hash digest rather than use the hex digest.

merino/jobs/relevancy_uploader/__init__.py
```diff
@@ -0,0 +1,255 @@
...
+        data: defaultdict[Category, list[dict[str, str]]] = defaultdict(list)
+        for row in rows:
+            categories = row["categories"].strip("[]").split(",")
+            for category in categories:
+                category_mapped = MAPPING.get(category, Category.Inconclusive)
+                if category_mapped != Category.Inconclusive:
+                    data[category_mapped].append(
+                        {
+                            "domain": md5(
+                                row["origin"].encode(), usedforsecurity=False
```
ncloudioj, line 97: Let's use the `domain` field rather than `origin`.

merino/jobs/relevancy_uploader/__init__.py
```diff
@@ -0,0 +1,255 @@
...
+
+    @classmethod
+    def csv_to_relevancy_data(
+        cls, csv_reader
+    ) -> defaultdict[Category, list[dict[str, str]]]:
+        """Read CSV file and extract required data for relevancy."""
+        rows = sorted(csv_reader, key=lambda x: x["categories"])
+        data: defaultdict[Category, list[dict[str, str]]] = defaultdict(list)
+        for row in rows:
+            categories = row["categories"].strip("[]").split(",")
```
ncloudioj, line 90: Looks like the `categories` field could be quoted if it has multiple values. Could you verify whether the csv reader would automatically unquote? If not, we will have to strip those quotes as well.

merino/jobs/relevancy_uploader/__init__.py
```diff
@@ -0,0 +1,255 @@
...
+    Inconclusive = 19
+
+
+RELEVANCY_RECORD_TYPE = "category_to_domains"
+
+# Mapping to unify categories across the sources
+MAPPING: dict[str, Category] = {
+    "Sports": Category.Sports,
+    "Economy & Finance": Category.Finance,
+    "Ecommerce": Category.Inconclusive,
```
bendk, line 49: Maybe this could be `Finance`

merino/jobs/relevancy_uploader/__init__.py
```diff
@@ -0,0 +1,255 @@
...
+    Health = 10
+    Hobbies = 11
+    Home = 12
+    News = 13
+    RealEstate = 14
+    Society = 15
+    Sports = 16
+    Tech = 17
+    Travel = 18
+    Inconclusive = 19
```
ncloudioj, line 40: SERP has `inconclusive` set as 0, let's make it the 1st enum variant as well here.

merino/jobs/relevancy_uploader/__init__.py
```diff
@@ -0,0 +1,255 @@
...
+
+
+RELEVANCY_RECORD_TYPE = "category_to_domains"
+
+# Mapping to unify categories across the sources
+MAPPING: dict[str, Category] = {
```
bendk, line 46: Nit: Could you what the keys mean?  Maybe the name could be something like `FOO_CATEGORY_TO_R2D2_CATEGORY` or something like that (filling in the blank for FOO, I've already forgotten where we're getting the categories from :rofl: )

merino/jobs/utils/chunked_rs_uploader.py
```diff
@@ -9,29 +9,31 @@
...
     @property
     def size(self) -> int:
-        return len(self.suggestions)
+        """Return the length of the data."""
+        return len(self.data)
 
 
 class ChunkedRemoteSettingsUploader:
-    """A class that uploads suggestions to remote settings. Suggestions are
+    """A class that uploads data to remote settings. Data is
```
ncloudioj, line 33:
```suggestion
    """A class that uploads data (e.g. suggestions or relevancy inputs) to remote settings. Data is
```

ncloudioj (changes requested):
Looks good, just a few comments on data processing.

merino/jobs/relevancy_uploader/__init__.py
```diff
@@ -0,0 +1,255 @@
...
+    "Forums": Category.Inconclusive,
+    "Entertainment": Category.Inconclusive,
+    "Clothing": Category.Fashion,
+    "Weather": Category.Inconclusive,
+    "Government": Category.Government,
+}
+
+
+class RelevancyData(BaseModel):
+    """Pydantic base model for remote settings relevancy data."""
```
bendk, line 80: I think there should be a Pydantict schema here.

merino/jobs/relevancy_uploader/__init__.py
```diff
@@ -0,0 +1,255 @@
...
+    Arts = 1
+    Autos = 2
+    Business = 3
+    Career = 4
+    Education = 5
+    Fashion = 6
+    Finance = 7
+    Food = 8
+    Government = 9
+    Health = 10
```
ncloudioj, line 31: Per legal's request, let's drop `Health` for now.

merino/jobs/relevancy_uploader/__init__.py
bendk: Also, maybe you could use `json.loads` to parse the categories field.  Just a suggestion, I'm not sure if that simplifies things or not.

merino/jobs/relevancy_uploader/__init__.py
bendk: Some ickiness seems fine for now.  Could you make a bug for the change and add a comment?

tests/unit/jobs/relevancy_uploader/test_chunked_relevancy_rs_uploader.py
```diff
@@ -0,0 +1,189 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+"""Unit tests for chunked_rs_uploader.py."""
```
bendk, line 5: Nit: I don't think this has the unit tests, is it support code?  In any case, please clarify what's here.

bendk (approved):
Looks great to me, although I don't understand everything that's going on here.  I mostly just had some suggestions for the docstrings/comments.  I'll mark this approve, because I don't think I need to re-review those changes.

merino/jobs/relevancy_uploader/__init__.py
ncloudioj: Hmm, "Ecommerce" is more around online commerce / shopping (amazon, ebay, etc.), whereas "Finance" is more general?

Thinking this again, I think we perhaps could leave those inconclusive categories as is and no need to map them if there are no obvious counterparts? Also, we could collect metrics about the inconclusive interest (as opposed to not collect as we planed) from client so that we could get some ideas about those inconclusive interests as well. So this means we could record one of the three types of metrics for each URL:
- A hit with a conclusive interest, which measures the "power" of the conclusive interests. 
- A hit with an inconclusive interest, which measures the "power" of the inconclusive interests. If it's too strong, we could extract more conclusive interests from this group. 
- A miss meaning that this domain is not on the classification list, The complement of this measures the "power" of the entire classification dataset.

Does that make sense?

merino/jobs/relevancy_uploader/__init__.py
ncloudioj: I know that'll be great. But I generated this csv from BigQuery and it didn't serialize the array in JSON but just into a string (or quoted). e.g. either `[Search Engines]` or `"[Search Engines,Information Technology]"`. So I had to do the above hack 😅

merino/jobs/relevancy_uploader/__init__.py
bendk: So, if we're thinking in terms of counters and vectors, this would mean we now track two inconclusive-style counts:
  - Inconclusive: we know that the URL doesn't fall into any of our categories
  - Unknown/Miss: we don't know if the URL falls into any of our categories

Also, you convinced me that ecommerce should map to inconclusive.

merino/jobs/relevancy_uploader/__init__.py
bendk: Ahh, that makes sense.  Nevermind about this one.

tests/unit/jobs/relevancy_uploader/test_chunked_relevancy_rs_uploader.py
tiftran: oopsies typo lemme fix

merino/jobs/relevancy_uploader/__init__.py
ncloudioj: Yup, two inconclusive counts. Take the 4K dataset as an example,

* Inconclusive (Others): the domain is on the 4K list, but it doesn't fall into any known categories
* Unknown: the domain is not on the list. This can be derived by (`total_input_urls - total_inconclusive_urls - total_conclusive_urls`)

This requires us to send all 4K domains to the client. Whereas if we don't care about the inconclusive count, we don't need to do that.

I am leaning towards including `total_inconclusive_urls`.

WDYT?

bendk: I like it it, let's to it.  For one thing, I think it will lead to better metrics.

merino/jobs/relevancy_uploader/__init__.py
tiftran: went with `UPLOAD_CATEGORY_TO_R2D2_CATEGORY`

merino/jobs/relevancy_uploader/__init__.py
tiftran: I believe there should be in the future when/if things get more complex, but considering we only have domains at the moment, it seemed a little too heavy handed to me. Currently the class only houses the csv -> data conversion method. I could have it not inherit the Pydantic base model?

merino/jobs/relevancy_uploader/__init__.py
```diff
@@ -0,0 +1,256 @@
...
+    dry_run: bool,
+    server: str,
+):
+    csv_reader = csv.DictReader(file_object)
+
+    # Generate the full list of domains before creating the chunked uploader
+    # so we can validate the source data before deleting existing records and
+    # starting the upload.
+    data = RelevancyData.csv_to_relevancy_data(csv_reader)
+    for category in data:
```
ncloudioj, line 237: Nit: at a first glance, I thought `data` was a `list` but it was actually a `defaultdict`. We could do the good old dict iteration here. 

```
for category, domains in data.items():
    # replace all the `data[category]` below with `domains`
```

tests/unit/jobs/relevancy_uploader/test_relevancy_data.csv
```diff
@@ -0,0 +1,4 @@
+rank,domain,host,origin,suffix,categories
+0,sports.com,sports0.sports.com,https://sports0.sports.com,com,[Sports]
+1,sports.com,sports1.sports.com,https://sports1.sports.com,com,[Sports]
+2,sports.com,sports2.sports.com,https://sports2.sports.com,com,[Sports]
```
ncloudioj, line 4: Can we add a few more test records to cover the cases of:
- A multi-category domain, e.g. `[Sports,News]`
- A domain of the `Inconclusive` category

ncloudioj (approved):
r+wc! Thanks!

PR #505, opened 2024-04-18
[DISCO-2781] set http cache-control header for weather response

Herraj:
## References

## PR Review Checklist

### DIFF
--- a/merino/configs/default.toml
+++ b/merino/configs/default.toml
@@ -39,6 +39,9 @@ query_timeout_sec = 0.2
 # Prevents a provider from being instantiated.
 disabled_providers = []
 
+# MERINO_RUNTIME__DEFAULT_SUGGESTIONS_REQUEST_TTL_SEC
+default_suggestions_request_ttl_sec = 300 # 5 mins
+
 
 [default.logging]
 # MERINO_LOGGING__LEVEL
--- a/merino/configs/default.toml
+++ b/merino/configs/default.toml
@@ -39,8 +39,8 @@ query_timeout_sec = 0.2
 # Prevents a provider from being instantiated.
 disabled_providers = []
 
-# MERINO_RUNTIME__DEFAULT_SUGGESTIONS_REQUEST_TTL_SEC
-default_suggestions_request_ttl_sec = 300 # 5 mins
+# MERINO_RUNTIME__DEFAULT_SUGGESTIONS_RESPONSE_TTL_SEC
+default_suggestions_response_ttl_sec = 300 # 5 mins
 
 
 [default.logging]
--- a/merino/providers/weather/backends/accuweather.py
+++ b/merino/providers/weather/backends/accuweather.py
@@ -134,7 +134,6 @@ class AccuweatherBackend:
     cached_location_key_ttl_sec: int
     cached_current_condition_ttl_sec: int
     cached_forecast_ttl_sec: int
-    cached_default_weather_report_ttl_sec: int
     metrics_client: aiodogstatsd.Client
     url_param_api_key: str
     url_postalcodes_path: str
--- a/merino/providers/weather/backends/accuweather.py
+++ b/merino/providers/weather/backends/accuweather.py
@@ -6,7 +6,7 @@ import hashlib
 import json
 import logging
 from enum import Enum
-from typing import Any, Callable, NamedTuple
+from typing import Any, Callable, NamedTuple, cast
 
 import aiodogstatsd
 from dateutil import parser
@@ -29,7 +29,7 @@ logger = logging.getLogger(__name__)
 PARTNER_PARAM_ID: str | None = settings.accuweather.get("url_param_partner_code")
 PARTNER_CODE: str | None = settings.accuweather.get("partner_code")
 
-# The Lua script to fetch the location key, current condition, and forecast for
+# The Lua script to fetch the location key, current condition, forecast, and a TTL for
 # a given country/postal code.
 #
 # Note:
@@ -46,6 +46,10 @@ PARTNER_CODE: str | None = settings.accuweather.get("partner_code")
 #     `[location_key, current_condition, forecast]`. The last two element can be `nil`
 #     if they are not present in the cache
 #   - If the location key is missing, it will return an empty array
+#   - TTL values are all integer type. Could be one of -2, -1, or a non-negative non-zero
+#     integer. The TTL returned is the smaller one of the two TTLs; current conditions and
+#     forecast. We are explicitly returning the -2 and -1 as a TTL value since those represent
+#     error values. See official Redis TTL docs for more info.
 LUA_SCRIPT_CACHE_BULK_FETCH: str = """
     local location_key = redis.call("GET", KEYS[1])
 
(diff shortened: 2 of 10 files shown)


merino/providers/weather/backends/accuweather.py
```diff
@@ -64,6 +64,32 @@
...
+    if current_conditions_ttl == -2 or forecast_ttl == -2 then
+        return -2
+    elseif current_conditions_ttl == -1 or forecast_ttl == -1 then
+        return -1
+    else
+        local smaller_ttl = math.min(current_conditions_ttl, forecast_ttl)
+        return smaller_ttl
+"""
+SCRIPT_ID_GET_MIN_TTL: str = "get_min_ttl"
+
```
Herraj, line 92: This is script is very similar to the above `bulk_fetch` script. It returns an integer. Depending on the keys, it could be `-2`, `-1`, or the `TTL`.

merino/providers/weather/backends/accuweather.py
```diff
@@ -444,6 +472,56 @@ async def as_awaitable(val: Any) -> Any:
...
+            return None
+        if ttl == -2:
+            logger.error(
+                f"Cannot get TTL for key {repr(cache_key)}. Key does not exist"
+            )
+            return None
+
+        # finally return the ttl
+        return ttl
+
```
Herraj, line 524: This method is also very similar to the above `get_weather_report` method. Returns an `int` or `None` depending on what the script returns.

merino/providers/weather/backends/protocol.py
```diff
@@ -70,6 +70,16 @@ async def get_weather_report(
...
+    async def get_ttl_for_weather_report(
+        self, geolocation: Location
+    ) -> int:  # pragma: no cover
+        """Get TTL for a weather report.
+
+        Raises:
+            BackendError: Category of error specific to provider backends.
+        """
+        ...
+
```
Herraj, line 82: implemented in `accuweather.py`

merino/providers/weather/backends/accuweather.py
ncloudioj: How about we move this ttl bit over to `bulk_fetch` and add a new return value for ttl?

merino/web/api_v1.py
```diff
@@ -199,6 +200,18 @@ async def suggest(
         if client_variants
         else [],
     )
+
+    # make sure the request received is for the accuweather provider
+    if len(search_from) == 1 and search_from[0].name == ProviderType.ACCUWEATHER:
```
ncloudioj, line 205: Or you could try pattern matching :)

```py
match search_from:
    case [AccuweatherProvider()]:  # A single element list with a particular provider
        #...
    case _:
        # other providers
```

merino/providers/weather/provider.py
```diff
@@ -63,6 +64,8 @@ async def query(self, srequest: SuggestionRequest) -> list[BaseSuggestion]:
         geolocation: Location = srequest.geolocation
         weather_report: WeatherReport | None = None
 
+        print(f"*********** IN QUERY: {srequest}")
```
ncloudioj, line 67: Forgot to remove the debugging printouts?

merino/web/api_v1.py
```diff
@@ -199,6 +200,30 @@ async def suggest(
...
+        #     "weather" in suggestion["custom_details"]
+        #     and "weather_report_ttl" in suggestion["custom_details"]["weather"]
+        # ):
+        #     weather_report_ttl = suggestion["custom_details"]["weather"][
+        #         "weather_report_ttl"
+        #     ]
+        #     break
+
+    if weather_report_ttl is not None:
+        response_headers = {"Cache-Control": f"public, max-age={weather_report_ttl}"}
```
ncloudioj, line 221: What do you think about using a default TTL (e.g. 300 seconds) if `weather_report_ttl` is None? If you agree, we can add a configuration for this to default.toml.

merino/providers/weather/backends/accuweather.py
```diff
@@ -649,3 +676,16 @@ def process_forecast_response(response: Any) -> dict[str, Any] | None:
             }
         case _:
             return None
+
+
+# TODO Should we make this a class method?
+def get_ttl_error_message(ttl: int) -> str | None:
+    # TODO add comment
+    ttl_errors = {
+        0: "Error getting TTL. Location Key not found.",
```
ncloudioj, line 685: It will return an empty array if the location key is missing. So I guess it shouldn't be possible to have a 0 ttl?

merino/web/api_v1.py
```diff
@@ -199,6 +200,30 @@ async def suggest(
         if client_variants
         else [],
     )
+
+    weather_report_ttl = None
+    # find in suggestions the ttl
+
+    print(f"----> response: {response}")
```
ncloudioj, line 207: Let's clean up those debugging artifacts.

ncloudioj (changes requested):

merino/providers/weather/backends/accuweather.py
```diff
@@ -439,6 +479,7 @@ async def as_awaitable(val: Any) -> Any:
                 city_name=location.localized_name,
                 current_conditions=current_conditions,
                 forecast=forecast,
+                ttl=min(current_conditions.ttl, forecast.ttl),
```
Herraj, line 482: This addresses the use case where we have to make a new accuweather request on a cache miss. This `ttl` value will always be set to `1800` given our current config values i.e:
```
current_condition_ttl_sec = 1800 # 1/2 hr

forecast_ttl_sec = 3600 # 1 hr
```

This is because when store the newly requested data in the cache, we do a `max()` of our default^ values and the `reponse_expiry` received from accuweather. The result is the default value for both ttls. And then we `min()` those two values here i.e 1800 vs 3600. 

We also set a default of ttl for the reponse of `300` seconds (see above `default.toml`) for cases when the ttl that is bubbled up is returned as `None` for reason / error.

Thoughts on this?

merino/web/api_v1.py
Herraj: Good suggestion! I've added a comment below regarding that.

merino/web/api_v1.py
```diff
@@ -199,6 +200,44 @@ async def suggest(
...
+        if weather_report_ttl is not None:
+            response_headers = {
+                "Cache-control": f"public, max-age={weather_report_ttl}"
+            }
+        else:
+            default_weather_report_ttl = (
+                accuweather_provider.backend.cached_default_weather_report_ttl_sec
+            )
+            response_headers = {
+                "Cache-control": f"public, max-age={default_weather_report_ttl}"
```
ncloudioj, line 232:
```suggestion
                "Cache-control": f"private, max-age={default_weather_report_ttl}"
```

Let's do private here as the weather suggestion may be customized (for specific locations) for the user.

ncloudioj (commented):
It looks great, would like to take another look at it once the tests turn green :)

merino/configs/default.toml
```diff
@@ -39,6 +39,9 @@ query_timeout_sec = 0.2
 # Prevents a provider from being instantiated.
 disabled_providers = []
 
+# MERINO_RUNTIME__DEFAULT_SUGGESTIONS_REQUEST_TTL_SEC
+default_suggestions_request_ttl_sec = 300 # 5 mins
+
```
Herraj, line 44: Do we need to add this to the the `cloudops-infra` repo?

merino/web/api_v1.py
```diff
@@ -199,7 +205,18 @@ async def suggest(
...
         else [],
     )
-    return JSONResponse(content=jsonable_encoder(response, exclude_none=True))
+
+    # response headers for cache-control directive
+    response_headers: dict[str, str] = {}
+
+    ttl = get_ttl_for_cache_control_header(search_from, suggestions)
+    if ttl is not None:
+        response_headers["Cache-control"] = f"private, max-age={ttl}"
```
ncloudioj, line 214: Let's always set this header to be consistent. Alternatively, we can replace all the `None`s in `get_ttl_for_cache_control_header()` so it'll always retrun a valid TTL.
```suggestion
    ttl = get_ttl_for_cache_control_header(search_from, suggestions) or DEFAULT_CACHE_CONTROL_TTL
    response_headers["Cache-control"] = f"private, max-age={ttl}"
```

ncloudioj (approved):
r+wc.

Thanks, this looks great!

PR #512, opened 2024-05-08
[DISCO-2781] Set http cache header for suggestions response

Herraj:
## References

JIRA: [DISCO-2781](https://mozilla-hub.atlassian.net/browse/DISCO-2781)

## Description
This PR is in lieu of this [original PR](https://github.com/mozilla-services/merino-py/pull/505). Since I broke prod with the original PR, this PR serves as a fix. [See sentry errors](https://mozilla.sentry.io/issues/?environment=prod&project=6622434&referrer=sidebar&statsPeriod=24h&utc=true).

I cherry picked the commit from the original one since most of the logic / work is the same except the bug fix. I've added comments below where I've added the bug fix.

I've tested it locally with different partial cache hit scenarios.

## PR Review Checklist

[DISCO-2781]: https://mozilla-hub.atlassian.net/browse/DISCO-2781?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/merino/providers/weather/backends/accuweather.py
+++ b/merino/providers/weather/backends/accuweather.py
@@ -6,7 +6,7 @@ import hashlib
 import json
 import logging
 from enum import Enum
-from typing import Any, Callable, NamedTuple
+from typing import Any, Callable, NamedTuple, cast
 
 import aiodogstatsd
 from dateutil import parser
@@ -29,7 +29,7 @@ logger = logging.getLogger(__name__)
 PARTNER_PARAM_ID: str | None = settings.accuweather.get("url_param_partner_code")
 PARTNER_CODE: str | None = settings.accuweather.get("partner_code")
 
-# The Lua script to fetch the location key, current condition, and forecast for
+# The Lua script to fetch the location key, current condition, forecast, and a TTL for
 # a given country/postal code.
 #
 # Note:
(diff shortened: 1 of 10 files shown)


merino/providers/weather/backends/accuweather.py
```diff
@@ -421,28 +471,44 @@ async def as_awaitable(val: Any) -> Any:
...
                 )
                 task_forecast = (
                     tg.create_task(self.get_forecast(location.key))
                     if forecast is None
-                    else as_awaitable(forecast)
+                    else as_awaitable(
+                        ForecastWithTTL(
+                            forecast=forecast, ttl=self.cached_forecast_ttl_sec
+                        )
+                    )
```
Herraj, line 488: This is the fix from the previous PR mentioned in the PR description above. In the case of a partial cache hit, i.e one of `current_conditions` or `forecast` is returned from the cache, we don't make a new accuweather request for it. That was causing the type of that variable to be `CurrentConditions` or `Forecast` and not `CurrentConditionsWithTTL` or `ForecastWithTTL`. 

We create our expected type and populate the `ttl` value as the default ttl value that is set when initializing this backend / provider. Now when we de-structure it as two values, it won't throw below.

tests/unit/providers/weather/backends/test_accuweather.py
```diff
@@ -697,11 +720,12 @@ def mock_register_script(script) -> Callable[[list, list], Awaitable[list]]:
 
 
 @pytest.mark.parametrize(
```
Herraj, line 722: Below is the test case that tests the partial cache hit scenario which I have mentioned above in the comment. I was incorrectly modifying the test behaviour 😞 

Now without modifying the core test logic and refactoring the source code above, it passes with all partial cache hit scenarios.

ncloudioj (approved):
🚀

PR #513, opened 2024-05-15
[DISCO-2806]: Add location completion via accuweather api

Herraj:
## References

JIRA: [DISCO-2806](https://mozilla-hub.atlassian.net/browse/DISCO-2806)

## Description
Adding functionality to provide location completion suggestions via Accuweather api by modifying our existing `/suggest` endpoint.

## Testing
- Make sure your `default.local.toml` has the correct settings and the `accuweather_city` provider's `cache` is set to `redis`. 
- Start the redis instance: `cd dev && docker-compose up`
- Start merino `make dev`

#### Backwards compatibility tests
This should return empty suggestions:
`curl -i -X 'GET' 'localhost:8000/api/v1/suggest?q=new&providers=accuweather&request_type=location'`

This should return weather suggestions:
`curl -i -X 'GET' 'localhost:8000/api/v1/suggest?q=&providers=accuweather&request_type=weather'`

This should also return weather suggestions:
`curl -i -X 'GET' 'localhost:8000/api/v1/suggest?q=&providers=accuweather'`

#### New provider tests
This should return location completion suggestions:
`curl -i -X 'GET' 'localhost:8000/api/v1/suggest?q=new&providers=accuweather_city&request_type=location`

This should return weather suggestions:
`curl -i -X 'GET' 'localhost:8000/api/v1/suggest?q=new&providers=accuweather_city&request_type=weather'`

This should return an error (400):
`curl -i -X 'GET' 'localhost:8000/api/v1/suggest?q=new&providers=accuweather_city&request_type=unsupported'`

## PR Review Checklist

[DISCO-2806]: https://mozilla-hub.atlassian.net/browse/DISCO-2806?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/merino/providers/weather/backends/accuweather.py
+++ b/merino/providers/weather/backends/accuweather.py
@@ -80,6 +80,7 @@ LUA_SCRIPT_CACHE_BULK_FETCH: str = """
     return {location_key, current_conditions, forecast, ttl}
 """
 SCRIPT_ID: str = "bulk_fetch"
+LOCATION_COMPLETION_REQUEST_TYPE: str = "autocomplete"
 
 
 class AccuweatherLocation(BaseModel):
@@ -194,7 +195,6 @@ class AccuweatherBackend:
         self.url_postalcodes_param_query = url_postalcodes_param_query
         self.url_current_conditions_path = url_current_conditions_path
         self.url_forecasts_path = url_forecasts_path
-        # TODO add location completion url path
         self.url_location_completion_path = url_location_completion_path
         self.url_location_key_placeholder = url_location_key_placeholder
 
@@ -624,34 +624,21 @@ class AccuweatherBackend:
             self.url_param_api_key: self.api_key,
         }
 
-        request_type: str = url_path.strip("/").split("/", 1)[0]
-
-        with self.metrics_client.timeit(f"accuweather.request.{request_type}.get"):
+        with self.metrics_client.timeit(
+            f"accuweather.request." f"{LOCATION_COMPLETION_REQUEST_TYPE}.get"
+        ):
             response: Response = await self.http_client.get(url_path, params=params)
             response.raise_for_status()
 
         processed_location_completions = process_location_completion_response(
             response.json()
-        )["locations"]
+        )
 
         location_completions = [
             LocationCompletion(**item) for item in processed_location_completions
         ]
 
         return location_completions
-
-        # response = await self.get_request(
-        #     self.url_location_completion_path.format(country_code=geolocation.country),
-        #     params={
-        #         "q": search_term,
-        #         "language": "en-us",
-        #         self.url_param_api_key: self.api_key,
-        #     },
-        #     process_api_response=process_location_completion_response,
-        #     cache_ttl_sec=0,
-        # )
-        #
-        # return response["locations"] if response else None
 
     async def shutdown(self) -> None:
         """Close out the cache during shutdown."""
@@ -1159,9 +1146,9 @@ def add_partner_code(
         return url
 
 
-def process_location_completion_response(response: Any) -> dict[str, Any]:
+def process_location_completion_response(response: Any) -> list[dict[str, Any]]:
     """Process the API response for location completion request."""
-    result = [
+    return [
         {
             "key": location["Key"],
             "localized_name": location["LocalizedName"],
@@ -1169,8 +1156,6 @@ def process_location_completion_response(response: Any) -> dict[str, Any]:
         }
         for location in response
     ]
-
-    return {"locations": result}
 
 
 def process_location_response(response: Any) -> dict[str, Any] | None:
--- a/merino/providers/weather/backends/accuweather.py
+++ b/merino/providers/weather/backends/accuweather.py
@@ -142,7 +142,6 @@ class AccuweatherBackend:
     url_postalcodes_param_query: str
     url_current_conditions_path: str
     url_forecasts_path: str
-    url_location_completion_path: str
     url_location_key_placeholder: str
     http_client: AsyncClient
 
@@ -160,7 +159,6 @@ class AccuweatherBackend:
         url_postalcodes_param_query: str,
         url_current_conditions_path: str,
         url_forecasts_path: str,
-        url_location_completion_path: str,
         url_location_key_placeholder: str,
     ) -> None:
         """Initialize the AccuWeather backend.
@@ -195,7 +193,6 @@ class AccuweatherBackend:
         self.url_postalcodes_param_query = url_postalcodes_param_query
         self.url_current_conditions_path = url_current_conditions_path
         self.url_forecasts_path = url_forecasts_path
-        self.url_location_completion_path = url_location_completion_path
         self.url_location_key_placeholder = url_location_key_placeholder
 
     def cache_key_for_accuweather_request(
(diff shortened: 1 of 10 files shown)


merino/providers/weather/backends/accuweather.py
```diff
@@ -605,6 +610,49 @@ async def get_forecast(self, location_key: str) -> ForecastWithTTL | None:
...
+        #     params={
+        #         "q": search_term,
+        #         "language": "en-us",
+        #         self.url_param_api_key: self.api_key,
+        #     },
+        #     process_api_response=process_location_completion_response,
+        #     cache_ttl_sec=0,
+        # )
+        #
+        # return response["locations"] if response else None
```
Herraj, line 654: I decided to not use our existing `get_request` method for now. The reason being, this method caches the request. However, right now we are not consuming from that cache so it's only one way and my fear is that we are filling it up for no good reason. 

I've pulled out some of the request logic from that method and making a direct req using the http client. Let me know your thoughts!

merino/providers/weather/backends/accuweather.py
```diff
@@ -625,6 +673,20 @@ def add_partner_code(
...
+    result = [
+        {
+            "key": location["Key"],
+            "localized_name": location["LocalizedName"],
+            "country": location["Country"]["LocalizedName"],
+        }
+        for location in response
+    ]
+
+    return {"locations": result}
```
Herraj, line 687: This might change, awaiting FE engineers' response as to what properties they'd like for a location.

merino/web/api_v1.py
```diff
@@ -64,6 +64,7 @@ async def suggest(
     sources: tuple[dict[str, BaseProvider], list[BaseProvider]] = Depends(
         get_providers
     ),
+    request_type: str | None = None,
```
Herraj, line 67: Thoughts on validating this query param? I was thinking if we can enforce a regex validation for it to be `location` or `weather`. Right now, it will only return locations when it is exactly `location`. For anything else (or omitted) it will return weather suggestions (given the providers is `accuweather`)

merino/providers/weather/backends/accuweather.py
```diff
@@ -191,6 +194,8 @@ def __init__(
         self.url_postalcodes_param_query = url_postalcodes_param_query
         self.url_current_conditions_path = url_current_conditions_path
         self.url_forecasts_path = url_forecasts_path
+        # TODO add location completion url path
```
ncloudioj, line 197: Can we drop this #TODO now?

merino/providers/weather/backends/accuweather.py
```diff
@@ -605,6 +610,49 @@ async def get_forecast(self, location_key: str) -> ForecastWithTTL | None:
...
+        url_path = self.url_location_completion_path.format(
+            country_code=geolocation.country
+        )
+        params = {
+            "q": search_term,
+            "language": "en-us",
+            self.url_param_api_key: self.api_key,
+        }
+
+        request_type: str = url_path.strip("/").split("/", 1)[0]
```
ncloudioj, line 627: I think we can use a constant (e.g. `autocomplete`) here for `request_type` as this method is for location completion only.

merino/providers/weather/backends/accuweather.py
```diff
@@ -625,6 +673,20 @@ def add_partner_code(
         return url
 
 
+def process_location_completion_response(response: Any) -> dict[str, Any]:
+    """Process the API response for location completion request."""
+    result = [
+        {
+            "key": location["Key"],
+            "localized_name": location["LocalizedName"],
+            "country": location["Country"]["LocalizedName"],
```
ncloudioj, line 682: Do we need to include the AdminCode (State/Province) as well?

merino/providers/weather/backends/accuweather.py
ncloudioj, line 687: Nit: we can just return the list w/o the wrapping.

merino/providers/weather/provider.py
```diff
@@ -62,34 +73,59 @@ def hidden(self) -> bool:  # noqa: D102
...
+            return [self.build_suggestion(location_completions)]
+        else:
             return []
 
-        return [
-            Suggestion(
-                title=f"Weather for {weather_report.city_name}",
-                url=weather_report.current_conditions.url,
+    def build_suggestion(
+        self, data: WeatherReport | list[LocationCompletion]
```
ncloudioj, line 101: Can we do?
```suggestion
        self, data: WeatherReport | LocationCompletionSuggestion
```

merino/web/api_v1.py
ncloudioj: Yes, we should validate this. Let's see if we can use FastAPI's builtin [validation facilities](https://fastapi.tiangolo.com/tutorial/query-params-str-validations/).

merino/providers/weather/provider.py
Herraj: umm that would be the return type no? did you mean change the return type of that method to 
`Suggestion | LocationCompletionSuggestion` from `BaseSuggestion`?

tests/integration/api/v1/suggest/test_suggest_weather.py
```diff
@@ -194,3 +296,44 @@ def test_providers_request_log_data_weather(
...
+    response = client.get(
+        url="/api/v1/suggest",
+        params={"q": "new", "providers": "weather", "request_type": "location"},
+    )
+
+    assert response.status_code == 200
+    result = response.json()
+    assert expected_suggestion == TypeAdapter(
+        list[LocationCompletionSuggestion]
+    ).validate_python(result["suggestions"])
```
Herraj, line 339: This test baffles me 🤔 This isn't a 100% accurate but I can't seem to figure out what the issue is. You'll see that on line `320` and `332`, the provider value is `weather` and not `accuweather`. The actual response has `accuweather`, gets set at provider initialization time. 

If I set the `providers` query param to `accuweather` (as the actual endpoint), it doesn't return any suggestions. And when it does, it sets the `provider` on the suggestion as `weather` and not as `accuweather`. I noticed the above weather suggestion has the same behaviour 🙃 not sure if it's an issue with the client mock or something else

merino/providers/weather/provider.py
ncloudioj: Oh, NVM, i just noticed `LocationCompletionSuggestion` is basically `list[LocationCompletion]` but didn't realize it's a subclass of `BaseSuggestion`.

merino/providers/weather/backends/accuweather.py
Herraj: Pending FE team's response here. Shouldn't be a big lift though!

merino/providers/weather/backends/protocol.py
```diff
@@ -26,6 +26,13 @@ def __init__(self, c: float | None = None, f: float | None = None):
         super().__init__(c=cc, f=ff)
 
 
+class LocationCompletionGeoDetails(BaseModel):
+    """Model for a location completion's country and administrative area attributes."""
+
+    id: str
+    localized_name: str
+
```
Herraj, line 34: Better name suggestion for this class?

merino/providers/weather/backends/protocol.py
```diff
@@ -26,6 +26,13 @@ def __init__(self, c: float | None = None, f: float | None = None):
         super().__init__(c=cc, f=ff)
 
 
+class LocationCompletionGeoDetails(BaseModel):
+    """Model for a location completion's country and administrative area attributes."""
```
ncloudioj, line 30: The model fields aren't obvious, maybe you can give an example in the docstring.

merino/web/api_v1.py
```diff
@@ -64,6 +65,7 @@ async def suggest(
     sources: tuple[dict[str, BaseProvider], list[BaseProvider]] = Depends(
         get_providers
     ),
+    request_type: Annotated[str | None, Query(pattern="^(location|weather)$")] = None,
```
ncloudioj, line 68: Nice!

ncloudioj (approved):
Looks great, thanks for putting this up together! 🚢

PR #539, opened 2024-06-19
[DISCO-2850] Fix option to delete records for jobs

### LINKED ISSUE #531: --delete-existing-records doesn't work properly
Because the default is `True`, if you add the `--delete-existing-records` flag to the command line it will toggle the value and turn it to `False`.  I just ran into this with the FS data and it was quite confusing.

Maybe we should change this to `--keep-existing-records` and make the default `False`.  If we did that, then we should also not read the default from the config, since if any config set this to `True` then we'd be in the same confusing situation.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/DISCO-2850)
┆Epic: 2024 Q1 Maintenance

tiftran:
## References

JIRA: [DISCO-2850](https://mozilla-hub.atlassian.net/browse/DISCO-2850)
fixes #531 
## Description
It was confusing when we had  `--delete-existing-records`, since when you add flag to the command line it will toggle the value and turn it to False, which isn't what we want...
Renamed it to `keep-existing-records` with a default to false instead

## PR Review Checklist

[DISCO-2850]: https://mozilla-hub.atlassian.net/browse/DISCO-2850?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/merino/jobs/amo_rs_uploader/__init__.py
+++ b/merino/jobs/amo_rs_uploader/__init__.py
@@ -42,10 +42,10 @@ collection_option = typer.Option(
     help="Remote settings collection ID",
 )
 
-delete_existing_records_option = typer.Option(
-    rs_settings.delete_existing_records,
-    "--delete-existing-records",
-    help="Delete existing records before uploading new records",
+keep_existing_records_option = typer.Option(
+    rs_settings.keep_existing_records,
+    "--keep-existing-records",
+    help="Keep existing records before uploading new records",
 )
 
 dry_run_option = typer.Option(
@@ -84,7 +84,7 @@ def upload(
     bucket: str = bucket_option,
     chunk_size: int = chunk_size_option,
     collection: str = collection_option,
-    delete_existing_records: bool = delete_existing_records_option,
+    keep_existing_records: bool = keep_existing_records_option,
     dry_run: bool = dry_run_option,
     record_type: str = record_type_option,
     score: float = score_option,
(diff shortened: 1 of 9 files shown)


merino/jobs/amo_rs_uploader/__init__.py
```diff
@@ -42,10 +42,10 @@
     help="Remote settings collection ID",
 )
 
-delete_existing_records_option = typer.Option(
-    rs_settings.delete_existing_records,
-    "--delete-existing-records",
-    help="Delete existing records before uploading new records",
+keep_existing_records_option = typer.Option(
+    rs_settings.keep_existing_records,
```
bendk, line 46: What do you think about hard-coding this one to `false` rather than making it a config.  The reason I saw that is if the config changed to `keep_existing_records = true`, then it would cause the same issues if a user specified it on the command line.

bendk (approved):
Looks good to me, I had a suggestion about hard-coding the value rather than taking it from the config.  I'll let you make the call there.

merino/jobs/amo_rs_uploader/__init__.py
tiftran: ah fair, okie doke

PR #540, opened 2024-06-21
[MC-1101] feat: Return scheduled corpus items from corpus graph

katerinachinnappan:
## References

JIRA: https://mozilla-hub.atlassian.net/browse/MC-1101

## Description
Make a GRAPHQL request to the curated-corpus-api graph to return scheduled corpus items for the current date. 

## PR Review Checklist


merino/main.py
```diff
@@ -39,6 +39,8 @@ async def lifespan(app: FastAPI):
     configure_sentry()
     await configure_metrics()
     await providers.init_providers()
+    # curated recommendations provider (corpus-api)
```
mmiermans, line 42: Nit: In this case I think the code itself is clear on what it does.
```suggestion
```

merino/curated_recommendations/corpus_backends/corpus_api_backend.py
```diff
@@ -0,0 +1,115 @@
...
+            headers=CorpusApiGraphConfig.HEADERS,
+        )
+
+        data = res.json()
+
+        # Map Corpus topic to SERP topic
+        for item in data["data"]["scheduledSurface"]["items"]:
+            item["corpusItem"]["topic"] = map_corpus_topic_to_serp_topic(
+                item["corpusItem"]["topic"]
+            )
```
mmiermans, line 109: In general, I like to avoid using a variable for two distinct types or purposes. We change `item["corpusItem"]["topic"]` from a corpus topic to a serp topic. This might lead to confusion, for example when a Sentry error gets logged, and `data` contains different topics than we received.

I can think of a few alternatives, but all come with pros and cons. I don't think any of them are clearly better than what you have now, so your call whether you want to leave it as-is or not.

Here's an alternative that explicitly assigns every field, using the `:=` operator to avoid having to repeat `scheduled_item["corpusItem"]`. The downside is that adding a new field to CorpusItem requires a change here.
```python
corpus_items = [
    CorpusItem(
        scheduledCorpusItemId=scheduled_item["id"],
        url=(corpus_item := scheduled_item["corpusItem"])["url"],
        title=corpus_item["title"],
        excerpt=corpus_item["excerpt"],
        topic=map_corpus_topic_to_serp_topic(corpus_item["topic"]),
        publisher=corpus_item["publisher"],
        imageUrl=corpus_item["imageUrl"]
    )
    for scheduled_item in data["data"]["scheduledSurface"]["items"]
]
```

Here's an alternative that packs and unpacks values in a dictionary. I don't think the `**` operators make it especially readable.
```python
corpus_items = [
    CorpusItem(
        **{
          **item["corpusItem"],
          "topic": map_corpus_topic_to_serp_topic(item["corpusItem"]["topic"])
        },
        scheduledCorpusItemId=item["id"]
    )
    for item in data["data"]["scheduledSurface"]["items"]
]
```

tests/integration/api/v1/curated_recommendations/test_curated_recommendations.py
```diff
@@ -1,46 +1,114 @@
...
+    # Mock the endpoint
     response = client.post("/api/v1/curated-recommendations", json={"locale": "en-US"})
+    data = response.json()
+
+    # Check if the mock response is valid
     assert response.status_code == 200
 
-    result = response.json()
-    assert result["recommendedAt"] == 1326511294000  # 2012-01-14 03:21:34 UTC
+    corpus_items = data["data"]
```
mmiermans, line 91: At this point, I would call them 'curated recommendations' for consistency.
```suggestion
    curated_recommendations = data["data"]
```

tests/unit/curated_recommendations/test_curated_recommendations.py
```diff
@@ -0,0 +1,42 @@
+"""Unit test for map_corpus_to_serp_topic."""
```
mmiermans, line 1: Looking at the project structure, I think Merino has a convention of mirroring the paths in the implementation and unit tests. Could you move/rename this file to `tests/unit/curated_recommendations/test_corpus_api_backend.py`?

tests/unit/curated_recommendations/test_curated_recommendations.py
```diff
@@ -0,0 +1,42 @@
...
+
+from merino.curated_recommendations.corpus_backends.corpus_api_backend import (
+    map_corpus_topic_to_serp_topic,
+)
+
+
+@pytest.mark.parametrize(
+    "topic", ["PARENTING", "CORONAVIRUS", "GAMING", "CAREER", "EDUCATION"]
+)
+def test_map_corpus_to_serp_topic_return_null(topic):
```
mmiermans, line 13:
```suggestion
def test_map_corpus_to_serp_topic_return_none(topic):
```

tests/integration/api/v1/curated_recommendations/test_curated_recommendations.py
```diff
@@ -1,46 +1,114 @@
...
+
+@pytest.fixture(name="corpus_backend")
+def fixture_mock_corpus_backend(corpus_http_client: AsyncMock) -> CorpusApiBackend:
+    """Mock corpus api backend."""
+    # Initialize the backend with the mock HTTP client
+    return CorpusApiBackend(http_client=corpus_http_client)
+
+
+@pytest.fixture
+def provider(corpus_backend):
```
mmiermans, line 56:
```suggestion
def provider(corpus_backend: CorpusApiBackend) -> CuratedRecommendationsProvider:
```

tests/integration/api/v1/curated_recommendations/test_curated_recommendations.py
```diff
@@ -1,46 +1,114 @@
...
+    """Mock corpus provider."""
+    return CuratedRecommendationsProvider(corpus_backend=corpus_backend)
+
+
+@pytest.fixture(autouse=True)
+def setup_providers(provider):
+    """Set up the corpus api provider"""
```
mmiermans, line 63:
```suggestion
    """Set up the curated recommendations provider"""
```

mmiermans (commented):
I believe all my comments are pretty minor. Great work getting here in half a week!

merino/curated_recommendations/__init__.py
```diff
@@ -1,5 +1,24 @@
...
+    CorpusApiBackend,
+)
+from merino.curated_recommendations.provider import CuratedRecommendationsProvider
+from merino.utils.http_client import create_http_client
 
-"""Module dedicated to providing curated recommendations to New Tab."""
+corpus_api_provider: dict[str, CuratedRecommendationsProvider] = {}
+
+
+async def init_providers() -> None:
```
efixler, line 15: Why is this async?

merino/curated_recommendations/corpus_backends/corpus_api_backend.py
```diff
@@ -0,0 +1,115 @@
...
+        # based on the timezone associated with the scheduled surface.
+        # This requirement is documented in the NewTab slate spec:
+        # https://getpocket.atlassian.net/wiki/spaces/PE/pages/2927100008/Fx+NewTab+Slate+spec
+        today = datetime.now()
+
+        body = {
+            "query": query,
+            "variables": {
+                "scheduledSurfaceId": "NEW_TAB_EN_US",
+                "date": today.strftime("%Y-%m-%d"),
```
efixler, line 92: Just to idiot check here with all of the tz related chatter: I can see the comment above but I can see no code asserting that the date is within any particular timezone. (I would furthermore hope that Merino is running in prod with a UTC clock). Are we sure that this date assertion will be "correct", by which I think is meant "the date in the US/ET timezone)?

merino/curated_recommendations/corpus_backends/protocol.py
```diff
@@ -22,18 +22,22 @@ class Topic(str, Enum):
...
-
     The corpus is the set of all curated items deemed recommendable.
     """
 
     scheduledCorpusItemId: str
     url: HttpUrl
     title: str
     excerpt: str
-    topic: Topic
+    topic: Topic | None = None
```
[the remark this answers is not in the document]
efixler: Question(s): I believe that this is the content item being returned to the browser, and the mapped topic being mapped by `CORPUS_TOPIC_TO_SERP_TOPIC_MAPPING`

I believe that we are meant to exclude item in unmapped topics from the NewTab entirely, right? That being the case, we would not need to handle a null topic return here.

merino/curated_recommendations/corpus_backends/corpus_api_backend.py
```diff
@@ -0,0 +1,115 @@
...
+)
+
+
+class CorpusApiGraphConfig:
+    """Corpus API Graph Config."""
+
+    CORPUS_API_PROD_ENDPOINT = "https://client-api.getpocket.com"
+    CORPUS_API_DEV_ENDPOINT = "https://client-api.getpocket.dev"
+    CLIENT_NAME = "merino-py"
+    CLIENT_VERSION = "0.1.0"  # TODO: get this from pyproject.tml
```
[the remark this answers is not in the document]
katerinachinnappan: that seems more reasonable

merino/curated_recommendations/corpus_backends/corpus_api_backend.py
katerinachinnappan: I agree with the pros & cons, I vote to leave it as is for now & refactor if needed in the future? i do like the second proposed solution @mmiermans but agree that the `**` isn't intuitive.

merino/curated_recommendations/corpus_backends/corpus_api_backend.py
katerinachinnappan: @efixler no assertions in this PR for now but I think theres a ticket to handle this coming up soon

merino/curated_recommendations/corpus_backends/protocol.py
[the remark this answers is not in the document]
katerinachinnappan: According to Scott, they don't have a convention on the FE & will react the same if we return `null`, `""`, `undefined` or omit the prop completely. I propose to return `null` instead of omitting the prop, so that the data structure represents the actual response.

merino/curated_recommendations/__init__.py
katerinachinnappan: @efixler I am not 100% sure but I think its because the providers are initialized in `main.py` in the `lifespan` func, which is an `asynccontextmanager`, it waits for each provider to get initialized ? This is how merino sets up their other providers

merino/curated_recommendations/corpus_backends/protocol.py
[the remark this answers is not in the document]
efixler: Again, why would we be sending these resources to the browser at all?

merino/curated_recommendations/corpus_backends/corpus_api_backend.py
efixler: @katerina can you pls add a todo line in the code here with a reference to the ticket? Thanks!

katerinachinnappan: @efixler should already be there with the last commit

merino/curated_recommendations/corpus_backends/protocol.py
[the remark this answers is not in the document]
mmiermans: Firstly, my bad for including topic mapping in this ticket. Clearly more complex than I thought.

The [requirements](https://docs.google.com/document/d/1ICCHi1haxR-jIi_uZ3xQfPmphZm39MOmwQh0BRTXLHA/edit#heading=h.v1n5b52kwor4) do not mention that recommendations without a mapped topic should be excluded. @katerinachinnappan, could you clarify with Amber and Jim whether they have a product requirement around this?

merino/curated_recommendations/__init__.py
mmiermans: Sorry, I missed this during my review. This function doesn't do any IO, so it shouldn't be `async`.

merino/curated_recommendations/__init__.py
```diff
@@ -9,16 +9,17 @@
...
+    _provider = CuratedRecommendationsProvider(
         corpus_backend=CorpusApiBackend(create_http_client(base_url=""))
     )
 
 
 def get_provider() -> CuratedRecommendationsProvider:
-    """Return the corpus api provider."""
-    return corpus_api_provider["corpus_provider"]
+    """Return the curated recommendations provider."""
+    return _provider  # noqa: F821
```
mmiermans, line 25:
```suggestion
    global _provider
    return _provider
```

merino/main.py
```diff
@@ -39,8 +39,7 @@ async def lifespan(app: FastAPI):
     configure_sentry()
     await configure_metrics()
     await providers.init_providers()
-    # curated recommendations provider (corpus-api)
-    await curated_recommendations.init_providers()
+    await curated_recommendations.init_provider()
```
mmiermans, line 42:
```suggestion
    curated_recommendations.init_provider()
```

merino/curated_recommendations/__init__.py
```diff
@@ -9,16 +9,17 @@
...
 from merino.utils.http_client import create_http_client
 
-corpus_api_provider: dict[str, CuratedRecommendationsProvider] = {}
+_provider: CuratedRecommendationsProvider
 
 
-async def init_providers() -> None:
-    """Initialize the corpus api provider."""
-    corpus_api_provider["corpus_provider"] = CuratedRecommendationsProvider(
+async def init_provider() -> None:
```
mmiermans, line 15:
```suggestion
def init_provider() -> None:
```

mmiermans (commented):
Only some minor fixes.

I'll create a separate ticket to check whether filtering inmapped topics is a requirement, so we can keep that out-of-scope for this PR.

merino/curated_recommendations/corpus_backends/protocol.py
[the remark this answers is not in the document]
mmiermans: Thanks for raising this question Eric. I've create [MC-1241](https://mozilla-hub.atlassian.net/browse/MC-1241) so we don't lose track of it, and we can unblock this PR.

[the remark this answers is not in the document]
jpetto: i believe @efixler is correct here - that the 11 topics chosen for personalization specifically _do_ map to SERP topics. i don't think it's a bad idea to have this conversation/tests around topics that _do not_ map, but it's probably a future-proofing and not directly tied to the MVP.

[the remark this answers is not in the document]
efixler: I don't see how we can both display a label for every item (a requirement AFAIK) and also send items without topics. Those two seem to be mutually exclusive. Jim should be able to answer this definitively.

I don't think it was a problem introducing topic mapping here, we just need to clarify the behaviors :)

mmiermans (approved):

merino/curated_recommendations/corpus_backends/protocol.py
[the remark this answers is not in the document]
efixler: (also I think Jim T can answer the question about what we display but I am fairly certain that we only display items in mapped topics)

merino/curated_recommendations/corpus_backends/corpus_api_backend.py
efixler: Thanks @katerinachinnappan . Maybe it's just that the scope of this PR was a little too big but it's a valid concern that we're loading in operations here that we know won't operate as intended. I am extra vigilant on this one because it's an issue that comes up repeatedly and one that we're currently dealing with on the Android side because the expectations are odd and the implementations are somewhat shaky.

Furthermore, the description/acceptance criteria for MC-1199 just refers to copying code, it would be better if it actually described the requirements / acceptance criteria (pointing to the existing code as a reference is fine)

I believe that the actual requirements are something like the following: 

In order to know what the right date to fetch for is, the server at this point in time, needs to make a calculation, considering both the timezone that it's running in, and the timezone of the slate that it's trying to fill, in order to pick the correct date.  This should be a method and should not make any assumptions about configuration.

PR #544, opened 2024-06-25
[DISCO-2833] Fakespot blocklists

bendk:
Added code in the fakespot model to filter out items based on a blocklist.

Changed the signature of `csv_to_suggestions` to return `list[typing.Self]`.  I think this is more correct since the method should return instances of the subclass.  It's also needed to prevent lint errors.

It's possible we end up using a different system for the experiment, but this was easy to implement so I figured we might as well just do it.  If we need  to throw away the code later, it's fine.

## References

## PR Review Checklist


ncloudioj (approved):
LGTM!

PR #546, opened 2024-06-26
[MC-1197] feat: Curated-recs: fetching the correct feed

katerinachinnappan:
## References

JIRA: https://mozilla-hub.atlassian.net/browse/MC-1197

## Description
Based on passed `locale` & `region`, derive `surface_id`, pass it to the GRAPHQL query & return scheduled corpus items for the correct feed. Most of the code is ported from recs-api, added some more unit tests to the individual funcs.

## PR Review Checklist


merino/curated_recommendations/corpus_backends/fake_backends.py
```diff
@@ -6,13 +6,14 @@
     CorpusBackend,
     CorpusItem,
     Topic,
+    RecommendationSurfaceId,
```
katerinachinnappan, line 9: keeping this fake_backend for now, unless we are OK to remove it in this PR

merino/curated_recommendations/corpus_backends/protocol.py
```diff
@@ -22,10 +22,19 @@ class Topic(str, Enum):
     TECH = ("tech",)
     TRAVEL = ("travel",)
 
-    @staticmethod
-    def values():
```
katerinachinnappan, line 26: this func is not used anymore, removing

merino/curated_recommendations/corpus_backends/protocol.py
```diff
@@ -22,10 +22,19 @@ class Topic(str, Enum):
...
-        """Map enum values & returns"""
-        return Topic._value2member_map_
+
+@unique
+class RecommendationSurfaceId(str, Enum):
```
katerinachinnappan, line 27: ported from REC api

merino/curated_recommendations/provider.py
```diff
@@ -74,9 +81,89 @@ def __init__(
     ) -> None:
         self.corpus_backend = corpus_backend
 
-    async def fetch(self) -> CuratedRecommendationsResponse:
+    @staticmethod
```
katerinachinnappan, line 84: all 3 funcs are ported from recs-api

merino/curated_recommendations/corpus_backends/protocol.py
```diff
@@ -22,10 +22,19 @@ class Topic(str, Enum):
...
+    """Defines the possible recommendation surfaces."""
+
+    HOME = "HOME"
```
jpetto, line 30: curious what this surface is - i don't think this will return any results from the corpus...

tests/unit/curated_recommendations/test_provider.py
```diff
@@ -0,0 +1,164 @@
...
+
+
+def test_derive_region_return_none():
+    """Testing the derive_region() method &
+    ensure if no match is found, return None
+    """
+    # if region is passed
+    assert CuratedRecommendationsProvider.derive_region("123", "123") is None
+    # if only locale is passed
+    assert CuratedRecommendationsProvider.derive_region("123") is None
```
jpetto, line 87: would it be worth testing a locale like "en" with no region here?

jpetto (commented):
a couple non-blocking questions. test coverage is looking good!

merino/curated_recommendations/corpus_backends/fake_backends.py
mmiermans: I'm fine to remove it. Git makes sure nothing is truly removed.

mmiermans (commented):
Agreed with Jonathan's suggestions. I don't have anything to add. Thanks for bringing this over!

merino/curated_recommendations/corpus_backends/protocol.py
mmiermans: It's Pocket Home

katerinachinnappan: so do we remove?

tests/unit/curated_recommendations/test_provider.py
katerinachinnappan: @jpetto this would be also `None` since `en` can be a language in many regions

mmiermans (approved):

merino/curated_recommendations/corpus_backends/protocol.py
mmiermans: I'm on the fence. I slightly lean towards keeping this value, because it is a valid value that this enum can take, even if we won't use it in this service, for now. If anyone feels more strongly that it should be removed, I'm fine with that, too.

mmiermans: Maybe the least we should do is add a comment:
```suggestion
    HOME = "HOME"  # Pocket Home
```

mmiermans: Sorry, thinking about this again, we should remove it. `HOME` is not a scheduled surface id, and we should probably make a distinction between the two. We should probably rename `RecommendationSurfaceId` to `ScheduledSurfaceId` as well.

mmiermans (approved):
Thanks for making this change. 🙇

PR #550, opened 2024-07-02
[MC-1208] feat: add caching to CorpusApiBackend

mmiermans:
## References

JIRA: [MC-1208](https://mozilla-hub.atlassian.net/browse/MC-1208)

## Description

Guarantee reliability of the Pocket Graph and Merino, even when Merino will be handling thousands of requests per second for curated recommendations. This isn't an issue today, because a CDN sits in front of the firefox-proxy-api to cache and [coalesce requests](https://cloud.google.com/cdn/docs/caching#request-coalescing).

1. The curated corpus api should be guarded against high request volumes. The [ScheduledSurface query](https://studio.apollographql.com/graph/pocket-client-api/variant/current/insights?insightsTab=operations&operationSearch=scheduledsurf&operationsPage=0&query=31f1c92b1287f1f5a8d9b85f4d6806f5ac85d91a&queryName=ScheduledSurface) currently handles 90 requests/min, and Merino shouldn't send it much more traffic than that. We can accomplish this by adding a cache. When this cache expires, we should send a single "coalesced" request to the backend, to prevent the [thundering herd problem](https://en.wikipedia.org/wiki/Thundering_herd_problem).
2. Merino should be guarded against requests piling up while we make a request to the curated-corpus-api. When we have a stale value available in cache, we should use it, instead of blocking requests until the backend responds.

## PR Review Checklist

[MC-1208]: https://mozilla-hub.atlassian.net/browse/MC-1208?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/merino/curated_recommendations/corpus_backends/corpus_api_backend.py
+++ b/merino/curated_recommendations/corpus_backends/corpus_api_backend.py
@@ -5,7 +5,7 @@ import random
 from datetime import datetime, timedelta
 import asyncio
 
-from httpx import AsyncClient, HTTPStatusError
+from httpx import AsyncClient, HTTPError
 
 from merino.curated_recommendations.corpus_backends.protocol import (
     CorpusBackend,
@@ -176,7 +176,7 @@ class CorpusApiBackend(CorpusBackend):
             # Fetch new data from the backend.
             try:
                 data = await self._fetch_from_backend(surface_id)
-            except HTTPStatusError as e:
+            except HTTPError as e:
                 logger.warning(f"Retrying CorpusApiBackend._fetch_from_backend once after {e}")
                 # Backoff prevents high API rate during downtime.
                 await asyncio.sleep(self._backoff_time.total_seconds())
--- a/merino/curated_recommendations/corpus_backends/corpus_api_backend.py
+++ b/merino/curated_recommendations/corpus_backends/corpus_api_backend.py
@@ -1,10 +1,11 @@
 """Corpus API backend for making GRAPHQL requests"""
 
+import logging
 import random
 from datetime import datetime, timedelta
 import asyncio
 
-from httpx import AsyncClient
+from httpx import AsyncClient, HTTPStatusError
 
 from merino.curated_recommendations.corpus_backends.protocol import (
     CorpusBackend,
@@ -13,6 +14,9 @@ from merino.curated_recommendations.corpus_backends.protocol import (
     ScheduledSurfaceId,
 )
 from merino.utils.version import fetch_app_version_from_file
+
+
+logger = logging.getLogger(__name__)
 
 
 class CorpusApiGraphConfig:
@@ -66,16 +70,20 @@ class CorpusApiBackend(CorpusBackend):
     # rate to the scheduledSurface query stays close to the historic rate of ~100 requests/minute.
     cache_time_to_live_min = timedelta(seconds=50)
     cache_time_to_live_max = timedelta(seconds=70)
+    # The backoff time is the time that is waited before retrying.
+    # fetch() makes a single retry attempt, so there's exponential backoff (for now).
+    _backoff_time = timedelta(seconds=0.5)
     _cache: dict[ScheduledSurfaceId, list[CorpusItem]]
     _expirations: dict[ScheduledSurfaceId, datetime]
     _locks: dict[ScheduledSurfaceId, asyncio.Lock]
-    _background_tasks = set()
+    _background_tasks: set[asyncio.Task]
 
     def __init__(self, http_client: AsyncClient):
         self.http_client = http_client
         self._cache = {}
         self._expirations = {}
         self._locks = {}
+        self._background_tasks = set()
 
     async def fetch(self, surface_id: ScheduledSurfaceId) -> list[CorpusItem]:
         """Fetch corpus items with stale-while-revalidate caching and request coalescing."""
@@ -166,7 +174,19 @@ class CorpusApiBackend(CorpusBackend):
                 return self._cache[cache_key]
 
             # Fetch new data from the backend.
-            data = await self._fetch_from_backend(surface_id)
+            try:
+                data = await self._fetch_from_backend(surface_id)
+            except HTTPStatusError as e:
+                logger.warning(f"Retrying CorpusApiBackend._fetch_from_backend once after {e}")
+                # Backoff prevents high API rate during downtime.
+                await asyncio.sleep(self._backoff_time.total_seconds())
+                # http errors are expected to be rare, so a single retry attempt probably suffices.
+                data = await self._fetch_from_backend(surface_id)
+            except Exception as e:
+                # Backoff prevents high API rate when an unforeseen error occurs.
+                await asyncio.sleep(self._backoff_time.total_seconds())
+                raise e
+
             self._cache[cache_key] = data
             self._expirations[cache_key] = self.get_expiration_time()
             return data
--- a/merino/curated_recommendations/corpus_backends/corpus_api_backend.py
+++ b/merino/curated_recommendations/corpus_backends/corpus_api_backend.py
@@ -1,6 +1,7 @@
 """Corpus API backend for making GRAPHQL requests"""
 
-from datetime import datetime
+from datetime import datetime, timedelta
+import asyncio
 
 from httpx import AsyncClient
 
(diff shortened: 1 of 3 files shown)


merino/curated_recommendations/corpus_backends/corpus_api_backend.py
```diff
@@ -115,3 +139,21 @@ async def fetch(self, surface_id: ScheduledSurfaceId) -> list[CorpusItem]:
...
+        async with self._locks[cache_key]:
+            # Check if the cache was updated while waiting for the lock.
+            if cache_key in self._cache and datetime.now() < self._expirations[cache_key]:
+                return self._cache[cache_key]
+
+            # Fetch new data from the backend.
+            data = await self._fetch_from_backend(surface_id)
+            self._cache[cache_key] = data
+            self._expirations[cache_key] = datetime.now() + self.cache_expiration_time
+            return data
```
mmiermans, line 159: Ideally, logic like this should be implemented in a library. However, the popular async caching library [aiocache](https://github.com/aio-libs/aiocache) does not support request coalescing nor a 'stale-while-revalidate' caching strategy. The [default Python caching decorators](https://docs.python.org/3/library/functools.html) don't even support async functions.

As a general rule, I don't start generalizing code until it's reused at least 2 times. When we want to re-use this caching pattern, we could implement decorators similar to aiocache.

tests/integration/api/v1/curated_recommendations/test_curated_recommendations.py
```diff
@@ -1,46 +1,41 @@
...
     # Mock the POST request response with the loaded json mock data
     mock_http_client.post.return_value = Response(
         status_code=200,
-        json=scheduled_surface_data,
-        request=Request(
-            method="POST",
-            url=CorpusApiGraphConfig.CORPUS_API_PROD_ENDPOINT,
-            headers=CorpusApiGraphConfig.HEADERS,
-            json={"locale": "en-US"},
-        ),
```
mmiermans, line 43: I removed the mocked request because we don't use it. Tests pass without it.

merino/curated_recommendations/corpus_backends/corpus_api_backend.py
```diff
@@ -54,14 +55,38 @@ def map_corpus_topic_to_serp_topic(topic: str) -> tuple[str] | None:
...
 
     async def fetch(self, surface_id: ScheduledSurfaceId) -> list[CorpusItem]:
+        """Fetch corpus items with stale-while-revalidate caching and request coalescing."""
+        now = datetime.now()
+        cache_key = surface_id
+
+        # If we have expired cached data, revalidate asynchronously without waiting for the result.
+        if cache_key in self._cache:
+            if now >= self._expirations[cache_key]:
+                asyncio.create_task(self._revalidate_cache(surface_id))  # noqa: should not 'await'
```
ncloudioj, line 82: You might want to save a reference to the task somewhere, otherwise, it might get GC'd. See details [here](https://docs.python.org/3/library/asyncio-task.html#asyncio.create_task).

merino/curated_recommendations/corpus_backends/corpus_api_backend.py
```diff
@@ -115,3 +139,21 @@ async def fetch(self, surface_id: ScheduledSurfaceId) -> list[CorpusItem]:
...
+
+        async with self._locks[cache_key]:
+            # Check if the cache was updated while waiting for the lock.
+            if cache_key in self._cache and datetime.now() < self._expirations[cache_key]:
+                return self._cache[cache_key]
+
+            # Fetch new data from the backend.
+            data = await self._fetch_from_backend(surface_id)
+            self._cache[cache_key] = data
+            self._expirations[cache_key] = datetime.now() + self.cache_expiration_time
```
ncloudioj, line 158: Is it likely for many cached items expire around the same time? If so, we could add some cache jitter to TTL to mitigate that.

merino/curated_recommendations/corpus_backends/corpus_api_backend.py
```diff
@@ -54,14 +55,38 @@ def map_corpus_topic_to_serp_topic(topic: str) -> tuple[str] | None:
...
     http_client: AsyncClient
 
+    cache_expiration_time = timedelta(minutes=1)
+    _cache: dict[ScheduledSurfaceId, list[CorpusItem]]
+    _expirations: dict[ScheduledSurfaceId, datetime]
+    _locks: dict[ScheduledSurfaceId, asyncio.Lock]
+
     def __init__(self, http_client: AsyncClient):
         self.http_client = http_client
+        self._cache = {}
```
ncloudioj, line 70: Any estimate about the size of this in-memory cache? If the footprint is non-trivial, we will need to increase the resource quota for Merino or consider using a distributed cache instead.

merino/curated_recommendations/corpus_backends/corpus_api_backend.py
mmiermans: Thanks for reviewing and sharing knowledge! Great that we're already seeing the benefits from collaborating! 💪

merino/curated_recommendations/corpus_backends/corpus_api_backend.py
```diff
@@ -95,7 +120,6 @@ async def fetch(self, surface_id: ScheduledSurfaceId) -> list[CorpusItem]:
             },
         }
 
-        """Echoing the query as the single suggestion."""
```
ncloudioj, line 98: Notice that it doesn't check the HTTP status below, it's a good idea to do so:

```
res.raise_for_status()
```

Also, have we set the timeout configurations for the HTTP client? The default timeouts of httpx are not ideal and we have been bitten by them before. There is a utility HTTP client [creator](https://github.com/mozilla-services/merino-py/blob/main/merino/utils/http_client.py) for that.

merino/curated_recommendations/corpus_backends/corpus_api_backend.py
```diff
@@ -54,14 +55,38 @@ def map_corpus_topic_to_serp_topic(topic: str) -> tuple[str] | None:
 class CorpusApiBackend(CorpusBackend):
     """Corpus API Backend hitting the curated corpus api
     & returning recommendations for current date & locale/region.
+    Uses an in-memory cache and request coalescing to limit request rate to the backend.
     """
 
     http_client: AsyncClient
 
+    cache_expiration_time = timedelta(minutes=1)
```
jpetto, line 63: what's our logic for choosing one minute here?

merino/curated_recommendations/corpus_backends/corpus_api_backend.py
```diff
@@ -115,3 +139,21 @@ async def fetch(self, surface_id: ScheduledSurfaceId) -> list[CorpusItem]:
...
         return curated_recommendations
+
+    async def _revalidate_cache(self, surface_id: ScheduledSurfaceId) -> list[CorpusItem]:
+        """Purge and update the cache for a specific surface."""
+        cache_key = surface_id
+
+        if cache_key not in self._locks:
+            self._locks[cache_key] = asyncio.Lock()
+
+        async with self._locks[cache_key]:
```
jpetto, line 150: my lack of familiarity with python is showing, but - will this lock automatically clean itself up/release itself?

jpetto (commented):
nice tests, and thanks for pinging me here. i'm learning a lot about python async and caching 😄

merino/curated_recommendations/corpus_backends/corpus_api_backend.py
mmiermans: The additional memory usage will be less than 4 MB. Would that be considered trivial? I see that memory consumption per pod is 30-65 MB historically. Where can I see how much memory is available?

merino/curated_recommendations/corpus_backends/corpus_api_backend.py
mmiermans: Yeah, the `with` statement lets a 'context manager' release the lock when we exit this block, even if an exception were to be raised.

I've found realpython.com to contain clearly written descriptions of Python concepts. Here's their article about the with statement: https://realpython.com/python-with-statement/#using-the-async-with-statement

merino/curated_recommendations/corpus_backends/corpus_api_backend.py
ncloudioj: 4MB is fine. Here are the Merino [configurations](https://github.com/mozilla-services/cloudops-infra/blob/master/projects/merino/k8s/charts/merinopy/values.yaml#L143-L147) for resources.

merino/curated_recommendations/corpus_backends/corpus_api_backend.py
mmiermans: 👍 We use the utility creator with its default timeout.

merino/curated_recommendations/corpus_backends/corpus_api_backend.py
mmiermans: No, not likely. I added some jitter anyway like you suggested, just in case.

merino/curated_recommendations/corpus_backends/corpus_api_backend.py
mmiermans: Good question. I added a comment as to why I chose a minute.

PR #560, opened 2024-07-11
[DISCO-2823] Refine Inconclusive interest data

tiftran:
## References

JIRA: [DISCO-2823](https://mozilla-hub.atlassian.net/browse/DISCO-2823)
Uploaded new mapping into Remote Settings Dev 🙂 
I think there's something funny with deleting existing records, will investigate as a diff issue.

## Description
Instead of using our own mapping of categories, allow the job to use a mapping json file of categories to domains instead.

## PR Review Checklist

[DISCO-2823]: https://mozilla-hub.atlassian.net/browse/DISCO-2823?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/merino/jobs/relevancy_uploader/__init__.py
+++ b/merino/jobs/relevancy_uploader/__init__.py
@@ -87,7 +87,7 @@ class RelevancyData:
 
     @classmethod
     def csv_to_relevancy_data(
-        cls, csv_reader, categories_mapping
+        cls, csv_reader, upload_categories
     ) -> defaultdict[Category, list[dict[str, str]]]:
         """Read CSV file and extract required data for relevancy in the structure
         [
@@ -98,18 +98,26 @@ class RelevancyData:
             <domain>: {<category> : <score>}
         }.
         """
-        unique_domains = set((row["domain"], row["categories"]) for row in csv_reader)
+        domain_categories = defaultdict(list)
         data: defaultdict[Category, list[dict[str, str]]] = defaultdict(list)
-        for domain, fallback_categories in unique_domains:
-            classified = classify_domain(domain, categories_mapping, data)
+
+        for row in csv_reader:
+            domain = row["domain"]
+            categories = row["categories"]
+            csv_categories = categories.strip("[]").split(",")
+            for csv_category in csv_categories:
+                domain_categories[domain].append(csv_category)
+
+        for domain, csv_categories in domain_categories.items():
+            classified = classify_domain(domain, upload_categories, data)
             if not classified:
-                classify_domain_with_fallback(domain, fallback_categories, data)
+                classify_domain_with_fallback(domain, csv_categories, data)
         return data
 
 
-def classify_domain(domain, categories_mapping, data) -> bool:
+def classify_domain(domain, upload_categories, data) -> bool:
     """Classify domain to its Category."""
-    categories = categories_mapping.get(domain, {})
+    categories = upload_categories.get(domain, {})
     classified = False
     if categories:
         for category_name, score in categories.items():
@@ -127,9 +135,8 @@ def classify_domain(domain, categories_mapping, data) -> bool:
 def classify_domain_with_fallback(domain, categories, data) -> None:
     """Classify the domain using UPLOAD_CATEGORY_TO_R2D2 mapping."""
     classified = False
-    fallbacks = categories.strip("[]").split(",")
     md5_hash = md5(domain.encode(), usedforsecurity=False).digest()
-    for category in fallbacks:
+    for category in set(categories):
         category_mapped = UPLOAD_CATEGORY_TO_R2D2_CATEGORY.get(category, Category.Inconclusive)
         if category_mapped != Category.Inconclusive:
             data[category_mapped].append({"domain": base64.b64encode(md5_hash).decode()})
--- a/merino/jobs/relevancy_uploader/__init__.py
+++ b/merino/jobs/relevancy_uploader/__init__.py
@@ -3,13 +3,13 @@
 import asyncio
 import base64
 import csv
+import json
 import io
 from collections import defaultdict
 from enum import Enum
 from hashlib import md5
 
 import typer
-from pydantic import BaseModel
 
 from merino.config import settings as config
 from merino.jobs.relevancy_uploader.chunked_rs_uploader import (
@@ -17,6 +17,7 @@ from merino.jobs.relevancy_uploader.chunked_rs_uploader import (
 )
 
 CLASSIFICATION_CURRENT_VERSION = 1
+CATEGORY_SCORE_THRESHOLD = 0.5
 
 
 class Category(Enum):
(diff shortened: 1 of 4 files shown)


Herraj (approved):
looks good to me!

merino/jobs/relevancy_uploader/__init__.py
```diff
@@ -47,60 +48,48 @@ class Category(Enum):
...
     @classmethod
-    def csv_to_relevancy_data(cls, csv_reader) -> defaultdict[Category, list[dict[str, str]]]:
+    def csv_to_relevancy_data(
+        cls, csv_reader, categories_mapping
+    ) -> defaultdict[Category, list[dict[str, str]]]:
         """Read CSV file and extract required data for relevancy in the structure
         [
             { "domain" : <base64 string> }
         ]
         """
```
bendk, line 63: Can you add some notes on how `categories_mapping` works?

merino/jobs/relevancy_uploader/__init__.py
```diff
@@ -47,60 +48,48 @@ class Category(Enum):
...
+                                data[category].append(
+                                    {"domain": base64.b64encode(md5_hash).decode()}
+                                )
+                            else:
+                                md5_hash = md5(
+                                    row["domain"].encode(), usedforsecurity=False
+                                ).digest()
+                                data[Category.Inconclusive].append(
+                                    {"domain": base64.b64encode(md5_hash).decode()}
+                                )
```
bendk, line 87: Take this with a grain of last since I'm not really sure how this is supposed to work, but ISTM that if there are multiple categories/scores in `categories_mapping` we should take the item with the highest score and use that, rather than process each item.

- If one score is above `CATEGORY_SCORE_THRESHOLD` and one is below, it feels weird to me to map a domain to a both a specific category and `Inconclusive`.
- If both scores are below, then I think the domain will be mapped to `Inconclusive` twice.
- If both scores are above, then it makes some sense to map it to both categories, but I think this might violate some of the assumptions that the Rust code makes (not sure on this one).

tiftran: Oh yes, we could have a case where a domain is labelled as a category and inconclusive and there would defs be weirdness around that. Also after talking to @ncloudioj will be investigating what optimal threshold would be and also falling back to the old mapping in the event the domain isn't in the `categories_mapping` file. So lemme try to rework this with all of this in mind 🙂

merino/jobs/relevancy_uploader/__init__.py
```diff
@@ -80,30 +79,65 @@ class Category(Enum):
...
-            for category in categories:
-                category_mapped = UPLOAD_CATEGORY_TO_R2D2_CATEGORY.get(
-                    category, Category.Inconclusive
-                )
-                md5_hash = md5(row["domain"].encode(), usedforsecurity=False).digest()
-                data[category_mapped].append({"domain": base64.b64encode(md5_hash).decode()})
+        for domain, fallback_categories in unique_domains:
+            classified = classify_domain(domain, categories_mapping, data)
+            if not classified:
+                classify_domain_with_fallback(domain, fallback_categories, data)
```
tiftran, line 106: I think this reads better now?

@bendk one of your feedback bullet points was "If both scores are above, then it makes some sense to map it to both categories, but I think this might violate some of the assumptions that the Rust code makes (not sure on this one)." 

I'm not sure if I'm interpreting this right, but [this](https://github.com/mozilla/application-services/blob/main/components/relevancy/src/db.rs#L114-L123) assumes that there might be multiple categories, right?

merino/jobs/relevancy_uploader/__init__.py
```diff
@@ -3,20 +3,21 @@
...
 import typer
-from pydantic import BaseModel
 
 from merino.config import settings as config
 from merino.jobs.relevancy_uploader.chunked_rs_uploader import (
     ChunkedRemoteSettingsRelevancyUploader,
 )
 
 CLASSIFICATION_CURRENT_VERSION = 1
+CATEGORY_SCORE_THRESHOLD = 0.3
```
tiftran, line 20: Decided upon 0.3 from looking at the number of inconclusive categories generated by each threshold and eyeballing the number of url -> questionable categories at each threshold

| Threshold |Number of Inconclusive |
| --- | --- |
| 0.1 | 200 |
| 0.2 | 965 |
| 0.3 | 1417 |
| 0.4 | 1814 |
| 0.5 | 2160 |
| Old mapping strategy | 2515 |

merino/jobs/relevancy_uploader/__init__.py
bendk: Seems like that should work.

At some point, maybe we want to use fractional components.  With the current code, it seems like URLs that have multiple categories are going to contribute more than ones with 1 category.  Maybe we should use a `0.5` value for both components to avoid that.  We also might want to allow different weights (`category_foo=0.7` and `category_bar=0.3`).

Lots to think about, but it seems like that's better for a follow-up than for this PR.

merino/jobs/relevancy_uploader/__init__.py
```diff
@@ -80,30 +79,65 @@ class Category(Enum):
...
             { "domain" : <base64 string> }
-        ]
+        ].
+        Categories are determined by a category_mapping file. Where its structure is like
+        {
+            <domain>: {<category> : <score>}
+        }.
         """
-        rows = sorted(csv_reader, key=lambda x: x["categories"])
+        unique_domains = set((row["domain"], row["categories"]) for row in csv_reader)
```
bendk, line 101: Now that categories is also included, I think this should be a dict that's keyed by `domain` rather than a set.  That way if a domain was somehow listed twice with different categories, we wouldn't process it twice.

merino/jobs/relevancy_uploader/__init__.py
```diff
@@ -80,30 +79,65 @@ class Category(Enum):
...
         data: defaultdict[Category, list[dict[str, str]]] = defaultdict(list)
-        for row in rows:
-            categories = row["categories"].strip("[]").split(",")
-            for category in categories:
-                category_mapped = UPLOAD_CATEGORY_TO_R2D2_CATEGORY.get(
-                    category, Category.Inconclusive
-                )
-                md5_hash = md5(row["domain"].encode(), usedforsecurity=False).digest()
-                data[category_mapped].append({"domain": base64.b64encode(md5_hash).decode()})
+        for domain, fallback_categories in unique_domains:
```
bendk, line 103: When I read the name `fallback_categories`, I didn't get what these were.  Also in `classify_domain_with_fallback`, they're just called `categories`.  Can we come up with a consistent name that clearly explains what they are?  I kind of like `upload_categories`, which matches the map name.  Maybe `csv_categories` is even more clear.

bendk (approved):
This looks great to me.  I had some minor comments, however you want to deal with them is fine with me -- I don't think I need to re-review that.

merino/jobs/relevancy_uploader/__init__.py
tiftran: oh you're right, the tuples wouldn't make same domains unique

PR #571, opened 2024-07-17
[MC-212] feat: Adding utm_source param to curated rec url

katerinachinnappan:
## References

JIRA: https://mozilla-hub.atlassian.net/browse/MC-1212

## Description
Adding `utm_source` param to the curated rec url returned from corpus API. Adding a new one or replacing existing one. Created a `SCHEDULED_SURFACE_ID_TO_UTM_SOURCE` dict to map to the appropriate `utm_source`. Using `urllib.parse` to add/replace `utm_source` in a url.

## PR Review Checklist

### DIFF
--- a/merino/curated_recommendations/corpus_backends/corpus_api_backend.py
+++ b/merino/curated_recommendations/corpus_backends/corpus_api_backend.py
@@ -7,6 +7,7 @@ from datetime import datetime, timedelta
 import asyncio
 
 from httpx import AsyncClient, HTTPError
+from urllib.parse import urlparse, urlencode, parse_qsl
 
 from merino.curated_recommendations.corpus_backends.protocol import (
     CorpusBackend,
@@ -65,6 +66,20 @@ CORPUS_TOPIC_TO_SERP_TOPIC_MAPPING = {
     "TRAVEL": Topic.TRAVEL,
 }
 
+"""
+Set utm_source for specific ScheduledSurfaceId.
+For ids not in the table, null is returned.
+"""
+SCHEDULED_SURFACE_ID_TO_UTM_SOURCE: dict[ScheduledSurfaceId, str] = {
+    ScheduledSurfaceId.NEW_TAB_EN_US: "pocket-newtab-en-us",
+    ScheduledSurfaceId.NEW_TAB_EN_GB: "pocket-newtab-en-gb",
+    ScheduledSurfaceId.NEW_TAB_EN_INTL: "pocket-newtab-en-intl",
+    ScheduledSurfaceId.NEW_TAB_DE_DE: "pocket-newtab-de-de",
+    ScheduledSurfaceId.NEW_TAB_ES_ES: "pocket-newtab-es-es",
+    ScheduledSurfaceId.NEW_TAB_FR_FR: "pocket-newtab-fr-fr",
+    ScheduledSurfaceId.NEW_TAB_IT_IT: "pocket-newtab-it-it",
+}
+
 
 def map_corpus_topic_to_serp_topic(topic: str) -> Topic | None:
     """Map the corpus topic to the SERP topic."""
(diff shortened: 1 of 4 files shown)


merino/curated_recommendations/corpus_backends/corpus_api_backend.py
```diff
@@ -66,6 +67,20 @@ def headers(self):
     "TRAVEL": Topic.TRAVEL,
 }
 
+"""
+Set utm_source for specific ScheduledSurfaceId.
+For ids not in the table, null is returned.
+"""
+SCHEDULED_SURFACE_ID_TO_UTM_SOURCE: dict[ScheduledSurfaceId, str] = {
```
katerinachinnappan, line 74: instead of porting the implementation from recs API, figured we already have this pattern for Topics, so we can use it for utm_source too. Good readability

merino/curated_recommendations/corpus_backends/corpus_api_backend.py
```diff
@@ -207,11 +244,18 @@ async def _fetch_from_backend(self, surface_id: ScheduledSurfaceId) -> list[Corp
...
+        utm_source = self.get_utm_source(surface_id)
+
         for item in data["data"]["scheduledSurface"]["items"]:
+            # Map Corpus topic to SERP topic
             item["corpusItem"]["topic"] = map_corpus_topic_to_serp_topic(
                 item["corpusItem"]["topic"]
             )
+            # Update url (add / replace utm_source query param)
+            item["corpusItem"]["url"] = self.update_url_utm_source(
+                item["corpusItem"]["url"], str(utm_source)
```
jpetto, line 257: do we always want to update the utm_source if the value of `utm_source` is `None`? if `utm_source` is `None`, would we want to overwrite an existing utm_source in the url?

tests/integration/api/v1/curated_recommendations/test_curated_recommendations.py
```diff
@@ -126,6 +128,75 @@ async def test_curated_recommendations():
...
+        # assert total of 80 items returned
+        assert len(corpus_items) == 80
+        # Assert all corpus_items have expected fields populated.
+        assert all(item["url"] for item in corpus_items)
+        assert all(item["publisher"] for item in corpus_items)
+        assert all(item["imageUrl"] for item in corpus_items)
+
+        # Assert 2nd returned recommendation has topic = None & all fields returned are expected
+        actual_recommendation: CuratedRecommendation = CuratedRecommendation(**corpus_items[2])
+        assert actual_recommendation == expected_recommendation
```
jpetto, line 170: you could probably assert that every recommendation has the expected `utm_source` value in the `url`, right?

jpetto (changes requested):
looks really good. i would say update the test to make sure all recommendations have the expected `utm_source` value in the one test, but otherwise lgtm!

merino/curated_recommendations/corpus_backends/corpus_api_backend.py
katerinachinnappan: Good question. I thought about that & requirements say to overwrite if the param already exists. So maybe there was an old utm source & the new one returns None, we would want to replace

jpetto: okay - as long as the product requirements are clear on it, all good by me. having a `None` value would be very, very unlikely if i'm understanding the system right 😅

katerinachinnappan: it would actually never happen (I _think_) because surface ID is an enum, and if it doesn't exist it won't even reach to the step for `utm_source`.

merino/curated_recommendations/corpus_backends/corpus_api_backend.py
```diff
@@ -124,6 +119,11 @@ def __init__(
         self._locks = {}
         self._background_tasks = set()
 
+    @staticmethod
+    def map_corpus_topic_to_serp_topic(topic: str) -> Topic | None:
```
katerinachinnappan, line 123: moving this inside the class as this seems the pattern we are using

merino/curated_recommendations/corpus_backends/corpus_api_backend.py
```diff
@@ -147,19 +147,20 @@ def update_url_utm_source(url: str, utm_source: str) -> str:
...
         """
         zones = {
-            "NEW_TAB_EN_US": "America/New_York",
-            "NEW_TAB_EN_GB": "Europe/London",
-            "NEW_TAB_EN_INTL": "Asia/Kolkata",  # Note: en-Intl is poorly named. Only India is currently eligible.
-            "NEW_TAB_DE_DE": "Europe/Berlin",
-            "NEW_TAB_ES_ES": "Europe/Madrid",
-            "NEW_TAB_FR_FR": "Europe/Paris",
-            "NEW_TAB_IT_IT": "Europe/Rome",
+            ScheduledSurfaceId.NEW_TAB_EN_US: "America/New_York",
```
katerinachinnappan, line 156: we already have the surface id enum defined, we should use it

jpetto (approved):
nice refactor!

PR #573, opened 2024-07-18
[load test: warn] [MC-1270] feat: Handle curated-corpus-api GraphQL errors

nina-py:
## References

JIRA: https://mozilla-hub.atlassian.net/browse/MC-1270

## PR Review Checklist


merino/curated_recommendations/corpus_backends/corpus_api_backend.py
```diff
@@ -245,23 +245,31 @@ async def _fetch_from_backend(self, surface_id: ScheduledSurfaceId) -> list[Corp
         res.raise_for_status()
         data = res.json()
 
+        # log to Sentry if GraphQL returned errors
+        if res.status_code == 200 and "errors" in data:
+            for error in data["errors"]:
+                logger.exception(error)
```
mmiermans, line 251: My bad.. I realize the Jira ticket is ambiguous about what I think should happen.

I don't think this response will have (partially) useful data if there is a GraphQL error, so I think it's better to raise an exception here, and not continue execution of the function.
```suggestion
            raise BackendError("curated-corpus-api returned GraphQL error(s) {data["errors"]}")
```
You can import `BackendError` using `from merino.exceptions import BackendError`.

In `_revalidate_cache` I would catch this BackendError along with the HTTPError, by changing `except BackendError as e:` to `except (HTTPError, BackendError) as e:`.

merino/curated_recommendations/corpus_backends/corpus_api_backend.py
```diff
@@ -245,23 +245,31 @@ async def _fetch_from_backend(self, surface_id: ScheduledSurfaceId) -> list[Corp
...
+                item["corpusItem"]["url"] = self.update_url_utm_source(
+                    item["corpusItem"]["url"], str(utm_source)
+                )
+
+            curated_recommendations = [
+                CorpusItem(**item["corpusItem"], scheduledCorpusItemId=item["id"])
+                for item in data["data"]["scheduledSurface"]["items"]
+            ]
+        else:
+            curated_recommendations = []
```
mmiermans, line 272: These changes can be reverted. This method should raise an exception if recommendations can't be updated. It is called by `_revalidate_cache`, which will continue to serve recommendations from the stale cache when exceptions happen.

tests/integration/api/v1/curated_recommendations/test_curated_recommendations.py
```diff
@@ -177,6 +184,34 @@ async def test_curated_recommendations_locales(locale):
...
+                status_code=200,
+                json=fixture_graphql_200ok_with_error_response,
+                request=fixture_request_data,
+            )
+
+        corpus_http_client.post = AsyncMock(side_effect=return_with_graphql_error)
+        response = await fetch_en_us(ac)
+
+        # GraphQL server returns a 200 OK...
+        assert response.status_code == 200
```
mmiermans, line 206: We don't want a 200 code to be returned when a fatal exception happens. Firefox clients are designed to gracefully handle 5xx errors from our endpoint, and returning clear error status codes improves our observability.

Instead, I would modify `test_single_request_multiple_failed_fetches` as follows, to confirm that we can recover from temporary GraphQL errors:
```python
    @freezegun.freeze_time("2012-01-14 00:00:00", tick=True, tz_offset=0)
    @pytest.mark.parametrize("error_type", ["graphql", "http"])
    @pytest.mark.asyncio
    async def test_single_request_multiple_failed_fetches(
        self,
        corpus_http_client,
        fixture_request_data,
        fixture_response_data,
        fixture_graphql_200ok_with_error_response,
        caplog,
        error_type,
    ):
        """Test that only a few requests are made to the curated-corpus-api when it is down."""
        async with AsyncClient(app=app, base_url="http://test") as ac:
            start_time = datetime.now()

            def temporary_downtime(*args, **kwargs):
                # Simulate the backend being unavailable for 0.2 seconds.
                if datetime.now() < start_time + timedelta(seconds=0.2):
                    if error_type == "graphql":
                        return Response(
                            status_code=200,
                            json=fixture_graphql_200ok_with_error_response,
                            request=fixture_request_data
                        )
                    elif error_type == "http":
                        return Response(status_code=503, request=fixture_request_data)
                else:
                    return Response(
                        status_code=200,
                        json=fixture_response_data,
                        request=fixture_request_data,
                    )

            corpus_http_client.post = AsyncMock(side_effect=temporary_downtime)

            # Hit the endpoint until a 200 response is received.
            while datetime.now() < start_time + timedelta(seconds=1):
                try:
                    result = await fetch_en_us(ac)
                    if result.status_code == 200:
                        break
                except HTTPStatusError:
                    pass

            assert result.status_code == 200

            # Assert that we did not send a lot of requests to the backend
            # call_count is 400+ for me locally when CorpusApiBackend._backoff_time is changed to 0.
            assert corpus_http_client.post.call_count == 2

            # Assert that a warning was logged with a descriptive message.
            warnings = [r for r in caplog.records if r.levelname == "WARNING"]
            assert len(warnings) == 1
            if error_type == "graphql":
                # TODO: Assert the expected warning is logged.
                assert (
                    "Could not find Scheduled Surface with id of \"NEW_TAB_EN_UX\"."
                ) in warnings[0].message
            elif error_type == "http":
                assert (
                    "Retrying CorpusApiBackend._fetch_from_backend once after "
                    "Server error '503 Service Unavailable'"
                ) in warnings[0].message
```
(untested code.)

Alternatively or additionally, you could assert that the right error was sent to Sentry, but I don't think that's necessary, and it's not done anywhere else in Merino. Recommendation-api does have examples of Sentry tests.

mmiermans (commented):
Good call on getting feedback early. I now realize the Jira ticket was ambiguous on several aspects. Hopefully these comments clarify what I think the desired behavior is. 🙏

tests/integration/api/v1/curated_recommendations/test_curated_recommendations.py
nina-py: Updated, thank you @mmiermans! ⭐

tests/integration/api/v1/curated_recommendations/test_curated_recommendations.py
```diff
@@ -358,6 +365,72 @@ async def test_single_request_multiple_fetches(self, corpus_http_client):
             # Assert that the results are the same
             assert all(results[0].json() == result.json() for result in results)
 
+    @freezegun.freeze_time("2012-01-14 00:00:00", tick=True, tz_offset=0)
+    @pytest.mark.parametrize("error_type", ["graphql", "http"])
+    @pytest.mark.asyncio
+    async def test_single_request_multiple_failed_fetches_with_graphql_error(
```
mmiermans, line 371: This was meant to replace `test_single_request_multiple_failed_fetches` below.

nina-py: Done!

merino/curated_recommendations/corpus_backends/corpus_api_backend.py
```diff
@@ -243,6 +244,10 @@ async def _fetch_from_backend(self, surface_id: ScheduledSurfaceId) -> list[Corp
         res.raise_for_status()
         data = res.json()
 
+        # log to Sentry if GraphQL returned errors
+        if res.status_code == 200 and "errors" in data:
+            raise BackendError(f"curated-corpus-api returned GraphQL error(s) {data["errors"]}")
```
efixler, line 249: Aren't these errors already logged in Sentry at the corpus step? Just wondering if it's really wothwhile to insert them again.

merino/curated_recommendations/corpus_backends/corpus_api_backend.py
```diff
@@ -281,7 +287,7 @@ async def _revalidate_cache(self, surface_id: ScheduledSurfaceId) -> list[Corpus
             try:
                 data = await self.retry_fetch(surface_id, cache_key)
                 return data
-            except (HTTPError, ValueError) as e:
+            except (BackendError, HTTPError, ValueError) as e:
```
efixler, line 290: cc @katerinachinnappan : I'm not clear here how these changes interact with the recent changes you've made to error handling and caching in the endpoint, probably worth a close look from you.

merino/curated_recommendations/corpus_backends/corpus_api_backend.py
efixler: Also, is GraphQL error catching (e.g. the 200 status that is an error) new to this PR? If that's the case, I'm inferring that we were only catching Gateway errors before but that these errors were passing through. (which I think would explain why all of the errors we do we from the GraphQL layer are 502s)

merino/curated_recommendations/corpus_backends/corpus_api_backend.py
```diff
@@ -291,7 +297,7 @@ async def _revalidate_cache(self, surface_id: ScheduledSurfaceId) -> list[Corpus
                 try:
                     data = await self.retry_fetch(surface_id, cache_key)
                     return data
-                except (HTTPError, ValueError) as e:
+                except (BackendError, HTTPError, ValueError) as e:
```
efixler, line 300: Why are we raising these as backend Errors and not HTTPErrors? I think (please confirm @katerinachinnappan ) that we are currently catching Apollo-sources errors as HTTPErrors. This seems appropriate since the Apollo/GraphQL is _not_ the backend here, it's a service that's being consumed. 

This may be an appropriate error, but I want to double check. Let's please be consistent about how we're raising errors that come from the corpus.

efixler (changes requested):
Please make sure to insert Corpus errors into the appropriate StatsD metric so they make their way into the [Grafana dashboard](https://earthangel-b40313e5.influxcloud.net/d/nqIyrq%5FSk/merino-py-curated-recommendations?orgId=1&var-environment=stagepy&var-canary=All&from=1721076381565&to=1721077470717).

You will need to insert at a fabricated code since 200 will look like an OK. 500 is probably fine.

merino/curated_recommendations/corpus_backends/corpus_api_backend.py
katerinachinnappan: we missed that corpus API / graphql sub-graphs mask errors, returns a 200 status code with an error message. In this case `HTTPError` would only get caught on the gateway level, the 502 we have seen. that also explains why [these errors](https://mozilla.sentry.io/issues/5623243151/?project=6622434&referrer=project-issue-stream) in Merino were not handled, because they weren't treated as `HTTPErrors`. I think we should still catch `HTTPError`, `ValueError`. Handling GraphQL errors with `BackendError` seems ok, though I would instead make a derived class `CorpusAPIBackendError` from `BackendError`:

`class CorpusAPIBackendError(BackendError)`. Merino does this already for another backend I think. & would raise & catch `CorpusAPIBackendError` to be a bit more specific.

katerinachinnappan (commented):
@nina-py ty for reminding us with this PR that graphql errors return 200 + error message. PR looks good, I would only suggest to create a corpus api backend error specific class, derived from Merino `BackendError`. & would prob be good to check that [graphql errors are tracked in StatsD metrics](https://github.com/mozilla-services/merino-py/blob/main/tests/integration/api/v1/curated_recommendations/test_curated_recommendations.py#L591)

merino/curated_recommendations/corpus_backends/corpus_api_backend.py
katerinachinnappan: we haven't been catching graphql errors yet, only `httperrors`, so the `httpError` would not get caught if corpus API responded with a 200 + error message.

nina-py:
Closing in favour of https://github.com/mozilla-services/merino-py/pull/621

PR #576, opened 2024-07-18
[MC-1201] feat: Ensure preferred topic is in top 2

katerinachinnappan:
## References

JIRA: https://mozilla-hub.atlassian.net/browse/MC-1201

## Description
Find first item with topic in `preferred_topics` & move it to the 2nd slot in list of returned recs.

## PR Review Checklist


merino/curated_recommendations/provider.py
```diff
@@ -221,6 +221,37 @@ def spread_publishers(
 
         return result_recs
 
+    @staticmethod
+    def boost_preferred_topic(
+        recs: list[CuratedRecommendation],
+        preferred_topics: list[Topic],
+        boostable_slot: int = 1,
+    ):
```
mmiermans, line 229:
```suggestion
    ) -> list[CuratedRecommendation]:
```

merino/curated_recommendations/provider.py
```diff
@@ -241,10 +272,14 @@ async def fetch(
...
-        recommendations = CuratedRecommendationsProvider.spread_publishers(
-            recommendations, spread_distance=3
-        )
+        # 2. Perform publisher spread on the recommendation set
+        recommendations = self.spread_publishers(recommendations, spread_distance=3)
+
+        # 1. Finally, perform preferred topics boosting if preferred topics are passed in the request
+        if curated_recommendations_request.topics:
+            recommendations = self.boost_preferred_topic(
+                recommendations, curated_recommendations_request.topics
```
mmiermans, line 281: A preferred topic should only be boosted up if the top 2 don't already contain a preferred topic.
```suggestion
            if not any(r.topic in curated_recommendations_request.topics for r in curated_recommendations_request[:2]):
                recommendations = self.boost_preferred_topic(
                recommendations, curated_recommendations_request.topics
```
^ Just meant to convey the basic idea. You can extract that logic into a method to improve readability. You may also want to extract 2 into a variable to avoid magic numbers.

tests/integration/api/v1/curated_recommendations/test_curated_recommendations.py
```diff
@@ -315,6 +315,25 @@ async def test_curated_recommendations_topics(self, topics):
...
+
+            assert response.status_code == 200
+            # assert total of 80 items returned
+            assert len(corpus_items) == 80
+            # assert 6th rec is now in 2nd slot
+            assert (
+                fixture_response_data["data"]["scheduledSurface"]["items"][5]["id"]
+                == corpus_items[1]["scheduledCorpusItemId"]
+            )  # noqa
+
```
mmiermans, line 336: Could you add tests that assert the order is not changed if the first or second recommendation is already preferred? Could be a parameterized test.

merino/curated_recommendations/provider.py
```diff
@@ -221,6 +221,37 @@ def spread_publishers(
...
+        """Boost a recommendation based on preferred topic(s) into `boostable_slot`.
+
+        :param recs: List of recommendations from which an item is boosted based on preferred topic(s).
+        :param preferred_topics: user's preferred topic(s)
+        :param boostable_slot: 0-based slot to boost an item into. Defaults to slot 1,
+        which is the second recommendation.
+        :return: CuratedRecommendations ranked based on a preferred topic, while otherwise preserving the order.
+        """
+        # get the first item found to boost based on the below condition starting after the 3rd item in the list.
+        # condition for boostable item: check if an item has a topic in the preferred_topics list.
```
mmiermans, line 239:
```suggestion
        # get the first item from the list with a preferred topic starting after the boostable_slot
```
I would refer to `boostable_slot` instead of 3rd, because the argument that may be given a different value in the future.

mmiermans (commented):
Thanks! Awesome to see both you and Jonathan implement ranking code. I think this is almost there.

merino/curated_recommendations/provider.py
```diff
@@ -221,6 +221,54 @@ def spread_publishers(
...
+        recs: list[CuratedRecommendation],
+        preferred_topics: list[Topic],
+    ) -> bool:
+        """Check if top 2 recommendations already have the preferred topics.
+
+        :param recs: List of recommendations
+        :param preferred_topics: user's preferred topic(s)
+        :return: bool
+        """
+        top_two_recs = recs[:2]  # slice operator, get the first two recs (index 0 & 1)
```
jpetto, line 235: `2` here might be nice to have in a config, or even as a parameter with a default like in `boost_preferred_topic` below.

merino/curated_recommendations/provider.py
```diff
@@ -221,6 +221,54 @@ def spread_publishers(
 
         return result_recs
 
+    @staticmethod
+    def is_boostable(
+        recs: list[CuratedRecommendation],
+        preferred_topics: list[Topic],
+    ) -> bool:
+        """Check if top 2 recommendations already have the preferred topics.
```
jpetto, line 229: would be nice to have a little more context here - why are we checking?

merino/curated_recommendations/provider.py
```diff
@@ -221,6 +221,54 @@ def spread_publishers(
...
+        # get the first item found to boost based on the below condition starting after the boostable_slot in the list.
+        # condition for boostable item: check if an item has a topic in the preferred_topics list.
+        boostable_rec = next(
+            (r for r in recs[boostable_slot + 1 :] if r.topic in preferred_topics),
+            None,
+        )
+
+        # if item to boost is found
+        if boostable_rec:
+            recs = copy(recs)  # Don't change the input, make a temp copy
```
jpetto, line 264: but isn't this changing the input by using the same variable name? it's making a copy of itself and storing it in the same container. 😅

tests/unit/curated_recommendations/test_provider.py
```diff
@@ -343,6 +344,100 @@ def test_spread_publishers_cannot_spread(self):
...
+
+    def test_boost_preferred_topic_one_topic(self):
+        """Should boost first item found with preferred topic to second slot."""
+        recs = self.generate_recommendations(
+            [Topic.TRAVEL, Topic.ARTS, Topic.SPORTS, Topic.FOOD, Topic.EDUCATION]
+        )
+        reordered_recs = CuratedRecommendationsProvider.boost_preferred_topic(
+            recs, [Topic.EDUCATION]
+        )
+        print(reordered_recs)
```
jpetto, line 379: can we drop this? guessing a debug statement...

tests/unit/curated_recommendations/test_provider.py
```diff
@@ -343,6 +344,100 @@ def test_spread_publishers_cannot_spread(self):
...
+        slot.
+        """
+        recs = self.generate_recommendations(
+            [Topic.TRAVEL, Topic.ARTS, Topic.SPORTS, Topic.FOOD, Topic.EDUCATION]
+        )
+        # career topic is not present in rec list, boost item with food topic to second slot
+        reordered_recs = CuratedRecommendationsProvider.boost_preferred_topic(
+            recs, [Topic.CAREER, Topic.FOOD]
+        )
+        print(reordered_recs)
```
jpetto, line 399: same - can we remove this?

tests/unit/curated_recommendations/test_provider.py
```diff
@@ -343,6 +344,100 @@ def test_spread_publishers_cannot_spread(self):
...
+        assert reordered_recs[1].topic == Topic.FOOD
+        assert reordered_recs[1].scheduledCorpusItemId == "4"
+
+    def test_boost_preferred_topic_no_preferred_topic_found(self):
+        """Don't reorder list of recs if no items with preferred topics are found."""
+        recs = self.generate_recommendations(
+            [Topic.POLITICS, Topic.ARTS, Topic.SPORTS, Topic.FOOD, Topic.PERSONAL_FINANCE]
+        )
+        reordered_recs = CuratedRecommendationsProvider.boost_preferred_topic(recs, [Topic.CAREER])
+        print(reordered_recs)
```
jpetto, line 414: and one more 😀

tests/integration/api/v1/curated_recommendations/test_curated_recommendations.py
```diff
@@ -315,6 +315,61 @@ async def test_curated_recommendations_topics(self, topics):
...
+            corpus_items = data["data"]
+
+            assert response.status_code == 200
+            # assert total of 80 items returned
+            assert len(corpus_items) == 80
+            # assert 6th rec is now in 2nd slot
+            assert (
+                fixture_response_data["data"]["scheduledSurface"]["items"][5]["id"]
+                == corpus_items[1]["scheduledCorpusItemId"]
+            )  # noqa
```
jpetto, line 335: instead of digging in to the returned array, what do you think about just making sure that `is_boostable` and `boost_preferred_topic` have been called?

you've got unit tests for those functions below that verify specific behavior - could be good enough in this test (and the one below) just to make sure those boost functions were/were not called as expected.

jpetto (changes requested):
nothing blocking except removing the `print` statements, but also added some suggestions for small improvements.

nice work!

merino/curated_recommendations/provider.py
katerinachinnappan: @jpetto this comes from recs api and from my understanding, this creates a shallow copy of the `recs` list. Any changes made to `recs` within this function do not affect the original input list.

katerinachinnappan: the comment might be confusing, I will change it

tests/integration/api/v1/curated_recommendations/test_curated_recommendations.py
katerinachinnappan: good call, I will update the test

merino/curated_recommendations/provider.py
jpetto: oh yes, because python passes by reference and not by value. this might read confusing to me as i'm used to javascript passing by value.

jpetto (approved):
nice work!

PR #577, opened 2024-07-19
[DISCO-2892] fix: accuweather to try to use all subdivisions

tiftran:
## References

JIRA: [DISCO-2892](https://mozilla-hub.atlassian.net/browse/DISCO-2892)

## Description
Try to use all the subdivisions from maxmindb, when the first one we try does not work.

## PR Review Checklist

[DISCO-2892]: https://mozilla-hub.atlassian.net/browse/DISCO-2892?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


merino/providers/weather/backends/accuweather.py
```diff
@@ -292,7 +293,8 @@ async def get_request(
             response.raise_for_status()
 
         if (response_dict := process_api_response(response.json())) is None:
-            self.metrics_client.increment(f"accuweather.request.{request_type}.processor.error")
+            if log_failure:
+                self.metrics_client.increment(f"accuweather.request.{request_type}.processor.error")
```
tiftran, line 297: this feels very jank, but i don't think we want to increment the metric, every time we're cycling through subdivisions and the api call returns empty

merino/middleware/geolocation.py
```diff
@@ -27,6 +27,7 @@ class Location(BaseModel):
     country_name: Optional[str] = None
     region: Optional[str] = None
     region_name: Optional[str] = None
+    subdivisions: Optional[list[str]] = None
```
ncloudioj, line 30: To be consistent, let's call it "alternative_regions".

merino/middleware/geolocation.py
```diff
@@ -38,6 +39,11 @@ def normalize_string(input_str) -> str:
     return unicodedata.normalize("NFKD", input_str).encode("ascii", "ignore").decode("ascii")
 
 
+def get_subdivisions(sub_divisions) -> list[str]:
+    """Get all subvisions iso codes."""
+    return [s.iso_code for s in sub_divisions]
```
ncloudioj, line 44: Shall we exclude `region` from this list?

merino/middleware/geolocation.py
```diff
@@ -74,6 +80,7 @@ async def __call__(self, scope: Scope, receive: Receive, send: Send) -> None:
                 country_name=record.country.names.get("en"),
                 region=record.subdivisions.most_specific.iso_code,
                 region_name=record.subdivisions.most_specific.names.get("en"),
+                subdivisions=get_subdivisions(record.subdivisions),
```
ncloudioj, line 83: I think the type of `subdivisions` should be `list[str]` instead of `Optional[list[str]]`.

merino/providers/weather/backends/accuweather.py
```diff
@@ -564,12 +576,20 @@ async def as_awaitable(val: Any) -> Any:
...
         # The cached report is incomplete, now fetching from AccuWeather.
         if location is None:
-            if country and city and region:
-                if (
-                    location := await self.get_location_by_geolocation(country, city, region)
-                ) is None:
-                    return None
-            else:
+            if country and city:
+                if region:
```
ncloudioj, line 580: Assuming `region` is not in `subdivisions`, I think you do the following to simplify this block. The reasoning: if `region` is None, `subdivisions` must be empty b/c `region` is the most specific subdivision, so no need to handle the case of `region is None but subdivision is not empty.

```py
    if country and city and region:
        for region in [region, *subdivisions]:
             ...
```

ncloudioj: Also, what do you think about logging the requests if it successfully fetched the location key by using an alternative subdivision? My understanding is that it's quite expensive (by pinging AccuWeather) to do try-and-error this way permanently. Rather, we should learn what subdivisions AccuWeather prefer for certain countries, so that Merino would just automatically use the alternative subdivision in the future.

Perhaps we can log the country code and the subdivision when an alternative subdivision was used to successfully fetch the location key. Over time, we can analyze those log entries to learn which countries should be treated differently and build a lookup table so that Merino can use to choose the best subdivision code for future calls.

merino/middleware/geolocation.py
tiftran: we could, but that would mean we would be trying to fetch location with the  `most_specific` subdivision first and then try the alternatives...... which is what you're proposing below with logging. kk I'm going to have to fiddle with the logic then

merino/middleware/geolocation.py
tiftran: actually, should it be list[str] when all other fields are optional?

merino/middleware/geolocation.py
```diff
@@ -74,6 +82,7 @@ async def __call__(self, scope: Scope, receive: Receive, send: Send) -> None:
                 country_name=record.country.names.get("en"),
                 region=record.subdivisions.most_specific.iso_code,
                 region_name=record.subdivisions.most_specific.names.get("en"),
+                alternative_regions=get_alternative_regions(record.subdivisions),
```
ncloudioj, line 85:
```suggestion
                alternative_regions=[s.iso_code for s in subdivisions if s != record.subdivisions.most_specific],
                # `record.subdivisions` should be always present, as a result, the type of `alternative_regions` should be `list[str]`.
```

merino/providers/weather/backends/accuweather.py
```diff
@@ -496,36 +500,47 @@ async def get_weather_report_with_geolocation(
         """
         country: str | None = geolocation.country
         region: str | None = geolocation.region
+        alternative_regions: list[str] | None = geolocation.alternative_regions
```
ncloudioj, line 503: `alternative_regions` should not be `None`, it will simplify the the for loop below.

ncloudioj (approved):
r+wc.

merino/providers/weather/backends/accuweather.py
tiftran: agreed but, https://github.com/mozilla-services/merino-py/pull/577#discussion_r1684887037?

ncloudioj: I think we should follow the actual type. An empty list means no alternative regions, no need to make it None. We can further refactor this when combining `region` and `alternative_regions` in the future.

PR #582, opened 2024-07-24
[DISCO-2893] Fallback to /locations/v1/cities/<country_code>/search w…

tiftran:
…hen regions return empty

## References

JIRA: [DISCO-2893](https://mozilla-hub.atlassian.net/browse/DISCO-2893)

## Description
There are cases where maxmind does not return any regions that accuweather expects, so we should fall back to just using the country_code endpoint in this case

## PR Review Checklist

[DISCO-2893]: https://mozilla-hub.atlassian.net/browse/DISCO-2893?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/merino/providers/weather/backends/accuweather.py
+++ b/merino/providers/weather/backends/accuweather.py
@@ -673,7 +673,9 @@ class AccuweatherBackend:
             response: dict[str, Any] | None = await self.get_request(
                 url_path,
                 params=self.get_location_key_query_params(city),
-                process_api_response=process_location_response,
+                process_api_response=process_location_response_with_country_and_region
+                if region
+                else process_location_response_with_country,
                 cache_ttl_sec=self.cached_location_key_ttl_sec,
                 log_failure=log_failure,
             )
@@ -796,7 +798,7 @@ class AccuweatherBackend:
     async def check_cache_for_weather(self, cache_key) -> list[bytes | None]:
         """Get cached weather data."""
         with self.metrics_client.timeit("accuweather.cache.fetch"):
-            cached_data = await self.cache.run_script(
+            cached_data: list = await self.cache.run_script(
                 sid=SCRIPT_ID,
                 keys=[cache_key],
                 # The order matters below. See `LUA_SCRIPT_CACHE_BULK_FETCH` for details.
@@ -806,8 +808,6 @@ class AccuweatherBackend:
                     self.url_location_key_placeholder,
                 ],
             )
-            if not isinstance(cached_data, list):
-                return []
             return cached_data
 
     async def shutdown(self) -> None:
@@ -854,7 +854,7 @@ def process_location_completion_response(response: Any) -> list[dict[str, Any]]:
     ]
 
 
-def process_location_response(response: Any) -> dict[str, Any] | None:
+def process_location_response_with_country_and_region(response: Any) -> dict[str, Any] | None:
     """Process the API response for location keys.
 
     Note that if you change the return format, ensure you update `LUA_SCRIPT_CACHE_BULK_FETCH`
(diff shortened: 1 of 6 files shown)


merino/providers/weather/backends/accuweather.py
```diff
@@ -782,6 +793,23 @@ async def get_location_completion(
...
+                keys=[cache_key],
+                # The order matters below. See `LUA_SCRIPT_CACHE_BULK_FETCH` for details.
+                args=[
+                    self.cache_key_template(WeatherDataType.CURRENT_CONDITIONS),
+                    self.cache_key_template(WeatherDataType.FORECAST),
+                    self.url_location_key_placeholder,
+                ],
+            )
+            if not isinstance(cached_data, list):
+                return []
```
tiftran, line 810: I did this to appease typing weirdness

ncloudioj: This should do the trick.
`cache_data: list = await self.cache...`

merino/providers/weather/backends/accuweather.py
```diff
@@ -658,17 +663,23 @@ async def get_location_by_geolocation(
         Reference:
             https://developer.accuweather.com/accuweather-locations-api/apis/get/locations/v1/cities/{countryCode}/{adminCode}/search
         """
+        if region:
+            url_path = self.url_cities_admin_path.format(country_code=country, admin_code=region)
+            log_failure = False
+        else:
+            url_path = self.url_cities_path.format(country_code=country)
+            log_failure = True
```
ncloudioj, line 671: This alone is not sufficient to handle the case where the upstream returns multiple locations for a give country/city as `process_location_response()` will only pick the first one. Some extra handling is needed for that.

ncloudioj (commented):
Looks good, just an edge case to handle.

Also, looks like the block of 592->609 is not tested, can you take a look at that?

merino/providers/weather/backends/accuweather.py
tiftran: right, we were talking about this yesterday, lemme fix

tiftran:
okay so i've handled the case where the fallback api call returns two locations.
I've pretty sure my tests _should_ cover 592->609 , I've checked for the logs in that block, the tests run, so im not sure why it's stating the block isn't covered?

merino/providers/weather/backends/accuweather.py
```diff
@@ -658,17 +663,25 @@ async def get_location_by_geolocation(
...
         try:
             response: dict[str, Any] | None = await self.get_request(
-                self.url_cities_path.format(country_code=country, admin_code=region),
+                url_path,
                 params=self.get_location_key_query_params(city),
-                process_api_response=process_location_response,
+                process_api_response=process_location_response_with_country_and_region
```
ncloudioj, line 676: Nit: let's move this to the if block on line 666.

merino/providers/weather/backends/accuweather.py
```diff
@@ -852,6 +880,31 @@ def process_location_response(response: Any) -> dict[str, Any] | None:
             return None
 
 
+def process_location_response_with_country(response: Any) -> dict[str, Any] | None:
+    """Process the API response for location keys from country code endpoint.
+
+    Note that if you change the return format, ensure you update `LUA_SCRIPT_CACHE_BULK_FETCH`
```
ncloudioj, line 886: Let's note here that it will process responses with a single returned location key for disambiguity.

ncloudioj (approved):
LGTM! Thanks!

PR #586, opened 2024-07-27
[MC-1366] fix: Return latest valid data from cache on error response

katerinachinnappan:
## References

JIRA: https://mozilla-hub.atlassian.net/browse/MC-1366

## Description
During a load test, for curated recs merino endpoint ,[ errors in Corpus API](https://pocket.sentry.io/issues/5331854131/?environment=production&query=start%3D2024-07-24T19%3A01%3A00%26end%3D2024-07-25T20%3A05%3A00%26groupStatsPeriod%3Dauto&referrer=alerts-related-issues-issue-stream) have occurred.

Update caching, do not invalidate cached response even if expired if API returns errors, continue to fallback on latest valid cached response.

## PR Review Checklist


merino/curated_recommendations/corpus_backends/corpus_api_backend.py
```diff
@@ -276,24 +277,42 @@ async def _revalidate_cache(self, surface_id: ScheduledSurfaceId) -> list[Corpus
             if cache_key in self._cache and datetime.now() < self._expirations[cache_key]:
                 return self._cache[cache_key]
 
-            # Fetch new data from the backend.
+            # Attempt to fetch new data from the backend
             try:
                 data = await self._fetch_from_backend(surface_id)
+                if not data:  # Check if the fetched data is valid
```
efixler, line 283: This block appears logically correct to me. It is, however, repeating the same logic twice. You might think about using a 2-iteration repeat loop to be a little DRYer.

Non-blocking comment but worth considering IMO.

tests/integration/api/v1/curated_recommendations/test_curated_recommendations.py
```diff
@@ -499,6 +499,53 @@ async def test_cache_returned_on_subsequent_calls(
                 assert new_data["recommendedAt"] > initial_data["recommendedAt"]
                 assert all("NEW" in item["title"] for item in new_data["data"])
 
+    @freezegun.freeze_time("2012-01-14 00:00:00", tick=True, tz_offset=0)
+    @pytest.mark.asyncio
+    async def test_valid_cache_returned_on_error(
+        self, corpus_http_client, fixture_request_data, caplog
+    ):
+        """Test that the cache does not cache error data even if expired & returns latest valid data from cache."""
+        with freezegun.freeze_time(tick=True) as frozen_datetime:
```
efixler, line 508: 👍

tests/integration/api/v1/curated_recommendations/test_curated_recommendations.py
```diff
@@ -499,6 +499,53 @@ async def test_cache_returned_on_subsequent_calls(
...
+            async with AsyncClient(app=app, base_url="http://test") as ac:
+                # First fetch to populate cache with good data
+                initial_response = await fetch_en_us(ac)
+                initial_data = initial_response.json()
+                assert initial_response.status_code == 200
+
+                # Simulate 503 error from Corpus API
+                corpus_http_client.post.return_value = Response(
+                    status_code=503,
```
jpetto, line 517: just to be clear - 503 is the response when the prisma connection pool issue comes up?

merino/curated_recommendations/corpus_backends/corpus_api_backend.py
jpetto: what circumstances would result in `data` being falsy? is this if corpus doesn't return anything at all?

merino/curated_recommendations/corpus_backends/corpus_api_backend.py
```diff
@@ -276,24 +277,42 @@ async def _revalidate_cache(self, surface_id: ScheduledSurfaceId) -> list[Corpus
...
+                    raise ValueError("_revalidate_cache: Response is invalid or empty")
+                # Update the cache with valid data
+                self._cache[cache_key] = data
+                self._expirations[cache_key] = self.get_expiration_time()
+                return data
             except HTTPError as e:
```
jpetto, line 289: also clarifying - what corpus circumstances would cause this exception to occur?

merino/curated_recommendations/corpus_backends/corpus_api_backend.py
```diff
@@ -276,24 +277,42 @@ async def _revalidate_cache(self, surface_id: ScheduledSurfaceId) -> list[Corpus
             if cache_key in self._cache and datetime.now() < self._expirations[cache_key]:
                 return self._cache[cache_key]
 
-            # Fetch new data from the backend.
+            # Attempt to fetch new data from the backend
             try:
                 data = await self._fetch_from_backend(surface_id)
+                if not data:  # Check if the fetched data is valid
+                    raise ValueError("_revalidate_cache: Response is invalid or empty")
```
jpetto, line 284: so this will fall through to the `except Exception as e:` line below? in that case, what is returned from this function?

jpetto (commented):
some clarifying questions 😄

merino/curated_recommendations/corpus_backends/corpus_api_backend.py
katerinachinnappan: this is a very rare case but in the case there is no data or its corrupt.this is more of a safety check

tests/integration/api/v1/curated_recommendations/test_curated_recommendations.py
katerinachinnappan: I think so, 503 is returned when the server can't handle incoming requests atm.

merino/curated_recommendations/corpus_backends/corpus_api_backend.py
katerinachinnappan: good catch. I added `ValueError` to the exception with `HTTPError`. `except Exception as e` will only happen if some kind of rare-case error occurs

merino/curated_recommendations/corpus_backends/corpus_api_backend.py
```diff
@@ -276,24 +277,32 @@ async def _revalidate_cache(self, surface_id: ScheduledSurfaceId) -> list[Corpus
...
-                # http errors are expected to be rare, so a single retry attempt probably suffices.
-                data = await self._fetch_from_backend(surface_id)
+                data = await self.retry_fetch(surface_id, cache_key)
+                return data
+            except (HTTPError, ValueError) as e:
+                logger.warning(
+                    f"Exception occurred on first attempt to fetch: "
+                    f"Retrying CorpusApiBackend._fetch_from_backend once after {e}"
+                )
+
```
jpetto, line 289: would it be worth sleeping for any amount of time between retries? maybe something to tweak later...

jpetto (approved):
nice refactor! i think this is good to roll out.

merino/curated_recommendations/corpus_backends/corpus_api_backend.py
```diff
@@ -303,3 +312,15 @@ def get_expiration_time() -> datetime:
             CorpusApiBackend.cache_time_to_live_max.total_seconds(),
         )
         return datetime.now() + timedelta(seconds=time_to_live_seconds)
+
+    async def retry_fetch(
```
efixler, line 316: I'm a little confused about the asynchronicity intent/relationship between this one here and the parent call. 

What I see here is that this function is declared `async` and the parent is calling it with an `await`, which effectively make this synchronous in this context. 

There may be reasons to to implement it this way, but I wanted to double-check what your intent is here because I _think_ the intent here is to run this series of operations synchronously.

katerinachinnappan: hmm _revalidate_cache was already `async` as it was calling the `async` function `_fetch_from_backend` which performs the actual request. I just moved the common code retrying `_fetch_from_backend` to `retry_fetch`. It is  `async` as it is now the one awaiting on `_fetch_from_backend`.

merino/curated_recommendations/corpus_backends/corpus_api_backend.py
```diff
@@ -276,24 +277,32 @@ async def _revalidate_cache(self, surface_id: ScheduledSurfaceId) -> list[Corpus
...
+                except (HTTPError, ValueError) as e:
+                    logger.warning(
+                        f"Retrying CorpusApiBackend._fetch_from_backend failed: {e}. "
+                        f"Returning latest valid cached data."
+                    )
+                    # Provide the latest valid cached data if available
+                    return self._cache.get(cache_key, [])
             except Exception as e:
                 # Backoff prevents high API rate when an unforeseen error occurs.
                 await asyncio.sleep(self._backoff_time.total_seconds())
```
efixler, line 303: I think what you're expressing here is essentially `don't back off the call interval unless the call fails twice`. Is that correct?

katerinachinnappan: basically this makes sure there is a delay before the exception propagates. otherwise if errors are retried too frequently , it could lead a further strain on the API.

efixler: I think in the above reply you are referring to the caller of `_revalidate_cache` as this exception block will only, I'm pretty sure, execute _after_ the code has slept between two retries, in the case where the HTTP Error | ValueError are encountered 2x in the block above.

PR #592, opened 2024-08-01
[load test: warn]MC-1316: curated-recommendations load test with random topics

katerinachinnappan:
## References

JIRA: https://mozilla-hub.atlassian.net/browse/MC-1316

## Description
Run a load test against the /api/v1/curated-recommendations endpoint with random topics query param.

## PR Review Checklist

### DIFF
--- a/tests/load/locustfiles/locustfile.py
+++ b/tests/load/locustfiles/locustfile.py
@@ -415,7 +415,7 @@ class MerinoUser(HttpUser):
     @task(weight=421)
     def curated_recommendations_locale(self) -> None:
         """Send request to get curated recommendations, specifying random locale & 0 topics."""
-        self._request_recommendations(CuratedRecommendationsRequest(locale=choice(list(Locale))))
+        self._request_recommendations(CuratedRecommendationsRequest(locale=choice(list(Locale))), "locale")
 
     @task(weight=74)
     def curated_recommendations_random_topics(self) -> None:
@@ -427,20 +427,23 @@ class MerinoUser(HttpUser):
                 topics=self.generate_random_arr_from_enum(
                     enum_values=list(Topic), array_length=num_topics
                 ),
-            )
+            ),
+            "topics"
         )
 
-    def _request_recommendations(self, data: CuratedRecommendationsRequest) -> None:
+    def _request_recommendations(self, data: CuratedRecommendationsRequest, query_param: str | None = None,) -> None:
         """Request recommendations from Merino for the given data.
 
         Args:
             data: CuratedRecommendationsRequest object containing request data
+            query_param: query param passed in body request
         """
         with self.client.post(
             url=CURATED_RECOMMENDATIONS_API,
             json=data.model_dump(),
             headers={"User-Agent": choice(DESKTOP_FIREFOX)},
             catch_response=True,
+            name=f"{CURATED_RECOMMENDATIONS_API}{(f'?{query_param}' if query_param else '')}",
         ) as response:
             if response.status_code == 0:
                 # Do not classify as failure
--- a/tests/load/locustfiles/locustfile.py
+++ b/tests/load/locustfiles/locustfile.py
@@ -415,7 +415,9 @@ class MerinoUser(HttpUser):
     @task(weight=421)
     def curated_recommendations_locale(self) -> None:
         """Send request to get curated recommendations, specifying random locale & 0 topics."""
-        self._request_recommendations(CuratedRecommendationsRequest(locale=choice(list(Locale))), "locale")
+        self._request_recommendations(
+            CuratedRecommendationsRequest(locale=choice(list(Locale))), "locale"
+        )
 
     @task(weight=74)
     def curated_recommendations_random_topics(self) -> None:
@@ -428,10 +430,14 @@ class MerinoUser(HttpUser):
                     enum_values=list(Topic), array_length=num_topics
                 ),
             ),
-            "topics"
+            "topics",
         )
 
-    def _request_recommendations(self, data: CuratedRecommendationsRequest, query_param: str | None = None,) -> None:
+    def _request_recommendations(
+        self,
+        data: CuratedRecommendationsRequest,
+        query_param: str | None = None,
+    ) -> None:
         """Request recommendations from Merino for the given data.
 
         Args:
--- a/tests/load/locustfiles/locustfile.py
+++ b/tests/load/locustfiles/locustfile.py
@@ -22,6 +22,7 @@ from locust.exception import StopUser
 from locust.runners import MasterRunner
 from pydantic import BaseModel
 
+from merino.curated_recommendations.corpus_backends.protocol import Topic
 from merino.providers.adm.backends.protocol import SuggestionContent
 from merino.providers.adm.backends.remotesettings import (
     RemoteSettingsBackend,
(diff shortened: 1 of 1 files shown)


tests/load/locustfiles/locustfile.py
```diff
@@ -413,10 +414,20 @@ def weather_suggestions(self) -> None:
         self._request_suggestions(query, providers, headers)
 
     @task(weight=495)
```
Trinaa, line 416: The curated_recommendations_local and curated_recommendations_topics tasks are each set with a `weight` of 49.5% here, putting our total to 149.5%, which I think would direct way too much load towards the curated recommendations endpoint. 

Could you make a change so that the weight of the two tasks adds up to 49.5%? Perhaps `247` and `248`? Or if you have estimates about the proportion of traffic that would be using locale vs topics you could split accordingly.

tests/load/locustfiles/locustfile.py
```diff
@@ -449,6 +460,17 @@ def _request_recommendations(self, data: CuratedRecommendationsRequest) -> None:
...
 
+    @staticmethod
+    def generate_random_arr_from_enum(enum_class, end_range: int):
+        """Generate an array of random enum values."""
+        # Get all possible values of the enum
+        enum_values = list(enum_class)
+        # Generate a random length for the array
+        array_length = randint(a=1, b=end_range)
+        # Create an array with random values from the enum
+        random_enum_array = [choice(enum_values) for _ in range(array_length)]
```
Trinaa, line 471: Just want to make sure that repetitions are okay in this array?

tests/load/locustfiles/locustfile.py
```diff
@@ -413,10 +414,20 @@ def weather_suggestions(self) -> None:
         self._request_suggestions(query, providers, headers)
 
     @task(weight=495)
-    def curated_recommendations(self) -> None:
-        """Send request to get curated recommendations."""
+    def curated_recommendations_locale(self) -> None:
+        """Send request to get curated recommendations, specifying random locale."""
         self._request_recommendations(CuratedRecommendationsRequest(locale=choice(list(Locale))))
```
Trinaa, line 419: Do you need to distinguish between the locale and topics requests in the load test reporting? At the moment `_request_recommendations` groups them together. If yes, the `name` parameter can be used, this is already done in `_request_suggestions`

Trinaa (commented):
@katerinachinnappan I listed a few concerns, let me know what you think. ES is still down, so if we want to run a load test with these changes we will have to rebase the 'no_wikipedia_task' branch.

tests/load/locustfiles/locustfile.py
efixler: I don't think we need to distinguish locale. Topic selection adds complexity/load to the request and hits some different code paths, I don't _think_ that's true for locale.

tests/load/locustfiles/locustfile.py
efixler: 100% here is 100% of what exactly?

tests/load/locustfiles/locustfile.py
jpetto: pretty sure we don't want repetitions here...

tests/load/locustfiles/locustfile.py
katerinachinnappan: well there is some code around boosting items when topics are passed, if no topics are passed then that code gets ignored

tests/load/locustfiles/locustfile.py
```diff
@@ -449,6 +460,17 @@ def _request_recommendations(self, data: CuratedRecommendationsRequest) -> None:
             # fields which will be reported as a failure in Locust's statistics.
             CuratedRecommendationsResponse(**response.json())
 
+    @staticmethod
+    def generate_random_arr_from_enum(enum_class, end_range: int):
+        """Generate an array of random enum values."""
+        # Get all possible values of the enum
+        enum_values = list(enum_class)
```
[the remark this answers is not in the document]
katerinachinnappan: good point. I think even instead of hardcoding `enum_values` in the method, I will get rid of `enum_class` & pass `enum_values`, that way allowing this method to be used by other calls if needed.

tests/load/locustfiles/locustfile.py
katerinachinnappan: I will use `random.sample`

tests/load/locustfiles/locustfile.py
katerinachinnappan: I can leave Mathijs test with random locales & maybe test the topics tests with en-us locale instead of choosing a random one every time?

katerinachinnappan: i've updated the random topics test to test for random# of topics between 1 & 4 as 4 is the max # of topics user can pass.

tests/load/locustfiles/locustfile.py
Trinaa: I am re-reading what I wrote and I think i can do better.

The weight translates to the probability of a task being chosen for execution by the user or in other words the amount of request traffic that would be directed to an endpoint. We try and have it add up to a base 10 number so that it is easier to calculate/understand the percentage.

**Example with the old distribution and an 8k request/s total average load test shape**

49.5% curated_recommendations --> 3.96k requests/s at peak
49.5% weather_suggestions --> 3.96k requests/s at peak
0.2% adm_suggestions --> 16 requests/s at peak
0.2% amo_suggestions --> 16 requests/s at peak
0.2% dynamic_wikipedia_suggestions --> 16 requests/s at peak
0.2% faker_suggestions --> 16 requests/s at peak
0.2% top_picks_suggestions --> 16 requests/s at peak

**Example with the distribution in this PR and an 8k request/s total average load test shape**

33.1% curated_recommendations_topics --> 2.65k/s at peak
33.1% curated_recommendations_locale --> 2.65k/s at peak
33.1% weather_suggestions --> 2.65k/s at peak
0.15% adm_suggestions --> 12/s at peak
0.15% amo_suggestions --> 12/s at peak
0.15% dynamic_wikipedia_suggestions --> 12/s at peak
0.15% faker_suggestions --> 12/s at peak
0.15% top_picks_suggestions --> 12/s at peak

If we check this in more than 66% of the traffic will be directed to curated_recommendations, but last time the teams got together my understanding is that we expect the traffic to be about 50/50 between recommendations and suggestions.

efixler: Thanks @Trinaa this makes sense.

@katerinachinnappan I think this means we want to follow @Trinaa's original suggestion, since it simulates server load distribution most accurately.

jpetto: this is incredibly helpful. ty!

tests/load/locustfiles/locustfile.py
Trinaa: Just for clarity, I'm talking about the final reporting structure. The way things are coded right now, the stats for curated recommendations will continue to be aggregated together in 1 row. 
Suggest for example have decided to split by provider. 

Example:

image

tests/load/locustfiles/locustfile.py
katerinachinnappan: i've updated the locale load test task weight to 421 & topics load test to 74. This is roughly 85% of allocated traffic for locale & remaining 15% for topics. Since we won't expect a lot of traffic for topics initially, lets start with these numbers.

tests/load/locustfiles/locustfile.py
katerinachinnappan: in this case we do want to differentiate, otherwise whats the point of having 2 tests, right?

ncloudioj (approved):
LGTM.

PR #593, opened 2024-08-07
[DISCO-2909] Collapse geo fields 'region'/'alternative_regions' into one

tiftran:
## References

JIRA: [DISCO-2909](https://mozilla-hub.atlassian.net/browse/DISCO-2909)

## Description
`region` was first introduced, we then realized we needed `alternative_regions` for certain countries for accuweather.
This is just to declutter and combine the two fields to `regions` where the first region in the list is the `most_specific` subdivision.

## PR Review Checklist

[DISCO-2909]: https://mozilla-hub.atlassian.net/browse/DISCO-2909?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


merino/middleware/geolocation.py
```diff
@@ -25,8 +25,8 @@ class Location(BaseModel):
 
     country: Optional[str] = None
     country_name: Optional[str] = None
-    region: Optional[str] = None
-    region_name: Optional[str] = None
+    regions: Optional[list[str]] = []
```
ncloudioj, line 28: Let's default to `None` for both regions and region_names to be consistent with other fields.

merino/middleware/geolocation.py
```diff
@@ -39,6 +39,22 @@ def normalize_string(input_str) -> str:
     return unicodedata.normalize("NFKD", input_str).encode("ascii", "ignore").decode("ascii")
 
 
+def get_regions(subdivisions) -> list[str]:
```
ncloudioj, line 42: Since we already have this helper, maybe we can handle all the edge cases here.

Per [geoip2](https://github.com/maxmind/GeoIP2-python/blob/e8bed66b51bef3860b547a85a4f109faedf5e7e3/geoip2/records.py#L535), `Subdivisions` is a subclass of `tuple`. `most_specific` is the last element of the tuple (the subdivisions are sorted in the ascending order w.r.t specificity). If `subdivisions` is empty, `most_specific` would be an empty `Subdivision` instead of `None`. Therefore, we can write the helper as follows:

```py
def get_regions(subdivisions) -> Optional[list[str]]:
    return [s.iso_code for s in reversed(subdivisions)] or None

def get_region_names(subdivisions) -> Optional[list[str]]:
    return [s.names.get("en") for s in reversed(subdivisions)] or None

# We can simply do this down below
    ...
    regions=get_regions(record.subdivisions)
    region_names=get_region_names(record.subdivisions)

```

ncloudioj (approved):
Looks good, r=wc.

merino/middleware/geolocation.py
tiftran: lol, somehow feels like cheating

PR #597, opened 2024-08-12
[DISCO-2945]: Handle get location by geolocation exceptions

Herraj:
## References

JIRA: [DISCO-2945](https://mozilla-hub.atlassian.net/browse/DISCO-2945)

## Description
We are seeing SSLErrors, EndOfStream and other errors for location endpoints for Accuweather. These errors are recent and unexpected. I've also reinforced the `current conditions` and `forecast` try-excepts too. These two aren't erroring out for now but I've added them as a precaution.

Following the existing pattern, I've added these excepts in the individual functions instead of the root `get_request` function. The `get_request` function would simply throw them up and the accuweather provider will catch them as `BackendError`s

## PR Review Checklist

[DISCO-2945]: https://mozilla-hub.atlassian.net/browse/DISCO-2945?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/merino/providers/weather/backends/accuweather.py
+++ b/merino/providers/weather/backends/accuweather.py
@@ -10,7 +10,6 @@ from enum import Enum
 from typing import Any, Callable, NamedTuple, cast
 
 import aiodogstatsd
-import httpx
 from dateutil import parser
 from httpx import URL, AsyncClient, HTTPError, InvalidURL, Response
 from pydantic import BaseModel, ValidationError
@@ -672,6 +671,11 @@ class AccuweatherBackend:
             raise AccuweatherError(
                 f"Unexpected location response from: {url_path}, city: {city}"
             ) from error
+        except Exception as exc:
+            raise AccuweatherError(
+                f"Unexpected error occurred when requesting location by geolocation from "
+                f"Accuweather: {exc.__class__.__name__}"
+            ) from exc
 
         return AccuweatherLocation(**response) if response else None
 
@@ -697,6 +701,11 @@ class AccuweatherBackend:
             raise AccuweatherError(
                 f"Unexpected current conditions response, Url: {self.url_current_conditions_path.format(location_key=location_key)}"
             ) from error
+        except Exception as exc:
+            raise AccuweatherError(
+                f"Unexpected error occurred when requesting current conditions from Accuweather:"
+                f" {exc.__class__.__name__}"
+            ) from exc
 
         return (
             CurrentConditionsWithTTL(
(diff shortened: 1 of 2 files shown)


merino/providers/weather/backends/accuweather.py
```diff
@@ -684,6 +683,11 @@ async def get_location_by_geolocation(
             raise AccuweatherError(
                 f"Unexpected location response from: {url_path}, city: {city}"
             ) from error
+        except Exception as exc:
+            raise AccuweatherError(
+                f"Unexpected error occurred when requesting location by geolocation from "
+                f"Accuweather: {exc}"
+            ) from exc
```
tiftran, line 690: because I made the mistake last time.... do we know what `exc` could spit out? specifically could it spit out the url path with the api key?

tests/unit/providers/weather/backends/test_accuweather.py
```diff
@@ -8,12 +8,14 @@
...
 import json
 import logging
+from ssl import SSLError
 from typing import Any, Awaitable, Callable, Optional, cast
 from unittest.mock import AsyncMock
 
 import pytest
 import freezegun
 from httpx import AsyncClient, HTTPError, Request, Response
+from anyio import EndOfStream
```
ncloudioj, line 18: I believe that `EndOfStream` is unintentionally leaked by `httpx`. Also, `anyio` is an indirect dependency from httpx, which might get dropped by `httpx` in the future. Let's not use it in our code, feel free to fake another exception (it's okay to use the ones not observed in prod) for testing.

merino/providers/weather/backends/accuweather.py
ncloudioj: Nice catch, Tif. To avoid that, I think we can only include the error class type for now, wdyt?

merino/providers/weather/backends/accuweather.py
```diff
@@ -787,9 +800,14 @@ async def get_location_completion(
             ):
                 response: Response = await self.http_client.get(url_path, params=params)
                 response.raise_for_status()
-        except httpx.HTTPError as exc:
-            logger.warning(f"Failed to get location completion from Accuweather: {exc}")
-            return None
+        except HTTPError as http_error:
+            raise AccuweatherError(
+                f"Failed to get location completion from Accuweather: " f"{http_error}"
```
ncloudioj, line 805: The `http_error` is likely to include the credentials, let's be extra mindful here as well.

merino/providers/weather/backends/accuweather.py
```diff
@@ -709,6 +713,11 @@ async def get_current_conditions(self, location_key: str) -> CurrentConditionsWi
             raise AccuweatherError(
                 f"Unexpected current conditions response, Url: {self.url_current_conditions_path.format(location_key=location_key)}"
             ) from error
+        except Exception as exc:
```
ncloudioj, line 716: Looks like those catch-all branches are not being tested, can you please check the test coverage report and fix them accordingly. Alternatively, I think it's fine to opt out the coverage check here as just intentionally catch all unexpected errors and convert them into `AccuweatherError`s.

ncloudioj (changes requested):
Looks good. A few minor changes are requested.

ncloudioj (approved):
Looks great, thanks!

PR #603, opened 2024-08-16
[DISCO-2910] Add mapping for successful region lookups

tiftran:
## References

JIRA: [DISCO-2910](https://mozilla-hub.atlassian.net/browse/DISCO-2910)

## Description
For successful calls to accuweather to get location key. We want to store it, so the next time the same city is queried, we don't have to cycle through all regions.

## PR Review Checklist

[DISCO-2910]: https://mozilla-hub.atlassian.net/browse/DISCO-2910?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


merino/providers/weather/backends/accuweather/pathfinder.py
```diff
@@ -53,7 +61,10 @@ async def explore(
     """
     for country, region, city in compass(location):
         res = await probe(country, region, city)
-        if res is not None:
+
+        if res not in (None, []):
```
tiftran, line 65: I believe when we fetch from the cache, we can get an empty list too... which was causing me some trouble

merino/providers/weather/backends/accuweather/pathfinder.py
```diff
@@ -8,6 +8,8 @@
 MaybeStr = Optional[str]
 Triplet = tuple[MaybeStr, MaybeStr, MaybeStr]
 
+SUCCESSFUL_REGIONS_MAPPING: dict[tuple[str, str], str | None] = {}
```
tiftran, line 11: @ncloudioj  just wanted to double check if i understood correctly when you say learn as it goes?

Given the assumption, i opted not to hardcode the countries of interest, technically at some point, they become irrelevant as the map grows?

ncloudioj: Ah, so your proposal will store all the successful triplets in the map. We can certainly try that for its simplicity, though two potential downsides I can see:
- This map will grow over time and will store many redundant records. For example, for all US/CA cities, we know that each city only has 1 region, so we could return "country/region/city" directly, no need to store/consult this map at all. I believe this is the case for other countries
- As `pathfinder.explore()` is called for both cache fetching and API requesting, either one succeeds, it will write to this map which is a bit wasteful. I think we can do that once instead (perhaps upon API request successes)

I was thinking about combining those heuristics (known & static) with other rules (unknown & dynamic) that it learns at the runtime in `pathfinder.compass()`. Something like:

```python
country = location.country
regions = location.regions
city = location.city

match (country, city):
    case ("US" | "CA" | ..., _):
        yield country, regions[0], city  # use the most specific region
    case ("IT" | "ES" | ..., _):
        yield country, regions[-1], city  # use the least specific region
    case other_static_rules:  # to be added later
        pass
    case (country, city) if (country, city) in SUCCESSFUL_REGIONS_MAPPING:  # dynamic rules we've learned
        yield country, S_R_MAPPING[(country, city)], city
    case _:  # Fall back to try all triplets
        if regions is not None:
            regions = [*regions, None]
        else:
            regions = [None]

        for regions in regions:
            yield country, region, city
```

WDYT?

tiftran: ooh much fancier

tests/unit/providers/weather/backends/test_accuweather.py
```diff
@@ -966,83 +966,16 @@ async def test_get_weather_report(
...
+@pytest.mark.parametrize(
+    "regions_mapping",
+    [
+        {("US", "SanFrancisco"), "CA"},
+        {("US", "SanFrancisco"), None},
+    ],
+    ids=["Region", "Fallback"],
+)
 @pytest.mark.asyncio
-async def test_get_weather_report_with_alternative_region(
```
tiftran, line 970: theoretically, since the region decision logic has moved out to a separate module, these cases are redundant?

merino/providers/weather/backends/accuweather/pathfinder.py
```diff
@@ -21,14 +23,28 @@ def compass(location: Location) -> Generator[Triplet, None, None]:
...
     # TODO(nanj): add more heuristics to here.
 
-    # Append None as the fallback since AccuWeather can take params w/o the region code
-    if location.regions is not None:
-        regions = [*location.regions, None]
+    country = location.country
+    regions = location.regions
+    city = location.city
+
+    if regions and country and city:
```
tiftran, line 30: we needa be sure regions is not None before looking at the US/CA or euro cases

merino/providers/weather/backends/accuweather/backend.py
```diff
@@ -60,7 +61,7 @@
     local location_key = redis.call("GET", KEYS[1])
 
     if not location_key then
-        return {}
+        return nil
```
tiftran, line 64: hold on, we had this convo before right, if i remember, nil will return an empty list right?

Herraj: I can't find the link but `nil` is returned as empty or if it's an element in list/array, it will truncate the rest after it. 

If you wan't python to receive `None`, you could return `false` instead of `nil`. I haven't tested returning `nil` on it's own, just as a part of a list / object

ncloudioj: Oh, I think I was suggesting to do this in `_fetch_from_cache()`.

```py
async def _fetch_from_cache(...):
     ...
     cached_data: list = await self.cache.run_script(...)
     # return cached_data
     return cached_data if cached_data else None
```

merino/providers/weather/backends/accuweather/backend.py
```diff
@@ -525,7 +526,7 @@ async def get_weather_report_with_location_key(self, location_key) -> WeatherRep
 
     async def _fetch_from_cache(
         self, country: str | None, region: str | None, city: str | None
-    ) -> list[bytes | None] | None:
+    ) -> list[bytes] | None:
```
ncloudioj, line 529: Can you confirm this? Like if the script return ["foo", null, "bar"] from Redis, will we get ["foo", "bar"] instead?

merino/providers/weather/backends/accuweather/backend.py
```diff
@@ -719,6 +720,9 @@ async def get_location_by_geolocation(
                 f"Accuweather: {exc.__class__.__name__}"
             ) from exc
 
+        # record the region that gave a location
+        if response and country and city:
+            SUCCESSFUL_REGIONS_MAPPING[(country, city)] = region
```
ncloudioj, line 725: I'd recommend not mutating the module variable this way as it will make it hard to reason about. We can provide getter&setter instead.

Also, I guess this will insert all positive triplets into the map including US/CA/IT etc. which would be a lot more to store. Shall we exclude those known patterns to save some space?

Alsox2, it'll be great to have a metric to gauge the map size for peace of mind. This can be done in a follow up.

merino/providers/weather/backends/accuweather/pathfinder.py
```diff
@@ -21,14 +23,28 @@ def compass(location: Location) -> Generator[Triplet, None, None]:
...
+            case ("US" | "CA", _):
+                yield country, regions[0], city  # use the most specific region
+            case ("IT" | "ES" | "GR", _):
+                yield country, regions[-1], city  # use the least specific region
+            case (country, city) if (
+                country,
+                city,
+            ) in SUCCESSFUL_REGIONS_MAPPING:  # dynamic rules we've learned
+                yield country, SUCCESSFUL_REGIONS_MAPPING[(country, city)], city
+
```
ncloudioj, line 41:
```suggestion

```

tests/unit/providers/weather/backends/test_pathfinder.py
```diff
@@ -0,0 +1,40 @@
...
+
+@pytest.mark.parametrize(
+    ("location", "expected_tuple"),
+    [
+        (Location(country="CA", regions=["BC"], city="Vancouver"), ("CA", "BC", "Vancouver")),
+        (Location(country="IT", regions=["MT", "77"], city="Matera"), ("IT", "77", "Matera")),
+        (Location(country="GB", regions=["ENG", "HWT"], city="London"), ("GB", "LND", "London")),
+        (Location(country="BR", regions=["DF"], city="Brasilia"), ("BR", "DF", "Brasilia")),
+        (Location(country="IE", regions=None, city="Dublin"), ("IE", None, "Dublin")),
+    ],
```
ncloudioj, line 24: Love these tests!

merino/providers/weather/backends/accuweather/backend.py
tiftran: oooh, nope this is an oopsies

merino/providers/weather/backends/accuweather/backend.py
```diff
@@ -719,6 +723,12 @@ async def get_location_by_geolocation(
                 f"Accuweather: {exc.__class__.__name__}"
             ) from exc
 
+        # record the region that gave a location
+        if response and country and city:
+            set_region_mapping(country, city, region)
+            self.metrics_client.gauge(
```
tiftran, line 729: increment felt weird

ncloudioj: Gauging this for every success would be a little unnecessary and expensive. I think we can spawn off a cron job and periodically gauge the map size in the background.

Sorry for not being clear with my previous comment. Thought about filing a ticket and noting that in detail, really should've noted that earlier.

That said, not need to do that in this patch. Could you file a follow-up ticket to track that? TIA!

merino/providers/weather/backends/accuweather/pathfinder.py
```diff
@@ -57,3 +72,21 @@ async def explore(
...
+def set_region_mapping(country: str, city: str, region: str | None):
+    """Set country, city, region into SUCCESSFUL_REGIONS_MAPPING
+    that don't fall in countries where region can be determined.
+
+    Params:
+      - country {str}: country code
+      - city {str}: city name
+      - region {str}: admin code
+    """
+    if country not in ("CA", "US", "IT", "ES", "GR"):
```
ncloudioj, line 91: This list will grow over time, let's define a constant for it.

```py
REGION_MAPPING_EXCLUSIONS: frozenset = frozenset(["CA", "ES", "GR", "IT", "US"])
```

merino/providers/weather/backends/accuweather/pathfinder.py
```diff
@@ -57,3 +72,21 @@ async def explore(
...
+
+
+def set_region_mapping(country: str, city: str, region: str | None):
+    """Set country, city, region into SUCCESSFUL_REGIONS_MAPPING
+    that don't fall in countries where region can be determined.
+
+    Params:
+      - country {str}: country code
+      - city {str}: city name
+      - region {str}: admin code
```
ncloudioj, line 89:
```suggestion
      - region {str | None}: region code
```

merino/providers/weather/backends/accuweather/pathfinder.py
```diff
@@ -57,3 +72,21 @@ async def explore(
             return res
 
     return None
+
+
+def get_region_mapping() -> dict[tuple[str, str], str | None]:
```
ncloudioj, line 77: Suggestion: let's also add `clear_region_mapping()` and then no need to patch `SUCCESSFUL_REGIONS_MAPPING` in the test down below. Just call `set_region_mapping()` at setup and `clear_region_mapping()` at teardown.

With the clear_map API, we could opportunistically reset the map (to ease the memory pressure) when the cron job detects an excessively large region mapping.

WDYT?

tests/unit/providers/weather/backends/test_pathfinder.py
```diff
@@ -0,0 +1,37 @@
...
+    ids=[
+        "Specific Region Country",
+        "Alternative Region Country",
+        "Successful Region Mapping Pair",
+        "Fallback with Region",
+        "Fallback No Region",
+    ],
+)
+def test_compass(location: Location, expected_tuple: Tuple, mocker: MockFixture) -> None:
```
ncloudioj, line 33: `mocker` can be dropped now.

tests/unit/providers/weather/backends/test_pathfinder.py
```diff
@@ -0,0 +1,37 @@
...
+    """Test country that returns the most specific region."""
+    set_region_mapping("GB", "London", "LND")
+
+    assert next(compass(location)) == expected_tuple
```
ncloudioj, line 37: Don't forget to reset the mapping. Otherwise, it could interfere with other tests.

merino/providers/weather/backends/accuweather/pathfinder.py
```diff
@@ -57,3 +72,26 @@ async def explore(
...
```
ncloudioj, line 77: Is this used anywhere? If not, we can drop it.

ncloudioj (approved):
Thanks! Just a few minor nits.

Pathfinder is now more useful and smarter \o/

merino/providers/weather/backends/accuweather/pathfinder.py
tiftran: ah you're right

PR #545, opened 2024-06-25
[load test: warn]: [DISCO-2324] Use redis testcontainers for integration tests

Herraj:
## References

JIRA: [DISCO-2324](https://mozilla-hub.atlassian.net/browse/DISCO-2324)

## Description
Refactor existing unit tests that mock `redis` to use the real `testcontainer` instance.

## PR Review Checklist

- [x] This PR conforms to the [Contribution Guidelines](https://github.com/mozilla-services/merino-py/blob/main/CONTRIBUTING.md)
- [x] The PR title starts with the JIRA issue reference, format `[DISCO-####]`, and has the same title (if applicable)
- [ ] `[load test: (abort|warn)]` keywords are applied (if applicable)
- [ ] [Documentation](https://github.com/mozilla-services/merino-py/tree/main/docs) has been updated (if applicable)
- [ ] [Functional and performance test](https://github.com/mozilla-services/merino-py/blob/main/docs/dev/testing.md) coverage has been expanded and maintained (if applicable)

[DISCO-2324]: https://mozilla-hub.atlassian.net/browse/DISCO-2324?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #496, opened 2024-03-27
test: add integration tests for GcsUploader using testcontainers library

Herraj:
## References

JIRA: [DISCO-2767](https://mozilla-hub.atlassian.net/browse/DISCO-2767)

## Description
Add integration tests for `GcsUploader` using `testcontainers` library

We had to switch to using `machine` executors for running the `integration-tests` check since the `testcontainers` test spin up a docker container at run time. This wasn't working if we were using the docker container to run those tests in. Surprisingly, setting up a python environment on a bare bones ubuntu VM and running the tests seems to be slightly faster than using a pre-built python docker image

In the `project.toml` file, you'll see that the new dev dependencies are added under `tool.poetry.group.dev.dependencies` and not `tool.poetry.dev.dependencies`. This is because the latter has been deprecated. Poetry wants us to use `...group...` now. Can make this change for the rest of the dev dependencies in the next PR.

## PR Review Checklist

- [x] This PR conforms to the [Contribution Guidelines](https://github.com/mozilla-services/merino-py/blob/main/CONTRIBUTING.md)
- [x] The PR title starts with the JIRA issue reference, format `[DISCO-####]`, and has the same title (if applicable)
- [ ] `[load test: (abort|warn)]` keywords are applied (if applicable)
- [x] [Documentation](https://github.com/mozilla-services/merino-py/tree/main/docs) has been updated (if applicable)
- [ ] [Functional and performance test](https://github.com/mozilla-services/merino-py/blob/main/docs/dev/testing.md) coverage has been expanded and maintained (if applicable)

[DISCO-2767]: https://mozilla-hub.atlassian.net/browse/DISCO-2767?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #606, opened 2024-08-26
[MC-1302] feat(curated-recommendations): add Thompson sampling [load test: warn]

mmiermans:
## References
JIRA: [MC-1302](https://mozilla-hub.atlassian.net/browse/MC-1302)

## Description
Curated recommendations are sorted based on [Thompson sampling](https://en.wikipedia.org/wiki/Thompson_sampling), before publisher spreading is applied.

Changes:
- extract curated recommendations models and rankers in separate modules to simplify provider module
- call Thompson sampling in provider
- get engagement from list of blobs, instead of a single one
- fix naming mismatch of impression_count and click_count with JSON in GCS
- turn GCS engagement parameters into settings

## PR Review Checklist


merino/curated_recommendations/engagement_backends/protocol.py
```diff
@@ -13,14 +13,14 @@ class Engagement(BaseModel):
...
-    impressions: int
+    click_count: int
+    impression_count: int
 
 
 class EngagementBackend(Protocol):
     """Protocol for Engagement backend that the provider depends on."""
 
-    def __getitem__(self, scheduled_corpus_item_id: str) -> Engagement:
+    def get(self, scheduled_corpus_item_id: str) -> Engagement | None:
```
mmiermans, line 23: Just a matter of style, but the `EngagementBackend[id]` syntax seemed like it could be confusing to me, on second thought.

merino/curated_recommendations/engagement_backends/gcs_engagement.py
```diff
@@ -32,102 +35,98 @@ def __init__(
...
+        start_date = (datetime.now(timezone.utc) - timedelta(days=1)).strftime("%Y%m%d%H")
+        start_offset = f"{self.blob_prefix}{start_date}"
+        blobs = list(bucket.list_blobs(start_offset=start_offset))
+
+        if not blobs:
+            logger.error("Curated recommendations engagement blobs not found for start_offset")
+            return
+
+        # Find the most recent blob by the last modified timestamp.
+        blob = max(blobs, key=lambda b: b.updated)
```
ncloudioj, line 105: Suggestion: the producer can duplicate the most recent file and name it "engagement_latest.json" so that Merino only needs to fetch that file. Optionally, the last modification timestamp can be used to tell whether the file gets updated compared to the last fetch.

merino/curated_recommendations/rankers.py
```diff
@@ -0,0 +1,43 @@
+"""Algorithms for ranking curated recommendations."""
+
+from merino.curated_recommendations.engagement_backends.protocol import EngagementBackend
+from merino.curated_recommendations.models import CuratedRecommendation
+from scipy.stats import beta
+
+DEFAULT_ALPHA_PRIOR = 188  # beta * P99 German NewTab CTR for 2023-03-28 to 2023-04-05 (1.5%)
```
ncloudioj, line 7: Would we ever use different priors for different environments? If so, these can be configured in the config files.

tests/integration/api/v1/curated_recommendations/engagement_backends/test_gcs_engagement.py
```diff
@@ -16,25 +16,47 @@
...
 
 @pytest.fixture
-def mock_gcs_blob(mock_gcs_bucket):
-    """Return a mock GCS Blob instance"""
-    blob = MagicMock(spec=Blob)
-    mock_gcs_bucket.blob.return_value = blob
-    return blob
+def mock_gcs_client(mock_gcs_bucket):
+    """Return a mock GCS Client instance"""
+    client = MagicMock(spec=Client)
```
ncloudioj, line 57: FYI, mocking services in tests could hide bugs inadvertently. Hence we'd encourage folks to use the actual services or emulators via Testcontainers ([ADR](https://github.com/mozilla-services/merino-py/blob/main/docs/adr/0003-test-containers.md)) for integration testing if possible.

See [here](https://github.com/mozilla-services/merino-py/pull/496) and [here](https://github.com/mozilla-services/merino-py/pull/545) for some examples.

ncloudioj (approved):
LGTM r=nanj.

tests/integration/api/v1/curated_recommendations/test_curated_recommendations.py
```diff
@@ -71,10 +77,50 @@ def fixture_mock_corpus_backend(corpus_http_client: AsyncMock) -> CorpusApiBacke
...
 
 
+class MockEngagementBackend(EngagementBackend):
+    """Mock class implementing the protocol for EngagementBackend."""
+
+    def get(self, scheduled_corpus_item_id: str) -> Engagement | None:
+        """Return random click and impression counts based on the scheduled corpus id."""
+        rng = np.random.default_rng(seed=int.from_bytes(scheduled_corpus_item_id.encode()))
+
+        # 50% chance of no engagement data being available.
```
mmiermans, line 87: This comment should be on line 96.

tests/integration/api/v1/curated_recommendations/test_curated_recommendations.py
```diff
@@ -71,10 +77,50 @@ def fixture_mock_corpus_backend(corpus_http_client: AsyncMock) -> CorpusApiBacke
...
+        if scheduled_corpus_item_id == "50f86ebe-3f25-41d8-bd84-53ead7bdc76e":
+            # Give one item 100% click-through rate to put it on top with high certainty.
+            return Engagement(
+                scheduled_corpus_item_id=scheduled_corpus_item_id,
+                click_count=1000000,
+                impression_count=1000000,
+            )
```
mmiermans, line 94: Exclude index 0 from linear regression.

merino/curated_recommendations/engagement_backends/gcs_engagement.py
```diff
@@ -32,102 +35,98 @@ def __init__(
...
+        return self._cache.get(scheduled_corpus_item_id)
 
     async def _update_task_async(self) -> None:
-        """Run _update_task in a thread to prevent blocking the event loop
-
-        A thread is used because the Google Cloud library does not support asyncio.
-        Alternatively, the gcloud-aio-storage library provides asyncio access to GCS.
-        """
-        await asyncio.to_thread(self._update_task)
+        """Run _update_engagement_task in a thread to prevent blocking the event loop"""
```
mmiermans, line 87: Link to GCS docs

merino/curated_recommendations/engagement_backends/gcs_engagement.py
```diff
@@ -32,102 +35,98 @@ def __init__(
...
+            data = blob.download_as_text()
+            engagement_data = self._parse_data(data)
+            self._cache = engagement_data
+            self.last_updated = blob.updated
+            self.metrics_client.gauge(
+                "recommendation.engagement.count", value=len(engagement_data)
+            )
+
+        # Report the staleness of the engagement data in seconds.
+        if self.last_updated != LAST_UPDATED_INITIAL_VALUE:
```
mmiermans, line 125: This fixes `datetime.min` not being comparable with `self.last_updated` because the latter has a timezone.

merino/configs/production.toml
```diff
@@ -26,3 +26,12 @@ gcs_bucket = "merino-images-prodpy"
...
 domain_data_source = "local"
+
+[production.curated_recommendations.gcs_engagement]
+# MERINO__CURATED_RECOMMENDATIONS__GCS_ENGAGEMENT__BUCKET_NAME
+# GCS bucket that contains aggregate engagement data
+bucket_name = "merino-airflow-data-prodpy"
+
+# MERINO__CURATED_RECOMMENDATIONS__GCS_ENGAGEMENT__GCP_PROJECT
+# GCP project name where the GCS bucket lives.
+gcp_project = "moz-fx-merino-prod-1c2f"
```
mmiermans, line 37: Docs on how to create a local config for development: https://github.com/mozilla-services/merino-py/blob/main/docs/dev/index.md#create-a-local-override

merino/curated_recommendations/models.py
```diff
@@ -0,0 +1,99 @@
+"""Curated Recommendations provider models"""
+
+import hashlib
+from enum import unique, Enum
+from typing import Annotated
+
+from pydantic import Field, model_validator, BaseModel
+
+from merino.curated_recommendations.corpus_backends.protocol import CorpusItem, Topic
```
mmiermans, line 9: Move Topic here.

merino/curated_recommendations/models.py
```diff
@@ -0,0 +1,99 @@
+"""Curated Recommendations provider models"""
```
mmiermans, line 1: Rename to protocol.py and document that these are the top-level models for the request and response.

merino/curated_recommendations/engagement_backends/gcs_engagement.py
mmiermans: Good prompt. That's what I expected we'd do, and maybe that wasn't communicated well with data engineering. I'll break it out into [MC-1437](https://mozilla-hub.atlassian.net/browse/MC-1437) so we're not blocked by Airflow changes.

merino/curated_recommendations/models.py
mmiermans: @katerinachinnappan Moving Topic here would result in a circular import. You asked a good question: What distinguishes `Topic` from `Locale`? I didn't think about it at the time, but `Topic` is a type that comes from the corpus backend, while `Locale` exists only at the provider input level.

merino/curated_recommendations/rankers.py
mmiermans: No, I don't expect we will. That feels like a good guideline for when something should be in a config file.

merino/curated_recommendations/models.py
katerinachinnappan: Ah, if circular import is occurring then yes, let’s keep it where it is

merino/curated_recommendations/engagement_backends/gcs_engagement.py
```diff
@@ -92,9 +92,11 @@ def _update_task(self):
         """Task to update the cache with the latest engagement data from GCS."""
         bucket = self.storage_client.bucket(self.bucket_name)
 
-        # Only list files from the last day (about 100) to reduce memory usage.
+        # Only list files from the last day (about 100) to reduce memory usage and requests to GCS.
         start_date = (datetime.now(timezone.utc) - timedelta(days=1)).strftime("%Y%m%d%H")
         start_offset = f"{self.blob_prefix}{start_date}"
+        # Filter results to blobs whose names are lexicographically equal to or after start_offset.
+        # https://cloud.google.com/python/docs/reference/storage/latest/google.cloud.storage.bucket.Bucket#google_cloud_storage_bucket_Bucket_list_blobs
```
katerinachinnappan, line 99: 👍

katerinachinnappan (approved):
👍 👍

PR #611, opened 2024-08-29
[DISCO-2962] - Add cron job to gauge size of the region look up map

tiftran:
## References

JIRA: [DISCO-2962](https://mozilla-hub.atlassian.net/browse/DISCO-2962)

## Description
We have a look up map to help with retrieving the correct region for a given city, made changes so we can see the size of the map periodically. Debated whether we should log the contents of the map, but decided to hold off, until we look at the size?

## PR Review Checklist

[DISCO-2962]: https://mozilla-hub.atlassian.net/browse/DISCO-2962?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/merino/configs/default.toml
+++ b/merino/configs/default.toml
@@ -218,6 +218,8 @@ connect_timeout_sec = 3.0
 # Global Redis settings. The weather provider optionally uses Redis to cache weather suggestions.
 # In the form of `redis://localhost:6379`.
 
+# MERINO_PROVIDERS__ACCUWEATHER__CRON_INTERVAL_SEC
+cron_interval_sec = 21600 # 6 hours
 
 [default.providers.accuweather.cache_ttls]
 # Cache TTLs for weather data.
--- a/merino/providers/weather/backends/accuweather/pathfinder.py
+++ b/merino/providers/weather/backends/accuweather/pathfinder.py
@@ -87,9 +87,9 @@ def set_region_mapping(country: str, city: str, region: str | None):
         SUCCESSFUL_REGIONS_MAPPING[(country, city)] = region
 
 
-def get_region_mapping() -> dict:
-    """Get SUCCESSFUL_REGIONS_MAPPING"""
-    return SUCCESSFUL_REGIONS_MAPPING
+def get_region_mapping_size() -> int:
+    """Get SUCCESSFUL_REGIONS_MAPPING size"""
+    return len(SUCCESSFUL_REGIONS_MAPPING)
 
 
 def clear_region_mapping() -> None:
--- a/merino/providers/weather/backends/accuweather/pathfinder.py
+++ b/merino/providers/weather/backends/accuweather/pathfinder.py
@@ -87,6 +87,11 @@ def set_region_mapping(country: str, city: str, region: str | None):
         SUCCESSFUL_REGIONS_MAPPING[(country, city)] = region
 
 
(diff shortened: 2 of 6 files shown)


merino/configs/default.toml
```diff
@@ -218,6 +218,8 @@ connect_timeout_sec = 3.0
 # Global Redis settings. The weather provider optionally uses Redis to cache weather suggestions.
 # In the form of `redis://localhost:6379`.
 
+# MERINO_PROVIDERS__ACCUWEATHER__CRON_INTERVAL_SEC
+cron_interval_sec = 21600 # 6 hours
```
tiftran, line 222: thoughts about this time interval?

merino/providers/weather/provider.py
```diff
@@ -61,15 +67,30 @@ def __init__(
...
+            condition=self._should_fetch,
+            task=self._fetch_mapping,
+        )
+        self.cron_task = asyncio.create_task(cron_job())
 
     def hidden(self) -> bool:  # noqa: D102
         return False
 
+    def _should_fetch(self) -> bool:
+        return True
```
tiftran, line 87: I looked at the other providers and they periodcally check for last fetch time, in the event the previous fetch failed, but that's related to suggestion/content, while this is more of a good to know singular number.

merino/providers/weather/provider.py
```diff
@@ -61,15 +67,30 @@ def __init__(
...
+
+    async def _fetch_mapping(self) -> None:
+        self.metrics_client.gauge(
+            name=f"providers.{self.name}.pathfinder.size", value=len(get_region_mapping())
+        )
+
```
Herraj, line 93: nit: Thoughts on renaming the function to `_fetch_mapping_size`?

merino/configs/default.toml
Herraj: I think that's good, should give us 4 measurements per day to start with.

Herraj (approved):
🗺️

merino/providers/weather/backends/accuweather/pathfinder.py
```diff
@@ -87,6 +87,11 @@ def set_region_mapping(country: str, city: str, region: str | None):
         SUCCESSFUL_REGIONS_MAPPING[(country, city)] = region
 
 
+def get_region_mapping() -> dict:
```
ncloudioj, line 90: The setter and the getter would behave very differently. Maybe call it `get_region_mapping_size()` and return the mapping size instead?

merino/providers/weather/provider.py
```diff
@@ -61,15 +67,30 @@ def __init__(
...
 
     def hidden(self) -> bool:  # noqa: D102
         return False
 
+    def _should_fetch(self) -> bool:
+        return True
+
+    async def _fetch_mapping_size(self) -> None:
+        self.metrics_client.gauge(
+            name=f"providers.{self.name}.pathfinder.size", value=len(get_region_mapping())
```
ncloudioj, line 91:
```suggestion
            name=f"providers.{self.name}.pathfinder.mapping.size", value=len(get_region_mapping())
```

ncloudioj (approved):
r+wc.

PR #629, opened 2024-09-12
feat(curated recommendations): [MC-1498] Prototype support for contextual content experiments [load test: warn]

nina-py:
## References

JIRA: https://mozilla-hub.atlassian.net/browse/MC-1498

## Description

This is a prototype of an additional feed of curated corpus recommendations alongside the one already being sent to New Tab surfaces. If an optional parameter (`"feeds": ["need_to_know"]`) is requested with the API call, the new feed of 5-10 items (10 for the current prototype with semi-mocked data) is returned. 

The shape of the API response becomes the following:

```
{
  recommendedAt: int
  data: recommendation[]
  feeds: {
    need_to_know: {
      recommendations: recommendation[],
      title: string // localized based on locale value sent in request
    }
  }
}
```

The full spec is available at https://mozilla-hub.atlassian.net/wiki/spaces/FPS/pages/975110262/API+Technical+Specification.

## PR Review Checklist

### DIFF
--- a/merino/curated_recommendations/protocol.py
+++ b/merino/curated_recommendations/protocol.py
@@ -93,7 +93,7 @@ class CuratedRecommendationsRequest(BaseModel):
     region: str | None = None
     count: int = 100
     topics: list[Topic | str] | None = None
-    feeds: str | None = None
+    feeds: list[str] | None = None
 
     @field_validator("topics", mode="before")
     def validate_topics(cls, values):
@@ -121,29 +121,22 @@ class CuratedRecommendationsRequest(BaseModel):
         return []
 
 
-class LocalizedString(BaseModel):
-    """A string value for a given locale"""
-
-    value: str
-    locale: Locale
-
-
 class CuratedRecommendationsBucket(BaseModel):
     """A ranked list of curated recommendations"""
 
     recommendations: list[CuratedRecommendation]
-    title: LocalizedString | None = None
+    title: str | None = None
 
 
 class CuratedRecommendationsFeed(BaseModel):
     """Multiple lists of curated recommendations for the main New Tab feed and also for experiments"""
 
-    general: CuratedRecommendationsBucket
-    experimental: CuratedRecommendationsBucket
+    need_to_know: CuratedRecommendationsBucket
 
 
 class CuratedRecommendationsResponse(BaseModel):
     """Response schema for a list of curated recommendations"""
 
     recommendedAt: int
-    data: CuratedRecommendationsFeed | list[CuratedRecommendation]
+    data: list[CuratedRecommendation]
+    feeds: CuratedRecommendationsFeed | None = None
--- a/merino/curated_recommendations/protocol.py
+++ b/merino/curated_recommendations/protocol.py
@@ -121,10 +121,18 @@ class CuratedRecommendationsRequest(BaseModel):
         return []
 
 
+class LocalizedString(BaseModel):
+    """A string value for a given locale"""
+
+    value: str
+    locale: Locale
+
+
 class CuratedRecommendationsBucket(BaseModel):
     """A ranked list of curated recommendations"""
 
     recommendations: list[CuratedRecommendation]
+    title: LocalizedString | None = None
 
 
 class CuratedRecommendationsFeed(BaseModel):
--- a/merino/curated_recommendations/protocol.py
+++ b/merino/curated_recommendations/protocol.py
@@ -93,6 +93,7 @@ class CuratedRecommendationsRequest(BaseModel):
     region: str | None = None
     count: int = 100
     topics: list[Topic | str] | None = None
+    include_experimental: bool | None = None
 
     @field_validator("topics", mode="before")
     def validate_topics(cls, values):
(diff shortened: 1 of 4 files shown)


merino/curated_recommendations/protocol.py
```diff
@@ -120,8 +121,24 @@ def validate_topics(cls, values):
...
+
+    need_to_know: CuratedRecommendationsBucket
+
+
 class CuratedRecommendationsResponse(BaseModel):
     """Response schema for a list of curated recommendations"""
 
     recommendedAt: int
     data: list[CuratedRecommendation]
+    feeds: CuratedRecommendationsFeed | None = None
```
nina-py, line 144: This part here worries me a little. With the optional `feeds` value in the response, we're now always sending ` "feeds": null` as part of _every_ API response. Does this break backward compatibility? If yes, a different solution is needed here.

merino/curated_recommendations/provider.py
```diff
@@ -107,39 +111,27 @@ def derive_region(locale: Locale, region: str | None = None) -> str | None:
...
-
+    def process_recommendations(
+        self,
+        recommendations: list[CuratedRecommendation],
+        surface_id: str,
+        topics: list[Topic | str] | None = None,
+    ):
+        """Apply additional processing to the list of recommendations
+        received from Curated Corpus API
+        """
```
nina-py, line 122: I moved the processing steps for recommendations into its own method so that it can be called for both the "classic" list of recommendations and also the new one with `need_to_know`-specific items taken out.

merino/curated_recommendations/protocol.py
jpetto: i'll check with scott in slack today

merino/curated_recommendations/provider.py
```diff
@@ -107,39 +111,27 @@ def derive_region(locale: Locale, region: str | None = None) -> str | None:
...
-        # Convert the CorpusItem list to a CuratedRecommendation list.
-        recommendations = [
-            CuratedRecommendation(
-                **item.model_dump(),
-                receivedRank=rank,
-            )
-            for rank, item in enumerate(corpus_items)
-        ]
-
+    def process_recommendations(
```
mmiermans, line 114: Would this name be more descriptive?
```suggestion
    def rank_main_recommendations(
```

merino/curated_recommendations/provider.py
```diff
@@ -153,11 +145,79 @@ async def fetch(
...
+            # to the main recommendations feed
+            general_feed = self.process_recommendations(
+                general_feed, surface_id, curated_recommendations_request.topics
+            )
+
+            # Provide a localized title string for the "Need to Know" feed.
+            localized_titles = {
+                ScheduledSurfaceId.NEW_TAB_EN_US: "Need to Know",
+                ScheduledSurfaceId.NEW_TAB_EN_GB: "Need to Know in British English",
+                ScheduledSurfaceId.NEW_TAB_DE_DE: "Need to Know auf Deutsch",
```
mmiermans, line 196: 😃 🇩🇪

merino/curated_recommendations/provider.py
```diff
@@ -153,11 +145,79 @@ async def fetch(
...
+
+            return CuratedRecommendationsResponse(
+                recommendedAt=self.time_ms(),
+                data=general_feed,
+                feeds=CuratedRecommendationsFeed(
+                    need_to_know=CuratedRecommendationsBucket(
+                        recommendations=need_to_know_feed, title=title
+                    ),
+                ),
+            )
```
mmiermans, line 208: I think it would improve readability to extract the need-to-know if block into a method.

tests/integration/api/v1/curated_recommendations/test_curated_recommendations.py
```diff
@@ -207,6 +214,60 @@ async def test_curated_recommendations():
...
+        assert all(item["tileId"] for item in feed)
+
+        # Make sure the top item in the `need_to_know` feed is the one we expect
+        # (Currently, #71 in the test data. Once `isTimeSensitive` prop is in use,
+        # this should be the top item in the list with `isTimeSensitive`=true)
+        assert feed[0]["title"] == "How Does a Therapist Stay Neutral?"
+
+        # Assert `need_to_know` stories have the correct rank.
+        # Currently, these stories keep their original rank (70+).
+        # Should this change, this assertion will need an update.
```
mmiermans, line 266: Good that you documented this here. I put a comment in confluence so we don't lose track of this question. I don't think it matters much for the prototype.

merino/curated_recommendations/protocol.py
jpetto: pinged you in the slack thread - looks like we're fine here. though potentially we could create a new response class, right? something like `CuratedRecommendationsResponseExperimental` - and then set either as the return type.

i'll leave that up to you.

merino/curated_recommendations/provider.py
jpetto: agree with mathijs here - i like `rank_recommendations` (as that makes it more general purpose should we end up ranking future feeds).

merino/curated_recommendations/provider.py
jpetto: agree with this. then we could write unit tests against it 😄

jpetto (approved):
a few places we could tighten up a bit, but none of those need to block/could be done in a follow up.

nice work and i love the comments!

merino/curated_recommendations/protocol.py
mmiermans: Agreed, I think this is fine. I like to double-check with a front-end engineer, but the answer I've gotten for Firefox New Tab has always been that extending the schema is OK for backward-compatibility.

If you do go the route of creating a new class, you can use inheritance to avoid duplicate code.

PR #635, opened 2024-09-23
[DISCO-3003] Add Accept-Language header

tiftran:
## References

JIRA: [DISCO-3003](https://mozilla-hub.atlassian.net/browse/DISCO-3003)

## Description
Suggest endpoint to parse the Accept-Language header and filter out languages to only the ones it support

## PR Review Checklist

[DISCO-3003]: https://mozilla-hub.atlassian.net/browse/DISCO-3003?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/merino/configs/default.toml
+++ b/merino/configs/default.toml
@@ -42,10 +42,6 @@ disabled_providers = []
 # MERINO_RUNTIME__DEFAULT_SUGGESTIONS_RESPONSE_TTL_SEC
 default_suggestions_response_ttl_sec = 300 # 5 mins
 
-# MERINO_RUNTIME__DEFAULT_LANGUAGES
-default_languages  = ["de", "en-CA", "en-GB", "en-US", "es-ES", "fr", "it", "pl", "pt-BR", "ru", "zh-CN"]
-
-
 [default.logging]
 # MERINO_LOGGING__LEVEL
 # Minimum level of logs that should be reported.
@@ -291,6 +287,9 @@ metrics_sampling_rate = 0.75
 
 # MERINO_ACCUWEATHER__PARTNER_CODE - partner_code (Not currently defined)
 # The partner code to append to URLs in the current conditions and forecast responses.
+
+# MERINO_ACCUWEATHER__DEFAULT_LANGUAGES
+default_languages  = ["de", "en-CA", "en-GB", "en-US", "es-ES", "fr", "it", "pl", "pt-BR", "ru", "zh-CN"]
 
 
 [default.providers.adm]
--- a/merino/configs/default.toml
+++ b/merino/configs/default.toml
@@ -42,6 +42,9 @@ disabled_providers = []
 # MERINO_RUNTIME__DEFAULT_SUGGESTIONS_RESPONSE_TTL_SEC
 default_suggestions_response_ttl_sec = 300 # 5 mins
 
+# MERINO_RUNTIME__DEFAULT_LANGUAGES
+default_languages  = ["de", "en-CA", "en-GB", "en-US", "es-ES", "fr", "it", "pl", "pt-BR", "ru", "zh-CN"]
+
 
 [default.logging]
 # MERINO_LOGGING__LEVEL
--- a/merino/configs/default.toml
+++ b/merino/configs/default.toml
@@ -76,6 +76,12 @@ client_variant_max = 10
 # in this case the total string length count for client variants.
 client_variant_character_max = 100
 
+
+# MERINO_WEB__API__V1__HEADER_CHARACTER_MAX
+# A non-negative integer that limits the string character length of a given header string,
+# in this case the total string length count for accept-language.
+header_character_max = 100
+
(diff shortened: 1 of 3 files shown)


merino/configs/default.toml
```diff
@@ -42,6 +42,9 @@ disabled_providers = []
 # MERINO_RUNTIME__DEFAULT_SUGGESTIONS_RESPONSE_TTL_SEC
 default_suggestions_response_ttl_sec = 300 # 5 mins
 
+# MERINO_RUNTIME__DEFAULT_LANGUAGES
+default_languages  = ["de", "en-CA", "en-GB", "en-US", "es-ES", "fr", "it", "pl", "pt-BR", "ru", "zh-CN"]
```
tiftran, line 46: I thought our "tier 1" locales would be a good start.

merino/web/api_v1.py
```diff
@@ -174,6 +175,7 @@ async def suggest(
         search_from = default_providers
 
     lookups: list[Task] = []
+    _ = get_accepted_languages(accept_language)
```
tiftran, line 178: since it's not being currently used.

gruberb: Then why calling it? Or at least, add a comment why it's not being used and still  being called here?

merino/web/api_v1.py
```diff
@@ -341,3 +343,22 @@ async def curated_content(
     [curated-topics-doc]: https://mozilla-hub.atlassian.net/wiki/x/LQDaMg
     """
     return await provider.fetch(curated_recommendations_request)
+
+
+def get_accepted_languages(languages: str) -> list[str]:
+    """Retrieve filtered list of languages that merino accepts."""
+    if languages:
```
ncloudioj, line 350: Some notes:
- Let's wrap the parsing block inside a try..except block to protect against malformed inputs
- Let's parse all the accept languages and leave the default_languages filtering to individual downstream consumers
- Optionally, we can decorate this parser with `@functools.lru_cache(maxsize=1000)` for memorization

ncloudioj (changes requested):

merino/web/api_v1.py
tiftran: gah, sorry i didn't finish my thought above. I was planning to have it consumed/used in a follow up PR. Yes agreed, was going to put it into SuggestionRequest, wanted to keep this PR tiny 😅

merino/web/api_v1.py
tiftran: Agreed that we should wrap the block in a try/except block, but not sure what exceptions that could pop up. `split` returns a list even if the split string does not exist. quality logic also handles an index error if the quality value doesn't exist.

merino/web/api_v1.py
```diff
@@ -67,7 +68,8 @@
 )
 async def suggest(
     request: Request,
-    q: str = Query(max_length=QUERY_CHARACTER_MAX),
+    q: Annotated[str, Query(max_length=QUERY_CHARACTER_MAX)],
+    accept_language: Annotated[str | None, Header()] = None,
```
ncloudioj, line 72:
```suggestion
    accept_language: Annotated[str | None, Header(max_length=HEADER_CHARACTER_MAX)] = None,
```

merino/web/api_v1.py
```diff
@@ -341,3 +343,25 @@ async def curated_content(
     [curated-topics-doc]: https://mozilla-hub.atlassian.net/wiki/x/LQDaMg
     """
     return await provider.fetch(curated_recommendations_request)
+
+
+@functools.lru_cache(maxsize=1000)
+def get_accepted_languages(languages: str) -> list[str]:
```
ncloudioj, line 349:
```suggestion
def get_accepted_languages(languages: str | None) -> list[str]:
```

tests/unit/web/test_api_v1.py
```diff
@@ -0,0 +1,24 @@
...
+    ("languages", "expected_filtered_languages"),
+    [
+        ("*", ["en-US"]),
+        ("en-US", ["en-US"]),
+        ("en-US,en;q=0.5", ["en-US", "en"]),
+        ("en-US,en;q=0.9,zh-CN;q=0.7", ["en-US", "en", "zh-CN"]),
+        ("en-CA;q=invalid", ["en-US"]),
+    ],
+)
+def test_get_accepted_languages(languages, expected_filtered_languages):
```
ncloudioj, line 22: Let's add a test case for the exception handling path.

ncloudioj (approved):
Looks good, r+wc.

tests/unit/web/test_api_v1.py
tiftran: ("en-CA;q=invalid", ["en-US"]), should technically cover the exception path?

ncloudioj: Right, sorry, I didn't realize that.

PR #639, opened 2024-09-26
[MC-1521] Use isTimeSensitive to populate Need to Know feed

nina-py:
## References

JIRA: https://mozilla-hub.atlassian.net/browse/MC-1521

## Description
Replace the current prototype code that builds the “need to know” feed with logic using isTimeSensitive.

- The `need_to_know` feed is now populated with time-sensitive items instead of the bottom 10 items from the general feed.
- The `receivedRank` values for items in the `need_to_know` feed now start at zero and are consecutive numbers - essentially these are array indexes, similar to the general feed with or without the new feed.

## PR Review Checklist

### DIFF
--- a/merino/curated_recommendations/provider.py
+++ b/merino/curated_recommendations/provider.py
@@ -186,11 +186,15 @@ class CuratedRecommendationsProvider:
         @return: A tuple with two re-ranked lists of curated recommendations and a localised
         title for the "Need to Know" heading
         """
-        # TODO: the "need_to_know" items will be filtered by a corpus item property `isTimeSensitive`.
-        #  While data is being prepared, take the bottom ten items from the original recommendations
-        #  list and use them instead for the prototype.
-        general_feed = recommendations[:-10]
-        need_to_know_feed = recommendations[-10:]
+        # Filter out all time-sensitive recommendations into the need_to_know feed
+        need_to_know_feed = [r for r in recommendations if r.isTimeSensitive]
+
+        # Update received_rank for need_to_know recommendations
+        for rank, rec in enumerate(need_to_know_feed):
+            rec.receivedRank = rank
+
+        # Place the remaining recommendations in the general feed
+        general_feed = [r for r in recommendations if r not in need_to_know_feed]
 
         # Apply all the additional re-ranking and processing steps
         # to the main recommendations feed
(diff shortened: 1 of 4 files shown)


tests/data/scheduled_surface.json
```diff
@@ -22,7 +22,7 @@
             "excerpt": "A rigorous experiment revealed that on a hot, dry day, drinking a hot beverage can help your body stay cool.",
             "topic": "FOOD",
             "publisher": "Smithsonian Magazine",
-            "isTimeSensitive": false,
+            "isTimeSensitive": true,
```
nina-py, line 25: Marked exactly ten stories as time-sensitive in our test data

tests/integration/api/v1/curated_recommendations/test_curated_recommendations.py
```diff
@@ -259,15 +260,8 @@ async def test_curated_recommendations_with_need_to_know_feed():
...
         # Make sure the top item in the `need_to_know` feed is the one we expect
-        # (Currently, #71 in the test data. Once `isTimeSensitive` prop is in use,
-        # this should be the top item in the list with `isTimeSensitive`=true)
-        assert feed[0]["title"] == "How Does a Therapist Stay Neutral?"
-
-        # Assert `need_to_know` stories have the correct rank.
-        # Currently, these stories keep their original rank (70+).
-        # Should this change, this assertion will need an update.
-        for i, item in enumerate(feed, 70):
-            assert item["receivedRank"] == i
```
nina-py, line 270: Took this out as we can't test this with our fixture data. However, this is being tested in unit tests in `test_provider.py` as part of this PR

merino/curated_recommendations/provider.py
```diff
@@ -186,11 +186,15 @@ def rank_need_to_know_recommendations(
...
-        need_to_know_feed = recommendations[-10:]
+        # Filter out all time-sensitive recommendations into the need_to_know feed
+        need_to_know_feed = [r for r in recommendations if r.isTimeSensitive]
+
+        # Update received_rank for need_to_know recommendations
+        for rank, rec in enumerate(need_to_know_feed):
+            rec.receivedRank = rank
+
+        # Place the remaining recommendations in the general feed
+        general_feed = [r for r in recommendations if r not in need_to_know_feed]
```
jpetto, line 197: curious why you went the `not in` route here. would `if not r.isTimeSensitive` be more efficient? or a single loop over the array populating the two different feeds?

jpetto (approved):
the small size of this PR speaks to how well built the prototype was 😄.

one non-blocking question on how you're going about populating the `general_feed`.

merino/curated_recommendations/provider.py
nina-py: I don't think either would be more efficient over the other as we end up filtering the same list twice either way. My logic was that this way the code is more readable. I'll have a think about a better solution!

PR #646, opened 2024-10-07
[DISCO-3004] weather location auto complete to use accept language

tiftran:
## References

JIRA: [DISCO-3004](https://mozilla-hub.atlassian.net/browse/DISCO-3004)

## Description
Weather location auto complete to use accept language and select the first accepted language that is in the default language list

## PR Review Checklist

[DISCO-3004]: https://mozilla-hub.atlassian.net/browse/DISCO-3004?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/merino/providers/base.py
+++ b/merino/providers/base.py
@@ -15,6 +15,7 @@ class SuggestionRequest(BaseModel):
 
     query: str
     geolocation: Location
+    languages: list[str] | None = None
     request_type: str | None = None
 
 
--- a/merino/providers/weather/backends/accuweather/backend.py
+++ b/merino/providers/weather/backends/accuweather/backend.py
@@ -15,7 +15,6 @@ from httpx import AsyncClient, HTTPError, Response
 from pydantic import BaseModel, ValidationError
 
 from merino.cache.protocol import CacheAdapter
-from merino.config import settings
 from merino.exceptions import BackendError, CacheAdapterError
 from merino.middleware.geolocation import Location
 from merino.providers.weather.backends.accuweather.pathfinder import (
@@ -36,6 +35,7 @@ from merino.providers.weather.backends.accuweather.utils import (
     process_current_condition_response,
     process_location_response_with_country,
     process_location_response_with_country_and_region,
+    get_language,
 )
 
 
@@ -179,14 +179,6 @@ class WeatherDataType(Enum):
 
     CURRENT_CONDITIONS = 1
     FORECAST = 2
-
-
-def get_language(requested_languages) -> str | None:
-    """Get first language that is in default_languages."""
-    valid_languages = set(settings.accuweather.default_languages)
-    return next(
-        (language for language in requested_languages if language in valid_languages), None
-    )
 
 
 class AccuweatherBackend:
--- a/merino/providers/weather/backends/accuweather/backend.py
+++ b/merino/providers/weather/backends/accuweather/backend.py
@@ -15,6 +15,7 @@ from httpx import AsyncClient, HTTPError, Response
 from pydantic import BaseModel, ValidationError
 
 from merino.cache.protocol import CacheAdapter
+from merino.config import settings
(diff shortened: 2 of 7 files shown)


merino/providers/weather/backends/accuweather/backend.py
```diff
@@ -179,6 +181,14 @@ class WeatherDataType(Enum):
     FORECAST = 2
 
 
+def get_language(requested_languages) -> str | None:
+    """Get first language that is in default_languages."""
+    valid_languages = set(settings.accuweather.default_languages)
+    return next(
+        (language for language in requested_languages if language in valid_languages), None
+    )
+
```
Herraj, line 190: nit: we should probably move this to the `utils.py` module

merino/providers/base.py
```diff
@@ -15,6 +15,7 @@ class SuggestionRequest(BaseModel):
 
     query: str
     geolocation: Location
+    languages: list[str] | None = None
```
Herraj, line 18: I noticed that the `get_accepted_languages` function in `api_v1` always returns a value (`["en-US"]`). Should we remove the `None` type here 🤔

Herraj (approved):
🚀

merino/providers/weather/backends/accuweather/backend.py
tiftran: ah yup 👍

merino/providers/base.py
tiftran: i initially had that, but then realized that would mean languages would be required for every SuggestionRequest object passed to every provider, but all providers aren't using it. If that makes sense?

Herraj: ah right, that `SuggestionRequest` type is used by all providers

ncloudioj: Good point. Currently, only `suggest/accuweather` needs this, once other endpoints share the same need in the future, we can extract this from `api_v1` and make it available for all the request handlers.

merino/providers/weather/backends/accuweather/utils.py
```diff
@@ -185,3 +185,11 @@ def process_forecast_response(response: Any) -> dict[str, Any] | None:
             }
         case _:
             return None
+
+
+def get_language(requested_languages) -> str | None:
```
ncloudioj, line 190: Let's type annotate `requested_languages` as well.

merino/providers/weather/backends/accuweather/utils.py
```diff
@@ -185,3 +185,11 @@ def process_forecast_response(response: Any) -> dict[str, Any] | None:
...
+    """Get first language that is in default_languages."""
+    valid_languages = set(settings.accuweather.default_languages)
```
ncloudioj, line 192: Getting a deeply nested config via Dynaconf has some overhead, plus this won't change anyway, we can store it in a const at the module level, e.g. `VALID_LANGUAGES: frozenset = frozenset(settings.accuweather.default_languages)`.

merino/providers/weather/backends/accuweather/utils.py
```diff
@@ -185,3 +185,11 @@ def process_forecast_response(response: Any) -> dict[str, Any] | None:
...
+    return next(
+        (language for language in requested_languages if language in valid_languages), None
```
ncloudioj, line 194: To be consistent, I think we should default to `en-US` here rather than None. Then you don't need that `if language:` check above.

ncloudioj (approved):
r+wc.

merino/providers/weather/backends/accuweather/utils.py
tiftran: eek, yup

PR #647, opened 2024-10-07
[DISCO-3023] Refactor out and centralize accuweather error strings

Herraj:
## References

JIRA: [DISCO-3023](https://mozilla-hub.atlassian.net/browse/DISCO-3023)

## Description
Refactoring out the accuweather backend error strings into a separate module for clarity, maintainability and flexibility.

## PR Review Checklist

[DISCO-3023]: https://mozilla-hub.atlassian.net/browse/DISCO-3023?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/merino/providers/weather/backends/accuweather/errors.py
+++ b/merino/providers/weather/backends/accuweather/errors.py
@@ -11,7 +11,7 @@ class AccuweatherErrorMessages(Enum):
     CACHE_WRITE_ERROR = "Something went wrong with storing to cache. Did not update cache."
     FAILED_WEATHER_REPORT = "Failed to fetch weather report: {exceptions}"
     UNEXPECTED_LOCATION_RESPONSE = "Unexpected location response from: {url_path}, city: {city}"
-    UNEXPECTED_GEOLOCATION_REQUEST_ERROR = "Unexpected error occurred when requesting location byeolocation from Accuweather: {exception_class_name}"
+    UNEXPECTED_GEOLOCATION_REQUEST_ERROR = "Unexpected error occurred when requesting location by geolocation from Accuweather: {exception_class_name}"
     UNEXPECTED_CURRENT_CONDITIONS_RESPONSE = (
         "Unexpected current conditions response, Url: {current_conditions_url}"
     )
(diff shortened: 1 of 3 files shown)


Herraj:
Let me know your thoughts on this pattern?

bendk (approved):
Seems good to me, having them all in one place makes it easier to keep the tone consistent.  It also seems like it would be easier to localize these.

The only downside I can think of is that you can't use f-strings, but that's not going to work with localized messages anyways.

merino/providers/weather/backends/accuweather/errors.py
```diff
@@ -0,0 +1,40 @@
...
+
+class AccuweatherError(BackendError):
+    """Error during interaction with the Accuweather api. Inherits the BackendError parent
+    class. On the provider level, this is caught as a BackendError type and logged.
+    """
+
+    def __init__(self, error_type: AccuweatherErrorMessages, **kwargs):
+        # Use the `format_message` method to get the formatted error message
+        message = error_type.format_message(**kwargs)
+        super().__init__(message)
```
Herraj, line 40: If we like this pattern, I can add this function to the `BackendError` class such that every child error class would need to implement it and follow the same pattern.

merino/providers/weather/backends/accuweather/errors.py
```diff
@@ -0,0 +1,40 @@
+"""Errors module that maintains all the accuweather specific error strings and the
+AccuweatherError class to format and wrap them as a BackendError child object
+"""
+
+from enum import Enum
+from merino.exceptions import BackendError
+
+
+class AccuweatherErrorMessages(Enum):
```
ncloudioj, line 9: We can also do `StrEnum`.

ncloudioj (approved):
I see the value of centralizing these errors and messages. One minor hitch is that now to locate a certain error by error message, we would need to find the corresponding message ID first and then locate that in the source code.

Let's give this a try for accuweather and see how folks feel about it. If it turns out to be successful, we can apply it to the entire repo.

Does that make sense?

PR #648, opened 2024-10-09
[DISCO-3006] add support for city based weather suggest

misaniwere:
## References

JIRA: [DISCO-3006](https://mozilla-hub.atlassian.net/browse/DISCO-3006)

## Description
This PR adds support for city based weather suggest. New query params (`city`, `region`, `country`) have been added to the suggest endpoint and when all 3 are provided with `weather` as `request_type`, accuweather should return weather suggestions based on those parameters. 

## PR Review Checklist

[DISCO-3006]: https://mozilla-hub.atlassian.net/browse/DISCO-3006?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


tests/unit/providers/weather/test_provider.py
```diff
@@ -72,31 +99,155 @@ def test_not_hidden_by_default(provider: Provider) -> None:
...
+                    "https://www.accuweather.com/en/us/boston-ma/02108/daily-weather-forecast/348735?lang=en-us"
+                ),
+                summary="Expect showery weather Sunday afternoon through Monday morning",
+                high=Temperature(c=20, f=68),
+                low=Temperature(c=13, f=56),
+            ),
+        )
+    ]
+
+    backend_mock.get_weather_report.return_value = report
```
tiftran, line 205: while this test would probably get line coverage, i don't think it effectively tests whether the Location object is overwritten. I would probably rather do something like `backend_mock.get_weather_report.assert_called_with( ....)` below.
What do you think?

bendk (commented):
Looks good to me, although I don't have enough context to really evaluate this one.  I left some suggestions about validating the params, but I'm not sure which is the best way.  Feel free to go with whatever you think is best.

tests/unit/providers/weather/test_provider.py
misaniwere: Hmm, so I assumed this test was effective in confirming the overwrite because geolocation is initially for San Fransisco but the weather report returned is for Boston which comes from the query params.

misaniwere: I'm also open to checking what the get weather report method was called with in addition to that but I wanna make sure I understand your comment about the location overwrite better

merino/providers/weather/provider.py
```diff
@@ -113,6 +117,16 @@ async def query(self, srequest: SuggestionRequest) -> list[BaseSuggestion]:
...
                 else:
+                    # pass in city, region and country into geolocation ONLY if they all exist in the params
+                    if srequest.city and srequest.region and srequest.country:
+                        geolocation = geolocation.model_copy(
+                            update={
+                                "city": srequest.city,
+                                "regions": [srequest.region],
+                                "country": srequest.country,
+                            }
+                        )
```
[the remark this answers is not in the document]
misaniwere: I agree, it's definitely better to give some sort of cue as to why the request isn't working as expected. I'll add a 400 👍🏾

tests/unit/providers/weather/test_provider.py
tiftran: im just thinking regardless of whatever params are in the call, since you set the return value for the mock, it doesn't really validate anything? does that make sense?

tiftran (approved):
lgtm

merino/providers/base.py
```diff
@@ -17,6 +17,9 @@ class SuggestionRequest(BaseModel):
     geolocation: Location
     languages: list[str] | None = None
     request_type: str | None = None
+    city: str | None = None
```
ncloudioj, line 20: For discussion: all the geo fileds are stored in `Location`, putting the new ones at the top level might be confusing. What do you think about adding them to `Location` and call them something different, e.g. `country/region/city_custom`?

WDYT? @misaniwere @tiftran

misaniwere: hmm i definitely think it'll be cleaner and more readable but I'm a little fuzzy on how we would be storing the custom city/region/country in this case. would this involve updating the geolocation in the request scope?

misaniwere: also, by top-level, do you mean the logic being in `provider.py` vs backend?

tiftran: I did think about adding additional fields to `Location` but then inside of the backend, you would still need to overwrite  `city`, `region`, `country` given how `pathfinder.explore` and `pathfinder.compass` works. Unless we want to modify those as well, which could be messy, given the time we have?

ncloudioj: > would this involve updating the geolocation in the request scope

No need to update the request scope, rather we can update the geolocation object when creating `SuggestionRequest` over [here](https://github.com/mozilla-services/merino-py/blob/main/merino/web/api_v1.py#L184).

> also, by top-level, do you mean the logic being in provider.py vs backend?

Yes, we normally try pushing business logic to the backend whenever makes sense.

> you would still need to overwrite city, region, country given how pathfinder.explore and pathfinder.compass works.

Not necessarily need to overwrite those. I think we can add a helper to `Location` to return the preferred triplet as follows:
```python
class Location(BaseModel):
    country: ...
    country_name: ...
    ...
    key: ...
    country_custom: Optional[str] = None
    regions_custom: Optional[list[str]] = None
    city_custom:: Optional[str] = None

    def preferred_location(self) -> (Optional[str], Optional[list[str]], Optional[str]):
         // If custom fileds are present then return them, otherwise return the regular ones
``` 

Then we can call this helper in pathfinder and other places.

> Unless we want to modify those as well, which could be messy, given the time we have?

Of course, feel free to land the patch as is so that the frontend team can try it out sooner. We can clean things up as a follow up later.

tiftran: oooh nice, makes sense to me. probs yes to it being a follow up. @misaniwere thoughts?

merino/web/api_v1.py
```diff
@@ -70,6 +70,9 @@
 async def suggest(
     request: Request,
     q: Annotated[str, Query(max_length=QUERY_CHARACTER_MAX)],
+    city: str | None = None,
+    country: str | None = None,
```
ncloudioj, line 74: Like other query parameters, let's add some validation for those new ones.

Country code follows [ISO 3166 Alpha-2](https://en.wikipedia.org/wiki/ISO_3166-1_alpha-2), so it should be exactly 2 characters. For `city` and `region`, we can use the generic cap for query params.

```
    country: Annotated[str, Query(max_length=2, min_length=2)],
    region: Annotated[str, Query(max_length=QUERY_CHARACTER_MAX)],
    city: Annotated[str, Query(max_length=QUERY_CHARACTER_MAX)],
``` 

Any request with invalid query params would be declined with HTTP 400.

ncloudioj (approved):
Looks good, r+wc.

merino/providers/base.py
misaniwere: makes sense to me too. i'll work on a follow up and merge this in for now

merino/web/api_v1.py
misaniwere: I'll include this in the follow up

PR #667, opened 2024-10-17
DISCO-3033 refactor city based suggest

misaniwere:
## References

JIRA: [DISCO-3033](https://mozilla-hub.atlassian.net/browse/DISCO-3033)

## Description
- Follow up refactor based on [this discussion](https://github.com/mozilla-services/merino-py/pull/648#discussion_r1801940670)
- This PR makes some improvements on the city-based suggest logic for overwriting geolocation for the accuweather provider. Mainly, moving logic away from the provider and adding a helper function to the `Location` model to return preferred location based on custom values. Also added some more validation to query params. 

## PR Review Checklist

[DISCO-3033]: https://mozilla-hub.atlassian.net/browse/DISCO-3033?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


merino/web/api_v1.py
```diff
@@ -190,10 +190,21 @@ async def suggest(
 
     lookups: list[Task] = []
     languages = get_accepted_languages(accept_language)
+
+    validate_suggest_custom_location_params(city, region, country)
+
+    geolocation = request.scope[ScopeKey.GEOLOCATION].model_copy(
```
ncloudioj, line 196: As of now, requests w/ the custom location is much fewer than those w/o it. So let's update it conditionally.

```python
if country and region and city:
    geolocation = request.scope[ScopeKey.GEOLOCATION].model_copy(...)
else:
    geolocation = request.scope[ScopeKey.GEOLOCATION]
    
```

merino/web/api_v1.py
```diff
@@ -392,3 +403,14 @@ def get_accepted_languages(languages: str | None) -> list[str]:
...
+
+
+def validate_suggest_custom_location_params(
+    city: Optional[str], region: Optional[str], country: Optional[str]
+):
+    """Validate that city, region & country params are either all present or all omitted."""
+    if any([country, region, city]) and not all([country, region, city]):
+        raise HTTPException(
+            status_code=400,
+            detail="Invalid query parameters: `city`, `region`, and `country` are required, but one or more are missing in the request.",
```
ncloudioj, line 415:
```suggestion
            detail="Invalid query parameters: `city`, `region`, and `country` are either all present or all omitted.",
```

ncloudioj (approved):
Looks good, thanks! r+wc.

merino/middleware/geolocation.py
```diff
@@ -24,14 +24,25 @@ class Location(BaseModel):
...
     city: Optional[str] = None
+    city_custom: Optional[str] = None
     dma: Optional[int] = None
     postal_code: Optional[str] = None
     key: Optional[str] = None
 
+    def prioritize_custom_location(self) -> "Location":
+        """Update city, region, and country with custom fields if all custom fields exist."""
+        if all([self.city_custom, self.regions_custom, self.country_custom]):
+            self.city = self.city_custom
```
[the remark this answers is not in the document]
misaniwere: this should be fine as it's not a property method but a helper method so the overwriting only happens when the function is invoked. currently it's only invoked within the `get_weather_report` function. this helps prevent us from having to do the overwrite in multiple weather related helper functions in the backend.

merino/web/api_v1.py
misaniwere: hmm since we're overwriting city, region & country down the line, I feel more comfortable working with the copy in this case. also we do some validation right before this so it either errors out before getting here if we have missing fields or the values are all `None` which is the default.

ncloudioj: Well, if all the three params are unspecified (i.e. `None`), would doing the update unconditionally create a new one with `None` country, city, and regions? As mentioned above, if we don't need to store the custom fields, we can do the following to make it always use a copy `Location`.

```py

if country and region and city:
    geolocation = request.scope[ScopeKey.GEOLOCATION].model_copy(update={"city": city, "regions": [region], country: country})
else:
    geolocation = request.scope[ScopeKey.GEOLOCATION].model_copy()
```

misaniwere: makes sense. I'll get rid of the custom fields since we're gonna be working with a copy. No need for the helper function either

PR #676, opened 2024-10-24
chore: Fix warnings emitted from tests

ncloudioj:
## References

JIRA: N/A

## Description
More than 1K warnings 🫢 emitted from our tests as of now, this cuts those down to 4 (~~which i couldn't figure out the fix atm~~).

Oh, I also noticed that `make bandit` took forever to run because of the WONDER file. Fixed that as well, now the `check` CI job runs much faster.

Update: the cause of the remaining 4 warnings is b/c we didn't fully specify the mock stubs of the http client. Actually, those two test cases are already covered elsewhere. So I went ahead remove them.  

## PR Review Checklist


tests/integration/providers/weather/backends/test_accuweather.py
```diff
@@ -880,23 +880,24 @@ async def test_get_weather_report_with_geolocation_when_cache_is_empty(
 
     assert report is None
 
-    @pytest.mark.asyncio
```
tiftran, line 883: omg took me forever to figure out, the diff is cause of the indenting

tiftran (approved):
WONDER file, eh? is that what we're calling it?

PR #678, opened 2024-10-25
[DISCO-3046] Add lat, long, radius to geo suggestion

tiftran:
## References

JIRA: [DISCO-3046](https://mozilla-hub.atlassian.net/browse/DISCO-3046)

## Description
Added Lat, Long and accuracy radius to geolocation

## PR Review Checklist


merino/middleware/geolocation.py
```diff
@@ -20,6 +20,14 @@
 logger = logging.getLogger(__name__)
 
 
+class Coordinates(BaseModel):
+    """Data model for coordinates for geolocation."""
+
+    latitude: Optional[float] = None
+    longitude: Optional[float] = None
+    radius: Optional[int] = None
```
tiftran, line 28: Maxminddb returns an int for radius, which is why i decided to leave it as an int

merino/providers/custom_details.py
```diff
@@ -17,6 +19,7 @@ class GeolocationDetails(BaseModel):
     country: str | None = None
     region: str | None = None
     city: str | None = None
+    location: Coordinates | None = None
```
tiftran, line 22: we already have `Location` as the model for Geolocation, any concerns terminology will get messy and confusing?

misaniwere: hmm it might be confusing. what do you think about using `coordinates` instead?

merino/providers/geolocation/provider.py
```diff
@@ -61,6 +61,7 @@ async def query(self, srequest: SuggestionRequest) -> list[BaseSuggestion]:
                         if srequest.geolocation.region_names
                         else None,
                         city=srequest.geolocation.city,
+                        location=srequest.geolocation.location,
```
misaniwere, line 64: can we add this field to the tests in `test_provider.py` as well?

misaniwere (approved):
LGTM :shipit:

merino/providers/custom_details.py
ncloudioj: Good point. Let's rename it to `coordinates` for the `Location` model, `location` is fine for the API response here.

ncloudioj (approved):

PR #718, opened 2024-11-25
[DISCO-3087] weather: log successful_region_mapping

tiftran:
## References

JIRA: [DISCO-3087](https://mozilla-hub.atlassian.net/browse/DISCO-3087)

## Description
This will allow us to see successful regions found by the path finder. A log will be emitting every 6 hours, so 4 logs/day.

## PR Review Checklist

[DISCO-3087]: https://mozilla-hub.atlassian.net/browse/DISCO-3087?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/merino/providers/weather/backends/accuweather/pathfinder.py
+++ b/merino/providers/weather/backends/accuweather/pathfinder.py
@@ -99,8 +99,13 @@ def set_region_mapping(country: str, city: str, region: str | None):
         SUCCESSFUL_REGIONS_MAPPING[(country, city)] = region
 
 
+def get_region_mapping() -> dict[tuple[str, str], str | None]:
+    """Get SUCCESSFUL_REGIONS_MAPPING."""
+    return SUCCESSFUL_REGIONS_MAPPING
+
+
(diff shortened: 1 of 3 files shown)


Herraj (approved):
🗺️

ncloudioj:
Oops, I have a belated comment.

What do you all think about changing the logging cadence from batching to on-site? Specifically, instead of dumping the entire map within one record, we can log a single successful mapping record whenever the pathfinder finds an alternative triplet. The main reason is that dealing with big payload log records is annoying and could be problematic for big maps. Whereas smaller individual records are much easier to handle, especially now that we can query and export them via SQL.

Let me know what you think.

tiftran:
@ncloudioj, makes sense I did consider it and thought maybe we want to save the number of logs 😅. Thinking about it again, long term view the number of new logs should decrease as we populate the successful_region_mapping by hand. Will file an issue and pick that up later

PR #719, opened 2024-11-26
[DISCO-3086] weather: add city skip list

tiftran:
## References

JIRA: [DISCO-3086](https://mozilla-hub.atlassian.net/browse/DISCO-3086)

## Description
There are cities that accuweather does not provide weather for. Instead of constantly requesting them, we should skip them. Cities are from taken from logs.

## PR Review Checklist

[DISCO-3086]: https://mozilla-hub.atlassian.net/browse/DISCO-3086?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/merino/providers/weather/backends/accuweather/backend.py
+++ b/merino/providers/weather/backends/accuweather/backend.py
@@ -611,7 +611,6 @@ class AccuweatherBackend:
             return None
 
         cached_data = cached_data if cached_data is not None else []
-
         self.emit_cache_fetch_metrics(cached_data)
         cached_report = self.parse_cached_data(cached_data)
 
@@ -656,15 +655,18 @@ class AccuweatherBackend:
 
         # The cached report is incomplete, now fetching from AccuWeather.
         if location is None:
+            skip_log = False
             try:
-                location, _ = await pathfinder.explore(
+                location, is_skipped = await pathfinder.explore(
                     geolocation, self.get_location_by_geolocation
                 )
+                skip_log = is_skipped
             except AccuweatherError as exc:
                 logger.warning(f"{exc}")
 
             if location is None:
-                logger.warning(f"Unable to find location for {country}/{city}")
+                if not skip_log:
+                    logger.warning(f"Unable to find location for {country}/{city}")
                 return None
 
         try:
--- a/merino/providers/weather/backends/accuweather/backend.py
+++ b/merino/providers/weather/backends/accuweather/backend.py
@@ -19,7 +19,6 @@ from merino.exceptions import CacheAdapterError
 from merino.middleware.geolocation import Location
 from merino.providers.weather.backends.accuweather.pathfinder import (
     set_region_mapping,
-    is_in_skip_cities_list,
 )
 from merino.providers.weather.backends.protocol import (
     CurrentConditions,
@@ -600,7 +599,7 @@ class AccuweatherBackend:
             return None
 
         try:
-            cached_data = await pathfinder.explore(
+            cached_data, is_skipped = await pathfinder.explore(
                 geolocation, self._fetch_from_cache, language=language
             )
         except CacheAdapterError as exc:
@@ -608,9 +607,12 @@ class AccuweatherBackend:
             self.metrics_client.increment("accuweather.cache.fetch.error")
             return None
 
+        if is_skipped:
+            return None
+
         cached_data = cached_data if cached_data is not None else []
-        if not is_in_skip_cities_list(country, city):
-            self.emit_cache_fetch_metrics(cached_data)
+
+        self.emit_cache_fetch_metrics(cached_data)
         cached_report = self.parse_cached_data(cached_data)
 
         return await self.make_weather_report(cached_report, weather_context)
@@ -655,13 +657,14 @@ class AccuweatherBackend:
         # The cached report is incomplete, now fetching from AccuWeather.
         if location is None:
             try:
-                location = await pathfinder.explore(geolocation, self.get_location_by_geolocation)
+                location, _ = await pathfinder.explore(
+                    geolocation, self.get_location_by_geolocation
+                )
             except AccuweatherError as exc:
                 logger.warning(f"{exc}")
 
             if location is None:
-                if not is_in_skip_cities_list(country, city):
-                    logger.warning(f"Unable to find location for {country}/{city}")
+                logger.warning(f"Unable to find location for {country}/{city}")
                 return None
 
         try:
--- a/merino/providers/weather/backends/accuweather/backend.py
+++ b/merino/providers/weather/backends/accuweather/backend.py
@@ -19,6 +19,7 @@ from merino.exceptions import CacheAdapterError
 from merino.middleware.geolocation import Location
 from merino.providers.weather.backends.accuweather.pathfinder import (
     set_region_mapping,
+    is_in_skip_cities_list,
 )
 from merino.providers.weather.backends.protocol import (
     CurrentConditions,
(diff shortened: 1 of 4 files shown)


merino/providers/weather/backends/accuweather/backend.py
```diff
@@ -658,7 +660,8 @@ async def as_awaitable(val: Any) -> Any:
                 logger.warning(f"{exc}")
 
             if location is None:
-                logger.warning(f"Unable to find location for {country}/{city}")
+                if not is_in_skip_cities_list(country, city):
```
tiftran, line 663: We should do something like the successful_regions_mapping, to automatically input cities into the skip list, but I'm going hold off until we have the distance calculation in place, so that we still have logs for those cases.

merino/providers/weather/backends/accuweather/pathfinder.py
```diff
@@ -37,7 +49,12 @@ def compass(location: Location) -> Generator[Triplet, None, None]:
         city = CITY_NAME_CORRECTION_MAPPING.get(city, city)
         match (country, city):
             case ("US" | "CA", _):
```
ncloudioj, line 51: Maybe we simplify this by:

```py
        match (country, city):
            case ("US" | "CA", _) if (country, regions[0], city) not in SKIP_CITIES_LIST:
                yield country, regions[0], city  # use the most specific region
            case ("US" | "CA", _):
                return
            ...
```

Then you don't need `if any((country, region, city))` in `explore()` b/c the generator is empty and it will enter the loop.

merino/providers/weather/backends/accuweather/backend.py
```diff
@@ -608,7 +609,8 @@ async def get_weather_report_with_geolocation(
             return None
 
         cached_data = cached_data if cached_data is not None else []
-        self.emit_cache_fetch_metrics(cached_data)
+        if not is_in_skip_cities_list(country, city):
```
ncloudioj, line 612: How about we add a boolean to `explore()`'s return value indicating whether or not the exploration was skipped? Then we don't need to check that again here.

merino/providers/weather/backends/accuweather/pathfinder.py
```diff
@@ -15,6 +15,18 @@
     "Middlebury (village)": "Middlebury",
     "TracadieSheila": "Tracadie Sheila",
 }
+SKIP_CITIES_LIST: frozenset = frozenset(
+    [
+        ("CA", "AB", "Sturgeon County"),
+        ("CA", "ON", "Kleinburg Station"),
+        ("CA", "ON", "North Park"),
+        ("CA", "QC", "Montreal West"),
```
ncloudioj, line 23: Those are good candidate for `CITY_NAME_CORRECTION_MAPPING`:
  "Montreal West" => "Montreal";
  "Kleinburg Station" => "Kleinburg";

merino/providers/weather/backends/accuweather/pathfinder.py
```diff
@@ -107,3 +125,8 @@ def get_region_mapping_size() -> int:
 def clear_region_mapping() -> None:
     """Clear SUCCESSFUL_REGIONS_MAPPING."""
     SUCCESSFUL_REGIONS_MAPPING.clear()
+
+
+def is_in_skip_cities_list(country: str | None, city: str | None) -> bool:
```
ncloudioj, line 130: Looks like this is only used by the weather beckend, see my previous comment. If we return a `is_skipped` boolean from `explore()`, this should be unneeded.

In case of you need it anyway, I'd recommend generating another set based on the original one rather than doing it for every call.

```py
SKIP_LOCATION_LIST = [
    ("", "", ""),
    ...
]

SKIP_LOCATION_SET: frozenset = frozenset(SKIP_LOCATION_LIST)

SKIP_COUNTRY_CITY_SET: forzenset = frozenset(coutry, city for country, _, city in SKIP_LCOATION_LIST)


```

ncloudioj (changes requested):

merino/providers/weather/backends/accuweather/backend.py
```diff
@@ -600,17 +599,20 @@ async def get_weather_report_with_geolocation(
...
-            cached_data = await pathfinder.explore(
+            cached_data, is_skipped = await pathfinder.explore(
                 geolocation, self._fetch_from_cache, language=language
             )
         except CacheAdapterError as exc:
             logger.error(f"Failed to fetch weather report from Redis: {exc}")
             self.metrics_client.increment("accuweather.cache.fetch.error")
             return None
 
+        if is_skipped:
```
tiftran, line 610: I realized there's no need to continue, if we know the tuple is in the skip list

merino/providers/weather/backends/accuweather/backend.py
```diff
@@ -599,15 +599,19 @@ async def get_weather_report_with_geolocation(
...
+            return None
```
tiftran, line 611: I realized that if we know city, region, country is in the skip list, there's no need to continue from this point

merino/providers/weather/backends/accuweather/backend.py
```diff
@@ -599,15 +599,19 @@ async def get_weather_report_with_geolocation(
...
+
         cached_data = cached_data if cached_data is not None else []
+
```
ncloudioj, line 614:
```suggestion
```

merino/providers/weather/backends/accuweather/backend.py
```diff
@@ -653,7 +657,9 @@ async def as_awaitable(val: Any) -> Any:
         # The cached report is incomplete, now fetching from AccuWeather.
         if location is None:
             try:
-                location = await pathfinder.explore(geolocation, self.get_location_by_geolocation)
+                location, _ = await pathfinder.explore(
```
ncloudioj, line 660: There is a logging statement down below, shall we check `is_skipped` for conditional logging?

merino/providers/weather/backends/accuweather/pathfinder.py
ncloudioj: This can be dropped now.

merino/providers/weather/backends/accuweather/pathfinder.py
```diff
@@ -75,15 +91,17 @@ async def explore(
     Raises:
       - Any exception raised from `probe`.
     """
-    for country, region, city in compass(location):
+    for country, region, city, is_skipped in compass(location):
+        if is_skipped:
+            return None, True
```
ncloudioj, line 96:
```suggestion
            return None, is_skipped
```

merino/providers/weather/backends/accuweather/pathfinder.py
```diff
@@ -32,32 +44,36 @@ def compass(location: Location) -> Generator[Triplet, None, None]:
     country = location.country
     regions = location.regions
     city = location.city
+    is_skipped = False
 
     if regions and country and city:
         city = CITY_NAME_CORRECTION_MAPPING.get(city, city)
         match (country, city):
+            case ("US" | "CA", _) if (country, regions[0], city) not in SKIP_CITIES_LIST:
```
ncloudioj, line 52: Just realized that this check is only applied to "US" and "CA" locations here, what if we move the skip-city check from here to "explore()" instead so `compass` doesn't need to worry about this any more?

```py
    for country, region, city in compass(location):
        if (country, region, city) in SKIP_CITIES_LIST:
            return None, True
       ...
```

ncloudioj (approved):
r+wc.

merino/providers/weather/backends/accuweather/pathfinder.py
tiftran: oh that makes things much nicer

PR #721, opened 2024-11-27
HNT-260: Hardcode localized topic section titles for US & Germany

katerinachinnappan:
## References

JIRA: https://mozilla-hub.atlassian.net/browse/HNT-260

## Description

Created a `LOCALIZED_TOPIC_SECTION_TITLES` dictionary in `corpus_backends/protocol` to hold the localized topic section titles for `new_tab_en_us` & `new_tab_de_de`.

EN-US strings: https://searchfox.org/mozilla-central/source/browser/locales/en-US/browser/newtab/newtab.ftl#392
DE-DE strings: https://github.com/mozilla-l10n/firefox-l10n/blob/main/de/browser/browser/newtab/newtab.ftl#L407

## PR Review Checklist

### DIFF
--- a/merino/curated_recommendations/corpus_backends/protocol.py
+++ b/merino/curated_recommendations/corpus_backends/protocol.py
@@ -41,6 +41,51 @@ class ScheduledSurfaceId(str, Enum):
     NEW_TAB_IT_IT = "NEW_TAB_IT_IT"
 
 
+# Hardcoded localized topic section titles (for now) for en-US & de-DE
+LocalizedTopicSectionTitles = dict[ScheduledSurfaceId, dict[str, str]]
+
+LOCALIZED_TOPIC_SECTION_TITLES: LocalizedTopicSectionTitles = {
+    # reference: https://searchfox.org/mozilla-central/source/browser/locales/en-US/browser/newtab/newtab.ftl#392
+    ScheduledSurfaceId.NEW_TAB_EN_US: {
+        "business": "Business",
+        "career": "Career",
+        "education": "Education",
+        "arts": "Entertainment",
+        "food": "Food",
+        "health": "Health",
+        "hobbies": "Gaming",
+        "finance": "Money",
+        "society-parenting": "Parenting",
+        "government": "Politics",
+        "education-science": "Science",
+        "society": "Life Hacks",
+        "sports": "Sports",
+        "tech": "Tech",
+        "travel": "Travel",
+        "home": "Home & Garden",
+    },
+    # reference: https://github.com/mozilla-l10n/firefox-l10n/blob/main/de/browser/browser/newtab/newtab.ftl#L407
+    ScheduledSurfaceId.NEW_TAB_DE_DE: {
+        "business": "Geschäftliches",
+        "career": "Karriere",
+        "education": "Bildung",
+        "arts": "Unterhaltung",
+        "food": "Essen",
+        "health": "Gesundheit",
+        "hobbies": "Gaming",
+        "finance": "Finanzen",
+        "society-parenting": "Erziehung",
+        "government": "Politik",
+        "education-science": "Wissenschaft",
+        "society": "Life-Hacks",
+        "sports": "Sport",
+        "tech": "Technik",
+        "travel": "Reisen",
+        "home": "Haus und Garten",
+    },
+}
+
+
 class CorpusItem(BaseModel):
     """Represents a scheduled item from our 'corpus'.
     The corpus is the set of all curated items deemed recommendable.
--- a/merino/curated_recommendations/corpus_backends/protocol.py
+++ b/merino/curated_recommendations/corpus_backends/protocol.py
@@ -41,51 +41,6 @@ class ScheduledSurfaceId(str, Enum):
     NEW_TAB_IT_IT = "NEW_TAB_IT_IT"
 
 
-# Hardcoded localized topic section titles (for now) for en-US & de-DE
-LocalizedTopicSectionTitles = dict[ScheduledSurfaceId, dict[str, str]]
-
-LOCALIZED_TOPIC_SECTION_TITLES: LocalizedTopicSectionTitles = {
-    # reference: https://searchfox.org/mozilla-central/source/browser/locales/en-US/browser/newtab/newtab.ftl#392
-    ScheduledSurfaceId.NEW_TAB_EN_US: {
-        "business": "Business",
-        "career": "Career",
-        "education": "Education",
-        "arts": "Entertainment",
-        "food": "Food",
-        "health": "Health",
-        "hobbies": "Gaming",
-        "finance": "Money",
-        "society-parenting": "Parenting",
-        "government": "Politics",
-        "education-science": "Science",
-        "society": "Life Hacks",
-        "sports": "Sports",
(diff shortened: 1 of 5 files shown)


merino/curated_recommendations/corpus_backends/protocol.py
```diff
@@ -41,6 +41,51 @@ class ScheduledSurfaceId(str, Enum):
...
+        "education-science": "Wissenschaft",
+        "society": "Life-Hacks",
+        "sports": "Sport",
+        "tech": "Technik",
+        "travel": "Reisen",
+        "home": "Haus und Garten",
+    },
+}
+
+
```
mmiermans, line 88: Given that some sections (Popular today, In the News) don't correspond to a topic, would it make sense to move this data to a separate file, for example `merino/curated_recommendations/localization.py`? Or do you think it would be better you keep localized topics separate from section titles not based on topics?

tests/integration/api/v1/curated_recommendations/test_curated_recommendations.py
```diff
@@ -1330,23 +1349,49 @@ async def test_curated_recommendations_with_sections_feed(self):
...
-            assert all(Layout(**feed["layout"]) for feed in feeds.values() if feed)
+            self.assert_section_feed_helper(data)
 
-            # Ensure "Today's top stories" is present with a valid date subtitle
-            top_stories_section = data["feeds"].get("top_stories_section")
-            assert top_stories_section is not None
-            assert top_stories_section["title"] == "Today's top stories"
-            assert top_stories_section["subtitle"] is not None
+            # Ensure that topic section titles fallback to English
+            assert data["feeds"]["arts"]["title"] == "Arts"
```
mmiermans, line 1394: Could you add a test asserting that we have a translation for every section title? Something like the following. (Untested code.)

```python
@pytest.mark.parametrize("surface_id", [
    ScheduledSurfaceId.NEW_TAB_EN_US,
    ScheduledSurfaceId.NEW_TAB_DE_DE,
])
def test_section_translations(surface_id):
    # Get all fields in CuratedRecommendationsFeed of type Section
    section_fields = [
        field_name
        for field_name, field_type in CuratedRecommendationsFeed.model_fields.items()
        if field_type.annotation is Section
    ]
    
    # Get the localized titles for the current surface_id
    localized_titles = LOCALIZED_TOPIC_SECTION_TITLES[surface_id]
    
    # Assert that each section field has a translation
    for field in section_fields:
        assert field in localized_titles, f"Missing translation for '{field}' in {surface_id}"
```

merino/curated_recommendations/provider.py
```diff
@@ -336,11 +338,16 @@ def get_sections(
...
+                    formatted_topic = rec.topic.replace("_", " ")
                     section = sections_by_topic[topic] = Section(
                         receivedFeedRank=len(sections_by_topic)
                         + 1,  # add 1 for top_stories_section
                         recommendations=[],
-                        title=rec.topic.replace("_", " ").capitalize(),
+                        # return the hardcoded localized topic section title
+                        # fallback on en-US topic title
+                        title=LOCALIZED_TOPIC_SECTION_TITLES.get(surface_id, {}).get(
+                            formatted_topic, formatted_topic.capitalize()
```
mmiermans, line 349: `replace("_", " ")` was meant to be temporary code to generate a human-readable title from identifiers. I think we should stop doing that as part of this PR, when we have actual strings. The keys of `LOCALIZED_TOPIC_SECTION_TITLES` should be section or topic identifiers without any string formatting.

mmiermans (changes requested):
Thanks for pulling this together! I've got a few (hopefully small) changes I think we should make.

tests/integration/api/v1/curated_recommendations/test_curated_recommendations.py
```diff
@@ -1330,23 +1349,49 @@ async def test_curated_recommendations_with_sections_feed(self):
...
-            assert len(data["data"]) == 0
+            self.assert_section_feed_helper(data)
 
-            # Check that all sections are present
-            feeds = data["feeds"]
-            assert (
-                len(feeds) > 5
-            )  # fixture data contains enough recommendations for many sections.
+    @pytest.mark.asyncio
+    async def test_curated_recommendations_with_sections_feed_de(self):
```
mmiermans, line 1355: I can reproduce the failure locally by repeating the test a couple times:
```suggestion
    @pytest.mark.asyncio
    @pytest.mark.parametrize('repeat', list(range(10)))
    async def test_curated_recommendations_with_sections_feed_de(self, repeat):
```

The order and presence of sections is not deterministic, so you can't make an assumption that they're there. I would evaluate the assertion only if the section is present. You might want cover a few more sections in addition to `arts` to ensure at least one assertion is evaluated.
```python
if data["feeds"]["arts"] is not None:
    assert data["feeds"]["arts"]["title"] == "Unterhaltung"
```

merino/curated_recommendations/corpus_backends/protocol.py
katerinachinnappan: I think moving this to a separate file makes more sense especially if we end up using pontoon path we will need separate files anyways

merino/curated_recommendations/provider.py
katerinachinnappan: shouldn't we still use `rec.topic.replace("_", " ")` as a default in the case there is no key/value found in the dict? we should cover the case where we fallback to english titles but the key/value is not available in the dict for whatever reason so we should work with `rec.topic`

merino/curated_recommendations/corpus_backends/protocol.py
```diff
@@ -41,6 +41,51 @@ class ScheduledSurfaceId(str, Enum):
...
+        "society-parenting": "Erziehung",
+        "government": "Politik",
+        "education-science": "Wissenschaft",
+        "society": "Life-Hacks",
+        "sports": "Sport",
+        "tech": "Technik",
+        "travel": "Reisen",
+        "home": "Haus und Garten",
+    },
+}
```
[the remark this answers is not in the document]
katerinachinnappan: Amy said she will try getting US and DE tmr (12.02), so will update this PR once that is established.

mmiermans (commented):
I only have one more comment. Besides that, it all looks good to me. 👍

merino/curated_recommendations/localization.py
```diff
@@ -0,0 +1,72 @@
...
+from merino.curated_recommendations.corpus_backends.protocol import ScheduledSurfaceId
+import logging
+
+logger = logging.getLogger(__name__)
+
+
+LocalizedTopicSectionTitles = dict[ScheduledSurfaceId, dict[str, str]]
+
+# Hardcoded localized topic section titles (for now) for en-US & de-DE
+LOCALIZED_TOPIC_SECTION_TITLES: LocalizedTopicSectionTitles = {
```
mmiermans, line 12: top-stories is not a topic section, so a better name might be:
```suggestion
LOCALIZED_SECTION_TITLES: LocalizedSectionTitles = {
```
I can also do this when adding the News Section later this week.

katerinachinnappan: updated this in my PR along with the German string for top-stories

PR #722, opened 2024-11-28
[DISCO-3091] weather: Add lat/long distance logic

tiftran:
## References

JIRA: [DISCO-3091](https://mozilla-hub.atlassian.net/browse/DISCO-3091)

## Description
For cases where we request location keys with just country and city and the response returns more than one location, we can use lat long to calculate the closer one. 

In the event of cases like North Park, ON and accuweather only returns North Park, SK and North Park, BC, I decided to have a distance threshold so that we don't accidentally pick the wrong location

## PR Review Checklist

[DISCO-3091]: https://mozilla-hub.atlassian.net/browse/DISCO-3091?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/merino/providers/weather/backends/accuweather/backend.py
+++ b/merino/providers/weather/backends/accuweather/backend.py
@@ -8,6 +8,7 @@ import orjson
 import logging
 from enum import Enum
 from typing import Any, Callable, NamedTuple, cast
+from functools import partial
 
 import aiodogstatsd
 from dateutil import parser
@@ -538,9 +539,16 @@ class AccuweatherBackend:
         return await self.make_weather_report(cached_report, weather_context)
 
     async def _fetch_from_cache(
-        self, country: str | None, region: str | None, city: str | None, language: str
+        self,
+        weather_context: WeatherContext,
     ) -> list[bytes | None] | None:
         """Fetch weather data from cache."""
+        geolocation = weather_context.geolocation
+        country = geolocation.country
+        city = geolocation.city
+        language = get_language(weather_context.languages)
+        region = weather_context.selected_region
+
         if country is None or city is None:
             return None
 
@@ -590,7 +598,6 @@ class AccuweatherBackend:
         geolocation: Location = weather_context.geolocation
         country: str | None = geolocation.country
         city: str | None = geolocation.city
-        language: str = get_language(weather_context.languages)
 
         if country is None or city is None:
             self.metrics_client.increment(
@@ -600,7 +607,7 @@ class AccuweatherBackend:
 
         try:
             cached_data, is_skipped = await pathfinder.explore(
-                geolocation, self._fetch_from_cache, language=language
+                weather_context, self._fetch_from_cache
             )
         except CacheAdapterError as exc:
             logger.error(f"Failed to fetch weather report from Redis: {exc}")
@@ -658,9 +665,8 @@ class AccuweatherBackend:
             skip_log = False
             try:
                 location, is_skipped = await pathfinder.explore(
-                    geolocation, self.get_location_by_geolocation
+                    weather_context, self.get_location_by_geolocation
                 )
-                skip_log = is_skipped
             except AccuweatherError as exc:
                 logger.warning(f"{exc}")
 
@@ -717,7 +723,7 @@ class AccuweatherBackend:
             return None
 
     async def get_location_by_geolocation(
-        self, country: str | None, region: str | None, city: str | None
+        self, weather_context: WeatherContext
     ) -> AccuweatherLocation | None:
         """Return location data for a specific country and city or None if
         location data is not found.
@@ -727,6 +733,11 @@ class AccuweatherBackend:
         Reference:
             https://developer.accuweather.com/accuweather-locations-api/apis/get/locations/v1/cities/{countryCode}/{adminCode}/search
         """
+        geolocation = weather_context.geolocation
+        country = geolocation.country
+        city = geolocation.city
+        region = weather_context.selected_region
+
         if country is None or city is None:
             return None
 
@@ -735,7 +746,10 @@ class AccuweatherBackend:
             processor = process_location_response_with_country_and_region
         else:
             url_path = self.url_cities_path.format(country_code=country)
-            processor = process_location_response_with_country
+            partial_process_func = partial(
+                process_location_response_with_country, geolocation.coordinates
+            )
+            processor = partial_process_func
 
         try:
             response: dict[str, Any] | None = await self.request_upstream(
--- a/merino/providers/weather/backends/accuweather/pathfinder.py
+++ b/merino/providers/weather/backends/accuweather/pathfinder.py
@@ -3,6 +3,7 @@
 from typing import Any, Awaitable, Callable, Generator, Optional
 
 from merino.middleware.geolocation import Location
+from merino.providers.weather.backends.protocol import WeatherContext
 
 MaybeStr = Optional[str]
 Triplet = tuple[MaybeStr, MaybeStr, MaybeStr]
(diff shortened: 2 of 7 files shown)


merino/providers/weather/backends/accuweather/backend.py
```diff
@@ -733,7 +742,10 @@ async def get_location_by_geolocation(
             processor = process_location_response_with_country_and_region
         else:
             url_path = self.url_cities_path.format(country_code=country)
-            processor = process_location_response_with_country
+            partial_process_func = partial(
+                process_location_response_with_country, geolocation.coordinates
+            )
+            processor = partial_process_func
```
Herraj, line 748: TIL about `partial`

merino/providers/weather/backends/accuweather/utils.py
```diff
@@ -93,7 +95,9 @@ def process_location_response_with_country_and_region(response: Any) -> dict[str
             return None
 
 
-def process_location_response_with_country(response: Any) -> dict[str, Any] | None:
+def process_location_response_with_country(
+    coordinates: Coordinates | None, response: Any
+) -> dict[str, Any] | None:
```
Herraj, line 100: Wondering if we should make a return type instead of having `dict[str, Any]`. Looks like this function always returns this object 🤔 
```
{
                "Key": key,
                "LocalizedName": localized_name,
                "AdministrativeArea": {"ID": administrative_area_id},
}
```

merino/providers/weather/backends/accuweather/utils.py
```diff
@@ -197,3 +203,59 @@ def get_language(requested_languages: list[str]) -> str:
...
+        try:
+            return {
+                "key": closest_location["Key"],
+                "localized_name": closest_location["LocalizedName"],
+                "administrative_area_id": closest_location["AdministrativeArea"]["ID"],
+            }
+        except KeyError:
+            return None
+    return None
+
```
Herraj, line 242: Same return type comment for this function.

Also, would the function below, `get_lat_long_distance` ever return a negative distance? If that's not the case then do we need the `if d < min_distance` part on line 226? Assuming the `d` returned will always be smaller than infinity (`math.inf`)?

Am  understanding this correctly 🤔

Herraj (approved):

merino/providers/weather/backends/accuweather/utils.py
tiftran: ah yup, sounds like a good idea

merino/providers/weather/backends/accuweather/utils.py
tiftran: i think the math.sqrt function returns a positive integer, so we should be good.

merino/providers/weather/backends/accuweather/pathfinder.py
```diff
@@ -29,7 +30,7 @@
 )
 
 
-def compass(location: Location) -> Generator[Triplet, None, None]:
+def compass(location: Location) -> Generator[str | None, None, None]:
```
ncloudioj, line 33: Let's update the docstring to reflect the change.

merino/providers/weather/backends/accuweather/utils.py
```diff
@@ -116,6 +120,8 @@ def process_location_response_with_country(response: Any) -> dict[str, Any] | No
                 "localized_name": localized_name,
                 "administrative_area_id": administrative_area_id,
             }
+        case [_, _, *_rest]:
```
ncloudioj, line 133: Hmm, what are the first two, why are they skipped? Can you add a comment for those?

tests/unit/providers/weather/backends/test_utils.py
```diff
@@ -0,0 +1,179 @@
...
+
+def test_process_location_response_with_country(
+    coordinates: Coordinates,
+    location_response: dict[str, Any],
+    expected_location_results: dict[str, Any],
+):
+    """Test location response with multiple locations are handled correctly."""
+    assert expected_location_results == process_location_response_with_country(
+        coordinates, location_response
+    )
```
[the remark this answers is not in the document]
tiftran: yeah that probs is a good idea, going to file a bug so we don't forget 🙂

merino/providers/weather/backends/accuweather/utils.py
```diff
@@ -197,3 +203,59 @@ def get_language(requested_languages: list[str]) -> str:
...
+            }
+        except KeyError:
+            return None
+    return None
+
+
+def get_lat_long_distance(lat1: float, long1: float, lat2: float, long2: float) -> float:
+    """Calculate distance between two coordinates via the Haversine formula."""
+    to_radian_ratio = math.pi / 180
+    lat1 = lat1 * to_radian_ratio
```
ncloudioj:
```py
lat1 = math.radians(lat1)
long1 = math.radians(long1)
...
...
```

merino/providers/weather/backends/accuweather/utils.py
```diff
@@ -197,3 +203,59 @@ def get_language(requested_languages: list[str]) -> str:
...
+    long1 = long1 * to_radian_ratio
+
+    lat2 = lat2 * to_radian_ratio
+    long2 = long2 * to_radian_ratio
+
+    d_lat = lat2 - lat1
+    d_long = long2 - long1
+
+    radicand = 1 - math.cos(d_lat) + (math.cos(lat1) * math.cos(lat2) * (1 - math.cos(d_long)))
+    sqrt = math.sqrt(radicand / 2)
```
ncloudioj, line 266: This might bite us later, let's give it another name. Even "s9rt" would be better.

merino/providers/weather/backends/accuweather/utils.py
tiftran: yup, was reading the docs and updated 🙂

merino/providers/weather/backends/accuweather/utils.py
tiftran: oooh I was trying to ensure that there's at least two elements

ncloudioj: Ah, i see. I think we can drop this arm and move `return get_closet_location_by_distance(...)` to the `case _` one, right?

tiftran: oh yup, you're right

merino/providers/weather/backends/accuweather/backend.py
```diff
@@ -538,9 +539,15 @@ async def get_weather_report_with_location_key(
         return await self.make_weather_report(cached_report, weather_context)
 
     async def _fetch_from_cache(
-        self, country: str | None, region: str | None, city: str | None, language: str
+        self,
+        weather_context: WeatherContext,
+        region,
```
ncloudioj, line 544: Hmm, why is `region` still needed given `weather_context` is added? Can we use the one from the context?

merino/providers/weather/backends/accuweather/pathfinder.py
ncloudioj: Can you update the "Returns" section as well?

tiftran: eek, yup

merino/providers/weather/backends/accuweather/backend.py
tiftran: weather context has `regions` which is all the regions, region is the one that the pathfinder gives, we could introduced `selected_region` in the weather context object?

ncloudioj: Sounds good. Let's do that. Having them both here is confusing and janky.

tiftran:
im sorry, squashed to rebase 🙈

PR #723, opened 2024-11-28
[DISCO-3088] weather: add more rules to country,city,region pathfinder

tiftran:
## References

JIRA: [DISCO-3088](https://mozilla-hub.atlassian.net/browse/DISCO-3088)

## Description
I pulled the logs after #718 was merged. Added some more countries to our pathfinder. 
This isn't reflective of all the countries/cities found in the logs, but I think it's a start. As a result with this though, if there is a city within those countries that don't follow the same region convention (like London). It'll fail and appear in the logs...
11 countries.... out of 195? 😅 Alternatively we could have another "wonder" file with the mapping, but I think that's too much.

## PR Review Checklist

[DISCO-3088]: https://mozilla-hub.atlassian.net/browse/DISCO-3088?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/merino/providers/weather/backends/accuweather/backend.py
+++ b/merino/providers/weather/backends/accuweather/backend.py
@@ -666,7 +666,9 @@ class AccuweatherBackend:
 
             if location is None:
                 if not skip_log:
-                    logger.warning(f"Unable to find location for {country}/{city}")
+                    logger.warning(
+                        f"Unable to find location for {country}/{city}/{geolocation.regions}"
+                    )
                 return None
 
         try:
--- a/merino/providers/weather/backends/accuweather/backend.py
+++ b/merino/providers/weather/backends/accuweather/backend.py
@@ -667,7 +667,7 @@ class AccuweatherBackend:
             if location is None:
                 if not skip_log:
                     logger.warning(
-                        f"Unable to find location for {country}/{city}/{geolocation.regions}"
+                        f"Unable to find location for {country}/{city}, regions: {geolocation.regions}"
                     )
                 return None
 
--- a/tests/unit/providers/weather/test_provider.py
+++ b/tests/unit/providers/weather/test_provider.py
@@ -18,7 +18,10 @@ from merino.config import settings
 from merino.middleware.geolocation import Location
 from merino.providers.base import BaseSuggestion, SuggestionRequest
 from merino.providers.custom_details import CustomDetails, WeatherDetails
-from merino.providers.weather.backends.accuweather.pathfinder import set_region_mapping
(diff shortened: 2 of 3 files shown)


merino/providers/weather/backends/accuweather/backend.py
```diff
@@ -658,7 +658,9 @@ async def as_awaitable(val: Any) -> Any:
                 logger.warning(f"{exc}")
 
             if location is None:
-                logger.warning(f"Unable to find location for {country}/{city}")
+                logger.warning(
+                    f"Unable to find location for {country}/{city}/{geolocation.regions}"
+                )
```
tiftran, line 663: I added regions back, since it's helpful.

tests/unit/providers/weather/test_provider.py
```diff
@@ -181,8 +184,9 @@ async def test_fetch_mapping(
     caplog.set_level(logging.INFO)
 
     assert len(statsd_mock.gauge.call_args_list) == 0
+    clear_region_mapping()
 
-    set_region_mapping("Canada", "Vancouver", "BC")
+    set_region_mapping("NL", "Andel", "NB")
```
tiftran, line 189: I realized the original test values wasn't reflective of what it should be

merino/providers/weather/backends/accuweather/backend.py
```diff
@@ -666,7 +666,9 @@ async def as_awaitable(val: Any) -> Any:
 
             if location is None:
                 if not skip_log:
-                    logger.warning(f"Unable to find location for {country}/{city}")
+                    logger.warning(
+                        f"Unable to find location for {country}/{city}/{geolocation.regions}"
+                    )
```
tiftran, line 671: I added regions back, since it's helpful.

ncloudioj: Cool, but it will include all the regions, is that desirable? Also let's do country/region/city.

ncloudioj (approved):
LGTM. No, wonder files are a sledge hammer. I'd bet 5 bucks that we won't need that for 2024 :)

merino/providers/weather/backends/accuweather/backend.py
tiftran: we don't know which region was used at this point, and potentially Merino might have cycled through all, so i think it is actually desirable to see all region options that were used or could've been used?

PR #755, opened 2025-01-16
[DISCO-3139] add provider for manifest endpoint

misaniwere:
## References

JIRA: [DISCO-3139](https://mozilla-hub.atlassian.net/browse/DISCO-3139)

## Description
Apologies for the large PR (most of it is file import changes due to the restructuring of the providers folder)
A few changes were made here to facilitate adding a provider for manifest:
- `BaseProvider` class along with the `__init__.py` folder that contains init methods for suggest providers have been moved into `providers/suggest/`
- The manifest provider files all live in `providers/manifest`
- Some test folder restructuring to reflect that as well
- Manifest config is now `default.manifest`. The provider tag is used to initialize the suggest providers currently so this was an easy/low effort way to decouple the manifest provider from the suggest providers.

## PR Review Checklist

[DISCO-3139]: https://mozilla-hub.atlassian.net/browse/DISCO-3139?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


misaniwere:
Manifest endpoint tests have been commented out in the mean time. I'll work with Herraj to get those added. Otherwise, looking forward to feedback :)

gruberb:
@misaniwere Can you create a follow up issue for the commented tests, or work with @Herraj to solve them in this PR?

merino/configs/default.toml
```diff
@@ -238,6 +238,28 @@ current_condition_ttl_sec = 1800 # 1/2 hr
...
+# the resync failure can be retried soon.
+cron_interval_sec = 60
+
+# MERINO_PROVIDERS__MANIFEST__RESYNC_INTERVAL_SEC
+# Time between re-syncs of the manifest file, in seconds. Defaults to 24 hours.
+resync_interval_sec = 86400
+
+# MERINO_PROVIDERS__MANIFEST__GCS_PROJECT
+# GCS project name that contains manifest data
+gcs_project = ""
```
gruberb, line 254: I assume we have to set the real project name here! The same goes for `gcs_bucket` and `gcs_blob_name`.

merino/main.py
```diff
@@ -38,11 +39,12 @@ async def lifespan(app: FastAPI):
...
     configure_sentry()
     await configure_metrics()
-    await providers.init_providers()
+    await suggest.init_providers()
+    await manifest.init_provider()
     curated_recommendations.init_provider()
     yield
     # Shut down providers and clean up.
-    await providers.shutdown_providers()
+    await suggest.shutdown_providers()
```
gruberb, line 47: And we don't need to shut down the `manifest` provider here?

merino/providers/manifest/__init__.py
```diff
@@ -0,0 +1,46 @@
...
+    start = timer()
+
+    provider = Provider(
+        backend=ManifestBackend(),
+        resync_interval_sec=settings.manifest.resync_interval_sec,
+        cron_interval_sec=settings.manifest.cron_interval_sec,
+    )
+
+    # initialize provider and record time
+    init_metric = "providers.initialize"
```
gruberb, line 30: I don't think we want to re-use the same metric name here for the `manifest` provider. The other providers are already using `providers.initialize`.

merino/providers/manifest/__init__.py
```diff
@@ -0,0 +1,46 @@
...
+
+        logger.info(
+            "Provider initialization completed",
+            extra={"provider": "manifest", "elapsed": timer() - start},
+        )
+
+
+def get_provider() -> Provider:
+    """Return manifest provider"""
+    print(f"DEBUG: provider={provider}")
```
gruberb, line 43: Remove the DEBUG statement here?

merino/providers/manifest/__init__.py
```diff
@@ -0,0 +1,46 @@
...
+    if provider is None:
+        raise ValueError("Provider has not been initialized.")
```
gruberb, line 45: Should we name the provider which couldn't be initialized?

merino/providers/manifest/backends/filemanager.py
```diff
@@ -0,0 +1,75 @@
...
+    gcs_bucket_path: str
+    gcs_project_path: str
+
+    def __init__(
+        self,
+        gcs_project_path: str,
+        gcs_bucket_path: str,
+        blob_name: str,
+    ) -> None:
+        """:param gcs_project_path: Google Cloud project (or path to credentials).
```
gruberb, line 28: Format, and also, it's not the path to the credentials.

merino/providers/manifest/backends/manifest.py
```diff
@@ -0,0 +1,63 @@
...
+
+    def __init__(self) -> None:
+        """Initialize the Manifest backend.
+        GCS configuration is read from `settings.manifest.*`:
+        - `gcs_project`
+        - `gcs_bucket`
+        - `gcs_blob_name`
+        """
+        # You could store more custom config here if needed,
+        # e.g., caching intervals or fallback flags.
```
gruberb, line 29: Remove generic comment

merino/providers/manifest/backends/manifest.py
```diff
@@ -0,0 +1,63 @@
...
+
+        result_code, manifest_data = remote_filemanager.get_file()
+
+        match GetManifestResultCode(result_code):
+            case GetManifestResultCode.SUCCESS:
+                logger.info("Manifest data loaded remotely from GCS.")
+                return (result_code, manifest_data)
+
+            case GetManifestResultCode.SKIP:
+                logger.debug("Manifest data was not updated (SKIP).")
```
gruberb, line 57: Why `debug` and not `info`?

gruberb (commented):
Looking good, thanks for taking on this huge refactor!

merino/providers/manifest/backends/filemanager.py
```diff
@@ -0,0 +1,64 @@
...
+        :param gcs_bucket_path: GCS bucket name to fetch from.
+        :param blob_name: Name of the blob in the GCS bucket.
+        """
+        self.gcs_client = GcsUploader(gcs_project_path,gcs_bucket_path, "")
+
+    def get_file(self) -> tuple[GetManifestResultCode, ManifestData | None]:
+        """Fetch the remote manifest file from GCS"""
+        try:
+            # bucket: Bucket = self.client.get_bucket(self.gcs_bucket_path)
```
gruberb, line 37: Why commented out?

misaniwere:
> @misaniwere Can you create a follow up issue for the commented tests, or work with @Herraj to solve them in this PR?

yup, already on it :) https://github.com/mozilla-services/merino-py/pull/755#issuecomment-2605768921

merino/main.py
misaniwere: ideally yes but I don't think there's any cleanup tasks we have to do for the manifest provider. Unless we want to clear the file? what do you think @gruberb

merino/providers/manifest/__init__.py
misaniwere: yeah i intentionally wanted to re-use this so we can track this provider alongside the rest. Any particular reason you'd suggest to separately track the initialization of this one? @gruberb

merino/providers/manifest/__init__.py
misaniwere: ooh nice catch!

merino/providers/manifest/backends/filemanager.py
misaniwere: oops, just copy pasted this from top picks. The refactor @Herraj and I are working on would most likely change this so we'll clean this up as well

merino/configs/default.toml
misaniwere: yeah, do we have the real project and bucket values to be updated here?

merino/main.py
gruberb: Don't we somehow have to cancel the cron job when we shut down the provider? But from gathering other providers, it seems we don't need to!

merino/providers/manifest/__init__.py
gruberb: Maybe I miss-understood how we collect metrics. In my mind, we will override this metric if we name it the same. Or does `statsd` automatically namespace this?

merino/configs/default.toml
gruberb: We use the same as `top_picks` (for now)

```
[stage.providers.top_picks]
# MERINO_PROVIDERS__TOP_PICKS__GCS_PROJECT
gcs_project = "moz-fx-merino-nonprod-ee93"

# MERINO_PROVIDERS__TOP_PICKS__GCS_BUCKET
gcs_bucket = "merino-images-stagepy"
```

merino/providers/manifest/__init__.py
misaniwere: apparently it doesn't override but aggregates them all which is also not what we want because there will be a misrepresentation - first metric is timing multiple providers vs second metric timing just one. in this case, I'm thinking to rename both metrics for more clarity. so suggest would be `providers.suggest.initialize` and manifest would be `providers.manifest.initialize`. We'd probably need to update grafana. What do you think? @gruberb

merino/main.py
misaniwere: you're right, we would need to cancel the cron task on shutdown. I'm not sure how the top picks provider gets away with not doing that though. any ideas @ncloudioj ?

merino/providers/manifest/__init__.py
gruberb: Or we leave the "manifest.initialize" out for now, because you also have a second metric: `lient.timeit_task(provider.initialize(), "providers.initialize.manifest")`.

misaniwere: sounds good! 👍🏾

merino/configs/default.toml
```diff
@@ -238,6 +238,28 @@ current_condition_ttl_sec = 1800 # 1/2 hr
...
+# the resync failure can be retried soon.
+cron_interval_sec = 60
+
+# MERINO_PROVIDERS__MANIFEST__RESYNC_INTERVAL_SEC
+# Time between re-syncs of the manifest file, in seconds. Defaults to 24 hours.
+resync_interval_sec = 86400
+
+# MERINO_PROVIDERS__MANIFEST__GCS_PROJECT
+# GCS project name that contains manifest data
+gcs_project = "moz-fx-merino-nonprod-ee93"
```
ncloudioj, line 254: I think both `gcs_project` and `gcs_bucket` should be left empty here rather than using the stage configs, you can specify them in "stage.toml" and "production.toml", respectively. You can see an example for "{env}.providers.top_picks.[gcs_project | gcs_bucket]".

I'd actually recommend extracting those two from `provider.top_picks` and making a top-level `{env.image_manifest}` so it can be shared by all the consumers. Though that'd require some more changes in this patch, so we can do that as a follow up if we want.

merino/configs/default.toml
```diff
@@ -238,6 +238,28 @@ current_condition_ttl_sec = 1800 # 1/2 hr
...
+
+# MERINO_PROVIDERS__MANIFEST__GCS_BUCKET
+# GCS bucket that contains manifest data files
+gcs_bucket = "merino-images-stagepy"
+
+# MERINO_PROVIDERS__MANIFEST__GCS_BLOB_NAME
+# Name of the file containing manifest data in GCS
+gcs_blob_name = "top_picks_latest.json"
```
ncloudioj, line 262: Feel free to use a constant for this rather than a configuration. As a naming convention, we won't be changing the name for this blob so it's very unlikely for us to re-configure it in the future.

merino/configs/testing.toml
```diff
@@ -64,6 +64,14 @@ domain_data_source = "local"
 # The backend of the provider. Either "elasticsearch" or "test".
 backend = "test"
 
+[testing.manifest]
+gcs_project = "test_gcp_uploader_project"
+gcs_bucket = "test_gcp_uploader_bucket"
+gcs_blob_name = "20250101120555_top_picks.json"
```
ncloudioj, line 70: Now that we'd always use `top_picks_latest.json`, not need to use a different one for testing.

merino/main.py
ncloudioj: In general, no need to cancel or disable the cron job if it's only doing read-only tasks such as loading data from GCS here. Explicit shutdown is needed for providers if they or their underlying backends run write tasks or are holding resources that require explicit shutdown such as Redis / ES clients or long-lived HTTP clients.

merino/providers/manifest/provider.py
```diff
@@ -0,0 +1,88 @@
...
+                    return None
+        except ManifestBackendError as err:
+            logger.error("Failed to fetch data from Manifest backend: %s", err)
+
+        finally:
+            self.data_fetched_event.set()
+
+    def _should_fetch(self) -> bool:
+        """Determine if we should fetch new data based on time elapsed."""
+        return bool((time.time() - self.last_fetch_at) >= self.resync_interval_sec)
```
ncloudioj, line 84:
```suggestion
        return (time.time() - self.last_fetch_at) >= self.resync_interval_sec
```

merino/providers/manifest/provider.py
```diff
@@ -0,0 +1,88 @@
...
+        cron_interval_sec: int,
+        name: str = "manifest",
+    ) -> None:
+        self.backend = backend
+        self.name = name
+        self.resync_interval_sec = resync_interval_sec
+        self.cron_interval_sec = cron_interval_sec
+        self.last_fetch_at = 0.0
+        self.manifest_data = ManifestData(domains=[])
+        self.data_fetched_event = asyncio.Event()
```
ncloudioj, line 42: Nice pick!

ncloudioj (changes requested):
Looks good. Just a few minor nits to be fixed.

merino/configs/default.toml
misaniwere: hmm good idea! to keep this already large PR from getting larger, I'll do this as a follow up

gruberb (approved):
Thanks for all the work @misaniwere (and the support @Herraj). Looking good! I think we all saw some possibilities for future clean ups, but I am good with leaving this for a follow-up PR! 

Let's squash the commits and get it in!

merino/providers/manifest/provider.py
```diff
@@ -0,0 +1,88 @@
...
+    ManifestData,
+)
+
+logger = logging.getLogger(__name__)
+
+
+class Provider:
+    """Provide access to in-memory manifest data fetched from GCS."""
+
+    manifest_data: ManifestData | None
```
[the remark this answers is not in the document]
misaniwere: sometimes we return `None` on skip or fail. We're only ever setting `self.manifest_data` when we have data but the linter does not allow `self.manifest_data = data` to be done (if we don't have `None` in the type annotation) if because `get_file` function sometimes returns `None`

PR #777, opened 2025-01-30
[DISCO-3067] Trial async gcs client

Herraj:
## References

JIRA: [DISCO-3067](https://mozilla-hub.atlassian.net/browse/DISCO-3067)

## Description
Adding an async gcs client in parallel to the existing sync one. This does not override the existing logic, only does the same task in parallel. The main goal is to see if the auth works fine.

## PR Review Checklist

[DISCO-3067]: https://mozilla-hub.atlassian.net/browse/DISCO-3067?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


merino/utils/gcs/async_gcs_client.py
```diff
@@ -0,0 +1,47 @@
...
+class AsyncGcsClient:
+    """Class that provides wrapper functions around gcloud-aio-storage functions. More functionality to be added."""
+
+    storage: Storage
+
+    def __init__(self) -> None:
+        try:
+            self.storage = Storage()
+        except Exception as ex:
+            raise RuntimeError(f"Failed to initialize async GCS client: {ex}")
```
Herraj, line 23: If there are any auth or any other issues in initializing this client, we would throw this exception which would be caught by the `provider.py` and will return `None`.

merino/utils/gcs/async_gcs_client.py
```diff
@@ -0,0 +1,47 @@
...
+
+    async def get_manifest_from_blob(
```
misaniwere, line 25: might be better to make this method more generic since this is the async client class and not necessarily related to the manifest provider. What do you think about renaming this to `get_blob_by_name` or something similar and have it return a generic `Blob` object?

misaniwere: then I think we could call the method in the manifest filemanager to retrieve the blob (in place of `get_file_by_name`)

merino/providers/manifest/provider.py
```diff
@@ -86,3 +86,12 @@ def _should_fetch(self) -> bool:
     def get_manifest_data(self) -> ManifestData | None:
         """Return manifest data"""
         return self.manifest_data
+
+    async def get_manifest_data_via_async_client(self) -> ManifestData | None:
+        """Return manifest data"""
+        try:
+            manifest_via_async = await self.backend.fetch_via_async_gcs_client()
+            return manifest_via_async
```
misaniwere, line 94: hmm should we be caching this result somewhere?

misaniwere (commented):
looking good! a few comments

merino/utils/gcs/async_gcs_client.py
Herraj: right, that's a better idea! I'm down to refactor this once I try it out with this and confirm it works. Basically I want to get this in and see if this works and then follow it up with a refactor into an util module. wdyt?

Can do it now, no biggie

misaniwere: oh i see, no worries then, you can do this in a follow up

PR #781, opened 2025-02-04
[DISCO-3067] Refactor async gcs client to be more generic and remove redundant logic

Herraj:
## References

JIRA: [DISCO-3067](https://mozilla-hub.atlassian.net/browse/DISCO-3067)

## Description
This is undoing what [this](https://github.com/mozilla-services/merino-py/pull/777) PR did, which was a trial PR. It removes the wrapper class.

## PR Review Checklist

[DISCO-3067]: https://mozilla-hub.atlassian.net/browse/DISCO-3067?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


merino/web/api_v1.py
```diff
@@ -341,15 +341,12 @@ async def get_manifest(
 
     with metrics_client.timeit("manifest.request.timing"):
         manifest_data = provider.get_manifest_data()
-        manifest_data_via_async_client = await provider.get_manifest_data_via_async_client()
```
Herraj, line 344: removing the duplicate call to download the same blob. We've verified in a previous PR that this client works.

merino/utils/gcs/async_gcs_client.py
```diff
@@ -1,18 +1,16 @@
...
 from gcloud.aio.storage import Storage
-from merino.providers.manifest.backends.protocol import ManifestData
 
 
 logger = logging.getLogger(__name__)
 
 
 class AsyncGcsClient:
-    """Class that provides wrapper functions around gcloud-aio-storage functions. More functionality to be added."""
+    """Class that provides wrapper functions around gcloud-aio-storage functions.
```
ncloudioj, line 11: Do we still need this wrapper? Looks like all it does is wrap the `download` call with some additional logging. Do we also use this to facilitate testing (e.g. dependency ingestion)?

ncloudioj: Also noticed that there was exception handling in `__init__()`. We don't normally handle exceptions in the constructor. If it fails in the rare cases, just let it fail. Those errors will be captured and reported by Sentry for further investigation. You can reference `utils/http_client.py` for that.

merino/utils/gcs/async_gcs_client.py
```diff
@@ -22,26 +20,20 @@ def __init__(self) -> None:
...
+                return object
 
-                return manifest
-        except JSONDecodeError:
-            await self.storage.close()
-            logger.error(f"Tried to load invalid json for blob: {blob_name}")
             return None
         except Exception as ex:
             await self.storage.close()
-            logger.error(f"Unexpected error when downloading blob: {ex}")
```
ncloudioj, line 46: Calling `self.storage.close()` inside of `get_object_from_bucket()` is problematic. If the consumer calls this function twice, the second one will fail b/c the client is already closed. You might want to add a `close()` to this client so the consumer can close it explicitly. Again, perhaps we should think if this wrapper is still necessary.

ncloudioj (approved):
Looks good. Will we do a follow up to swap the sync GCS client with the async replacement? Specifically, we can drop the `asyncio.to_thread()` bit as we can use the async client to interface with GCS now.

PR #784, opened 2025-02-05
DISCO-3202 - Enrich the curated-recommendations endpoint with icon-urls

gruberb:
## References

JIRA: 
- [DISCO-3204](https://mozilla-hub.atlassian.net/browse/DISCO-3204)
- [DISCO-3204](https://mozilla-hub.atlassian.net/browse/DISCO-3203)
- [DISCO-3204](https://mozilla-hub.atlassian.net/browse/DISCO-3205)

## Description
Enriching the `/curated-recommendations` endpoint with a `iconUrl` field for each article recommendation. The `iconUrl` is a `HttpUrl`, taken from the `Manifest` provider, and if the favicon for the domain is available, we attach it to a recommendation. Otherwise, we return an empty string. 

In the background, the `Manifest` is updated once a week, and therefore the icons will be updated in this interval. Which means newly added domains to the `Manifest` will be available to this endpoint at that time.

## PR Review Checklist

[DISCO-3204]: https://mozilla-hub.atlassian.net/browse/DISCO-3204?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[DISCO-3204]: https://mozilla-hub.atlassian.net/browse/DISCO-3204?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[DISCO-3204]: https://mozilla-hub.atlassian.net/browse/DISCO-3204?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/merino/providers/manifest/provider.py
+++ b/merino/providers/manifest/provider.py
@@ -3,6 +3,9 @@
 import asyncio
 import time
 import logging
+from urllib.parse import urlparse
+
+from pydantic import HttpUrl
 
 from merino.providers.manifest.backends.filemanager import GetManifestResultCode
 from merino.utils import cron
@@ -20,6 +23,7 @@ class Provider:
     """Provide access to in-memory manifest data fetched from GCS."""
 
     manifest_data: ManifestData | None
+    domain_lookup_table: dict[str, int]
     cron_task: asyncio.Task
     resync_interval_sec: int
     cron_interval_sec: int
@@ -39,6 +43,7 @@ class Provider:
         self.cron_interval_sec = cron_interval_sec
         self.last_fetch_at = 0.0
         self.manifest_data = ManifestData(domains=[])
+        self.domain_lookup_table = {}
         self.data_fetched_event = asyncio.Event()
 
         super().__init__()
@@ -63,8 +68,11 @@ class Provider:
             result_code, data = await self.backend.fetch()
 
             match GetManifestResultCode(result_code):
-                case GetManifestResultCode.SUCCESS:
+                case GetManifestResultCode.SUCCESS if data is not None:
                     self.manifest_data = data
+                    self.domain_lookup_table = {
+                        domain.domain: idx for idx, domain in enumerate(data.domains)
+                    }
                     self.last_fetch_at = time.time()
 
                 case GetManifestResultCode.SKIP:
@@ -86,3 +94,27 @@ class Provider:
     def get_manifest_data(self) -> ManifestData | None:
         """Return manifest data"""
         return self.manifest_data
+
+    def get_icon_url(self, url: str | HttpUrl) -> str | None:
+        """Get icon URL for a domain.
+
+        Args:
+            url: Full URL to extract domain from (string or HttpUrl)
+
+        Returns:
+            Icon URL if found, None otherwise
+        """
+        try:
+            url_str = str(url)
+            # Remove www. and get the domain
+            domain = urlparse(url_str).netloc.replace("www.", "")
+            # Strip TLD by taking first part of domain
+            base_domain = domain.split(".")[0] if "." in domain else domain
+
+            idx = self.domain_lookup_table.get(base_domain, -1)
+            if idx >= 0 and self.manifest_data is not None:
+                return self.manifest_data.domains[idx].icon
+
+        except Exception as e:
+            logger.warning(f"Error getting icon for URL {url}: {e}")
+        return None
--- a/tests/integration/api/v1/curated_recommendations/test_curated_recommendations.py
+++ b/tests/integration/api/v1/curated_recommendations/test_curated_recommendations.py
@@ -46,6 +46,8 @@ from merino.curated_recommendations.protocol import (
 )
 from merino.curated_recommendations.protocol import CuratedRecommendation
 from merino.main import app
+from merino.providers.manifest import get_provider as get_manifest_provider
+from merino.providers.manifest.backends.protocol import Domain
 from tests.integration.api.conftest import fakespot_feed
 
 
@@ -121,6 +123,15 @@ def extended_expiration_corpus_backend(
     )
 
 
+@pytest.fixture(autouse=True)
+def setup_manifest_provider(manifest_provider):
+    """Set up the manifest provider dependency"""
+    app.dependency_overrides[get_manifest_provider] = lambda: manifest_provider
+    yield
+    if get_manifest_provider in app.dependency_overrides:
+        del app.dependency_overrides[get_manifest_provider]
+
+
 @pytest.fixture(name="corpus_provider")
 def provider(
     corpus_backend: CorpusApiBackend,
@@ -140,7 +151,7 @@ def provider(
 
 
 @pytest.fixture(autouse=True)
-def setup_providers(corpus_provider):
+def setup_suggest_providers(corpus_provider):
     """Set up the curated recommendations provider"""
     app.dependency_overrides[get_provider] = lambda: corpus_provider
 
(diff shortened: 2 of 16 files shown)


merino/providers/manifest/__init__.py
```diff
@@ -3,13 +3,13 @@
 from timeit import default_timer as timer
 import logging
 from merino.utils import metrics
-from merino.providers.manifest.provider import Provider
+from merino.providers.manifest.provider import ManifestProvider
```
misaniwere, line 6: Since this is in the manifest folder, it should be fine to name it `Provider` and then we can import it as 

`from merino.providers.manifest.provider import Provider as ManifestProvider`

merino/services/__init__.py
```diff
@@ -0,0 +1 @@
+"""Services package for Merino application."""
```
ncloudioj, line 1: Personally, I find it a bit confusing to introduce this services concept into Merino as the Merino web app is already a service itself. Usually we'd refer to those running units as tasks or jobs instead of services in Merino.

As for this `IconEnricher`, I feel like it's a little bit over-engineered. Specifically, its main functionality perhaps should serve as an icon URL lookup "service", i.e. given a domain, return a icon URL if found. I think it'd better off letting consumer decide how to use the icon URL instead of doing the enrichment here for them, doing so would inevitably introduce dependencies and implementation details from the higher level components.

There is also an ergonomic nit: for each icon lookup, the consumer needs to create a new instance of `IconEnrichService`. I think we can simply add `get_icon_url()` to `ManifestProvider` so that consumers can use it directly.

Let me know what you think.

merino/providers/manifest/provider.py
```diff
@@ -86,3 +87,25 @@ def _should_fetch(self) -> bool:
...
+            Domain object if found, None otherwise
+        """
+        if not self.manifest_data or not self.manifest_data.domains:
+            return None
+
+        # Clean up domain
+        domain = domain.lower().strip()
+
+        # Find matching domain
+        for domain_info in self.manifest_data.domains:
```
ncloudioj, line 107: Scanning this domain list (w/ thousands of items) for each lookup would be costly. I think we can make a lookup table "{domain => icon-url}" based on the original manifest to speed things up.

merino/services/__init__.py
gruberb: Fair point. I had multiple versions, and somehow, I didn't like putting the logic how to add the icon urls to the feeds directly in the endpoint. But yeah, I get. 

I refactored the PR and removed the `service`, and added the lines inside the endpoint directly.

merino/providers/manifest/provider.py
gruberb: Great, yes, did not think about that! Thanks for pointing it out!

merino/providers/manifest/provider.py
```diff
@@ -86,3 +94,23 @@ def _should_fetch(self) -> bool:
...
+        Returns:
+            Icon URL if found, None otherwise
+        """
+        try:
+            url_str = str(url)
+            domain = urlparse(url_str).netloc.replace("www.", "")
+            idx = self.lookup_table.get(domain)
+            if idx is not None and self.manifest_data is not None:
+                return self.manifest_data.domains[idx].icon
+
```
misaniwere, line 113:
```suggestion
idx = self.lookup_table.get(domain, -1)  # Default to -1 if domain is not found
if idx >= 0 and self.manifest_data is not None:
    return self.manifest_data.domains[idx].icon

```
this might be better since idx is an int

merino/providers/manifest/provider.py
```diff
@@ -20,6 +23,7 @@ class Provider:
     """Provide access to in-memory manifest data fetched from GCS."""
 
     manifest_data: ManifestData | None
+    lookup_table: dict[str, int]
     cron_task: asyncio.Task
```
misaniwere, line 27: nit: for clarity, what do you think about calling this `domain_lookup_table`

merino/web/api_v1.py
```diff
@@ -320,7 +321,28 @@ async def curated_content(
 
     [curated-topics-doc]: https://mozilla-hub.atlassian.net/wiki/x/LQDaMg
     """
-    return await provider.fetch(curated_recommendations_request)
+    response = await provider.fetch(curated_recommendations_request)
+
+    # Add icons to recommendations
+    for rec in response.data:
```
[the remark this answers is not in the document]
Herraj: I agree, we can append that property here where we build out the list of `CorpusItem`s

https://github.com/mozilla-services/merino-py/blob/5ae463c8460a83dd3923535f6ab858fa5b7dbc5f/merino/curated_recommendations/corpus_backends/corpus_api_backend.py#L325

merino/providers/manifest/provider.py
```diff
@@ -86,3 +94,23 @@ def _should_fetch(self) -> bool:
...
+        try:
+            url_str = str(url)
+            domain = urlparse(url_str).netloc.replace("www.", "")
+            idx = self.domain_lookup_table.get(domain, -1)
+            if idx >= 0 and self.manifest_data is not None:
+                return self.manifest_data.domains[idx].icon
+
+        except Exception as e:
+            logger.warning(f"Error getting icon for URL {url}: {e}")
+        return None
```
Herraj, line 116: Nice clean logic with the look up table and this!

Herraj: I know this function is being tested indirectly by the integration test but wdyt about adding a couple of unit tests (happy / unhappy path) for this function? 🤔

Herraj (approved):
🖼️

merino/providers/manifest/provider.py
gruberb: Yupp, good idea! Added a few tests (and found that our manifest fixture was slightly wrong due this - so very much appreciated you pointing it out :) )

tests/integration/api/v1/curated_recommendations/test_curated_recommendations.py
```diff
@@ -121,6 +125,16 @@ def extended_expiration_corpus_backend(
...
 
 
+@pytest.fixture(autouse=True)
+def setup_manifest_provider(manifest_provider):
+    """Set up the manifest provider dependency"""
+    app.dependency_overrides[get_manifest_provider] = lambda: manifest_provider
+    yield
+    # Clean up after test
+    if get_manifest_provider in app.dependency_overrides:
+        del app.dependency_overrides[get_manifest_provider]
```
mmiermans, line 135: Can't this be removed, because the manifest provider is already set in the main `corpus_backend` fixture? https://github.com/mozilla-services/merino-py/blob/31448f3635351318887502aa94ae7e20a9bfe5d3/tests/integration/api/v1/curated_recommendations/corpus_backends/fixtures.py#L68-L75

mmiermans: If not, then could you rename `setup_providers` to distinguish between these two fixtures that both set up providers?

tests/integration/api/v1/curated_recommendations/test_curated_recommendations.py
```diff
@@ -1529,3 +1543,94 @@ def mock_post_by_days_ago(*args, **kwargs):
...
+        corpus_backend=backend,
+        extended_expiration_corpus_backend=ExtendedExpirationCorpusBackend(
+            backend=backend, engagement_backend=MockEngagementBackend()
+        ),
+        engagement_backend=MockEngagementBackend(),
+        prior_backend=ConstantPrior(),
+        fakespot_backend=MockFakespotBackend(),
+    )
+
+    app.dependency_overrides[get_provider] = lambda: provider_instance
```
mmiermans, line 1586: Is there a need to set up a custom `CuratedRecommendationsProvider` in this test? Or can this be removed? The `setup_providers` fixtures automatically sets up one. I believe you should still be able to modify the `manifest_provider` that's used by it, because fixtures get recreated for every function by default.

mmiermans (commented):
Only have some questions related to testing. The rest looks good. Thanks for making this change! 👍

mmiermans:
> Since this would add a new filed to the recommendation endpoint, let's confirm that the API consumer(s) are ready to consume (or ignore) it without affecting anything there. I think @mmiermans should be a blocking reviewer for this patch.

We set the expectation that adding attributes is not a breaking change, but that clients (Desktop and Android for now) will be notified of any changes regardless, to avoid surprises. If they're not already aware, could you send them a Slack message before merging this?

tests/integration/api/v1/curated_recommendations/test_curated_recommendations.py
gruberb: Good catch - I removed it!

tests/integration/api/v1/curated_recommendations/test_curated_recommendations.py
gruberb: We need this reset here for another test - I put it in the list to investigate if we can make this more efficient! I renamed `setup_providers` to `setup_suggest_providers`.

tests/integration/api/v1/curated_recommendations/test_curated_recommendations.py
```diff
@@ -140,7 +151,7 @@ def provider(
 
 
 @pytest.fixture(autouse=True)
-def setup_providers(corpus_provider):
+def setup_suggest_providers(corpus_provider):
```
mmiermans, line 154:
```suggestion
def setup_curated_recommendations_provider(corpus_provider):
```

mmiermans (approved):

PR #785, opened 2025-02-07
chore: more weather pathfinder improvements

tiftran:
## References

## PR Review Checklist

### DIFF
--- a/merino/providers/suggest/weather/backends/accuweather/pathfinder.py
+++ b/merino/providers/suggest/weather/backends/accuweather/pathfinder.py
@@ -1,12 +1,12 @@
 """Pathfinder - a utility to reconcile geolocation distinctions between MaxmindDB and AccuWeather."""
 
+import unicodedata
 from typing import Any, Awaitable, Callable, Generator, Optional
 
 from merino.middleware.geolocation import Location
 from merino.providers.suggest.weather.backends.protocol import WeatherContext
 
 MaybeStr = Optional[str]
-Pair = tuple[MaybeStr, MaybeStr]
 
 SUCCESSFUL_REGIONS_MAPPING: dict[tuple[str, str], str | None] = {("GB", "London"): "LND"}
 REGION_MAPPING_EXCLUSIONS: frozenset = frozenset(
@@ -14,11 +14,9 @@ REGION_MAPPING_EXCLUSIONS: frozenset = frozenset(
 )
 CITY_NAME_CORRECTION_MAPPING: dict[str, str] = {
     "Baie Ste. Anne": "Baie-Sainte-Anne",
-    "Banhā": "Banha",
     "Bogota D.C.": "Bogota",
     "Centro Municipality": "Centro",
     "Chihuahua City": "Chihuahua",
-    "Chūōku": "Chuoku",
     "Collingwood": "Collingwood Corner",
     "Dawson City": "Dawson",
     "Délı̨ne": "Deline",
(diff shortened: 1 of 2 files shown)


merino/providers/suggest/weather/backends/accuweather/pathfinder.py
```diff
@@ -88,7 +63,17 @@
...
 
-def compass(location: Location) -> Generator[Pair, None, None]:
+def normalize_string(input_str) -> str:
+    """Normalize string with special chcarcters"""
+    return unicodedata.normalize("NFKD", input_str).encode("ascii", "ignore").decode("ascii")
+
+
+def remove_locality_suffix(city: str) -> str:
+    """Remove either city or municipality suffix from a city name"""
+    return re.sub(r"\s*(city|municipality)$", "", city, flags=re.IGNORECASE).strip()
```
ncloudioj, line 73: Hmm, sure this is beneficial? I can think of "Mexico City", "Panama City", "New York City", "Salt Lake City", "Oklahoma City", "Quebec City", "Guatemala City", "Ho Chi Minh City", etc.

tiftran: So I did tested out "New York City", "Quebec City" and they were okay, but you're right the others may or may not work. There are a couple of cities in South America that will benefit. Given that I think it would make sense to have the suffix removed as another city name to try if the original fails.

ncloudioj: Ah, I see. If this doesn't regress other locations, we can give it a try.

You can compile this regex (`re.compile()`) as a constant so it won't be compiled on the fly for each removal.

tiftran: k so i now have it so that we try the original, normalized, then without the city/municipality suffix

merino/providers/suggest/weather/backends/accuweather/pathfinder.py
```diff
@@ -98,31 +84,29 @@ def compass(location: Location) -> Generator[Pair, None, None]:
...
     """
-    # TODO(nanj): add more heuristics to here.
-
     country = location.country
     regions = location.regions
     city = location.city
 
     if regions and country and city:
-        corrected_city = CITY_NAME_CORRECTION_MAPPING.get(city, city)
+        corrected_city = CITY_NAME_CORRECTION_MAPPING.get(city, remove_locality_suffix(city))
```
ncloudioj, line 92: Since `remove_locality_suffix()` is already handled in `explore()`, do we still need it here? Doing the suffix removal in both places seems pretty hard to follow to me. Is there a way for us to not do suffix removing here?

merino/providers/suggest/weather/backends/accuweather/pathfinder.py
```diff
@@ -149,19 +133,26 @@ async def explore(
...
-            return None, True
-
-        weather_context.selected_region = region
-        geolocation.city = city
-
-        res = await probe(weather_context)
-
-        if res is not None:
-            return res, is_skipped
+    city = geolocation.city
```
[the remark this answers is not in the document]
tiftran: oh that's fancy

merino/providers/suggest/weather/backends/accuweather/pathfinder.py
tiftran: ah yup, you're right

merino/providers/suggest/weather/backends/accuweather/pathfinder.py
```diff
@@ -98,31 +92,28 @@ def compass(location: Location) -> Generator[Pair, None, None]:
...
       - region string that could be None.
     """
-    # TODO(nanj): add more heuristics to here.
-
     country = location.country
     regions = location.regions
     city = location.city
 
     if regions and country and city:
-        corrected_city = CITY_NAME_CORRECTION_MAPPING.get(city, city)
```
ncloudioj, line 108: Hmm, is this a mis-kill? `CITY_NAME_CORRECTION_MAPPING` is still in use, eh?

ncloudioj (approved):
Looks good, r+wc.

PR #790, opened 2025-02-12
[HNT-374] Add interest picker

mmiermans:
## References

- JIRA: [HNT-374](https://mozilla-hub.atlassian.net/browse/HNT-374)
- [Interest picker API spec](https://mozilla-hub.atlassian.net/wiki/spaces/FPS/pages/1287553025/Inline+Interest+Picker+-+API+Specification)
- [Product plan](https://docs.google.com/document/d/1dTHDMNhxRre0ubsjBDWQJ7Z_X-J9EBiP9noRyLla7EI/edit?tab=t.0)

## Description
Return an 'Interest Picker' from the curated recommendation endpoint, to allow clients to follow sections.

image

## PR Review Checklist

[HNT-374]: https://mozilla-hub.atlassian.net/browse/HNT-374?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/merino/curated_recommendations/interest_picker.py
+++ b/merino/curated_recommendations/interest_picker.py
@@ -3,7 +3,7 @@
 import random
 
 from merino.curated_recommendations.protocol import (
-    CuratedRecommendationsResponse,
+    CuratedRecommendationsFeed,
     InterestPickerSection,
     InterestPicker,
 )
@@ -12,26 +12,29 @@ from merino.curated_recommendations.protocol import (
 MIN_INITIALLY_VISIBLE_SECTION_COUNT = 3  # Minimum number of sections that are shown by default.
 MIN_INTEREST_PICKER_COUNT = 8  # Minimum number of items in the interest picker.
 
+def create_interest_picker(
+        feed: CuratedRecommendationsFeed
+) -> tuple[CuratedRecommendationsFeed, InterestPicker | None]:
+    """Set the interest picker on the given feed.
 
-def apply_interest_picker(response: CuratedRecommendationsResponse) -> None:
-    """Set the interest picker on the given response.
-
-    This method processes the sections in the response's feeds to determine which sections
+    This method processes the sections in the feed to determine which sections
     should be initially visible and which should be hidden. It sets `isInitiallyVisible` to True
-    for the first `MIN_INITIALLY_VISIBLE_SECTION_COUNT` sections and for follows sections.
+    for the first `MIN_INITIALLY_VISIBLE_SECTION_COUNT` sections and for followed sections.
     It then checks if there are at least MIN_INTEREST_PICKER_COUNT hidden sections available.
     If there are not enough hidden sections, all sections are made visible and no interest
     picker is created. Otherwise, the hidden sections are added to the interestPicker.sections.
 
     Args:
-        response (CuratedRecommendationsResponse): The API response containing sections.
-    """
-    # If there are no feeds in the response, nothing to do.
-    if not response.feeds:
-        return
+        feed (CuratedRecommendationsFeed): The feed containing sections.
 
+    Returns:
+        tuple[CuratedRecommendationsFeed, InterestPicker | None]:
+            A tuple where the first element is the updated feed and the second element is the
+            constructed InterestPicker. If there are not enough hidden sections, the InterestPicker
+            is not created and None is returned as the second element.
+    """
     # Get all available sections sorted by the order in which they should be shown.
-    sections = sorted(response.feeds.get_sections(), key=lambda tup: tup[0].receivedFeedRank)
+    sections = sorted(feed.get_sections(), key=lambda tup: tup[0].receivedFeedRank)
 
     # The first MIN_INITIALLY_VISIBLE_SECTION_COUNT and followed sections must be visible.
     visible_count = 0
@@ -53,7 +56,7 @@ def apply_interest_picker(response: CuratedRecommendationsResponse) -> None:
     if len(picker_sections) < MIN_INTEREST_PICKER_COUNT:
         for section, _ in sections:
             section.isInitiallyVisible = True
-        return
+        return feed, None
 
     # Randomize the positioning of the interest picker within a range. If any section is followed,
     # then this range shifts down by 1. Note that the rank is 0-indexed, so rank 1 is the 2nd row.
@@ -65,15 +68,15 @@ def apply_interest_picker(response: CuratedRecommendationsResponse) -> None:
     # Renumber the receivedFeedRank of all sections such that they increment by 1,
     # and that the section following the picker has a rank equal to interest_picker_rank + 1.
     new_rank = 0
-    for section, section_id in sections:
+    for section, _ in sections:
         if new_rank == interest_picker_rank:
             # Skip the rank for the interest picker.
             new_rank += 1
         section.receivedFeedRank = new_rank
         new_rank += 1
 
-    # Create and attach the interestPicker to the response.
-    response.interestPicker = InterestPicker(
+    # Create and attach the interestPicker.
+    interest_picker = InterestPicker(
         receivedFeedRank=interest_picker_rank,
         title="Follow topics to personalize your feed",
         subtitle=(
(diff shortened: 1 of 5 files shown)


merino/curated_recommendations/interest_picker.py
```diff
@@ -0,0 +1,86 @@
...
+    InterestPickerSection,
+    InterestPicker,
+)
+
+
+MIN_INITIALLY_VISIBLE_SECTION_COUNT = 3  # Minimum number of sections that are shown by default.
+MIN_INTEREST_PICKER_COUNT = 8  # Minimum number of items in the interest picker.
+
+
+def apply_interest_picker(response: CuratedRecommendationsResponse) -> None:
```
jpetto, line 16: what do you think about passing in the result of `response.feeds.get_sections()` here instead of the entire response?

that would allow skipping the `if not response.feeds:` check, and set this function up to return a value instead of modifying the `response` in place. this would also make testing a bit easier, as you can test the output of different (simpler) inputs.

merino/curated_recommendations/interest_picker.py
```diff
@@ -0,0 +1,86 @@
...
+
+    # Renumber the receivedFeedRank of all sections such that they increment by 1,
+    # and that the section following the picker has a rank equal to interest_picker_rank + 1.
+    new_rank = 0
+    for section, section_id in sections:
+        if new_rank == interest_picker_rank:
+            # Skip the rank for the interest picker.
+            new_rank += 1
+        section.receivedFeedRank = new_rank
+        new_rank += 1
```
jpetto, line 73: i'm unclear on the purpose of this `for` block. i don't see `sections` used anywhere below. is there some by reference variable updating happening?

maybe this would be more clear if this function returned a value?

tests/integration/api/v1/curated_recommendations/test_curated_recommendations.py
```diff
@@ -1463,6 +1464,30 @@ async def test_sections_feed_titles(self, locale, expected_titles):
...
+                    "feeds": ["sections"],
+                    "enableInterestPicker": enable_interest_picker,
+                },
+            )
+            data = response.json()
+
+            if enable_interest_picker:
+                assert data["interestPicker"] is not None
+            else:
+                assert data.get("interestPicker") is None
```
jpetto, line 1489: just curious - why is one assertion using `.get` and the other isn't?

merino/curated_recommendations/interest_picker.py
mmiermans: Yeah, I don't love passing in a complex object and modifying it in place.

The interest picker logic determines `response.sections[].isInitiallyVisible` and `response.interestPicker`. This function could accept the result of `response.feeds.get_sections()` and return an `InterestPicker`, but it would still need to modify the sections in place. Do you think that's preferable?

I could also extract 2 or 3 small functions to make `apply_interest_picker` more readable. Would that help?

merino/curated_recommendations/interest_picker.py
mmiermans: `section.receivedFeedRank = new_rank` updates the rank of the sections, because the interest picker has been inserted between them. I can change it to return a `CuratedRecommendationsFeed` and `InterestPicker`.

tests/integration/api/v1/curated_recommendations/test_curated_recommendations.py
mmiermans: Good catch! This property is always present, so `get` is not necessary.

merino/curated_recommendations/interest_picker.py
```diff
@@ -0,0 +1,92 @@
...
+        (section, section_id)
+        for (section, section_id) in sections
+        if not section.isInitiallyVisible
+    ]
+
+    # If there are not enough hidden sections, then show all sections without an interest picker.
+    if len(picker_sections) < MIN_INTEREST_PICKER_COUNT:
+        for section, _ in sections:
+            section.isInitiallyVisible = True
+        return feed, None
```
jpetto, line 60: as there is quite a bit of logic in this one function, my advice here is to break this out into multiple functions that have small responsibilities. something like:

- set `isInitiallyVisible` on sections (basically lines 41-47 and 57-60). accepts `sections` and returns `sections`. doesn't do anything with interest picker.
- create an `InterestPicker`. accepts `sections` and returns `InterestPicker` or `None`.
- update `receivedFeedRank` for `sections`. accepts `InterestPicker` and returns `sections`. if `InterestPicker` is `None`, just returns `sections` unmodified.

an approach like this encapsulates functionality into smaller, dedicated functions that are more easily unit tested.

jpetto: if this really has to be out today, you could consider shipping as-is and refactoring immediately after.

merino/curated_recommendations/interest_picker.py
```diff
@@ -0,0 +1,107 @@
...
+    visible_count = 0
+    for section, _ in sections:
+        if section.isFollowed or visible_count < MIN_INITIALLY_VISIBLE_SECTION_COUNT:
+            section.isInitiallyVisible = True
+            visible_count += 1
+        else:
+            section.isInitiallyVisible = False
+
+
+def _get_interest_picker_rank(sections: list[tuple]) -> int:
```
jpetto, line 86: nice! i think it would be great to have unit tests on this.

merino/curated_recommendations/interest_picker.py
```diff
@@ -0,0 +1,107 @@
...
+    Marks the first MIN_INITIALLY_VISIBLE_SECTION_COUNT and followed sections as visible;
+    others are hidden.
+    """
+    visible_count = 0
+    for section, _ in sections:
+        if section.isFollowed or visible_count < MIN_INITIALLY_VISIBLE_SECTION_COUNT:
+            section.isInitiallyVisible = True
+            visible_count += 1
+        else:
+            section.isInitiallyVisible = False
```
jpetto, line 83: i think this logic belongs here:

```
if len(picker_sections) < MIN_INTEREST_PICKER_COUNT:
        for section, _ in sections:
            section.isInitiallyVisible = True
```

but instead of checking the length of `picker_sections`, you check the number of sections that have `isInitiallyVisible = false`.

also - why can't this return `sections`? i'm confused why `sections` is always being modified by reference...

jpetto: after this function completes, you can use the result to determine if you need to create an `InterestPicker` or not.

merino/curated_recommendations/interest_picker.py
```diff
@@ -0,0 +1,107 @@
...
+    """Return a randomized rank for the interest picker.
+
+    If any section is followed, choose a random int in [2, 4]; otherwise, in [1, 3].
+    """
+    if any(section.isFollowed for section, _ in sections):
+        return random.randint(2, 4)
+    return random.randint(1, 3)
+
+
+def _renumber_sections(sections: list[tuple], picker_rank: int) -> None:
```
jpetto, line 96: can this return a value instead of modifying `sections` by reference?

merino/curated_recommendations/provider.py
```diff
@@ -468,6 +469,10 @@ async def fetch(
                 curated_recommendations_request.sections, response.feeds
             )
 
+        if curated_recommendations_request.enableInterestPicker and response.feeds:
+            interest_picker = create_interest_picker(response.feeds.get_sections())
+            response.interestPicker = interest_picker
```
jpetto, line 474: this is probably a framework/app assumption i'm not familiar with, but just to check - it's okay to set `response.interestPicker = None`?

jpetto (approved):
one small question, but otherwise this looks good! thank you for breaking out the functions and the really good unit tests!

merino/curated_recommendations/provider.py
mmiermans: Yeah, that's fine. The [default value](https://github.com/mozilla-services/merino-py/blob/f39ee18850c128bc148c7e826611abfb6845ccf7/merino/curated_recommendations/protocol.py#L311) is None.

PR #791, opened 2025-02-12
[HNT-405] Remove blocked topics from sections

mmiermans:
## References

JIRA: [HNT-405](https://mozilla-hub.atlassian.net/browse/HNT-405)

## Description
If a user blocks a section using the Follow/Block feature (e.g. Food, Politics), we remove recommendations with those topics from sections, including 'Popular Today'.

## PR Review Checklist

[HNT-405]: https://mozilla-hub.atlassian.net/browse/HNT-405?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/merino/curated_recommendations/provider.py
+++ b/merino/curated_recommendations/provider.py
@@ -312,7 +312,9 @@ class CuratedRecommendationsProvider:
         return general_feed, need_to_know_feed
 
     @staticmethod
-    def exclude_recommendations_from_blocked_sections(recommendations, requested_sections):
+    def exclude_recommendations_from_blocked_sections(
+        recommendations, requested_sections
+    ) -> list[CuratedRecommendation]:
         """Return recommendations which topic doesn't match any blocked section"""
         blocked_section_ids = {rs.sectionId for rs in requested_sections if rs.isBlocked}
         return [
--- a/merino/curated_recommendations/provider.py
+++ b/merino/curated_recommendations/provider.py
@@ -311,6 +311,14 @@ class CuratedRecommendationsProvider:
         general_feed = [r for r in recommendations if not r.isTimeSensitive]
         return general_feed, need_to_know_feed
 
+    @staticmethod
+    def exclude_recommendations_from_blocked_sections(recommendations, requested_sections):
+        """Return recommendations which topic doesn't match any blocked section"""
+        blocked_section_ids = {rs.sectionId for rs in requested_sections if rs.isBlocked}
+        return [
+            r for r in recommendations if not r.topic or r.topic.value not in blocked_section_ids
+        ]
+
     async def get_sections(
         self,
         recommendations: list[CuratedRecommendation],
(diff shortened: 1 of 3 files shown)


merino/curated_recommendations/provider.py
```diff
@@ -323,6 +323,13 @@ async def get_sections(
         min_fallback_recs_per_section = 1
         top_stories_count = 6
 
+        # Recommendations whose topic matches a blocked section should not be shown.
+        if request.sections:
+            blocked_section_ids = {rs.sectionId for rs in request.sections if rs.isBlocked}
+            recommendations = [
+                r for r in recommendations if r.topic and r.topic.value not in blocked_section_ids
+            ]
```
jpetto, line 331: i think it would be nice to abstract this to a helper function with a few unit tests.

- verify it's only removing recommendations matching a blocked section
- verify it does nothing if no sections are marked as blocked
- verify it handles recommendations without topics as expected (is the expectation that a recommendation without a topic be excluded here? also how do recommendations without a topic get in?)

jpetto: also - do you think it's worth skipping the list comprehension (i think that's what it's called?) on line 329 if `blocked_section_ids` has a length of 0?

tests/integration/api/v1/curated_recommendations/test_curated_recommendations.py
```diff
@@ -1417,6 +1417,34 @@ async def test_curated_recommendations_with_sections_feed_boost_followed_section
...
+                },
+            )
+            data = response.json()
+
+            # assert that none of the recommendations has a blocked topic.
+            for _, feed in data["feeds"].items():
+                if feed:
+                    for recommendation in feed["recommendations"]:
+                        assert recommendation["topic"] not in blocked_topics
+
```
jpetto, line 1447: just to say - if the new logic in the above file is fully tested in a unit test, the only line of code this function would really be testing (that isn't already tested) is line 327 in the file above:

`if request.sections:`

for peace of mind this test isn't bad though.

jpetto (changes requested):
i think it's worth it to cover more scenarios in some unit tests.

merino/curated_recommendations/provider.py
mmiermans: Good question. I was curious what the performance impact was, so I did a quick benchmark:
```python
import timeit
setup_code = '''
class DummyTopic:
    def __init__(self, value):
        self.value = value
class DummyRec:
    def __init__(self, topic):
        self.topic = topic
# Create 10,000 dummy recommendations.
# For this benchmark, each recommendation has a DummyTopic.
recommendations = [DummyRec(DummyTopic("dummy")) for _ in range(1000)]
# blocked_section_ids is empty.
blocked_section_ids = set()
'''
stmt = '''
filtered = [r for r in recommendations if not r.topic or r.topic.value not in blocked_section_ids]
'''
# Run the list comprehension 1000 times and measure the total time.
iterations = 1000
total_time = timeit.timeit(stmt=stmt, setup=setup_code, number=iterations)
avg_time_ms = (total_time / iterations) * 1000
print(f"Average time per iteration: {avg_time_ms:.3f} ms")
```
> Average time per iteration: 0.028 ms

Currently we have a few hundred recommendations. At the cost of a few microseconds, I would lean towards letting list comprehension handle the case of an empty list.

PR #792, opened 2025-02-13
ideas on breaking up functionality

jpetto:
ideas!


mmiermans:
Thanks for the ideas! The code is more readable because of them.

PR #800, opened 2025-02-19
[DISCO-3247] fix: weather pathfinder to use region when None

tiftran:
## References

JIRA: [DISCO-3247](https://mozilla-hub.atlassian.net/browse/DISCO-3247)

## Description
We tried to be fancy and it sort of bit us. We can't iterate over a map twice, so I swapped the two for loops. so we can keep map. In theory there's only a small number of edge cases where the first iteration is not successful and we would have more iterations then if the for loop order was swapped

## PR Review Checklist

[DISCO-3247]: https://mozilla-hub.atlassian.net/browse/DISCO-3247?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/merino/providers/suggest/weather/backends/accuweather/pathfinder.py
+++ b/merino/providers/suggest/weather/backends/accuweather/pathfinder.py
@@ -155,21 +155,19 @@ async def explore(
     for city in cities:
         if city in explored_cities:
             continue
-        else:
-            explored_cities.append(city)
-            for region in compass(weather_context.geolocation):
-                if country and city and (country, region, city) in SKIP_CITIES_MAPPING:
-                    # increment since we tried to look up this combo again.
-                    increment_skip_cities_mapping(country, region, city)
+        explored_cities.append(city)
+        for region in compass(weather_context.geolocation):
+            if country and city and (country, region, city) in SKIP_CITIES_MAPPING:
+                # increment since we tried to look up this combo again.
+                increment_skip_cities_mapping(country, region, city)
+                return None, True
 
-                    return None, True
+            weather_context.selected_region = region
+            weather_context.selected_city = city
+            res = await probe(weather_context)
 
-                weather_context.selected_region = region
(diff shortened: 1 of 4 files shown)


tests/unit/providers/suggest/weather/backends/test_pathfinder.py
```diff
@@ -66,109 +66,31 @@ def test_compass(location: Location, expected_region_and_city: str) -> None:
 
 
 @pytest.mark.parametrize(
-    ("weather_context", "expected_calls"),
```
tiftran, line 69: I removed the actual list of weather contexts, as it only returns the state of the weather context after all the calls and it's actually not useful

merino/providers/suggest/weather/backends/accuweather/pathfinder.py
```diff
@@ -150,26 +150,26 @@ async def explore(
...
+                    increment_skip_cities_mapping(country, region, city)
+
+                    return None, True
+
+                weather_context.selected_region = region
+                weather_context.selected_city = city
+                res = await probe(weather_context)
+
+                if res is not None:
+                    return res, is_skipped
```
ncloudioj, line 172:
```suggestion
          explored_cities.append(city)
          for region in compass(weather_context.geolocation):
              if country and city and (country, region, city) in SKIP_CITIES_MAPPING:
                  # increment since we tried to look up this combo again.
                  increment_skip_cities_mapping(country, region, city)
                  return None, True
      
              weather_context.selected_region = region
              weather_context.selected_city = city
              res = await probe(weather_context)      
              if res is not None:
                  return res, is_skipped
```

ncloudioj (approved):
Good catch.

ncloudioj (approved):

PR #804, opened 2025-02-21
[DISCO-3228] Add metrics and logging for curated recommendations favicons

Herraj:
## References

JIRA: [DISCO-3228](https://mozilla-hub.atlassian.net/browse/DISCO-3228)

## Description
Add metrics and logging for curated recommendations favicons hits and misses. This will allow us to better understand the success percentage and be able to add the missing icon urls manually for the urls that had a miss.

## PR Review Checklist

[DISCO-3228]: https://mozilla-hub.atlassian.net/browse/DISCO-3228?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/merino/curated_recommendations/corpus_backends/corpus_api_backend.py
+++ b/merino/curated_recommendations/corpus_backends/corpus_api_backend.py
@@ -7,6 +7,7 @@ from datetime import datetime, timedelta
 from urllib.parse import urlparse, urlencode, parse_qsl
 from zoneinfo import ZoneInfo, ZoneInfoNotFoundError
 from dataclasses import dataclass
+from pydantic import HttpUrl
 
 import aiodogstatsd
 from httpx import AsyncClient, HTTPError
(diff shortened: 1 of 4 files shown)


merino/curated_recommendations/corpus_backends/corpus_api_backend.py
```diff
@@ -330,6 +334,14 @@ async def _fetch_from_backend(
...
             if icon_url := self.manifest_provider.get_icon_url(item["corpusItem"]["url"]):
                 item["corpusItem"]["iconUrl"] = icon_url
+                self.metrics_client.increment("corpus_item.manifest_provider.icon_url.hit")
+            else:
+                corpus_items_no_icons.append(item["corpusItem"]["url"])
+                self.metrics_client.increment("corpus_item.manifest_provider.icon_url.miss")
+
+        logger.info(
+            f"Curated Recommendations with missing icon urls: {len(corpus_items_no_icons)}, {corpus_items_no_icons}"
+        )
```
Herraj, line 344: 1- Should we do a separate log for each missed item or logging an array / list is fine?

2- Naming of the metric. This function already emits metrics but those are prefixed `corpus_api...`. Since this isn't related with the corpus api itself, thought I'd go with this.

gruberb: I think logging the list and the naming is fine. The only thing I am thinking of - Will we emit the same metrics over and over again? AFAIK, the cache is by date? So new feeds/items could be from the same URL, and therefore we would list the URLs again. But I guess this doesn't matter. And for the count: We would always count the same URLs, which I guess is also fine, because we want to know each day how many items we cannot serve.

gruberb (approved):

PR #813, opened 2025-03-04
[DISCO-3294] chore: update more cities for weather

tiftran:
## References

JIRA: [DISCO-3294](https://mozilla-hub.atlassian.net/browse/DISCO-3294)

## Description
Because we're cycling through city names, normalized city name, city name without locality suffix times # of regions for that city, we're making a lot of requests we don't need to. I've split `REGION_MAPPING_EXCLUSIONS` into two sets instead and added the countries that we know. In addition added the cities in those countries that require the city to be None.

## PR Review Checklist

[DISCO-3294]: https://mozilla-hub.atlassian.net/browse/DISCO-3294?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/merino/providers/suggest/weather/backends/accuweather/pathfinder.py
+++ b/merino/providers/suggest/weather/backends/accuweather/pathfinder.py
@@ -96,9 +96,11 @@ SKIP_CITIES_MAPPING: dict[tuple[str, str | None, str], int] = {
     ("US", "UT", "Hill Air Force Base"): 0,
 }
 
+# Countries that use the most specific region to retrieve weather
 KNOWN_SPECIFIC_REGION_COUNTRIES: frozenset = frozenset(
     ["AR", "AU", "BR", "CA", "CN", "DE", "FR", "GB", "IN", "MX", "NZ", "PL", "PT", "RU", "US"]
 )
+# Countries that use the least specific region to retrieve weather
 KNOWN_REGION_COUNTRIES: frozenset = frozenset(["IT", "ES", "GR"])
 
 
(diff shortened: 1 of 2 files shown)


tests/unit/providers/suggest/weather/backends/test_pathfinder.py
```diff
@@ -32,7 +32,7 @@
             "77",
         ),
         (
-            Location(country="AR", regions=["B", "5"], city="La Plata"),
+            Location(country="AA", regions=["B", "5"], city="La Plata"),
```
tiftran, line 35: Changed this to a fake country code, since as we add more countries to our lists, this test will fail.

merino/providers/suggest/weather/backends/accuweather/pathfinder.py
```diff
@@ -89,6 +96,11 @@
     ("US", "UT", "Hill Air Force Base"): 0,
 }
 
+KNOWN_SPECIFIC_REGION_COUNTRIES: frozenset = frozenset(
```
ncloudioj, line 99: We're getting constants defined here, let's add some descriptions to them for better readability.

tests/unit/providers/suggest/weather/backends/test_pathfinder.py
ncloudioj: This can be a comment for future reference.

ncloudioj (approved):

PR #846, opened 2025-03-19
DISCO-3360 Parallelize favicon processing for the nav-suggestions job

gruberb:
## References

JIRA: [DISCO-3360](https://mozilla-hub.atlassian.net/browse/DISCO-3360)

## PR Review Checklist

[DISCO-3360]: https://mozilla-hub.atlassian.net/browse/DISCO-3360?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/merino/jobs/navigational_suggestions/domain_metadata_extractor.py
+++ b/merino/jobs/navigational_suggestions/domain_metadata_extractor.py
@@ -1,5 +1,7 @@
 """Extract domain metadata from domain data"""
 
+import asyncio
+import itertools
 import logging
 from typing import Any, Optional
 from urllib.parse import urljoin, urlparse
@@ -7,13 +9,13 @@ from urllib.parse import urljoin, urlparse
 import requests
 from pydantic import BaseModel
 from robobrowser import RoboBrowser
+from typing import cast
 
 from merino.utils.gcs.models import Image
 from merino.jobs.navigational_suggestions.utils import (
     REQUEST_HEADERS,
     TIMEOUT,
-    FaviconDownloader,
-    requests_get,
+    AsyncFaviconDownloader,
 )
 
 logger = logging.getLogger(__name__)
@@ -40,11 +42,13 @@ class Scraper:
     MANIFEST_SELECTOR: str = 'link[rel="manifest"]'
 
     browser: RoboBrowser
+    request_client: AsyncFaviconDownloader
 
     def __init__(self) -> None:
         session: requests.Session = requests.Session()
         session.headers.update(REQUEST_HEADERS)
         self.browser = RoboBrowser(session=session, parser="html.parser", allow_redirects=True)
+        self.request_client = AsyncFaviconDownloader()
 
     def open(self, url: str) -> Optional[str]:
         """Open the given url for scraping.
@@ -78,8 +82,8 @@ class Scraper:
             ],
         )
 
-    def scrape_favicons_from_manifest(self, manifest_url: str) -> list[dict[str, Any]]:
-        """Scrape favicons from manifest of an already opened URL.
+    async def scrape_favicons_from_manifest(self, manifest_url: str) -> list[dict[str, Any]]:
+        """Scrape favicons from manifest of an already opened URL asynchronously.
 
         Args:
             manifest_url: URL of the manifest file
@@ -88,15 +92,19 @@ class Scraper:
         """
         result = []
         try:
-            response: Optional[requests.Response] = requests_get(manifest_url)
+            response = await self.request_client.requests_get(manifest_url)
             if response:
-                result = response.json().get("icons")
+                try:
+                    json_data = await response.json()
+                    result = json_data.get("icons", [])
+                except AttributeError as e:
+                    logger.info(f"Exception: {e} while parsing icons from manifest {manifest_url}")
         except Exception as e:
             logger.info(f"Exception: {e} while parsing icons from manifest {manifest_url}")
         return result
 
-    def get_default_favicon(self, url: str) -> Optional[str]:
-        """Return the default favicon for the given url.
+    async def get_default_favicon(self, url: str) -> Optional[str]:
+        """Return the default favicon for the given url asynchronously.
 
         Args:
             url: URL to scrape for favicon at default location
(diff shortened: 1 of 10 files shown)


merino/jobs/navigational_suggestions/utils.py
```diff
@@ -26,43 +28,74 @@
...
 
-class FaviconDownloader:
-    """Download favicon from the web"""
+class AsyncFaviconDownloader:
+    """Download favicon from the web asynchronously"""
 
-    def download_favicon(self, url: str) -> Image | None:
-        """Download the favicon from the given url.
+    def __init__(self) -> None:
+        self.session: Optional[aiohttp.ClientSession] = None
```
misaniwere, line 35: we seem to already be using `AsyncClient` from `httpx` across the application. I'm wondering if it's possible to use that here as well to maintain convention?

gruberb: Ah, good point! Thank you! Switched it out for `httpx` 👍

merino/jobs/navigational_suggestions/domain_metadata_extractor.py
```diff
@@ -316,37 +334,49 @@ def _is_domain_blocked(self, domain: str, suffix: str) -> bool:
...
+    async def _process_domains_async(
+        self, domains_data: list[dict[str, Any]], favicon_min_width: int
+    ) -> list[dict[str, Optional[str]]]:
+        """Process all domains concurrently"""
+        tasks = [
+            self._process_single_domain(domain_data, favicon_min_width)
+            for domain_data in domains_data
+        ]
+        logger.info(f"Processing {len(domains_data)} domains concurrently")
+        return await asyncio.gather(*tasks)
```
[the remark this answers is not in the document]
ncloudioj: Another question: looks like all the `asycnio.gather()` calls in this patch are not explicitly handling exceptions raised by the tasks, meaning that if anyone throws, `gather()` will immediately return without waiting for other tasks. Is that desirable everywhere? FWIW, there is an option `return_exceptions = True` which can be used to change that behavior.

merino/jobs/navigational_suggestions/utils.py
```diff
@@ -26,43 +30,76 @@
...
-    """Download favicon from the web"""
+class AsyncFaviconDownloader:
+    """Download favicon from the web asynchronously"""
 
-    def download_favicon(self, url: str) -> Image | None:
-        """Download the favicon from the given url.
+    def __init__(self) -> None:
+        self.session: Optional[AsyncClient] = None
+
+    async def init_session(self) -> None:
```
ncloudioj, line 39: Since the `session` is essentially an HTTP client, which can be re-used for different download tasks instead of creating a new one for each download task as it's known to be expensive, I'd suggest dropping `init_session` and `close_session` by just holding an HTTP client permanently for each `AsyncFaviconDownloader` instance. This client will be dropped automatically when the Merino job exits.

merino/jobs/navigational_suggestions/domain_metadata_extractor.py
```diff
@@ -215,18 +224,55 @@ def _extract_favicons(self, scraped_url: str) -> list[dict[str, Any]]:
...
+                    chunk_urls = manifest_urls[i : i + chunk_size]
+
+                    scraped_favicons_list = await asyncio.gather(*chunk, return_exceptions=True)
+
+                    # Filter out exceptions from results with explicit cast
+                    filtered_scraped_favicons_list: list[list[dict[str, Any]]] = []
+                    for result in scraped_favicons_list:
+                        if isinstance(result, Exception):
+                            # Log the specific exception that occurred during manifest processing
+                            logger.info(f"Exception during manifest processing: {result}")
```
ncloudioj, line 253:
```suggestion
                            logger.wraning(f"Exception during manifest processing: {result}")
```

merino/jobs/navigational_suggestions/domain_metadata_extractor.py
```diff
@@ -215,18 +224,55 @@ def _extract_favicons(self, scraped_url: str) -> list[dict[str, Any]]:
...
+                            filtered_scraped_favicons_list.append([])
+                        else:
+                            from typing import cast
```
ncloudioj, line 256: Let's move this import to the module header.

merino/jobs/navigational_suggestions/domain_metadata_extractor.py
```diff
@@ -236,56 +282,63 @@ def _extract_favicons(self, scraped_url: str) -> list[dict[str, Any]]:
...
-                else:
-                    logger.info(f"Masked SVG favicon {favicon} found; skipping it")
+
+        try:
+            # Process favicons in chunks to limit concurrent connections
+            chunk_size = 20
+            all_favicon_images = []
+
+            for i in range(0, len(urls), chunk_size):
+                chunk_urls = urls[i : i + chunk_size]
```
ncloudioj, line 299: FYI: `itertools` comes with a function `batched` for this if you like. 
```suggestion
            for chunk_urls in itertools.batched(urls, chunk_size):
```

merino/jobs/navigational_suggestions/domain_metadata_extractor.py
```diff
@@ -236,56 +282,63 @@ def _extract_favicons(self, scraped_url: str) -> list[dict[str, Any]]:
...
+                if image is None or "image/" not in image.content_type:
+                    continue
+
+                try:
+                    width = self._get_favicon_smallest_dimension(image)
+                    if width > best_favicon_width:
+                        best_favicon_url = urls[i]
+                        best_favicon_width = width
+                except Exception as e:
+                    logger.info(f"Exception {e} for favicon {favicon}")
```
ncloudioj, line 325:
```suggestion
                    logger.warning(f"Exception {e} for favicon {favicon}")
```

merino/jobs/navigational_suggestions/domain_metadata_extractor.py
```diff
@@ -236,56 +282,63 @@ def _extract_favicons(self, scraped_url: str) -> list[dict[str, Any]]:
...
+        finally:
```
ncloudioj, line 326: This can be omitted if there is nothing to be done in "finally". Also, looks like the outer try block doesn't have the except block, if that's intended we can simply remove the outer try...except... as a whole.

ncloudioj (approved):
LGTM. r=wc.

merino/jobs/navigational_suggestions/domain_metadata_extractor.py
gruberb: I think this was a left over from all the refactoring the past few days! Thanks for catching it!

PR #851, opened 2025-03-24
(feat)HNT-466: boost followed sections based on followedAt

katerinachinnappan:
## References

JIRA: https://mozilla-hub.atlassian.net/browse/HNT-466

## Description
* Added new `followedAt` field to sections. This field is to be sent by the client in `UTC`.
* Adjusted `boost_followed_sections` to used `followedAt` date & decide which sections get a higher boost:
    * Created new `is_section_recently_followed` helper to check if a section is "recently followed" (followed within a week)
    * If a section is recently followed, it gets a higher boost, followed by the rest of the followed sections. Relative order is preserved.
    * Nothing changes for unfollowed / blocked sections.
* Updated tests to check boosting based on `followedAt`.

## PR Review Checklist


merino/curated_recommendations/rankers.py
```diff
@@ -149,12 +150,35 @@ def boost_preferred_topic(
...
+    # Return followed_at in UTC timezone (in case client sent in different timezone)
+    followed_at = followed_at.astimezone(timezone.utc)
+
+    # Get current UTC time with dropping the microseconds
+    current_time = datetime.now(timezone.utc).replace(microsecond=0)
+    # drop microseconds from followed_at for accurate comparison
+    followed_at = followed_at.replace(microsecond=0)
+
+    # Return true if recently followed (<=7) false if > 7
+    return current_time - followed_at <= timedelta(days=7)
```
mmiermans, line 172: Why does dropping microseconds make the comparison more accurate?
```suggestion
    return datetime.now(timezone.utc) - followed_at <= timedelta(days=7)
```

merino/curated_recommendations/protocol.py
```diff
@@ -84,6 +85,7 @@ class SectionConfiguration(BaseModel):
     sectionId: str
     isFollowed: bool
     isBlocked: bool
+    followedAt: datetime | None = None
```
mmiermans, line 88: I'm curious how FastAPI documents this type in [localhost:8000/docs](http://localhost:8000/docs). The docs should make clear what format and timezone `followedAt` is expected to be. Do we require UTC time? Or accept any timezone? Consider adding a description:
```suggestion
    followedAt: datetime | None = Field(
        default=None,
        description="Timestamp when the section was followed. Must be in ISO 8601 format with timezone, e.g. '2024-03-24T12:34:56Z'."
    )
```
(Untested code.)

merino/curated_recommendations/protocol.py
```diff
@@ -239,6 +241,7 @@ class Section(BaseModel):
     layout: Layout
     isFollowed: bool = False
     isBlocked: bool = False
+    followedAt: datetime | None = None
```
mmiermans, line 244: Same comment as for the request `followedAt` parameter.

merino/curated_recommendations/rankers.py
```diff
@@ -199,18 +226,38 @@ def boost_followed_sections(
...
+        for section in recently_followed_sections:
+            section.receivedFeedRank = current_received_feed_rank
+            current_received_feed_rank += 1
+
+        for section in other_followed_sections:
             section.receivedFeedRank = current_received_feed_rank
             current_received_feed_rank += 1
 
-        # 8. Assign new rank (starting from last rank value assigned to a followed section)
+        # 10. Assign new rank (starting from last rank value assigned to a followed section)
```
mmiermans, line 260: The logic in this function feels more complicated than it needs to be. I asked ChatGPT if this code could be simplified by using a single sort function, and this is what it came up with:
```python
def boost_followed_sections(
    req_sections: list[SectionConfiguration], feeds: CuratedRecommendationsFeed
) -> CuratedRecommendationsFeed:
    """Boost followed sections to the very top, right after top_stories_section.
    Received feed rank for top_stories_section should always stay 0.
    Received feed rank for followed_sections should follow top_stories_section.
    Most recently followed sections (followed within 1 week) should be boosted higher.
    Unfollowed sections should be ranked after followed_sections, and relative order should be preserved.

    :param req_sections: List of Section configurations
    :param feeds: CuratedRecommendationsFeed object
    :return: updated CuratedRecommendationsFeed with boosted followed sections (if found)
    """
    followed_sections_info = {
        section.sectionId: section.followedAt for section in req_sections if section.isFollowed
    }

    blocked_section_ids = [section.sectionId for section in req_sections if section.isBlocked]

    # 3. Update isBlocked based on blocked_section_ids
    # For now, we will only update isBlocked value on a Section
    # The client-side will handle the actual blocking action
    if blocked_section_ids:
        for section_id in blocked_section_ids:
            section = feeds.get_section_by_topic_id(section_id)
            if section:
                section.isBlocked = True

    # 4. Update isFollowed & followedAt based on followed_sections_info
    if followed_sections_info:
        for section_id, followed_at in followed_sections_info.items():
            section = feeds.get_section_by_topic_id(section_id)
            if section:
                section.isFollowed = True
                section.followedAt = followed_at

        # 5. Collect followed & unfollowed sections
        all_sections = []
        for section_id in feeds.model_fields_set:
            section = getattr(feeds, section_id)
            if section:
                if section_id == "top_stories_section":
                    # top_stories_section is always on the top
                    section.receivedFeedRank = 0
                    continue
                all_sections.append(section)

        # 6–10. Sort sections using a composite key:
        #  1. isFollowed (followed first)
        #  2. is_section_recently_followed (recently followed first)
        #  3. followedAt (most recent first)
        #  4. original rank for unfollowed (to preserve order)
        all_sections.sort(
            key=lambda section: (
                0 if section.isFollowed else 1,  # 1st sort order
                0 if is_section_recently_followed(section.followedAt) else 1,  # 2nd sort order
                -(section.followedAt.timestamp() if section.followedAt else 0),  # 3rd sort order
                section.receivedFeedRank or 0,  # 4th sort order
            )
        )

        current_received_feed_rank = 1
        for section in all_sections:
            section.receivedFeedRank = current_received_feed_rank
            current_received_feed_rank += 1

    return feeds
```
What do you think?

You could also consider splitting off some of these code blocks as functions.

tests/integration/api/v1/curated_recommendations/test_curated_recommendations.py
```diff
@@ -1436,8 +1438,18 @@ async def test_curated_recommendations_with_sections_feed_boost_followed_section
                     "locale": "en-US",
                     "feeds": ["sections"],
                     "sections": [
-                        {"sectionId": "sports", "isFollowed": True, "isBlocked": False},
-                        {"sectionId": "arts", "isFollowed": True, "isBlocked": False},
+                        {
+                            "sectionId": "sports",
+                            "isFollowed": True,
+                            "isBlocked": False,
+                            "followedAt": "2025-03-17T12" ":00:00Z",
```
mmiermans, line 1445:
```suggestion
                            "followedAt": "2025-03-17T12:00:00Z",
```

tests/integration/api/v1/curated_recommendations/test_curated_recommendations.py
```diff
@@ -1436,8 +1438,18 @@ async def test_curated_recommendations_with_sections_feed_boost_followed_section
...
+                        },
+                        {
+                            "sectionId": "arts",
+                            "isFollowed": True,
+                            "isBlocked": False,
+                            "followedAt": "2025-03-10T12:00" ":00Z",
```
mmiermans, line 1451:
```suggestion
                            "followedAt": "2025-03-10T12:00:00Z",
```

tests/unit/curated_recommendations/test_rankers.py
```diff
@@ -356,6 +360,36 @@ def test_boost_preferred_topic_no_reorder(self):
         assert recs == not_reordered_recs
 
 
+class TestIsSectionRecentlyFollowed:
+    """Unit tests for is_section_recently_followed"""
+
+    @freezegun.freeze_time("2025-03-20 12:00:00", tick=True, tz_offset=0)
```
mmiermans, line 366: Why set `tick=True`? I think you want [tick](https://github.com/spulec/freezegun?tab=readme-ov-file#tick-argument) to be false for all these tests, which is the default. When set to true, the time keeps ticking during the test, which you sometimes need if there's a sleep call.
```suggestion
    @freezegun.freeze_time("2025-03-20 12:00:00", tz_offset=0)
```

tests/integration/api/v1/curated_recommendations/test_curated_recommendations.py
```diff
@@ -1447,18 +1459,26 @@ async def test_curated_recommendations_with_sections_feed_boost_followed_section
...
                 assert data["feeds"]["sports"]["receivedFeedRank"] in [1, 2]
+                assert data["feeds"]["sports"]["followedAt"]
+
+            # in the case both sections are present, sports is recently followed & needs to have higher rank
+            if data["feeds"]["arts"] is not None and data["feeds"]["sports"] is not None:
+                assert data["feeds"]["sports"]["receivedFeedRank"] == 1
+                assert data["feeds"]["arts"]["receivedFeedRank"] == 2
 
             # Assert no errors were logged
             errors = [r for r in caplog.records if r.levelname == "ERROR"]
```
mmiermans, line 1484: Could you add a test to cover what happens when the client sends invalid time formats?
- Missing timezone
- Not in ISO format
- Integer unix time

merino/curated_recommendations/rankers.py
katerinachinnappan: @mmiermans i added this because the tests weren't passing & now I realize this is because of `tick=true`. agreed that we should keep the microseconds. also i do like keeping `datetime.now(timezone.utc)` in `current_time` variable for readability.

merino/curated_recommendations/rankers.py
katerinachinnappan: the single sort using composite key makes sense, i will take a closer look

mmiermans (approved):
Looks great! 😻

PR #866, opened 2025-03-27
DISCO-3411 Save resources on parsing favicons in the nav-suggestion job

gruberb:
## References

JIRA: [DISCO-3411](https://mozilla-hub.atlassian.net/browse/DISCO-3411)

This PR re-writes how the Navigational Suggestion job works. We recently started adding more custom domains to the already existing domains we query each week inside BigQuery. The custom domains (https://github.com/mozilla-services/merino-py/blob/main/merino/jobs/navigational_suggestions/custom_domains.py) grew to over 1.5k entries.

This caused the job to error out on Airflow, most likely due to increased resource consumption. The old way was to query all domains for all possible favicons, download and compare the quality of these favicons for each side, and at the end of the run, upload them and create the `top-picks.json` file and upload it to a Google Cloud bucket.

We then introduced an async version of this job with this PR: https://github.com/mozilla-services/merino-py/pull/846 Which sped up the processing, but still used the same in-memory resources.

This PR tries to address these issues by:
- Don't query the domains for all possible favicons, but limit it to 5 (because we download and compare all of them)
- Upload the favicon right away to Google Cloud instead of keeping them in memory
- Delete all fetched favicons after uploading them to Google Cloud right away

Therefore, this PR replaces the methods where we get a list of favicons, and downloading all of them before uploading them to GCS, and uploading a good fit right away (https://github.com/mozilla-services/merino-py/pull/866/files#diff-3f65e67d8441adc349f59daf0c480b0b16b2826ffd35c0dd993a151326c34c94L77-L93).

Instead of `get_best_favicon`, we have now the method `_process_favicon` (https://github.com/mozilla-services/merino-py/pull/866/files#diff-d9ccfdf46db463d509aaf3ae5dee144bc11d765352784192d49d81e27f4ffeafL316) where we extract possible links of the fetched HTML of the queried domain, do the same quality and type comparison, and upload a match right away.

This PR doesn't change how we find favicons or compare them, and implements a more streamlined approach to storing and uploading them to save memory resources.

## PR Review Checklist

[DISCO-3411]: https://mozilla-hub.atlassian.net/browse/DISCO-3411?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/merino/jobs/navigational_suggestions/domain_metadata_extractor.py
+++ b/merino/jobs/navigational_suggestions/domain_metadata_extractor.py
@@ -9,8 +9,8 @@ from urllib.parse import urljoin, urlparse
 import requests
 from pydantic import BaseModel
 from robobrowser import RoboBrowser
-from typing import cast
 
+from merino.jobs.navigational_suggestions.domain_metadata_uploader import DomainMetadataUploader
 from merino.utils.gcs.models import Image
 from merino.jobs.navigational_suggestions.utils import (
     REQUEST_HEADERS,
@@ -65,11 +65,9 @@ class Scraper:
             logger.info(f"Exception: {e} while opening url {url}")
             return None
 
-    def scrape_favicon_data(self, url: str) -> FaviconData:
+    def scrape_favicon_data(self) -> FaviconData:
         """Scrape the favicon data for an already opened URL.
 
-        Args:
-            url: URL to scrape for favicon data
         Returns:
             FaviconData: Favicon data for a URL
         """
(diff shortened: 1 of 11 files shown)


merino/jobs/navigational_suggestions/domain_metadata_extractor.py
```diff
@@ -225,24 +223,35 @@ def _get_favicon_smallest_dimension(self, image: Image) -> int:
...
         try:
-            favicon_data: FaviconData = self.scraper.scrape_favicon_data(scraped_url)
+            # Get the most common favicon sources first
+            favicon_data: FaviconData = self.scraper.scrape_favicon_data()
 
-            for favicon in favicon_data.links:
+            # First priority: process standard link icons (usually higher quality)
+            for favicon in favicon_data.links[:max_icons]:
                 favicon_url = favicon["href"]
                 if self._is_problematic_favicon_url(favicon_url):
```
jrconlin, line 241: Really stupid question.
Are we doing size checks for icon sizes as well? 
I see the `data:` one is being handled for stuff like https://evilonastick.com/huge.html but there's no reason I couldn't just make that a `<link rel="icon" href="...">` pretty easily.

jrconlin (commented):
One thing I prefer to do is be a bit verbose in either the commit message or the bug (ideally, the commit message since that's what can be read by the larger community).

IIUC, this PR does a couple of things. 
1) It modifies the `navigational_suggestions` doc to note that we may pull from non-origin domain servers (like CDNs). If that's the case, we might want to specify that. We might also want to safe-guard against malicious injections. I can think of a number of things like `paypa1.com` using PayPal's favicon origin.)
2) A few method renames to better abstract the function description. (e.g. "metadata_extractor" vs. just "favicons") 
3) reduce redundant passed data
4) (the biggest one) preferentially select the "best" icons. 

Is this correct or am i missing something else?

gruberb:
Thanks for the review @jrconlin. I updated the PR description. Since we increased the amount of domains we are processing, it is no longer feasible to compare all possible favicons from a website (which can be up to 15 and more in some cases). We would have to download and compare all of them. In most cases, when finding a SVG is good enough, and if not, the first 4-5 extraction links we find in a HTML for a website which point to a favicon are good enough. 

This PR mainly addresses the issue that we want to (and have to) increase the amount of domains we process, and keep the resource management low (the current pod this airflow job is runnig on is being constantly killed by k8s).

merino/jobs/navigational_suggestions/domain_metadata_extractor.py
gruberb: We currently just compare the `width` of the found favicons. We focused on making this Airflow job more versatile and usable by other teams, and there are loads of improvements we still can and want to do. This might be one more on the list!

jrconlin (approved):

PR #872, opened 2025-03-28
fix: reduce logs for nav-suggestions job

gruberb:
## References

Reduce the amount of logs we accumulate to not to overload the Kubernetes pod.

## PR Review Checklist

### DIFF
--- a/merino/jobs/navigational_suggestions/domain_metadata_extractor.py
+++ b/merino/jobs/navigational_suggestions/domain_metadata_extractor.py
@@ -61,8 +61,7 @@ class Scraper:
         try:
             self.browser.open(url, timeout=TIMEOUT)
             return str(self.browser.url)
-        except Exception as e:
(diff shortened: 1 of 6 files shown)


tiftran (approved):
Just a question, if we're reducing the number of logs, we are also reducing the visibility of unsuccessful domains or favicons, are there any negative ramifications? 

I understand it's a job, so in the event that something goes wrong, we could re add logs to help debug

gruberb:
That's correct! We have a ticket we are working on shortly (https://mozilla-hub.atlassian.net/browse/DISCO-3422). The current logs were quite verbose, and frankly, not very helpful. They just mentioned what was already happening, and based on the current logs, we couldn't easily tell major issues, like: Summary of websites without favicons, websites with certain SSL errors etc. 

So I want to shrink it down, since the pod seems to struggle keeping up with the logs as well. This ticket mentioned should figure out what we need and want, and then implement it in a sufficient manner!

PR #874, opened 2025-04-01
[DISCO-3393] Suggest: Weather 204 when there's no weather for location

tiftran:
## References

JIRA: [DISCO-3393](https://mozilla-hub.atlassian.net/browse/DISCO-3393)

## Description
When no weather report isn't returned in the event that we don't get anything from accuweather we want to return 204.
If no weather report is returned due to a BackendError we want to return 200. 
This way, there's a way to differentiate diff scenarios so that clients can retry for the latter case.

## PR Review Checklist

[DISCO-3393]: https://mozilla-hub.atlassian.net/browse/DISCO-3393?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/merino/providers/suggest/weather/provider.py
+++ b/merino/providers/suggest/weather/provider.py
@@ -8,6 +8,7 @@ import aiodogstatsd
 from fastapi import HTTPException
 from pydantic import HttpUrl
 
+from merino.providers.suggest.weather.backends.accuweather.errors import MissingLocationKeyError
 from merino.utils import cron
 from merino.exceptions import BackendError
 from merino.middleware.geolocation import Location
@@ -41,6 +42,31 @@ class Suggestion(BaseSuggestion):
     current_conditions: CurrentConditions
     forecast: Forecast
     placeholder: Optional[bool] = None
+
+
+# A sentinel suggestion indicating that no location key was found for the given location.
+NO_LOCATION_KEY_SUGGESTION: Suggestion = Suggestion(
+    title="N/A",
+    url=HttpUrl("https://merino.services.mozilla.com"),
+    city_name="",
+    region_code="",
+    current_conditions=CurrentConditions(
+        url=HttpUrl("https://merino.services.mozilla.com"),
+        icon_id=0,
+        summary="",
+        temperature=Temperature(),
+    ),
+    forecast=Forecast(
+        url=HttpUrl("https://merino.services.mozilla.com"),
+        summary="",
+        high=Temperature(),
+        low=Temperature(),
+    ),
+    provider="",
+    is_sponsored=False,
+    score=0,
+    placeholder=True,
+)
 
 
 class LocationCompletionSuggestion(BaseSuggestion):
@@ -130,7 +156,6 @@ class Provider(BaseProvider):
         weather_report: WeatherReport | None = None
         location_completions: list[LocationCompletion] | None = None
         weather_context = WeatherContext(geolocation, languages)
-        has_error = False
         try:
             with self.metrics_client.timeit(f"providers.{self.name}.query.backend.get"):
                 if is_location_completion_request:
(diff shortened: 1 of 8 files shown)


merino/web/api_v1.py
```diff
@@ -221,6 +223,14 @@ async def suggest(
             for task in completed_tasks
         )
     )
+    if request_type == "weather" and not suggestions:
+        return Response(status_code=204)
```
tiftran, line 227: `ORJSONResponse` requires `content` as a required parameter. It doesn't check whether the value is `None` which is why we had the weird `Response content longer than Content-Length` error

merino/web/api_v1.py
```diff
@@ -221,6 +223,14 @@ async def suggest(
...
+
+    if request_type == "weather":
+        # remove placeholder weather suggestion
+        suggestions = [
+            s for s in suggestions if not (s.provider == WEATHER_PROVIDER and hasattr(s, "placeholder") and s.placeholder)
+        ]
```
tiftran, line 233: I flip-flopped on how to do this well, it still feels clunky. I ended returning a placeholder suggestion when a BackendError occurs and it gets filtered here.

merino/providers/suggest/weather/provider.py
```diff
@@ -179,6 +185,25 @@ def build_suggestion(
                 locations=data,
             )
 
+    def build_placeholder_error_weather_suggestion(self) -> Suggestion:
```
ncloudioj, line 188: It's a shame that there is no clearer way than using a dummy suggestion to indicate "No Content" from a provider in `api_v1.py`. Though instead of building a dummy suggestion for each `BackendError` with a somewhat obscure `placeholder` field in `Suggestiion`. I'd propose doing the following:
- The `WeatherBackend` protocol has `get_weather_report(self, ...) -> WeatherReport | None`, does the return value `None` always mean there is no location key for the given `WeatherContext`? If not, I think we can define a new exception `MissingLocationKeyError` in `protocol.py` and raise it from the backend whenever it couldn't find the location key from the upstream.
- Then we can define a sentinel suggestion and return it whenever `self.backend.get_weather_report(weather_context)` raise a `MissingLocationKeyError` in the provider.
- In `api_v1.py`, we can check that sentinel suggestion and short-circuit it with a 204 accordingly.  

```py
# Define this in `suggest/weather/provider.py`

# A sentinel suggestion indicating that no location key was found for the given location.
NO_LOCATION_KEY_SUGGESTION: Suggestion = Suggestion(
    title="N/A",
    url=self.dummy_url,
    city_name="",
    ...
)

try:
    with self.metrics_client(...):
        ...
except MissingLocationKeyError:
    return [NO_LOCATION_KEY_SUGGESTION]
except BackendError as backend_error:
    logger.warning(backend_error)

# Lastly, in `api_v1.py`
if (
    request_type == "weather"
    and len(suggestions) == 1
    and suggestions[0] is NO_LOCATION_KEY_SUGGESTION
):
    return Response(status_code=204)
```

What do you think?

tiftran: ooh i like this better, but I think we would still need to check/filter the suggestion out for the general case where request_type isn't specified

ncloudioj: Good point. This makes me wonder if we could generalize this for other providers as well? For example, to differentiate between "no suggestions for this query/location" and "we have to return an empty list because something goes wrong with this provider". For the former case, each provider would return a single-entry list with the no-content sentinel. Then in `api_v1.py`, we can do:

```py
if all([suggestion is NO_CONTENT_SENTINAL for suggestion in suggestions]):
    return Response(status_code=204)
```

That requires a lot of changes to other providers and we haven't identified any practical use cases other than AccuWeather. So we can handle the AccuWeather for now and leave the generalization for future iterations if needed.

ncloudioj (approved):
LGTM, thanks!

PR #875, opened 2025-04-01
Bug 1957309: Remove location modifier from Yelp data

dadaa:
## References

## PR Review Checklist

### DIFF
--- a/merino/jobs/csv_rs_uploader/yelp.py
+++ b/merino/jobs/csv_rs_uploader/yelp.py
@@ -9,7 +9,6 @@ from merino.jobs.csv_rs_uploader.base import BaseSuggestion
 FIELD_SUBJECTS = "subjects"
 FIELD_PRE_MODIFIERS = "pre-modifiers"
 FIELD_POST_MODIFIERS = "post-modifiers"
-FIELD_LOCATION_MODIFIERS = "location-modifiers"
 FIELD_LOCATION_SIGNS = "location-signs"
 FIELD_YELP_MODIFIERS = "yelp-modifiers"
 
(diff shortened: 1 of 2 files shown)


0c0w3 (approved):
Nice, thanks!

@Herraj This is part of some changes and simplifications we're making to offline Yelp suggestions, see https://bugzilla.mozilla.org/show_bug.cgi?id=1957309 and https://bugzilla.mozilla.org/show_bug.cgi?id=1953667

Herraj (approved):

0c0w3 (changes requested):
@dadaa I'll go ahead and request changes here based on my comment in https://github.com/mozilla/application-services/pull/6678. Sorry again for not realizing that sooner.

dadaa:
Thank you very much, Drew!
[PR 6678](https://github.com/mozilla/application-services/pull/6678) can handle old struct format `{ keyword, need_location }` and new string format. Thus, as the struct format can be handled regardless of applying the PR, we should use struct format for a while (until some versions up). I will change to do that.

0c0w3 (commented):
This means that from now on, location signs will always have `needLocation = false`, which means old clients will now think that location signs never need a location. That will break them because [they will only include `find_loc` if `need_location` is true](https://github.com/mozilla/application-services/blob/21b2ede7c53de47766bbdd417e9c30da88ed8a8a/components/suggest/src/yelp.rs#L454-L456), right?

If I'm not mistaken, we don't need to make any changes to this uploader job at all. The RS data can continue to use `{ keyword, needLocation }` objects for location signs. And the new Rust code can just ignore `needLocation` and treat all location signs as modifiers.

There are other approaches too (which we've discussed) but that seems like the simplest.

dadaa:
Yeah, as you said, the current script also works forthe  new dataset (I added locatoin-modifiers field to the csv)
I will close this PR. Thanks!

PR #880, opened 2025-04-08
fix: add psutil to dependencies

gruberb:
## References

Adding `psutil` to the dependencies.

## PR Review Checklist

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -21,7 +21,7 @@ COPY . $APP_HOME
 
 # Install maxmind db and app dependencies. Clean up build tools after
 RUN apt-get update && \
-    apt-get install --yes build-essential libmaxminddb0 libmaxminddb-dev && \
+    apt-get install --yes build-essential libmaxminddb0 libmaxminddb-dev procps && \
     uv sync --frozen --no-cache --no-dev --no-group load && \
     apt-get remove --yes build-essential && \
     apt-get -q --yes autoremove && \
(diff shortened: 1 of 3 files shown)


pyproject.toml
```diff
@@ -29,6 +29,7 @@ dependencies = [
     "gcloud-aio-storage>=9.3.0,<10",
     "ua-parser>=1.0,<2.0",
     "redis[hiredis]>=5.2.1,<6",
+    "psutil>=7.0.0",
```
tiftran, line 32: not sure, but can it be under the dependency group `jobs` below?

gruberb: But we do this `uv sync --frozen --no-cache --no-dev --no-group load` inside the Docker container. So we won't install it. Then we would have to add it inside the Docker container I assume?

Dockerfile
```diff
@@ -21,8 +21,8 @@ COPY . $APP_HOME
 
 # Install maxmind db and app dependencies. Clean up build tools after
 RUN apt-get update && \
-    apt-get install --yes build-essential libmaxminddb0 libmaxminddb-dev && \
-    uv sync --frozen --no-cache --no-dev --no-group load && \
+    apt-get install --yes build-essential libmaxminddb0 libmaxminddb-dev procps && \
+    uv sync --frozen --no-cache --no-dev --group jobs --group load && \
```
Herraj, line 25: ah we shouldn't install the `load` group here for the main docker file. I think it would've installed the `jobs` and main dependencies. But lets just add back the `--no-group load`

gruberb: So we just do `--no-group load`, which installs `dev` and `jobs` by default?

Herraj: yep because the previously it was fine with the other job dependencies. I think `uv sync` installs all groups, unless you specifically exclude a group.

PR #887, opened 2025-04-10
add GCP token file to dockerignore

fkiriakos07:
## References

Preventing the same issue reported in this bug: https://bugzilla.mozilla.org/show_bug.cgi?id=1959182


ncloudioj (approved):
Thanks!

PR #890, opened 2025-04-11
[DISCO 2053] feat: Add a circuit breaker for the AccuWeather provider

ncloudioj:
## References

JIRA: [DISCO-2053](https://mozilla-hub.atlassian.net/browse/DISCO-2053)

## Description
This adds a circuit breaker to the AccuWeather provider which guards against two potential integration failures:
- Redis
- The AccuWeather endpoint

With this, should any of the above experiences interruptions (e.g. service downtime or high latency), the circuit breaker will help Merino to gracefully handle those without affecting the rest of the service. I've done some local benchmarks, the overhead is negligible. We will see how it performs in load tests and production.

Please see in-source documents for more implementation details. The patch was split into several smaller commits to facilitate the code review.

Will do a follow up to add metrics monitoring the circuit break status.     

## PR Review Checklist

[DISCO-2053]: https://mozilla-hub.atlassian.net/browse/DISCO-2053?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/merino/providers/suggest/base.py
+++ b/merino/providers/suggest/base.py
@@ -169,6 +169,16 @@ class BaseProvider(ABC):
         """
         ...
 
+    def validate(self, srequest: SuggestionRequest) -> None:  # pragma: no cover
+        """Validate the request. Raise an `HTTPException` for any validation errors.
+
+        Args:
+          - `srequest`: the suggestion request.
+        Raise:
+          - `HTTPException` for any validation errors.
+        """
+        return
+
     def normalize_query(self, query: str) -> str:  # pragma: no cover
         """Normalize the query string when passed to the provider.
         Each provider can extend this class given its requirements. Can be used to
--- a/merino/providers/suggest/weather/provider.py
+++ b/merino/providers/suggest/weather/provider.py
@@ -9,8 +9,8 @@ from fastapi import HTTPException
 from pydantic import HttpUrl
 
 from merino.providers.suggest.weather.backends.accuweather.errors import MissingLocationKeyError
+from merino.governance.circuitbreakers import WeatherCircuitBreaker
 from merino.utils import cron
-from merino.exceptions import BackendError
 from merino.middleware.geolocation import Location
 from merino.providers.suggest.base import BaseProvider, BaseSuggestion, SuggestionRequest
 from merino.providers.suggest.custom_details import CustomDetails, WeatherDetails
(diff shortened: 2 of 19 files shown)


merino/web/api_v1.py
```diff
@@ -191,16 +191,18 @@ async def suggest(
...
     languages = get_accepted_languages(accept_language)
+    geolocation = refine_geolocation_for_suggestion(request, city, region, country)
 
     validate_suggest_custom_location_params(city, region, country)
 
     for p in search_from:
         srequest = SuggestionRequest(
             query=p.normalize_query(q),
-            geolocation=refine_geolocation_for_suggestion(request, city, region, country),
+            geolocation=geolocation,
```
misaniwere, line 201: 👌🏾nice

merino/providers/suggest/weather/provider.py
```diff
@@ -141,15 +141,23 @@ async def _fetch_mappings(self) -> None:
...
 
+    @WeatherCircuitBreaker(name="weather")  # Expect `AccuweatherError` and `BackendError`
+    async def query(self, srequest: SuggestionRequest) -> list[BaseSuggestion]:
+        """Provide weather suggestions.
+
+        All the `AccuweatherError` errors, raised from the backend, are intentionally
+        unhandled in this function to drive the circuit breaker. Those exceptions will
+        eventually be propagated to the provider consumer (i.e. the API handler) and be
+        handled there.
+        """
```
Herraj, line 160: curious: Is the `@WeatherCircuitBreaker...` decorator how the library recommends to implement circuit breaker? Almost like a wrapper for the entry point function.

merino/providers/suggest/base.py
```diff
@@ -169,6 +169,16 @@ async def query(self, srequest: SuggestionRequest) -> list[BaseSuggestion]:  # p
...
+    def validate(self, srequest: SuggestionRequest) -> None:  # pragma: no cover
+        """Validate the request. Raise an `HTTPException` for any validation errors.
+
+        Args:
+          - `srequest`: the suggestion request.
+        Raise:
+          - `HTTPException` for any validation errors.
+        """
+        return
+
```
Herraj, line 181: I'm assuming the other suggest providers have to implement this too now?

tests/integration/api/v1/suggest/test_suggest_weather.py
```diff
@@ -271,6 +291,65 @@ def test_suggest_backend_error_weather_report_returns_empty(
...
+                "/api/v1/suggest?provider=accuweather&q=weather&request_type=weather"
+            )
+
+            assert response.status_code == 200
+            result = response.json()
+            assert expected_suggestion == TypeAdapter(list[Suggestion]).validate_python(
+                result["suggestions"]
+            )
+
+
```
Herraj, line 352: great test!

Herraj (approved):
🚀

merino/providers/suggest/weather/provider.py
ncloudioj: Not necessarily, it also provides a function decorator which achieves the same. The main benefit with this is that we can further customize the breakers individually and define all of them in a single module for better discoverability.

merino/providers/suggest/base.py
ncloudioj: Yup, it provides a default no-op implementation so that each provider can override it to implement its own validation. The other reason is to decouple the query request validation from the query execution to facilitate the circuit breaking, otherwise it'd be awkward for us to distinct query validation errors and backend errors.

With this change, we can do query validation prior to query execution so that the failures with the former will raise HTTP 400 errors and Merino won't even bother running those queries.

PR #891, opened 2025-04-14
DISCO-3440 Fix wrong title in manifest production

gruberb:
## References

JIRA: [DISCO-3440](https://mozilla-hub.atlassian.net/browse/DISCO-3440)

## Description
We were facing race conditions in the `top-picks` file creation. When getting the title from the scraper in each batch, the one domain which was processed last was overwriting the previous domains. So each batch just hat one `title` - the one from the last domain processed in that batch.

This didn't happen for all the batches though.

This PR is:
- removing the global scraper inside `_process_domains`
- creating a local `Scraper()` for each `_process_single_domain` task
- Is reducing the batch size from`50` to `25` to be able to clean up resources more often and keep the RAM down

The maximum RAM used (locally) is around 650MB. 

## PR Review Checklist

[DISCO-3440]: https://mozilla-hub.atlassian.net/browse/DISCO-3440?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/merino/jobs/navigational_suggestions/domain_metadata_extractor.py
+++ b/merino/jobs/navigational_suggestions/domain_metadata_extractor.py
@@ -2,6 +2,8 @@
 
 import asyncio
 import logging
+import contextvars
+
 from typing import Any, Optional
 from urllib.parse import urljoin, urlparse
 
@@ -44,6 +46,13 @@ class Scraper:
     browser: StatefulBrowser
     request_client: AsyncFaviconDownloader
     parser: str = "html.parser"
+
+    async def __aenter__(self):
+        return self
+
+    async def __aexit__(self, exc_type, exc_val, exc_tb):
+        self.close()
+        return False
 
     def __init__(self) -> None:
         session: requests.Session = requests.Session()
@@ -137,30 +146,17 @@ class Scraper:
         try:
             if hasattr(self.browser, "session") and self.browser.session:
                 self.browser.session.close()
+
             if hasattr(self.browser, "_browser"):
                 self.browser._browser = None
+
             self.browser.close()
         except Exception as ex:
             logger.warning(f"Error occurred when closing scraper session: {ex}")
 
-    def reset(self) -> None:
-        """Reset the scraper by closing current connections and initializing a new browser session."""
-        try:
-            # Close the current browser session
-            self.close()
 
-            # Reinitialize the browser with a fresh session
-            session: requests.Session = requests.Session()
-            self.browser = StatefulBrowser(
-                session=session,
-                soup_config={"features": self.parser},
-                raise_on_404=False,
-                user_agent=REQUEST_HEADERS["User-Agent"],
-            )
-
-            # Don't recreate the request_client as it will be reset separately
-        except Exception as ex:
-            logger.warning(f"Error occurred when resetting scraper: {ex}")
+# Create a context variable for the current scraper
+current_scraper: contextvars.ContextVar[Scraper] = contextvars.ContextVar("current_scraper")
 
 
 class DomainMetadataExtractor:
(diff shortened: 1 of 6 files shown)


merino/jobs/navigational_suggestions/domain_metadata_extractor.py
```diff
@@ -45,6 +45,13 @@ class Scraper:
     request_client: AsyncFaviconDownloader
     parser: str = "html.parser"
 
+    async def __aenter__(self):
```
gruberb, line 48: To be able to use the `Scraper` in a context manager, I had to add these two function definitions.

merino/jobs/navigational_suggestions/domain_metadata_extractor.py
```diff
@@ -45,6 +45,13 @@ class Scraper:
...
+        return self
+
+    async def __aexit__(self, exc_type, exc_val, exc_tb):
```
gruberb, line 51: This will automatically close the scraper and clean it up once the context is closed.

merino/jobs/navigational_suggestions/domain_metadata_extractor.py
```diff
@@ -136,32 +143,19 @@ def close(self) -> None:
         """Close the scraper browser session properly"""
         try:
             if hasattr(self.browser, "session") and self.browser.session:
+                # Close all adapters explicitly
+                for adapter in self.browser.session.adapters.values():
```
ncloudioj, line 147: Can you share some context about why this needs to be done here? Especially on why only to close the adapters of the session instead of closing the session as a whole.

merino/jobs/navigational_suggestions/domain_metadata_extractor.py
```diff
@@ -243,15 +237,18 @@ def _fix_url(self, url: str) -> str:
         return url
 
     async def _extract_favicons(
-        self, scraped_url: str, max_icons: int = 5
+        self,
+        scraped_url: str,
+        scraper: Scraper,
```
ncloudioj, line 242: `DomainMetadataExtractor` already has a `scraper` instance member, does this change make that one obsoleted?

ncloudioj: Looks like all the methods of this class now take a `Scaper` as an argument. If we have to use a local `Scaper` to better manage its state rather than reusing the instance member, consider using a [`contextvar`](https://docs.python.org/3/library/contextvars.html#module-contextvars) to ease the argument passing bits.

merino/jobs/navigational_suggestions/domain_metadata_extractor.py
gruberb: I think this is "leftover" from my local testing. Somehow the scraper uses a lot of RAM, and I wanted to see if it really deletes all the stored values or not. I think we can go to the previous simple session close.

merino/jobs/navigational_suggestions/domain_metadata_extractor.py
gruberb: I like the `contextvar`. It means a bit more complicated testing, but I think it's worth it. Maybe we can even do it better in the future.

merino/jobs/navigational_suggestions/domain_metadata_extractor.py
ncloudioj: @gruberb Can you take a look at this comment as well?

ncloudioj (approved):
LGTM, thanks!

merino/jobs/navigational_suggestions/domain_metadata_extractor.py
```diff
@@ -243,15 +236,22 @@ def _fix_url(self, url: str) -> str:
...
+        self,
+        scraped_url: str,
+        max_icons: int = 5,
     ) -> list[dict[str, Any]]:
         """Extract a limited number of favicons for an already opened url"""
         self._current_base_url = scraped_url
         favicons: list[dict[str, Any]] = []
 
+        scraper = current_scraper.get()
+        if not scraper:
```
ncloudioj, line 248: I think it should be fine to not check the return value for each `current_scaper.get()` in this class, as `current_scaper.set()` should be guaranteed to happen before all the get calls, which is verified by the tests as well.

PR #894, opened 2025-04-14
fix: Use a regular context manager for the Scraper

ncloudioj:
## References

JIRA: N/A

## PR Review Checklist

### DIFF
--- a/tests/unit/jobs/navigational_suggestions/test_domain_metadata_extractor.py
+++ b/tests/unit/jobs/navigational_suggestions/test_domain_metadata_extractor.py
@@ -1353,7 +1353,7 @@ async def test_process_single_domain_unreachable(mock_domain_metadata_uploader, 
         )
 
     # Verify the exact expected result
-    assert result == {"url": None, "title": None, "icon": None, "domain": None}
+    assert result == {"url": None, "title": "", "icon": "", "domain": ""}
 
 
 @pytest.mark.asyncio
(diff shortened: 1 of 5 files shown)


tests/unit/jobs/navigational_suggestions/test_domain_metadata_extractor.py
```diff
@@ -1353,7 +1353,7 @@ async def test_process_single_domain_unreachable(mock_domain_metadata_uploader,
         )
 
     # Verify the exact expected result
-    assert result == {"url": None, "title": None, "icon": None, "domain": None}
+    assert result == {"url": None, "title": "", "icon": "", "domain": ""}
```
ncloudioj, line 1356: This is the only failing test that i had to revert. @gruberb would need some help from you for a more proper fix :).

gruberb:
Oh, good catch! I haven't tried `__exit__` instead of `__aexit__`.

tests/unit/jobs/navigational_suggestions/test_domain_metadata_extractor.py
gruberb: Hm, that's weird.... I now wonder why `None` was valid before. I a domain is blocked, we currently return empty Strings for `title`, `icon` and `domain` and `None` for the URL, like you fixed. Now looking through it, I wonder why it didn't before?

gruberb (approved):

PR #896, opened 2025-04-15
[DISCO-3357] Wiki Data is sent to new collection

tiftran:
## References

JIRA: [DISCO-3557](https://mozilla-hub.atlassian.net/browse/DISCO-3557)

## Description
This reuses existing scripts from `quicksuggest-rs` that are modified to be used to upload wikipedia suggestions for en, fr, de, it and pl. I omitted moving over the blocklist, as I was not sure whether it could be public.
I've uploaded suggestions that would be outputted from these scripts in remote settings dev as well, for review.

Scripts will need to be changed to conform to our existing rs-uploaders as a follow up.
## PR Review Checklist


merino/jobs/wikipedia_offline_uploader/upload-data.sh
```diff
@@ -0,0 +1,87 @@
...
+      "pl") fe="env.locale == 'pl'" ;;
+      *) error "Unsupported language specified" ;;
+    esac
+
+    echo "{\"type\": \"wikipedia\", \"filter_expression\": \"${fe}\"}"
+
+    $DRY_RUN curl -X POST ${SERVER}/buckets/${WORKSPACE}/collections/${OFFLINE_CID}/records/${NAME}/attachment \
+        -H 'Content-Type:multipart/form-data' \
+        -F attachment=@${file} \
+        -F data="{\"type\": \"wikipedia\", \"filter_expression\": \"${fe}\"}" \
```
tiftran, line 80: I flipped flopped with this. AMP uses countries as it is country specific. Wikipedia is language specific. While locale requires multiple locales to be listed for a given language like english, it is a finite small amount of locales that requires this, and this would not require client changes, compared to using `language` which isn't an official filter?

ncloudioj (commented):
Thanks for moving things over and adding the support for other locales!

While we can use those scripts for the time being to support the upcoming rollout, moving forward, I have a couple of suggestions for future iterations.

- Let's rewrite `fetch-wikipedia-*.sh` in Python. I opted for Bash scripts when I wrote them to get the job done in hope that we'd throw them away soon, but it was hacky and awkward to use them along with other Python ones. Converting them into Python would be much easier to write and maintain IMO.
- Let's drop `upload-data.sh` to favor the existing RS data uploaders.
- We can provide a CLI that glues all the tasks together, from downloading Wiki top articles, selecting the TopN pages, make suggestions, and to uploading to RS. No need to execute those tasks individually.

Let me know what you think.

ncloudioj:
Cool. All those three rewrites/refactors can be done as follow ups. Once the content moderation bits get moved in, we can merge this to get things rolling for now. Thanks!

ncloudioj (approved):
LGTM.

Can you convert `page_ignore.json` to `page_ignore.csv`, and move it along with `dynamic_wikipedia_blocklist.csv` into a `moderation` folder?

merino/jobs/wikipedia_offline_uploader/fetch-wikipedia-top-pages.sh
```diff
@@ -0,0 +1,125 @@
...
+    ;;
+  -l | --language)
+  language="${2}"
+  case "${language}" in
+  en | fr | de | it | pl)
+    shift
+    shift
+    ;;
+  *)
+    echo "Unknown access-method: ${access}. Should be one of 'en', 'fr', 'de', 'it', and 'pl'"
```
0c0w3, line 64: Should be language, not access-method?

merino/jobs/wikipedia_offline_uploader/make_suggestions.py
```diff
@@ -0,0 +1,168 @@
...
+
+# A set to store all the keywords observed thus far.
+SEEN_KEYWORDS: Set[str] = set([])
+
+# The leading words to be skipped
+SKIP_WORDS: Pattern = re.compile(r"(\d+|the|an|a)$")
+
+# The directory for adM provided suggestions. This is used to load all the
+# keywords of the in-flight sponsored suggestions.
+ADM_SOURCE_DIR = "../source-data/"
```
0c0w3, line 87: Small nit: Seems like we use "AMP" these days, so maybe change aDM -> AMP?

merino/jobs/wikipedia_offline_uploader/make_suggestions.py
```diff
@@ -0,0 +1,168 @@
...
+        if len(keywords) >= MIN_KEYWORDS:
+            for keyword in keywords:
+                SEEN_KEYWORDS.add(keyword)
+            yield {
+                "id": 0,  # FIXME: assign a suggestion ID.
+                "url": f"{URL_PREFIX.format(language=language)}{quote(title)}",
+                "iab_category": "5 - Education",
+                "icon": ICON_ID,
+                "advertiser": "Wikipedia",
+                "title": f"Wikipedia - {title_string}",
```
0c0w3, line 124: The name "Wikipedia" itself is localized, so we should use the proper localized name in the suggestion title, ideally directly from the data rather than being hardcoded here per language. I checked the new languages and here's the breakdown:

* fr: Wikipédia
* pl: Wikipedii
* de, it: Wikipedia (same as English)

merino/jobs/wikipedia_offline_uploader/upload-data.sh
0c0w3: Yeah using locale should be good. 👍

0c0w3 (commented):
Thanks Tif!

merino/jobs/wikipedia_offline_uploader/make_suggestions.py
0c0w3: > pl: Wikipedii

Hmm [the pl main page](https://pl.wikipedia.org/wiki/Wikipedia:Strona_g%C5%82%C3%B3wna) does use "Wikipedii" in some places but "Wikipedia" in others, including the logo. Individual article titles also use "Wikipedia", so we should probably use that after all.

fr Wikipedia seems to consistently use Wikipédia.

annasob:
Did we file the follow-ups ??

tiftran:
@annasob https://mozilla-hub.atlassian.net/browse/DISCO-3399 should covers the changes needed

PR #899, opened 2025-04-16
[DISCO-3457]: feat: Add metrics for circuit breakers

ncloudioj:
## References

JIRA: [DISCO-3457](https://mozilla-hub.atlassian.net/browse/DISCO-3457)

## PR Review Checklist

[DISCO-3457]: https://mozilla-hub.atlassian.net/browse/DISCO-3457?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/docs/data.md
+++ b/docs/data.md
@@ -204,4 +204,4 @@ When requesting a manifest file, we record the following metrics.
 
 The following metrics are recorded for service governance monitoring.
 
-- `governance.circuits.<circuit-breaker-name>` - A gauge to instrument the status for each circuit breaker in Merino. The value can be any of "closed", "open", and "half-open". A tag of `{"failure_count": count}` is also attached.
+- `governance.circuits.<circuit-breaker-name>` - A gauge to instrument the failure count for each "open" circuit breaker.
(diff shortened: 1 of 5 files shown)


Herraj (approved):
⏲️

Herraj:
Should we keep the cron job interval time the same as circuit breaker interval time or do a little bit longer? Let's maybe a second or two? Do you think we'll run into a case where it stops governing before the circuit breaker status changes? 🤔

ncloudioj:
> Should we keep the cron job interval time the same as circuit breaker interval time or do a little bit longer? Like maybe a second or two longer? Do you think we'll run into a case where it stops governing before the circuit breaker status changes? 🤔

Good point. Indeed, this might not be able to instrument those short-lived "open" circuits, my rationale is that those cases might not be the main concerns to us anyway as the circuit will be "closed" right away anyway (i.e. no actions we can take). So I ended up choosing a larger interval.

Alternatively, we can make the emission conditional, i.e. only emit if there is non-closed circuits. Then we can run the cron more often w/o incurring the overhead and recording lots of metrics that we don't care about.

Which one do you prefer?

Herraj:
> > Should we keep the cron job interval time the same as circuit breaker interval time or do a little bit longer? Like maybe a second or two longer? Do you think we'll run into a case where it stops governing before the circuit breaker status changes? 🤔
> 
> Good point. Indeed, this might not be able to instrument those short-lived "open" circuits, my rationale is that those cases might not be the main concerns to us anyway as the circuit will be "closed" right away anyway (i.e. no actions we can take). So I ended up choosing a larger interval.
> 
> Alternatively, we can make the emission conditional, i.e. only emit if there is non-closed circuits. Then we can run the cron more often w/o incurring the overhead and recording lots of metrics that we don't care about.
> 
> Which one do you prefer?

Hmm yeah really your call here. Only emitting conditionally does make sense to me though.

PR #900, opened 2025-04-17
[DISCO-3253] Fix gcs client auth error for curated recommendations

Herraj:
## References

JIRA: [DISCO-3253](https://mozilla-hub.atlassian.net/browse/DISCO-3253)

## Description
We were seeing errors about the gcs client initializing failing to find any credentials. We fixed this error in the `gcs_uploader.py` module. The reason this occurs is that it is looking for an explicit `credentials` argument but if none is passed, it tries to pull the ADC file from the google project. However, when running locally, we don't have an ADC file hence adding this check.

Screenshot 2025-04-17 at 11 37 18 AM

## PR Review Checklist

[DISCO-3253]: https://mozilla-hub.atlassian.net/browse/DISCO-3253?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/merino/curated_recommendations/__init__.py
+++ b/merino/curated_recommendations/__init__.py
@@ -27,6 +27,7 @@ from merino.utils.http_client import create_http_client
 from merino.utils.synced_gcs_blob import SyncedGcsBlob
 
 from merino.providers.manifest import get_provider as get_manifest_provider
+from merino.utils.gcs import initialize_storage_client
 
 logger = logging.getLogger(__name__)
 
@@ -38,7 +39,9 @@ def init_engagement_backend() -> EngagementBackend:
     try:
         metrics_namespace = "recommendation.engagement"
         synced_gcs_blob = SyncedGcsBlob(
-            storage_client=Client(settings.curated_recommendations.gcs.gcp_project),
+            storage_client=initialize_storage_client(
+                destination_gcp_project=settings.curated_recommendations.gcs.gcp_project
+            ),
             metrics_client=get_metrics_client(),
             metrics_namespace=metrics_namespace,
             bucket_name=settings.curated_recommendations.gcs.bucket_name,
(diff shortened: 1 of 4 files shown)


merino/curated_recommendations/__init__.py
```diff
@@ -33,12 +34,25 @@
...
+def initialize_storage_client(destination_gcp_project: str) -> Client:
+    """Initialize a Google Cloud Storage client with production or anonymous credentials if in non-production/staging environment"""
+    if settings.runtime.skip_gcp_client_auth:
+        # for production and staging envs we don't have to explicitly pass the credentials
+        #  as it picks up the ADC file automatically
+        return Client(destination_gcp_project)
+    else:
+        # if not using anonymous credentials in dev & testing envs, this will throw
+        return Client(destination_gcp_project, credentials=AnonymousCredentials())  # type: ignore
+
```
Herraj, line 46: this function is copied over from `gcs_uploader.py` module.

jpetto: would it make sense to have a single instance of this function to be shared? or is it better to have a copy?

jpetto (approved):
ty for fixing!

merino/curated_recommendations/__init__.py
Herraj: Good idea! But I'll wait for the 3rd instance where we have to use this function to refactor it into a util function 😅

ncloudioj: No need to wait for another use case. Since we already have `merino/utils/gcs`, I'd suggest moving this over `merino/utils/gcs/__init__.py` and use it for `gcs_uploader` and `curate_recommendations`.

ncloudioj (approved):
Thanks for fixing it up! r+wc.

PR #919, opened 2025-05-08
New AMP Suggestion storage structures

tiftran:
## Description
Just thought I put up what I have so far.

After investigation: I found that suggestions across form-factors are quite similar. Aside from `click_url`, `impression_url` and `url`, but there are about ~40 suggestions that have differences in addition to those fields.

I thought it might be a good idea to store them like the following
```
core_suggestion_data = {
    123: {"id": "123", "title": "Item A", ....},
    ...
}

overrides = {
    456: {
        "desktop": {"url": "https://example.com/url", "click_url": "https://example.com/click_url"},
        "phone": {"url": "https://example.com/mobile-url", "click_url": "https://example.com/mobile-url"},
         ...
    },
 
    ...
}
```

Each query will be associated with an suggestion_id which then can be used to build the suggestion data, but retrieve the core components and then adding the overrides to it.

There are about ~30 suggestions with the same ID that have different keywords, and I think we could do the same with keywords too.

This may not be the most compressed way to store data, but if we're okay with this, we could iterate as we go

## PR Review Checklist

### DIFF
--- a/merino/configs/default.toml
+++ b/merino/configs/default.toml
@@ -159,7 +159,7 @@ bucket = "main"
 # MERINO_REMOTE_SETTINGS__COLLECTION
 # The collection to use for Remote Settings providers if not specified in provider config.
 # Ex: "quicksuggest".
-collection = "quicksuggest"
+collection = "quicksuggest-amp"
 
 # The collection to use for uploading fakespot suggestions if not specified on the command line
 collection_fakespot = "fakespot-suggest-products"
@@ -179,6 +179,10 @@ dry_run = false
 # MERINO_REMOTE_SETTINGS__SCORE
 # Default score to set in suggestions uploaded to remote settings.
 score = 0.25
+
+# MERINO_REMOTE_SETTINGS__COUNTRIES
+# Default countries used to filter out records from remote settings.
+countries = ["US"]
 
 # Global Redis settings. The weather provider optionally uses Redis to cache weather suggestions.
 [default.redis]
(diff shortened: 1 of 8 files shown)


ncloudioj:
I think the overall structure is a good start, the main risk is perhaps that this assumes the `suggestion-id` is stable across form factors. If that's not enforced by the partner (it was not requested by us either), Merino could wind up serving suggestions with mismatched properties. Is there another filed(s) that can be used as the "join key", e.g. "advertiser"?

A few other open questions:
- The partner has noted us to merge "phone" and "tablet" on their end, so it's very likely to have "desktop" and "mobile" form factors. Would that make the proposal less effective?
- With this approach, how would we handle other dimensions, such as country?

tiftran:
Ooh so i did consider that` id` may not be stable across all form factors, which is why I don't plan to have a strict set of fields for `core_suggestion_data`. Rather when Merino reads all the records and attachments. For  suggestions with the same id, compute the fields they have in common before constructing `core_suggestion_data` From the existing data, ids seems to roughly match up, but yes this wouldn't be effective if there's a drastic change in the data

For shrinking to just `mobile` and `desktop` rather than 3 form factors. This would be less effective as we're shrinking two sets of similar data to one and a bit rather than 3? but there would still be shrinkage?

For countries... I assume suggestions would be different between countries with very few overlap

I think it would look something like 
```
{
  "US": {
    "core_suggestion_data": {...}
    ....
  }
  "FR": {
    "core_suggestion_data": {...}
    ....
  }
....
}
```?

PR #910, opened 2025-04-30
[DISCO-3479] Switch Merino to use quicksuggest-amp collection

tiftran:
## References

JIRA: [DISCO-3479](https://mozilla-hub.atlassian.net/browse/DISCO-3479)

## Description
This should move merino to fetch records from the new collection and should keep existing functionality all the same. This would help us half way to getting merino to support multi country and form factor

## PR Review Checklist

_Put an `x` in the boxes that apply_

- [ ] This PR conforms to the [Contribution Guidelines](https://github.com/mozilla-services/merino-py/blob/main/CONTRIBUTING.md)
- [ ] The PR title starts with the JIRA issue reference, format example `[DISCO-####]`, and has the same title (if applicable)
- [ ] `[load test: (abort|skip|warn)]` keywords are applied to the last commit message (if applicable)
- [ ] [Documentation](https://github.com/mozilla-services/merino-py/tree/main/docs) has been updated (if applicable)
- [ ] [Functional and performance test](https://github.com/mozilla-services/merino-py/blob/main/docs/dev/testing.md) coverage has been expanded and maintained (if applicable)

[DISCO-3479]: https://mozilla-hub.atlassian.net/browse/DISCO-3479?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #929, opened 2025-05-15
SNG-2522 - Use `quicksuggest-other` as the default RS collection

0c0w3:
The default RS collection changed to `quicksuggest-amp` in #910, but that breaks jobs that aren't related to AMP since they use `quicksuggest-other`, unless you override the value on the command line.

I think it makes sense to default to `quicksuggest-other` since there's only one major site in Merino that uses the AMP collection, and then we can add a new config value called `collection_amp` for AMP.

This also updates sites that use the old `quicksuggest` collection, let me know if that's not right.

## References

https://mozilla-hub.atlassian.net/browse/SNG-2522

## PR Review Checklist


ncloudioj (approved):
Thanks!

PR #933, opened 2025-05-20
[DISCO-3517] weather: add FIPs -> iso region mapping

tiftran:
## References

JIRA: [DISCO-3517](https://mozilla-hub.atlassian.net/browse/DISCO-3517)

## Description
Geoname mostly provides FIPS region codes. This PR adds the FIPS mapping for DE, GB, IT, and PL. This also introduces params `admin1` and `admin2`, which would be where the FIPS region codes come from.

## PR Review Checklist

[DISCO-3517]: https://mozilla-hub.atlassian.net/browse/DISCO-3517?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/MC-1725)


merino/providers/suggest/weather/backends/accuweather/pathfinder.py
```diff
@@ -212,9 +233,85 @@
     ("US", "UT", "Hill Air Force Base"): 0,
 }
 
+# mapping from https://dev.maxmind.com/geoip/whats-new-in-geoip2/#iso-3166-2-fips-10-4-and-country-subdivisions
+FIPS_ISO_MAPPING = {
```
tiftran, line 237: I omitted FR since the mapping doesn't map to ISO and one of the geonames admincodes is actually iso

ncloudioj: This can be attached as a comment to the code.

merino/providers/suggest/weather/backends/accuweather/pathfinder.py
```diff
@@ -248,7 +356,7 @@ def compass(location: Location) -> Generator[MaybeStr, None, None]:
       - region string that could be None.
     """
     country = location.country
-    regions = location.regions
+    regions = get_fips_region_mapping(country, location.regions)
```
ncloudioj, line 359: For those countries that are no in `FIPS_ISO_MAPPING.keys()`, they don't need to go through this loop.

```py
FIPS_ISO_MAPPING_COUNTRIES: frozenset =  frosenset(FIPS_ISO_MAPPING.keys())

regions = location.regions
if country and country in FIPS_ISO_MAPPING_COUNTRIES and regions:
    # country and regions are guaranteed not None, so get_fips_region_mapping(country: str, regions: list[str])
    regions = get_fips_region_mapping(country, regions)
```

0c0w3 (commented):
Thanks Tif. Geonames defines up to four admin divisions. I don't know how common it is to use all four, but there are geonames that have them, and Firefox should send them all to Merino IMO. Would that be all right? I don't know if we want individual `admin{level}` params or one `admin` param that's a comma-separated string, or something else -- I don't have a strong opinion.

merino/utils/api/query_params.py
```diff
@@ -35,21 +35,46 @@ def get_accepted_languages(languages: str | None) -> list[str]:
...
     """Validate that city, region & country params are either all present or all omitted."""
-    if any([country, region, city]) and not all([country, region, city]):
-        logger.warning(
-            "HTTP 400: invalid query parameters: `city`, `region`, and `country` are either all present or all omitted."
-        )
-        raise HTTPException(
-            status_code=400,
-            detail="Invalid query parameters: `city`, `region`, and `country` are either all present or all omitted.",
-        )
+    if any([city, region, country, admin1, admin2]):
```
ncloudioj, line 45: It's getting harder to comprehend the validation criteria by following through the code, can you outline the criteria in the function docstring?

merino/utils/api/query_params.py
```diff
@@ -35,21 +35,46 @@ def get_accepted_languages(languages: str | None) -> list[str]:
...
+            logger.warning(
+                "HTTP 400: Invalid query parameters: either `region` or `admin1` and/or `admin2` can be present."
+            )
+            raise HTTPException(
+                status_code=400,
+                detail="Invalid query parameters: either `region` or `admin1` and/or `admin2` can be present.",
+            )
+
+        if region:
+            return city, [region], country
```
ncloudioj, line 67: Validation code should not modify the inputs and output a new thing IMO, we can use another function for this.

merino/web/api_v1.py
```diff
@@ -104,6 +106,12 @@ async def suggest(
     - `region`: [Optional] Subdivision code. E.g. : “NY”. If provided,
         Accuweather provider returns weather suggestions based on this region. Note: If provided,
         `city` and `country` must also be provided to successfully return weather suggestions.
+    - `admin1`: [Optional] A subdivision code. E.g. : “NY”. If provided,
+        Accuweather provider tries to return weather suggestions based on this region. Note: If provided,
+        `city` and `country` must also be provided to successfully return weather suggestions.
+    - `admin2`: [Optional] A subdivision code. E.g. : “NY”. If provided,
```
ncloudioj, line 112: The API docs for admin1 and admin2 are identical, can you add more color in terms of how the two should be specified and if both present, which one will be used?

ncloudioj (changes requested):

merino/utils/api/query_params.py
```diff
@@ -59,18 +79,20 @@ def refine_geolocation_for_suggestion(
...
         country (Optional[str]): The name of the country to include in the geolocation, if available.
 
     Returns:
         Location: A Location object with refined geolocation data based on provided parameters.
     """
     geolocation: Location = request.scope[ScopeKey.GEOLOCATION].model_copy()
-    if country and region and city:
-        geolocation = request.scope[ScopeKey.GEOLOCATION].model_copy(
-            # temp measure as FX is not providing the expected region for GB. [DISCO-3507]
+    regions_list = regions.split(",") if regions else [region] if region else []
```
ncloudioj, line 89: Hmm, i can't tell what this "if_else_if_else" does. Can we split it into two for readability?

merino/web/api_v1.py
```diff
@@ -104,6 +105,10 @@ async def suggest(
     - `region`: [Optional] Subdivision code. E.g. : “NY”. If provided,
         Accuweather provider returns weather suggestions based on this region. Note: If provided,
         `city` and `country` must also be provided to successfully return weather suggestions.
+    - `regions`: [Optional] A list of subdivision code containing potentially FIP and ISO codes.
```
ncloudioj, line 108:
```suggestion
    - `regions`: [Optional] A list of subdivision code containing potentially FIPS and ISO codes.
```

ncloudioj: Suggestion: can we not add `regions` and just re-use `region`? So it can take either a single region or a comma separate region list. While the singular naming might not be ideal but it'll be less confusing to have two optional params for this endpoint IMO. Would this cause any difficulty elsewhere?

merino/utils/api/query_params.py
tiftran: fair 😅

ncloudioj (approved):
Looks good, thanks!

0c0w3:
lgtm too but it looks like this is missing the GB FIPS -> ISO mapping?

0c0w3 (approved):
Thanks!

PR #936, opened 2025-05-21
[DISCO-3481] - AMP: fetch and query mods for EU expansion

tiftran:
## References

JIRA: [DISCO-3481](https://mozilla-hub.atlassian.net/browse/DISCO-3481)

## Description
This allows merino to intake different suggestions for specific countries and form_factors.

Testing it out:
enforce geolocation to US or GB
set User-Agent to either desktop or phone.
make a query: US: ex `ama` `mirrorless` etc GB: `gro` -> `groupon`

## PR Review Checklist


merino/providers/suggest/adm/provider.py
```diff
@@ -162,33 +162,42 @@ def serp_categories(domain: str) -> list[Category]:
...
-                "is_sponsored": is_sponsored,
-                "icon": self.suggestion_content.icons.get(res.get("icon", MISSING_ICON_ID)),
-                "score": self.score,
-            }
-            return [
-                (
-                    SponsoredSuggestion(**suggestion_dict)
-                    if is_sponsored
-                    else NonsponsoredSuggestion(**suggestion_dict)
+        form_factor = srequest.user_agent.form_factor if srequest.user_agent else None
```
tiftran, line 165: the thought just occurred to me: do we want to default `other` to `desktop` and `tablet` to `phone`?

merino/providers/suggest/adm/backends/remotesettings.py
```diff
@@ -240,15 +243,24 @@ async def get_suggestions(
                                     base=attachment_host,
                                     url=record["attachment"]["location"],
                                 )
-                            )
+                            ),
+                            name=f"{record["country"]}_{record["form_factor"]}",
```
tiftran, line 247: @ncloudioj  i realize we never ripped this out when we switched over to the new collection. was this what you meant when you say "processing those records concurrently"? Apologies if im misunderstanding

merino/providers/suggest/adm/backends/remotesettings.py
```diff
@@ -123,27 +123,30 @@ async def fetch(self) -> SuggestionContent:
...
+                    icons_in_use.add(suggestion.icon)
+                    full_keywords_tuples = suggestion.full_keywords
+                    begin = 0
+                    for full_keyword, n in full_keywords_tuples:
+                        for query in keywords[begin : begin + n]:
+                            # Note that for adM suggestions, each keyword can only be
+                            # mapped to a single suggestion.
+                            if country not in suggestions:
+                                suggestions[country] = {}
+                            suggestions[country][(segment, query)] = (result_id, fkw_index)
```
ncloudioj, line 145: How about this index layout? The main reasoning is that `keywords` is a high cardinality set, combining it with the segment as the lookup key would significantly inflate the index size whereas the proposed layout won't suffer that. 

```py
INDEX_TYPE = dict[str, dict[tuple[str], tuple[int, int]]]
#                 ^ keyword   ^ segment
suggestions: dict[str, INDEX_TYPE]

# e.g.
{
    "US": {
        "kw-01": {
            ("desktop",): (0, 0),
            ("phone",): (1, 1),
        },
        "kw-02": {
            ("desktop",): (2, 2),
            ("phone",): (3, 3),
        },
        ...
    }
}
```

Also, we can define a numeric enum for `form_factor` such as `DESKTOP = 0; PHONE = 1` and use them for indexing. The numeric values take half of the size as string counterparts, which would save a lot given that we will have hundreds of thousands of those in memory.

merino/providers/suggest/adm/backends/remotesettings.py
ncloudioj: Yup, that's what I meant. Glad that we didn't rip it out.

I'd recommend moving `f"{record["country"]_{record["form_factor"]}}"` to a separate utility function as we might forget to update this when we add more dimensions to the segment in the future, which could cause issues.

merino/providers/suggest/adm/backends/remotesettings.py
```diff
@@ -240,15 +243,24 @@ async def get_suggestions(
...
-        suggestions: list[KintoSuggestion] = []
+        suggestions: dict[str, dict] = {}
+
         for task in tasks:
-            suggestions.extend(await task)
+            result_suggestions = await task
+            country, *segment = task.get_name().split("_")
+            segment_key = tuple(segment)
+            if country not in suggestions:
+                suggestions[country] = {}
```
ncloudioj, line 260: You can use a `defaultdict` to simplify this.

merino/providers/suggest/adm/backends/remotesettings.py
```diff
@@ -110,7 +124,7 @@ async def fetch(self) -> SuggestionContent:
         Raises:
             RemoteSettingsError: Failed request to Remote Settings.
         """
-        suggestions: dict[str, tuple[int, int]] = {}
+        suggestions: defaultdict[str, IndexType] = defaultdict(lambda: IndexType())
```
ncloudioj, line 127: Two scoops of defaultdict can eliminate two if-statements below. 
```suggestion
        suggestions: defaultdict[str, IndexType] = defaultdict(lambda: defaultdict(dict))
```

merino/providers/suggest/adm/backends/remotesettings.py
```diff
@@ -123,27 +137,33 @@ async def fetch(self) -> SuggestionContent:
...
+                    begin = 0
+                    for full_keyword, n in full_keywords_tuples:
+                        for query in keywords[begin : begin + n]:
+                            # Note that for adM suggestions, each keyword can only be
+                            # mapped to a single suggestion.
+                            if country not in suggestions:
+                                suggestions[country] = {}
+
+                            if query not in suggestions[country]:
+                                suggestions[country][query] = {}
```
ncloudioj, line 161: Defaultdict will handle this for us.
```suggestion
```

ncloudioj (approved):
LGTM, thanks!

PR #950, opened 2025-06-04
[DISCO-3551] Weather: to return localized cities

tiftran:
## References

JIRA: [DISCO-3551](https://mozilla-hub.atlassian.net/browse/DISCO-3551)

## Description
When weather is requested via geolocation, we don't retrieve the localized city name.
These changes are made, such that if a non language is requested,
Merino tries to look up whether the geolocation has the localized city name (maxmindb supports the locales listed in `MAXMIND_SUPPORTED_LOCALE`)
If that fails, it will try to request it from accuweather and cache for future request.

## PR Review Checklist


merino/providers/suggest/weather/backends/accuweather/backend.py
```diff
@@ -486,6 +515,93 @@ async def get_weather_report(self, weather_context: WeatherContext) -> WeatherRe
...
+            ) from exc
+
+        self.emit_cache_fetch_metrics(cached_data, skip_report=True)
+
+        weather_data = self.parse_cached_data(cached_data)
+        if weather_data.location:
+            return weather_data.location.localized_name
+
+        return await self._fetch_city_name_from_upstream(url, params)
+
```
Herraj, line 556: I would maybe add a comment for `line 551` to clarify that we are using the `parse_cached_data` to parse the cached byte data data. Which returns the `WeatherData` type. 

I just got confused because the `cached_data` variable gets the localized location data but then "turns into" `weather_data` variable 😅

Herraj (approved):
Looks good! 🚀

ncloudioj (commented):
A few thoughts I'd like to share for discussion:
- Unlike other things (location key & weather info) that we have to request from the partner, localized locations are relatively stable hence it's totally viable to self serve that w/ relying on a third party IMO.
- If that sounds reasonable, a few options we can consider:
  - Maxmind can be used as a short-term solution to cover various locales of our current concerns.
  - Geonames also comes with localized location strings, and we're already using it for weather. Depending on the dataset size, we can either load it in-memory in Merino or load it into a distributed store such as Redis and let Merino query against that instead  of a third party.

Let me know what you think.

ncloudioj (changes requested):

merino/providers/suggest/weather/backends/accuweather/backend.py
```diff
@@ -486,6 +486,26 @@ async def get_weather_report(self, weather_context: WeatherContext) -> WeatherRe
 
         return await self.get_weather_report_with_geolocation(weather_context)
 
+    async def get_localized_city_name(
+        self, location: AccuweatherLocation, weather_context: WeatherContext
+    ) -> str | None:
+        """Get city name based on specified language."""
+        geolocation = weather_context.geolocation
+        language = get_language(weather_context.languages)
```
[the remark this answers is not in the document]
tiftran: I don't think that's what we want. We have  "pt-BR",  "zh-CN" in the `MAXMIND_SUPPORTED_LOCALE` list. Though thinking about it we probs want `es-MX` to use `es`

merino/providers/suggest/weather/backends/accuweather/backend.py
```diff
@@ -486,6 +486,31 @@ async def get_weather_report(self, weather_context: WeatherContext) -> WeatherRe
...
+    @staticmethod
+    def get_localized_city_name(
+        location: AccuweatherLocation, weather_context: WeatherContext
+    ) -> str | None:
+        """Get city name based on specified language."""
+        geolocation = weather_context.geolocation
+        language = get_language(weather_context.languages)
+        normalized_lang = (
+            "es" if language.startswith("es") else "en" if language.startswith("en") else language
+        )
```
tiftran, line 498: does this make it slightly better?

ncloudioj: Yup, it's good for now. Big fan of if-chaining, eh? I'd put "en" in front of "es" though.

merino/middleware/geolocation.py
```diff
@@ -39,6 +41,7 @@ class Location(BaseModel):
     postal_code: Optional[str] = None
     key: Optional[str] = None
     coordinates: Optional[Coordinates] = None
+    city_names: Optional[dict[str, str]] = None
```
ncloudioj, line 44: `record.city.names` is of type `dict[str, str]`.
```suggestion
    city_names: dict[str, str] = {}
```

merino/providers/suggest/weather/backends/accuweather/backend.py
```diff
@@ -486,6 +486,31 @@ async def get_weather_report(self, weather_context: WeatherContext) -> WeatherRe
...
+
+        # ensure city was not overridden, if so city_names do not contain what we need
+        if (
+            geolocation
+            and geolocation.city_names
+            and location.localized_name == geolocation.city_names.get("en")
+            and normalized_lang in MAXMIND_SUPPORTED_LOCALE
+        ):
+            return geolocation.city_names.get(normalized_lang)
+        return None
```
ncloudioj, line 512: Since `geolocation` is a required field of `WeatherContext` and so is `city_names` to `geolocation`, I think this block can be reduced to:
```suggestion
        return geolocation.city_names.get(normalized_lang)
```

ncloudioj (approved):

merino/providers/suggest/weather/backends/accuweather/backend.py
tiftran: I think this condition is still needed `location.localized_name == geolocation.city_names.get("en")`, but yup, will simplify

PR #962, opened 2025-06-16
weather: add source param

tiftran:
## References

## Description
Just wanted to know if people had initial thoughts about adding source into the partner code param.
So merino already has a similarly named query param `sources` which I think is used for the manifest endpoint.
This may lead to some confusion... but I think desktop already made the uplift to use `source` as the param...

I also chose to rebuild the url right before constructing `WeatherReport`. My reasoning is that we probs don't want to cache with the forecast/current conditions info with the partner code source urls. This would mean doubling the number of forecast/current conditions to cache.

## PR Review Checklist


merino/providers/suggest/weather/backends/accuweather/backend.py
```diff
@@ -963,3 +973,19 @@ async def shutdown(self) -> None:
...
+        if partner_code_param and settings.accuweather.get("partner_code"):
+            parsed_url = urlparse(str(weather_model.url))
+            query_params = parse_qs(parsed_url.query)
+            query_params[partner_code_param] = [
+                val + append_str for val in query_params[partner_code_param]
+            ]
+
+            modified_query_params = urlencode(query_params, doseq=True)
+            modified_url = urlunparse(parsed_url._replace(query=modified_query_params))
+            return weather_model.model_copy(update={"url": modified_url})
```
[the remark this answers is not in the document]
tiftran: yeah this is convoluted... lol, i think there's a simpler more read-able way to do this.

merino/web/api_v1.py
```diff
@@ -85,6 +85,7 @@ async def suggest(
     country: Annotated[str | None, Query(max_length=2, min_length=2)] = None,
     region: Annotated[str | None, Query(max_length=QUERY_CHARACTER_MAX)] = None,
     city: Annotated[str | None, Query(max_length=QUERY_CHARACTER_MAX)] = None,
+    source: Annotated[str | None, Query(max_length=QUERY_CHARACTER_MAX)] = None,
```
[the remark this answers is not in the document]
tiftran: there should only be two valid source param values, so yes, I think we can add validation

merino/providers/suggest/weather/backends/accuweather/utils.py
```diff
@@ -187,3 +188,14 @@ def get_language(requested_languages: list[str]) -> str:
     return next(
         (language for language in requested_languages if language in VALID_LANGUAGES), "en-US"
     )
+
+
+def update_weather_url_with_suggest_partner_code(weather_model: CurrentConditions | Forecast):
+    """Update weather model urls to use suggest partner code."""
+    partner_code_param = settings.accuweather.get("url_param_partner_code")
```
ncloudioj, line 195: Calling `settings.accuweather.get()` repeatedly can be expensive, we can define constants for those to mitigate that. See `PARNTER_PARAM_ID` and `PARNTER_CODE`.

merino/web/api_v1.py
```diff
@@ -85,6 +86,7 @@ async def suggest(
     country: Annotated[str | None, Query(max_length=2, min_length=2)] = None,
     region: Annotated[str | None, Query(max_length=QUERY_CHARACTER_MAX)] = None,
     city: Annotated[str | None, Query(max_length=QUERY_CHARACTER_MAX)] = None,
+    source: Annotated[WEATHER_SOURCE_TYPE | None, Query(max_length=QUERY_CHARACTER_MAX)] = None,
```
ncloudioj, line 89: Setting this to `None` by default might be problematic as it will blow up `if source == URLBAR_REQUEST_SOURCE` elsewhere. We can make it a literal with three options ["urlbar", "newtab", "unknown"] and default to "unknown" if unspecified.

merino/providers/suggest/weather/backends/accuweather/utils.py
```diff
@@ -187,3 +188,14 @@ def get_language(requested_languages: list[str]) -> str:
...
+    if partner_code_param:
+        modified_url = URL(str(weather_model.url)).copy_set_param(
+            partner_code_param, settings.accuweather.get("url_param_partner_code_ffsuggest_value")
+        )
+        return weather_model.model_copy(update={"url": modified_url})
```
ncloudioj, line 200: We can simply do `weather_model.url = modified_url` to save a copy. Also, I'd recommend taking both `current_conditions` and `forecast` as two separate arguments and do the updating for both.

merino/configs/default.toml
```diff
@@ -361,6 +361,10 @@ url_forecasts_path = "/forecasts/v1/daily/1day/{location_key}.json"
 # Note that this is the name of the partner code parameter, not the partner code itself.
 url_param_partner_code = "partner"
 
+# MERINO_ACCUWEATHER__URL_PARAM_PARTNER_CODE_FFSUGGEST_VALUE
+# partner code for accuweather request from firefox suggest
+url_param_partner_code_ffsuggest_value = "web_mozilla_ffsuggest_adc"
```
ncloudioj, line 366: can you also define the other partner code here? Ideally, we should drop the envvar in cloudops-infra.

ncloudioj (approved):
Looks good, just a few minor nits.

merino/providers/suggest/weather/backends/accuweather/utils.py
```diff
@@ -187,3 +191,20 @@ def get_language(requested_languages: list[str]) -> str:
...
+        cc_modified_url = URL(str(current_conditions.url)).copy_set_param(
+            PARTNER_PARAM_ID, PARTNER_FFSUGGEST_CODE
+        )
+        f_modified_url = URL(str(forecast.url)).copy_set_param(
+            PARTNER_PARAM_ID, PARTNER_FFSUGGEST_CODE
+        )
+        current_conditions.url = HttpUrl(str(cc_modified_url))
+        forecast.url = HttpUrl(str(f_modified_url))
+        return current_conditions, forecast
+    return current_conditions, forecast
```
Herraj, line 210: suggestion(nit | non-blocking): we could flip this to do an early return e.g `if not PARTNER_PARAM_ID ... return ...` `else: .. do stuff`. But this is just nit picking 😅

merino/web/api_v1.py
```diff
@@ -71,6 +71,7 @@
 CLIENT_VARIANT_CHARACTER_MAX = settings.web.api.v1.client_variant_character_max
 HEADER_CHARACTER_MAX = settings.web.api.v1.header_character_max
 WEATHER_PROVIDER = settings.providers.accuweather.backend
+WEATHER_SOURCE_TYPE = Literal["urlbar", "newtab", "unknown"]
```
Herraj, line 74: nice!

Herraj (approved):
🚀

PR #970, opened 2025-06-24
HNT-780: fix sections layout

katerinachinnappan:
## References

JIRA: https://mozilla-hub.atlassian.net/browse/HNT-780

## Description
In some cases, seeing the same layout for 4 consecutive sections

In particular, we want to avoid to avoid the first 2 sections from having the same layout. There have been examples of Popular Today and the 2nd section both using the large card layout. 

* Fix this by sorting ranked sections by `receievedFeedRank` & then apply layout cycling.
* Apply the 6 card layout to all sections before pruning, the ensure we only return sections that can accommodate the biggest layout.

## PR Review Checklist

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/MC-1762)

### DIFF
--- a/merino/curated_recommendations/sections.py
+++ b/merino/curated_recommendations/sections.py
@@ -73,7 +73,7 @@ def map_section_item_to_recommendation(
 
 
 def map_corpus_section_to_section(
-    corpus_section: CorpusSection, rank: int, layout: Layout = layout_4_medium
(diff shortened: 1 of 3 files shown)


jpetto (approved):
the layout logic being sought here is a bit opaque to me - as is an inherent problem in having client layouts tightly coupled with API logic. but the code looks error free and i assume does what it needs to do in terms of telling the client how to layout the data.

PR #988, opened 2025-07-18
[GENAI-1485] add minimum interest vector value for ranking sections

mlplyler:
## References

## Description
The greedy ranker assumed a sparse interest vector where only topics/sections with clicks were present, and therefore, greedily moved all those sections to the top of the sections feed. Sections without a click would default to the order from the legacy ranking systems. The interest vector is now dense and all values appear with Laplace noise added. Without this change, the sections are ranked based on (CTR+noise). If CTR is zero, the ranking for that section is based on only noise. Without this change, sections are only ranked using the greedy ranker and there is no way to default to the legacy section Thompson sampling.

A couple notes: there are a few places in the code now that use the noise values in the local model definition, but they are hard-coded values. We should change these in the future to be programmatic.

## PR Review Checklist

- [x ] The PR title starts with the JIRA issue reference, format example `[DISCO-####]`, and has the same title (if applicable)
- [ skip] `[load test: (abort|skip|warn)]` keywords are applied to the last commit message (if applicable)
- [ N/A] [Documentation](https://github.com/mozilla-services/merino-py/tree/main/docs) has been updated (if applicable)
- [N/A ] [Functional and performance test](https://github.com/mozilla-services/merino-py/blob/main/docs/dev/testing.md) coverage has been expanded and maintained (if applicable)

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/MC-1780)

### DIFF
--- a/tests/unit/curated_recommendations/test_rankers.py
+++ b/tests/unit/curated_recommendations/test_rankers.py
@@ -648,26 +648,29 @@ class TestGreedyPersonalizedSectionRanker:
         # extract titles and build InferredInterests
         sec_titles = [sec for sec in sections]
         personal_sections = [sec_titles[i] for i in [4, 10, 13, 15]]
-        personal_interests = InferredInterests({k: i + 1 for i, k in enumerate(personal_sections)})
+        personal_interests = InferredInterests({k: i for i, k in enumerate(personal_sections)})
         # store original order of sections not in inferredInterests
         original_order = sorted(sections, key=lambda x: sections[x].receivedFeedRank)
-        original_order = [k for k in original_order if k not in personal_sections]
+        original_order = [
+            k
+            for k in original_order
+            if k not in personal_sections or personal_interests.root[k] < 0.0092
+        ]
         # rerank the sections
         reranked_sections = greedy_personalized_section_rank(
             sections=sections, personal_interests=personal_interests, epsilon=0.0
         )
-        print("original order", original_order)
-        print("personal_sections", personal_sections)
-        print(
-            "new order",
-            sorted(reranked_sections, key=lambda x: reranked_sections[x].receivedFeedRank),
-        )
+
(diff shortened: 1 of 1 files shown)


merino/curated_recommendations/rankers.py
```diff
@@ -170,7 +170,12 @@ def greedy_personalized_section_rank(
     ordered_sections = sorted(sections, key=lambda x: sections[x].receivedFeedRank)
 
     ## order of personal preferences
-    ptopics = [k for k, v in personal_interests.root.items() if isinstance(v, float)]
+    ### only keeps a value if above first coarse threshold. this
+    ### is because there is noise added to every value in client and
+    ## we do not want to rank on the noise
+    ptopics = [
+        k for k, v in personal_interests.root.items() if isinstance(v, float) and v >= 0.0092
```
rolf-moz, line 177: This constant is directly related to the noise_scale defined in fake_local_model.py. If you can define them both as constants in the same place I think that will improve maintenance of the values.

You could also make one a function of the other (thresh = 4*noise)

tests/unit/curated_recommendations/test_rankers.py
```diff
@@ -651,23 +651,26 @@ def test_expected_ranking(self):
         personal_interests = InferredInterests({k: i for i, k in enumerate(personal_sections)})
         # store original order of sections not in inferredInterests
         original_order = sorted(sections, key=lambda x: sections[x].receivedFeedRank)
-        original_order = [k for k in original_order if k not in personal_sections]
+        original_order = [
+            k
+            for k in original_order
+            if k not in personal_sections or personal_interests.root[k] < 0.0092
```
rolf-moz, line 657: use constant from above

rolf-moz (approved):
Thanks Mitchell!

PR #989, opened 2025-07-21
DISCO-3612 Match subdomains and TLDs when getting favicon urls

gruberb:
## References

JIRA: [DISCO-3612](https://mozilla-hub.atlassian.net/browse/DISCO-3612)

## Description
The HNT team is facing difficulties when trying to get the right favicons. We currently just match for `domain`, so `businessinsider.com` and `businessinsider.es` are both trying to get the favicon for the domain `businessinsider`.

This PR is changing the `domain_lookup_table` to a `url_lookup_table`. We add entries as `{subdomain}.{domain}.{tld}` instead of `{domain}`.

## PR Review Checklist

[DISCO-3612]: https://mozilla-hub.atlassian.net/browse/DISCO-3612?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/MC-1781)


merino/providers/manifest/provider.py
```diff
@@ -74,8 +74,9 @@ async def _fetch_data(self) -> None:
             match GetManifestResultCode(result_code):
                 case GetManifestResultCode.SUCCESS if data is not None:
                     self.manifest_data = data
-                    self.domain_lookup_table = {
-                        domain.domain: idx for idx, domain in enumerate(data.domains)
+                    self.url_lookup_table = {
```
mmiermans, line 77: I would continue to call it `domain_lookup_table`. The keys are (full) domains, not URLs.

merino/providers/manifest/provider.py
```diff
@@ -97,29 +98,53 @@ def get_manifest_data(self) -> ManifestData | None:
...
             Icon URL if found, None otherwise
         """
         try:
             url_str = str(url)
-            # Use tldextract to correctly handle all domain patterns
-            base_domain = tldextract.extract(url_str).domain
+            input_full_domain = self._extract_full_domain(url_str)
+
+            # Look for exact full domain match
+            idx = self.url_lookup_table.get(input_full_domain, -1)
```
mmiermans, line 114: Without adding all subdomains to the custom domains list, this would result in a lot of misses, wouldn't it? If we would need to add all subdomains/tld combinations, it may increase the number of custom domains significantly.

What do you think about trying these in order?
1. The full domain
2. The domain + tld, without subdomain
3. The second level domain (original behavior)
```suggestion
            input_full_domain = self._extract_full_domain(url_str)
            input_without_subdomain = f"{ext.domain}.{ext.suffix}"

            # Look for exact full domain match
            idx = self.url_lookup_table.get(input_full_domain) or \
                  self.url_lookup_table.get(input_without_subdomain) \
                  self.url_lookup_table.get(ext.domain, -1)
```
Another advantage of this is that we should be able to make this change without altering the custom domains list yet.

merino/providers/manifest/provider.py
```diff
@@ -97,29 +98,53 @@ def get_manifest_data(self) -> ManifestData | None:
...
+                    domain_for_metrics = tldextract.extract(url_str).domain
                     self.metrics_client.increment(
-                        "manifest.invalid_icon_url", tags={"domain": base_domain}
+                        "manifest.invalid_icon_url", tags={"domain": domain_for_metrics}
                     )
         except Exception as e:
             logger.warning(f"Error getting icon for URL {url}: {e}")
         return None
+
+    def _extract_full_domain(self, url: str) -> str:
```
mmiermans, line 129: Should we also call something like this method in `scripts/import_domains.sh`?

mmiermans (changes requested):
Thanks for taking this up! I proposed adding fallbacks. Without those, or changes to `custom_domains.py`, I believe we'd miss a lot of favicons.

merino/providers/manifest/provider.py
gruberb: I don't think that's needed? Maybe I am getting your intention (or mine?) wrong, but this has nothing to do with the custom domains list. This provider is separate from the Navigational Suggestions job. The end result from the Navigational Suggestion job is the `top_picks.json` file, which this provider will, in an interval, read and store in-memory. 

Every entry will have a full `url` (https://www.abc.com), even if the `CUSTOM_DOMAINS` and `CUSTOM_FAVICONS` don't. They are not meant to, because they just feed information into the job.

The `extract_full_domain` will generate `abc.com` out of `https://www.abc.com` and `go.abc.com` out of `https://go.abc.com`.

merino/providers/manifest/provider.py
gruberb: Hm, I thought about it. But this here is solely used by the Manifest provider. This PR has nothing to do with the Navigational Suggestions job. Maybe this is where the confusion comes from?

merino/providers/manifest/provider.py
mmiermans: You're right, I don't think this is needed.

I created https://github.com/mozilla-services/merino-py/pull/989, to ensure we continue to serve icons for domains such as sports.yahoo.com or edition.cnn.com. If you merge that into this branch, I think it's good-to-go.

mmiermans (approved):
Thanks for doing this work! Sorry I was a bit confused in my initial review. With the addition of subdomains, this is good-to-go, if we run the Airflow job shortly after it's live.

PR #992, opened 2025-07-24
GENAI-1559 Rescale priors for experiment based content based on relative experiment size

rolf-moz:
## References

JIRA: [GENAI-1559](https://mozilla-hub.atlassian.net/browse/GENAI-1559)

## Description
As part of the subtopics experiment, we are sending more targeted subtopic content to a small set of users (3% of US). This is leading to thompson sampling using unbalanced data to rank content in the experiment. By rescaling the interactions on the subtopics only, we hope to have less subtopic content in top stories.

This is implemented in a somewhat generic way so we can support multiple experiments at the same time with different sizes. Eventually we can have a GCS backend that dynamically computes the relative size.

image

## PR Review Checklist

- [x ] This PR conforms to the [Contribution Guidelines](https://github.com/mozilla-services/merino-py/blob/main/CONTRIBUTING.md)
- [x ] The PR title starts with the JIRA issue reference, format example `[DISCO-####]`, and has the same title (if applicable)

[GENAI-1559]: https://mozilla-hub.atlassian.net/browse/GENAI-1559?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/MC-1785)


merino/curated_recommendations/prior_backends/experiment_rescaler.py
```diff
@@ -0,0 +1,46 @@
...
+        return section_id in LOCALIZED_SECTION_TITLES[SurfaceId.NEW_TAB_EN_US]
+
+    def is_experiment_story(self, rec: CuratedRecommendation):
+        """Story is part of an experiment"""
+        return rec.corpusItemId in self.experiment_additional_content_id
+
+    def rescale(self, rec: CuratedRecommendation, opens: float, no_opens: float):
+        """Update open and non-open values based on whether item is unique to the experiment"""
+        if self.is_experiment_story(rec):
+            return opens / self.experiment_relative_size, no_opens / self.experiment_relative_size
```
mlplyler, line 44: i think this scaling makes sense. Where you able to dig up the old analysis/sims? I think we can reasonably sim this to double check, but maybe not worth the time if you can "educated-guess and check".

rolf-moz: No... I did test this with some print statements etc as a sanity check.

I'll look at your notebook and matt's work in a bit to see if any of it is applicable.

merino/curated_recommendations/prior_backends/experiment_rescaler.py
```diff
@@ -0,0 +1,46 @@
...
+        data.setdefault('target_region', "EN_US")
+        data.setdefault('experiment_relative_size', SUBSECTION_EXPERIMENT_PERCENT)
+        super().__init__(**data)
+
+    def model_post_init(self, __context) -> None:
+        """Complete setup for experiment by looking up items in/out of experiment"""
+        self.experiment_additional_content_id = set()
+        for section_name, section in self.cur_recs.items():
+            if not self.is_legacy_section(section_name):
+                self.experiment_additional_content_id.update([rec.corpusItemId for rec in section.recommendations])
```
mmiermans, line 31: Initially I was a bit confused by this class being a Pydantic model, and there being a post init step. Would this be able to replace `SubsectionsExperimentRescaler`?
- Add an `_is_in_experiment` on `CuratedRecommendation`, and initalize that when creating the sections.
- Pass in an `experiment_relative_size` float to `thompson_sampling`.
- `rescale` can be a plain function instead of a class.

mmiermans (commented):
I think this is doing what we'd expect. We should have some tests for this, and I'm happy to pair on those if you like.

I was initially thinking of scaling the prior, but scaling the engagement would work as well.

merino/curated_recommendations/sections.py
```diff
@@ -301,12 +318,15 @@ async def get_sections(
...
 
+    rescaler = SubsectionsExperimentRescaler() if subtopic_experiment_enabled else None
+
     # 5. Rank all corpus recommendations globally by engagement to build top_stories_section
     all_ranked_corpus_recommendations = thompson_sampling(
         all_corpus_recommendations,
         engagement_backend=engagement_backend,
         prior_backend=prior_backend,
         region=region,
+        rescaler=rescaler,
```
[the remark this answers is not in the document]
rolf-moz: It isn't required because the subtopics are already segregated, but I'll add it because priors are tuned to larger impression counts.

merino/curated_recommendations/protocol.py
```diff
@@ -140,6 +140,7 @@ class CuratedRecommendation(CorpusItem):
     """Extends CorpusItem with additional fields for a curated recommendation"""
 
     __typename: TypeName = TypeName.RECOMMENDATION
+    experiment_flags: Annotated[set[str] | None, Field(default_factory=set, exclude=True)] = None
```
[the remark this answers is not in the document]
rolf-moz: I thought so but I had trouble with the linting. I think I have to accept none so it an be defaulted to a set.

PR #999, opened 2025-07-30
[DISCO-3573] Add image_url to polygon suggest response

misaniwere:
## References

JIRA: [DISCO-3573](https://mozilla-hub.atlassian.net/browse/DISCO-3573)

## PR Review Checklist


merino/providers/suggest/finance/backends/protocol.py
```diff
@@ -40,3 +64,47 @@ async def get_ticker_summary(self, ticker: str) -> TickerSummary | None:  # prag
...
+        """
+        ...
+
+    async def upload_ticker_images(
+        self, tickers: list[str], prefix: str = "tickers"
+    ) -> dict[str, str]:
+        """Download and upload images for a list of ticker symbols.
+        Uses content hash to deduplicate and skips upload if destination blob already exists.
+        """
+        ...
```
Herraj, line 90: (nit): the name of the method and description is a little confusing. Judging from the name it would seem this method only uploads.

merino/providers/suggest/finance/backends/protocol.py
```diff
@@ -19,6 +29,20 @@ class TickerSummary(BaseModel):
...
     todays_change_perc: str
     query: str
+    image_url: str
+
+
+class FinanceManifest(BaseModel):
+    """Model for the manifest file content"""
+
+    tickers: Dict[str, HttpUrl]
+
```
Herraj, line 39: (non-blocking) question: wondering if this could just be a `dict` type instead of a class 🤔

merino/providers/suggest/finance/provider.py
```diff
@@ -38,11 +55,70 @@ def __init__(
...
+                    logger.error("Failed to fetch manifest data from finance backend.")
+                    return None
+        except FinanceBackendError as err:
+            logger.error("Failed to fetch manifest data from finance backend: %s", err)
+        except Exception as e:
+            logger.exception(f"Unexpected error in cron job 'fetch_manifest': {e}")
+
+        finally:
+            self.data_fetched_event.set()
+
```
Herraj, line 109: (non-blocking) question: Is there a case where we get `SUCESS` and `data` **is** `None`?

merino/providers/suggest/finance/backends/polygon/backend.py
```diff
@@ -79,6 +100,148 @@ async def fetch_ticker_snapshot(self, ticker: str) -> Any | None:
...
+                gcs_bucket_path=settings.image_gcs.gcs_bucket,
+                blob_name=GCS_BLOB_NAME,
+            )
+
+            success = await filemanager.upload_file(manifest)
+            if not success:
+                logger.error("polygon manifest upload failed.")
+        except Exception as e:
+            logger.error(f"Error building/uploading manifest: {e}")
+
```
Herraj, line 244: (non-blocking) suggestion: I notice that we we are intializing `PolygonFilemanager` twice and inside other methods in this module. Do you think it is possible to just initialize it once, maybe in this class' constructor and use it as a class variable instead of re-instantiating in the methods?

Herraj (commented):
Just did a first pass for source code. I'll do a second pass again later today and will review the tests too. Looking good so far!

merino/providers/suggest/finance/backends/protocol.py
misaniwere: it could but i think the class makes it more readable and more accurately describes its purpose

merino/providers/suggest/finance/provider.py
Herraj: follow up -- ok so maybe we can simplify the logic on line 94 to just check for `SUCCESS` code since data would always be `not None` when it is  a success.

merino/providers/suggest/finance/provider.py
```diff
@@ -85,6 +163,12 @@ def build_suggestion(self, data: TickerSummary) -> BaseSuggestion:
             custom_details=custom_details,
         )
 
+    def get_image_url_for_ticker(self, ticker: str) -> str:
+        """Return the GCS url from the manifest for a given ticker symbol"""
+        if self.manifest_data:
+            return str(self.manifest_data.tickers.get(ticker.upper(), ""))
+        return ""
```
Herraj, line 170: This would `None` as we talked about already!

tests/unit/providers/suggest/finance/test_provider.py
```diff
@@ -120,3 +127,73 @@ async def test_query_ticker_summary_not_returned(
...
+def test_get_image_url_for_ticker_not_found(provider: Provider):
+    """Test that get_image_url_for_ticker returns an empty string
+    when the ticker is not in the manifest.
+    """
+    provider.manifest_data = FinanceManifest(
+        tickers={"GOOGL": "https://cdn.example.com/googl.png"}
+    )
+
+    result = provider.get_image_url_for_ticker("AAPL")
+    assert result == ""
```
Herraj, line 161: I guess this will be set to assert on `None`.

merino/providers/suggest/finance/provider.py
```diff
@@ -66,6 +142,8 @@ async def query(self, srequest: SuggestionRequest) -> list[BaseSuggestion]:
                 if (ticker_summary := await self.backend.get_ticker_summary(ticker)) is None:
                     return []
                 else:
+                    image_url = self.get_image_url_for_ticker(ticker)
+                    ticker_summary.image_url = image_url
```
Herraj, line 146: (non-blocking): follow up -- It would be nice to have the `ticker_summary.image_url` set in the `build_ticker_summary` method in the main backend method. But I understand that would require moving the flow and method calling logic down to backend level instead of provider level, which I believe is non-trivial refactoring 😅

Herraj (changes requested):
I think one minor blocking change required

merino/providers/suggest/finance/provider.py
Herraj: correct me if Im wrong but this would need to return `None` because in the validate method above we set it via this method and it could still potentially set it as an ` "" ` 🤔

Herraj (approved):

PR #1009, opened 2025-08-05
[DISCO-3624] Optimize upload job to run based on timestamp

misaniwere:
## References

JIRA: [DISCO-3624](https://mozilla-hub.atlassian.net/browse/DISCO-3624)

## Description
This PR improves the polygon provider's resync logic by introducing accurate tracking of the last successful manifest upload time using the GCS timestamp. This prevents unnecessary manifest uploads during provider restarts or redeployments by using the actual last modified time of the manifest in GCS, ensuring we only re-upload after the resync_interval has passed.

Changes:
- Updated the `PolygonFilemanager.get_file()` method to return the Blob.updated timestamp.
- Used the GCS blob's `updated` metadata field as the source of truth for `last_upload_at`.
- `last_upload_at` is now initialized during ` _fetch_manifest()` using the fetched manifest’s blob timestamp.
- Upload jobs still update `last_upload_at` based on `time.time()` after a successful upload.

## PR Review Checklist

[DISCO-3624]: https://mozilla-hub.atlassian.net/browse/DISCO-3624?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/MC-1803)

### DIFF
--- a/merino/providers/suggest/finance/backends/polygon/backend.py
+++ b/merino/providers/suggest/finance/backends/polygon/backend.py
@@ -208,7 +208,9 @@ class PolygonBackend(FinanceBackend):
             logger.error(f"Failed to build FinanceManifest due to validation error: {e}")
             return FinanceManifest(tickers={})
 
-    async def fetch_manifest_data(self) -> tuple[GetManifestResultCode, FinanceManifest | None]:
+    async def fetch_manifest_data(
+        self,
+    ) -> tuple[GetManifestResultCode, FinanceManifest | None, float | None]:
         """Fetch manifest data from GCS through the filemanager."""
         return await self.filemanager.get_file()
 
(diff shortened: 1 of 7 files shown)


misaniwere:
> This looks good!
> 
> I'm wondering if we can do it without modifying the existing `get_file()` method 🤔. Have something like a class variable in the `filemanager` e.g. `last_uploaded_at` and have a separate get method for it. We could set that `last_uploaded_at` value every time we do an upload or call the `get_file()`. Basically separating it from the `get_file()` method. Thoughts?

hmm that would require an additional call to GCS and since we already have access to the blob in `get_file`, it seems more straightforward to use that opportunity to retrieve the timestamp. If we make a new method, we would basically need to fetch the blob/make the same call get_file is already doing

PR #1019, opened 2025-08-07
[DISCO-3631] ETF keywords to tickers mapping

Herraj:
## References

JIRA: [DISCO-3631](https://mozilla-hub.atlassian.net/browse/DISCO-3631)

## Description
- Adding mappings for keywords and their corresponding ETF tickers. 
- Updating `query()` logic to support tickers in list format (0 - 3 max) instead of a single ticker list.
- Refactoring backend methods.
- Refactoring util meathods.

### Request flow when a query hits the provider's `query()` method:

```text
Flow: Ticker Summary Building

[Provider]
   |
   |-- Extract tickers from query
   v
[Backend]
   |
   |-- Get snapshots for tickers
   |-- Return successful snapshots
   v
[Provider]
   |
   |-- For each snapshot:
   |     - Get image URL
   |     - Build ticker summary
   v
[Ticker Summaries]
   |
   |-- Build Finance suggestions from ticker summaries
   v
[Finance Suggestions]
```

### How to test locally
1. Make sure you have this config set in your `default.local.toml`:

```
[development.polygon]
api_key = <polygon-api-key-available-in-team-1password-vault>
```

2. Check out this branch and in your terminal run `make dev`.
3. In another terminal tab, run this cURL command:

```
# for 's&p 500' query:
curl --request GET \
  --url 'http://localhost:8000/api/v1/suggest?q=s%2526p%20500'

```
Try different keywords and stock/ETF tickers from the lists in this change set.

## PR Review Checklist

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/MC-1813)

[DISCO-3631]: https://mozilla-hub.atlassian.net/browse/DISCO-3631?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


merino/providers/suggest/finance/backends/polygon/etf_ticker_company_mapping.py
```diff
@@ -0,0 +1,319 @@
+"""Module containing JSON object with ETF ticker symbol to company name mapping for the top indices."""
+
+DOW_JONES_ETF_COMPANY = {
```
ncloudioj, line 3: Suggestion: we can use a list of strings or a [StrEnum](https://docs.python.org/3/library/enum.html#enum.StrEnum) to define those entries. Keeping those unused values in memory is kinda wasteful, we can keep them as comments instead.

```py
DOW_JONES_ETF: list[str] = [
    "DIA",  # SPDR Dow Jones Industrial Average ETF Trust
    "DJD",  # Invesco Dow Jones Industrial Average Dividend ETF
    ...
]

# Or via StrEnum, I personally prefer this: 

from enum import StrEnum, auto, unique

@unique
class DowJonesETF(StrEnum):
    DIA = auto()  # SPDR Dow Jones Industrial Average ETF Trust
    DJD = auto()  # Invesco Dow Jones Industrial Average Dividend ETF
    ...

DOW_JONES_ETF: list[str] = [entry.value.upper() for entry in DowJonesETF]
```

merino/providers/suggest/finance/backends/polygon/keyword_ticker_mapping.py
```diff
@@ -195,130 +205,130 @@
...
-    "stock indices": None,
-    "stock market": None,
-    "stock market futures": None,
-    "stock market news": None,
-    "stock market stocks": None,
-    "stock market today": None,
-    "stock prices": None,
-    "stocks": None,
-    "venture capital": None,
+    "100 nasdaq": NASDAQ_POPULAR_3_ETF,
```
ncloudioj, line 208: This mapping seems quite dense and might be hard to extend and maintain in the future. Alternatively, we can keep the mapping the other way around and use it to dynamically create this lookup table as an inverted index.

```py

ETF_TABLE: dict[set, list[str]] = {
    NASDAQ_POPULAR_3_ETF: [ ... ]  # all the keywords,
    ...
    DOW_JONES_POPULAR_3_ETF: [ ... ]  #  all the keywords,
}

KEYWORDS_TO_ETF: dict[str, set] = {}
for etf, keywords in ETF_TABLE.items():
    KEYWORDS_TO_ETF.update((keyword, etf) for keyword in keywords)
```

merino/providers/suggest/finance/backends/polygon/stock_ticker_company_mapping.py
```diff
@@ -2,7 +2,7 @@
 
 # Source of truth for ticker symbol and company name mapping.
 # NOTE: Should be treated as read-only and runtime.
-_TICKER_COMPANY = {
+ALL_STOCK_TICKER_COMPANY_MAPPING = {
```
ncloudioj, line 5: Ditto, we can save some memory by not storing those values if they're unused.

merino/providers/suggest/finance/backends/polygon/backend.py
```diff
@@ -223,7 +229,7 @@ async def build_and_upload_manifest_file(self) -> None:
         - Uploads the manifest JSON file to the GCS bucket.
         """
         try:
-            url_map = await self.bulk_download_and_upload_ticker_images(list(TICKERS))
+            url_map = await self.bulk_download_and_upload_ticker_images(list(ALL_STOCK_TICKERS))
```
ncloudioj, line 232: Be careful, this might take excessive amount of time to run and could prevent other Merino tasks from running. We need metrics to instrument this and optimize it if necessary.

merino/providers/suggest/finance/backends/polygon/utils.py
```diff
@@ -17,26 +20,27 @@
 
 logger = logging.getLogger(__name__)
 
-# Extracting just the ticker symbols into a separate set.
-TICKERS = frozenset(_TICKER_COMPANY.keys())
+# NOTE: Treat as read-only.
+# This is the comprehnsive list of all the tickers to company mapping
+# for all the tickers we support. Stock and ETF.
+STOCK_AND_ETF_TICKER_COMPANY_MAPPING = (
+    ALL_STOCK_TICKER_COMPANY_MAPPING | ALL_ETF_TICKER_COMPANY_MAPPING
```
ncloudioj, line 27: Are `ALL_STOCK_TICKER_COMPANY_MAPPING` and `ALL_ETF_TICKER_COMPANY_MAPPING` used elsewhere? If not, there will be many duplicate mappings stored as constants and it'll be wasteful as they're large maps.

ncloudioj (commented):
Looks good overall. Added a few suggestions for data modeling for your consideration. Let me know what you think.

merino/providers/suggest/finance/backends/polygon/etf_ticker_company_mapping.py
```diff
@@ -0,0 +1,319 @@
...
+
+NI225_POPULAR_3_ETF_COMPANY = {
+    "EWJ": "iShares MSCI Japan ETF",
+    "FLJP": "Franklin FTSE Japan ETF",
+    "DXJ": "WisdomTree Japan Hedged Equity Fund",
+}
+
+ALL_ETF_TICKER_COMPANY_MAPPING = (
+    DOW_JONES_ETF_COMPANY
+    | DOW_JONES_ETF_COMPANY
```
ncloudioj, line 302:
```suggestion
    DOW_JONES_ETF_COMPANY
```

merino/providers/suggest/finance/backends/polygon/utils.py
```diff
@@ -17,26 +20,27 @@
...
-        return None
+    # Check if the symbol provided is a supported ticker. Stock or ETF.
+    return symbol.upper() in ALL_TICKERS
 
 
 def lookup_ticker_company(ticker: str) -> str:
-    """Get the ticker company."""
-    return _TICKER_COMPANY[ticker.upper()]
+    """Get the ticker company for ticker symbol. Stock or ETF."""
+    return STOCK_AND_ETF_TICKER_COMPANY_MAPPING[ticker.upper()]
```
ncloudioj, line 43: As noted in another thread, let's consolidate `_is_valid_ticker()` and `lookup_ticker_company()` so that we can use a single map to rule them all.

```py
def get_ticker_company(ticker: str) -> str | None:
    return STOCK_AND_ETF_TICKER_COMPANY_MAPPING.get(ticker.upper())

# When it returns None, then the ticker is missing(invalid). We can short circuit at the calling side. 
```

merino/providers/suggest/finance/backends/polygon/utils.py
ncloudioj: With this and those frozensets, we are storing 4 copies of all the ticker/company pairs in memory. Let's try consolidating these maps to avoid the overhead.

ncloudioj (approved):
Thanks! r+wc.

Let's resolve those optimizations separately as followups.

PR #1053, opened 2025-09-03
[DISCO-3673] Yelp: Add exact match category keywords

tiftran:
## References

JIRA: [DISCO-3673](https://mozilla-hub.atlassian.net/browse/DISCO-3673)

## Description
Added logic to extract keywords. Keyword should grow later, once we get the finalized set.
This also enables the yelp provider.

## PR Review Checklist

[DISCO-3673]: https://mozilla-hub.atlassian.net/browse/DISCO-3673?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/MC-1848)


merino/providers/suggest/yelp/backends/yelp.py
```diff
@@ -86,23 +87,26 @@ async def store_in_cache(self, cache_key: str, data: dict) -> None:
...
+            if api_result is not None:
+                await self.store_in_cache(cache_key, api_result)
 
-        # Store in cache if successful
-        if api_result is not None:
-            await self.store_in_cache(cache_key, api_result)
+            return api_result
 
-        return api_result
+        return None
```
Herraj, line 109: (question): I forgot if the `debug` logs show in gcs?

(non-blocking)[suggestion]: Should we add a `try except` or a `response.raise_for_status()` in case the api call fails?

merino/providers/suggest/yelp/backends/yelp.py
```diff
@@ -149,6 +153,23 @@ def process_response(response: Any) -> dict | None:
...
+    def match_and_extract_search_term(search_term: str) -> str | None:
+        """Match and extract search term."""
+        stripped = search_term.casefold().strip()
+
+        # If it ends with a location keyword, strip it
+        for loc_kw in LOCATION_KEYWORDS:
+            if stripped.endswith(loc_kw):
+                stripped = stripped[: -len(loc_kw)].rstrip()
+                break  # only strip once
+
```
Herraj, line 166: (question): We are removing the `loc_kw` substring from the `search_term` on line 164 right?

merino/providers/suggest/yelp/backends/yelp.py
```diff
@@ -149,6 +153,23 @@ def process_response(response: Any) -> dict | None:
...
+        # Now check against categories
+        if any(stripped == kw.casefold() for kw in CATEGORIES):
+            return stripped
+
+        return None
+
```
Herraj, line 172: (question): For the line 168 logic, do we have to loop through the `CATEGORIES` keywords? Can we do something like
```
...stripped.casefold() in CATEGORIES...
```

Or am I understanding this bit incorrectly 😅

tiftran: oh yup

merino/providers/suggest/yelp/backends/yelp.py
Herraj: Yep that works!

merino/providers/suggest/yelp/backends/yelp.py
```diff
@@ -149,6 +153,23 @@ def process_response(response: Any) -> dict | None:
             logger.warning(f"Yelp business response json has incorrect shape: {response}")
             return None
 
+    @staticmethod
+    def match_and_extract_search_term(search_term: str) -> str | None:
+        """Match and extract search term."""
+        stripped = search_term.casefold().strip()
+
+        # If it ends with a location keyword, strip it
+        for loc_kw in LOCATION_KEYWORDS:
```
ncloudioj, line 162: Is it the case that we believe the location keywords will be much fewer than categories so we check them first? If so, that seems reasonable to me. My suggestion would be that we can combine all those location keywords into a regex so we'd only need to match the input against the regex once instead of looping thru all the location keywords. `regex.match()` will return the matched string and you can use that for suffix stripping.

PR #1093, opened 2025-10-01
[DISCO-3716] Fix bug and persist flight number set

misaniwere:
## References

JIRA: [DISCO-3716](https://mozilla-hub.atlassian.net/browse/DISCO-3716)

## Description
I just did a test run of the `fetch_schedules` job for flight numbers in Airflow. The good news is, it was able to fetch about 60k flight numbers. Bad news is, when it got to the last page, we hit an error and lost all that good data.

This PR fixes a bug in the  job where responses containing `"links": null` caused an `AttributeError` and aborted the task before flight numbers were persisted. It also ensures that partially collected flight numbers are always stored, even if the job encounters unexpected responses.

### What happened
- The code used data.get("links", {}).get("next", None) to follow pagination.
- When FlightAware returned `"links": null, data.get("links", {})` evaluated to `None` (because the key existed but was explicitly `null`), and calling `.get()` on `None` raised an exception.
- This happened after thousands of flights were already collected, but because the job crashed, none of them were stored in GCS.

### Fix
- Updated the job to safely handle the case where `links` is `None`
- Wrapped the CLI entrypoint with a finally block so that `store_flight_numbers(...)` is called even if the fetch loop fails, guaranteeing persistence of whatever flight numbers were collected.
- Added unit and integration tests.

## PR Review Checklist

### DIFF
--- a/merino/jobs/flightaware/__init__.py
+++ b/merino/jobs/flightaware/__init__.py
@@ -23,15 +23,22 @@ def fetch_and_store():
     logger.info("Starting flight schedules pipeline...")
 
     base_url = settings.flightaware.base_url
+    flight_number_set = set()
+    api_call_count = 0
 
     try:
         with httpx.Client(base_url=base_url) as client:
             flight_number_set, api_call_count = fetch_schedules.fetch_schedules(client)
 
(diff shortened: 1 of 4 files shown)


tiftran (approved):
such a subtle bug lol

PR #1094, opened 2025-10-01
[DISCO-3691] Add caching for ticker snapshots

Herraj:
## References

JIRA: [DISCO-3691](https://mozilla-hub.atlassian.net/browse/DISCO-3691)

## Description
**This PR adds the initial cache setup to support ticker snapshots in Redis. Included changes:**
- Added a Redis Lua script for bulk fetching ticker snapshots with their TTLs.
- Implemented cache key generation helper for ticker symbols.
- Added method to write snapshots into the cache with TTL.
- Added method to get snapshots with TTL and parse results back from cache.
- Added initial integration testcontainers tests for redis caching

**Notes:**
- The request-handling logic is not yet hooked up to the cache.
- Added TODOs in the code for error handling, metrics, and improved TTL warming strategy.

**Next steps:**
- Wire the cache logic into the main request flow.
- Address the TODOs for retries, metrics, and TTL handling.
- Expand integration tests to cover more cache scenarios.

## PR Review Checklist

[DISCO-3691]: https://mozilla-hub.atlassian.net/browse/DISCO-3691?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/merino/configs/default.toml
+++ b/merino/configs/default.toml
@@ -636,12 +636,12 @@ backend = "polygon"
 # MERINO_PROVIDERS__POLYGON__CACHE
 # The store used to finance data. Either `redis` or `none`.
 # If `redis`, the global Redis settings must be set. See redis.server.
-cache = "redis"
+cache = "none"
 
 # MERINO_PROVIDERS__POLYGON__CACHE_DB
 # The db within the redis instance that the clients should connect to.
 # Default is `0`. See redis.py.
-cache_db = 1
+cache_db = 0
 
 # MERINO_PROVIDERS__POLYGON__ENABLED_BY_DEFAULT
 # Whether this provider is enabled by default.
--- a/merino/configs/default.toml
+++ b/merino/configs/default.toml
@@ -636,7 +636,12 @@ backend = "polygon"
 # MERINO_PROVIDERS__POLYGON__CACHE
 # The store used to finance data. Either `redis` or `none`.
 # If `redis`, the global Redis settings must be set. See redis.server.
-cache = "none"
+cache = "redis"
+
+# MERINO_PROVIDERS__POLYGON__CACHE_DB
+# The db within the redis instance that the clients should connect to.
+# Default is `0`. See redis.py.
+cache_db = 1
 
 # MERINO_PROVIDERS__POLYGON__ENABLED_BY_DEFAULT
 # Whether this provider is enabled by default.
@@ -676,13 +681,10 @@ circuit_breaker_failure_threshold = 10
 circuit_breaker_recover_timeout_sec = 30
 
 [default.providers.polygon.cache_ttls]
-# Cache TTLs for finance data.
+# Cache TTLs for Polygon ticker snapshots.
 
-# MERINO_PROVIDERS__POLYGON__CACHE_TTLS__STOCK_TICKER_TTL_SEC
-stock_ticker_ttl_sec = 900 # 15 minutes
-
-# MERINO_PROVIDERS__POLYGON__CACHE_TTLS__INDEX_TICKER_TTL_SEC
-index_ticker_ttl_sec = 900 # 15 minutes
+# MERINO_PROVIDERS__POLYGON__CACHE_TTLS__TICKER_TTL_SEC
+ticker_ttl_sec = 300 # 5 minutes
 
 [default.polygon]
 # MERINO_POLYGON__API_KEY
--- a/merino/providers/suggest/finance/backends/polygon/backend.py
+++ b/merino/providers/suggest/finance/backends/polygon/backend.py
@@ -1,9 +1,9 @@
 """A wrapper for Polygon API interactions."""
 
+import itertools
 import hashlib
 import logging
 import aiodogstatsd
-from datetime import timedelta
 from httpx import AsyncClient, Response, HTTPStatusError
 import orjson
 from pydantic import HttpUrl, ValidationError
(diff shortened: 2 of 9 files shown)


tests/unit/providers/suggest/finance/backends/test_polygon.py
```diff
@@ -67,6 +68,8 @@ def fixture_polygon_parameters(
         "url_single_ticker_snapshot": URL_SINGLE_TICKER_SNAPSHOT,
         "url_single_ticker_overview": URL_SINGLE_TICKER_OVERVIEW,
         "gcs_uploader": mock_gcs_uploader,
+        "cache": mocker.MagicMock(),
+        "ticker_ttl_sec": TICKER_TTL_SEC,
     }
```
Herraj, line 73: Fixing broken unit test fixtures.

merino/configs/default.toml
```diff
@@ -636,7 +636,12 @@ backend = "polygon"
 # MERINO_PROVIDERS__POLYGON__CACHE
 # The store used to finance data. Either `redis` or `none`.
 # If `redis`, the global Redis settings must be set. See redis.server.
-cache = "none"
+cache = "redis"
+
+# MERINO_PROVIDERS__POLYGON__CACHE_DB
+# The db within the redis instance that the clients should connect to.
+# Default is `0`. See redis.py.
+cache_db = 1
```
Herraj, line 644: Will confirm with the team that no other provider is using this db. Accuweather uses the default (`0`).

ncloudioj: Let's continue to use the default database (i.e "db 0") for all cache data across the providers. While there are some benefits with that, there is one major downside are worth noting:
- Each Redis client (including its connection pools) can only open one DB, meaning that Merino would need to maintain multiple clients if we were to use mutiple DBs, which would prevent us from reusing the same client & connections pool to the cache. High connection counts are a known risk that we'd like to avoid. 

I recommended us to use separate dbs for each cache dataset, but after a second thought, doing so would cause more trouble than benefits, sorry about that!

merino/configs/default.toml
```diff
@@ -636,7 +636,12 @@ backend = "polygon"
 # MERINO_PROVIDERS__POLYGON__CACHE
 # The store used to finance data. Either `redis` or `none`.
 # If `redis`, the global Redis settings must be set. See redis.server.
-cache = "none"
+cache = "redis"
```
ncloudioj, line 639: Does it make sense to specify this in production.toml instead? So we can still run Merino without running Redis.

merino/providers/suggest/finance/backends/polygon/backend.py
```diff
@@ -36,11 +40,35 @@
...
+#   - The script expects each key to contain a JSON-serialized ticker snapshot string.
+#   - For every existing key, it returns the snapshot value followed by its TTL (in seconds).
+#   - Keys that are missing or expired are skipped and not included in the result.
+LUA_SCRIPT_CACHE_BULK_FETCH_TICKERS: str = """
+local result = {}
+for _, k in ipairs(KEYS) do
+  local snapshot = redis.call('GET', k)
+  if snapshot ~= false then
+    local ttl = redis.call('TTL', k)
+    if ttl ~= -2 then
```
ncloudioj, line 55: Reading the other comment for AccuWeather TTL fetch:

>  If the forecast and current_conditions TTLs are a non-positive value (-1 or -2),

Maybe we can do `if ttl > 0 then`?

merino/providers/suggest/finance/backends/polygon/backend.py
```diff
@@ -36,11 +40,35 @@
...
+#
+# Note:
+#   - The script expects each key to contain a JSON-serialized ticker snapshot string.
+#   - For every existing key, it returns the snapshot value followed by its TTL (in seconds).
+#   - Keys that are missing or expired are skipped and not included in the result.
+LUA_SCRIPT_CACHE_BULK_FETCH_TICKERS: str = """
+local result = {}
+for _, k in ipairs(KEYS) do
+  local snapshot = redis.call('GET', k)
+  if snapshot ~= false then
```
ncloudioj, line 53:
```suggestion
  if snapshot then
```

merino/providers/suggest/finance/backends/polygon/backend.py
```diff
@@ -258,8 +319,69 @@ async def build_and_upload_manifest_file(self) -> None:
...
+    async def store_snapshots_in_cache(self, snapshots: list[TickerSnapshot]) -> None:
+        """Store a list of ticker snapshots in the cache.
+
+        Each snapshot is serialized to JSON and written under a generated cache key
+        for its ticker symbol, with a configured TTL applied.
+        """
+        if len(snapshots) == 0:
+            return
+
+        for snapshot in snapshots:
```
ncloudioj, line 331: How many snapshots do we expect to write on average? If it's more than two digits, we can do `cache.mset` to save the network overhead.

merino/providers/suggest/finance/backends/polygon/backend.py
```diff
@@ -258,8 +319,69 @@ async def build_and_upload_manifest_file(self) -> None:
...
+        # [snapshot_1, ttl_1, snapshot_2, ttl_2, ...]
+        if (len(cached_data) % 2) != 0:
+            return []
+
+        result: list[Optional[Tuple[TickerSnapshot, int]]] = []
+
+        snapshots_bytes = cached_data[0::2]  # values at even indexes
+        ttls_bytes = cached_data[1::2]  # values at odd indexes
+
+        for snapshot, ttl in zip(snapshots_bytes, ttls_bytes):
```
ncloudioj, line 362: itertools is your friend.

```py
for snapshot, ttl in itertools.batched(result, n=2):
```

merino/providers/suggest/manager.py
```diff
@@ -201,6 +191,20 @@ def _create_provider(provider_id: str, setting: Settings) -> BaseProvider:
                 enabled_by_default=setting.enabled_by_default,
             )
         case ProviderType.POLYGON:
+            cache = (
+                RedisAdapter(
+                    *create_redis_clients(
```
ncloudioj, line 196: Non-issue; just a note: We could possibly share the same client across all the providers to keep the connection numbers low for each Merino pod.

ncloudioj (approved):
Looks good, just a few minor nits.

PR #1096, opened 2025-10-02
[DISCO-3736] Calculate flight status and prioritization

misaniwere:
## References

JIRA: [DISCO-3736](https://mozilla-hub.atlassian.net/browse/DISCO-3736)

## Description
This PR uses a flight's lifecycle time fields to calculate it's current high level status. It also populates the `time_left_minutes` and `delayed` fields.

### Changes
- Flight status calculation (derive_flight_status)
   - Added stable backend-calculated statuses (Scheduled, Delayed, En Route, Arrived, Cancelled, Unknown).
   - Added `is_delayed` helper with a 15-minute threshold for flagging delays.
- Time left calculation (calculate_time_left): computes minutes remaining until arrival.
- Prioritization adjustments (pick_best_flights)
   - Arrived and cancelled flights are only included if they occurred within 1 hour (was previously 2 hours).

### How to QA
1. Enable the provider:
   * Remove `"flightaware"` from the `disabled_providers` list in `default.toml`.
2. Configure the API key:
   * Replace the `"api_key"` field in `default.toml` with a real FlightAware API key.
3. Find a flight:
   * Use Flightaware's [flight finder ](https://www.flightaware.com/live/findflight/) to look up flights that have arrived or are en route with future instances (within the next 24 hours)
   * Add your chosen flight number to the `temp_cache` in the provider.
4. Call the endpoint:
   * Hit the Merino suggest endpoint with the flight code as the query and `"flightaware"` as the provider.
5. Verify behavior:
   * The query method should only respond to valid flight queries that exist in the temp cache and match a valid flight code pattern
   * The prioritization method should return flights in the right order and with the right status.
   * The `time_left_minutes` should match what's on the flightaware live page.
   * The `url` should point to a valid flightaware page for the correct flight.
   * Bonus points if you're able to find a delayed flight to test

## PR Review Checklist

### DIFF
--- a/merino/providers/suggest/flightaware/backends/utils.py
+++ b/merino/providers/suggest/flightaware/backends/utils.py
@@ -3,10 +3,10 @@
 import datetime
 import logging
 import re
-from typing import Any
 
 from pydantic import HttpUrl
 from merino.providers.suggest.flightaware.backends.protocol import (
+    AirlineDetails,
     AirportDetails,
     FlightScheduleSegment,
     FlightStatus,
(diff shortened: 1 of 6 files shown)


merino/providers/suggest/flightaware/backends/utils.py
```diff
@@ -65,12 +66,21 @@ def derive_flight_status(flight: dict) -> FlightStatus:
...
+    - If timezone is None, return a UTC datetime (default).
+    - If timezone is provided, return the datetime converted to that timezone.
+    """
     if not timestamp:
         return None
     try:
-        return datetime.datetime.fromisoformat(timestamp.replace("Z", "+00:00"))
+        dt = datetime.datetime.fromisoformat(timestamp.replace("Z", "+00:00"))
+        if timezone:
+            return dt.astimezone(ZoneInfo(timezone))
```
ncloudioj, line 82: `ZoneInfo(tz)` will raise an `ZoneInfoNotFoundError` for unrecognized time zones. Since `timezone` is from outside, let's be more cautious here.

ncloudioj (approved):

PR #1150, opened 2025-11-07
fix: Look for the elastic search credentials in all the places

jrconlin:
Issue DISCO-3806

## References

## Description

The `elastic.py` code is designed to be shared between suggest and jobs, meaning that there are a number of locations where the credentials might live. We need to search those in order to find the best value, since they may not always be defined in settings files. 

## PR Review Checklist

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/MC-1944)

### DIFF
--- a/merino/providers/suggest/sports/backends/sportsdata/common/elastic.py
+++ b/merino/providers/suggest/sports/backends/sportsdata/common/elastic.py
@@ -3,12 +3,10 @@
 import copy
 import json
 import logging
-import os
 from abc import abstractmethod, ABC
 from datetime import datetime, timezone
 from typing import Any, Final
 
-from dynaconf.base import LazySettings
 from elasticsearch import (
     AsyncElasticsearch,
     BadRequestError,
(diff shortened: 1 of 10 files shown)


merino/providers/suggest/sports/backends/sportsdata/common/elastic.py
```diff
@@ -179,11 +181,51 @@ def __init__(
...
+        Remember, this code is shared between suggest and jobs, so the credentials
+        can hide in lots of places.
+        """
+        if not lsettings:
+            lsettings = settings
+        if not dsn:
+            dsn = (
+                # Use the environment vars first, since the toml file may have
+                # "http://localhost:9200"
+                os.environ.get("MERINO_ELASTICSEARCH_DSN")
```
ncloudioj: I don't think it's a good idea to load envvars directly since the dynaconf settings would do that already for Merino and we should treat those as the single source of truth for configurations. If "http://localhost:9200" gets loaded in production, that means it's not properly set and we should fix that instead.

The configurations for sports seem funky and I guess one potential reason might be we're passing a `LazySettings` argument instead of using the original settings directly. Dynaconf does a lot of things dynamically, I think the safer way is to load the relevant configurations and pass their values into the consumer instead of the whole `LazySettings` object. That's what we've been doing throughout the code base.

```py
# Load secrets directly from settings
dsn = settings.providers.sports.es.dsn or settings.providers.wikipedia.es_url
api_key = settings.providers.sports.es.api_key or settings.providers.wikipedia.es_api_key

store = ElasticDataStore(dsn, api_key)
```

ncloudioj (changes requested):

jrconlin:
Ok, refactored a bunch of this to use explicit setting values where possible and to use passed `LazySettings` less. I'm still passing them down for eventual use by the `Sport` subclasses, but will look at ways to make those use a more standard set of settings that can be derived higher.

ncloudioj (approved):
Looks good, thanks!

PR #1158, opened 2025-11-11
Fix/disco 3806 es startup

jrconlin:
## References

## Description
* Correct logging to be more accurate
* Remove missed kickstart code from `query`
* Use a class to fetch out the Elasticsearch Credentials from potential settings candidates.

Issue DISCO-3806

## PR Review Checklist

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/MC-1952)

### DIFF
--- a/merino/providers/suggest/sports/backends/sportsdata/common/elastic.py
+++ b/merino/providers/suggest/sports/backends/sportsdata/common/elastic.py
@@ -229,6 +229,10 @@ class ElasticCredentials:
 
     def validate(self) -> bool:
         """Return if the credentials were located and valid."""
+        logger = logging.getLogger(__name__)
+        api_key = self.api_key or "None"
+        dsn = self.dsn or "None"
+        logger.info(f"{LOGGING_TAG} Elastic API key: [{api_key[:4]}...] DSN:[{dsn[:10]}]")
         return (self.api_key is not None and len(self.api_key) > 0) and (
             self.dsn is not None and len(self.dsn) > 0
         )


merino/providers/suggest/sports/backends/sportsdata/common/elastic.py
```diff
@@ -163,38 +164,121 @@ def get_index_settings(dsn: str = settings.providers.sports.es.dsn) -> dict[str,
...
-        self.client = AsyncElasticsearch(dsn, api_key=api_key)
+        """Create a core instance of elastic search.
+        NOTE: You need to call `.startup()` BEFORE performing any
+        functions with ElasticDataStore. `startup()` may raise an exception
+        if the credentials are not present or if there is any other connection
+        issue, so it needs to be protected by exception handling or isolated
+        into provider exclusive function.
+        """
+        logger = logging.getLogger(__name__)
+        if not credentials.dsn:
```
jrconlin: Yes, this should just call `credentials.validate()` but I wanted to identify the error.

merino/jobs/sportsdata_jobs/__init__.py
Herraj: Should we undo this or also catch `SportsDataError`? I added the generic `Exception` type as a stop gap yesterday 😅

merino/providers/suggest/sports/backends/sportsdata/common/elastic.py
```diff
@@ -163,41 +166,130 @@ def get_index_settings(dsn: str = settings.providers.sports.es.dsn) -> dict[str,
...
+            except AttributeError:
+                # fail to next candidate
+                pass
+        if not self.api_key:
+            try:
+                self.api_key = settings.wikipedia_indexer.es_api_key
+            except AttributeError:
+                # remember to call `.validate()` to ensure valid
+                pass
+
```
Herraj: Thoughts on collapsing the checks on line `200` and `206` into something like:

```
if not self.dsn:
  ...
  self.dsn = settings.providers.wikipedia.es_url or settings.wikipedia_indexer.es_url
  ...
```

merino/providers/suggest/sports/backends/sportsdata/common/elastic.py
```diff
@@ -163,41 +166,130 @@ def get_index_settings(dsn: str = settings.providers.sports.es.dsn) -> dict[str,
...
+        issue, so it needs to be protected by exception handling or isolated
+        into provider exclusive function.
+        """
+        logger = logging.getLogger(__name__)
+        if not credentials.dsn:
+            logger.error(f"{LOGGING_TAG} Missing DSN")
+        if not credentials.api_key:
+            logger.error(f"{LOGGING_TAG} Missing API Key")
+        self.credentials = credentials
+        self.client = None
```
Herraj: Regarding the comment on line `252`, do you think it would be useful to add a class boolean flag that allows methods to be called only if it is a successful startup? 

I'm assuming it'll be quiet a bit of refactoring.

Herraj (commented):
LGTM!

merino/providers/suggest/sports/backends/sportsdata/common/elastic.py
jrconlin: Thinking about this, i'm not sure that would work.

The problem is that we're trying to handle `AttributeError` that can be generated by `LazySettings` as it walks through it's constructions. For `providers`, we have `settings.providers.᠁`, but for `jobs` we don't have that, so when this gets called by the `sportsdata_jobs` the `AttirbuteError` would be raised on the first condition, before it had a chance to look at `settings.wikipedia_indexer.᠁`, right? I possibly could use something like `settings.get('providers','wikipedia_indexer')` but that complicate the rest of the pathing.

ncloudioj (approved):

PR #1167, opened 2025-11-13
Fix/disco 3806 next

jrconlin:
## References

## Description
There was a potential issue where an "scheduled" game might overlap the current game. For "all" sports, that's more likely to be a real thing. Remove the safeguards.

Issue DISCO-3806

## PR Review Checklist

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/MC-1961)

### DIFF
--- a/merino/providers/suggest/sports/backends/sportsdata/common/elastic.py
+++ b/merino/providers/suggest/sports/backends/sportsdata/common/elastic.py
@@ -662,7 +662,7 @@ class SportsDataStore(ElasticDataStore):
                 # If only show the next upcoming game.
                 if status.is_scheduled():
                     # if there is no "next" game, or if the "next" game is later than this one,
-                    # display the most immediate upcoming event. (if there's )
+                    # display the most immediate upcoming event.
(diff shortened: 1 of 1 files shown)


tests/unit/providers/suggest/sports/backends/common/test_elastic.py
```diff
@@ -214,8 +214,8 @@ async def test_search_event_hits(
                 "status": "InProgress",
             },
             "next": {
-                "date": 1758801600,
-                "es_score": 0.8,
+                "date": 1758715200,
```
annasob: nit this should use datetime.date or some equivalent to auto calculate yesterday today and future dates

annasob (approved):

tests/unit/providers/suggest/sports/backends/common/test_elastic.py
jrconlin: I think because we're using `@freezegun.freeze_time("2025-09-22T12:00:00Z")` at the top, we can presume that the time is a static value, so looking at the returned result should be fine. We already set the value as `now + {2,3} * 86400` when we set the values.

PR #1238, opened 2026-01-21
[HNT-1161] Add hourly forecasts

Herraj:
## References

JIRA: [HNT-1161](https://mozilla-hub.atlassian.net/browse/HNT-1161)

## Description
This change adds support for serving hourly forecasts via Accuweather. 
_NOTE_: The change set is big but it is mostly because of tests and fixtures. The business logic part is not that big. I've added comments below to help you with the review. 

I will also be adding additional integration tests in a follow up change.

### High level logic flow

- We will try to read from cache all of the values (including hourly forecasts) even if hourly forecasts or not requested.
- If hourly forecasts are not requested --> return the weather report as usual.
- If hourly forecasts are requested _and_ found in cache --> return weather report with cached hourly forecast.
- if hourly forecasts are requested but _not_ found in cache --> make an api call for hourly forecasts and return weather report with all the values.

### Testing
Make sure you have all the configs set in your `deafult.local.toml`.

Pull this branch locally and run:
```
cd dev/ && docker compose up

// in a new terminal tab
make dev

// once the service is up and running, try these commands

// this should return current conditions, forecast and 5 hourly forecasts
curl localhost:8000/api/v1/suggest?providers=accuweather&q&request_type=weather&forecast_hours=5

// this should return just current conditions and forecast (existing behaviour)
curl localhost:8000/api/v1/suggest?providers=accuweather&q&request_type=weather


// request param validation error 
curl localhost:8000/api/v1/suggest?providers=accuweather&q&request_type=weather&forecast_hours=4

curl localhost:8000/api/v1/suggest?providers=accuweather&q&request_type=weather&forecast_hours=6
```

## PR Review Checklist

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/MC-2031)

[HNT-1161]: https://mozilla-hub.atlassian.net/browse/HNT-1161?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/merino/providers/suggest/weather/backends/accuweather/backend.py
+++ b/merino/providers/suggest/weather/backends/accuweather/backend.py
@@ -51,10 +51,9 @@ from merino.providers.suggest.weather.backends.accuweather.errors import (
 
 logger = logging.getLogger(__name__)
 
-# TODO @herraj update comment
 
-# The Lua script to fetch the location key, current condition, forecast, and a TTL for
-# a given country/region/city.
+# The Lua script to fetch the location key, current condition, forecast, hourly_forecasts
+#  and a TTL for a given country/region/city.
 #
 # Note:
 #   - The script expects a JSON serialized string as the value of the location key,
@@ -63,12 +62,13 @@ logger = logging.getLogger(__name__)
 #   - The cache key for the country/region/city should be provided through `KEYS[1]`
 #   - The cache key templates for current conditions and forecast should be provided
 #     through `ARGV[1]` and `ARGV[2]`
+#   - The cache key template for hourly forecast through `ARGV[3]`
 #   - The placeholder for location key (i.e. `self.url_location_key_placeholder`)
-#     is passed via `ARGV[3]`
+#     is passed via `ARGV[4]`
 #   - If the location key is present in the cache, it uses the key to fetch the current
-#     conditions and forecast for that key in the cache. It returns a 4-element array
-#     `[location_key, current_condition, forecast, ttl]`. The `current_condition` and `forecast` are `nil`
-#     if they are not present in the cache
+#     conditions and forecast for that key in the cache. It returns a 5-element array
+#     `[location_key, current_condition, forecast, hourly_forecasts, ttl]`
+#     The `current_condition`, `forecast` and `hourly_forecasts` are `nil` if they are not present in the cache
 #   - If the location key is missing, it will return an empty array
 #   - If the current_conditions and forecast TTLs are a non-positive value (-1 or -2),
 #     it will return ttl as false, which is translated to None type in app code.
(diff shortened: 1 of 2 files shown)


merino/web/api_v1.py
```diff
@@ -97,6 +97,7 @@ async def suggest(
     client_variants: str | None = Query(default=None, max_length=CLIENT_VARIANT_CHARACTER_MAX),
     sources: tuple[dict[str, BaseProvider], list[BaseProvider]] = Depends(get_suggest_providers),
     request_type: Annotated[str | None, Query(pattern="^(location|weather)$")] = None,
+    forecast_hours: Annotated[int | None, Query(ge=5, le=5)] = None,
```
Herraj: This is a hacky way to enforce only accepted value i.e `5`. This is probably okay for now since the client side is aware of this. We might relax this constraint in the future.

tiftran: I tried `Annotated[Literal(5) | None, Query()]` thinking it would be nicer, but then it started complaining the param was a str not an int. sooo i don't have anything better either....

merino/configs/default.toml
```diff
@@ -338,6 +338,10 @@ current_condition_ttl_sec = 1800 # 1/2 hr
 # MERINO_PROVIDERS__ACCUWEATHER__CACHE_TTLS__FORECAST_TTL_SEC
 forecast_ttl_sec = 3600 # 1 hr
 
+# MERINO_PROVIDERS__ACCUWEATHER__CACHE_TTLS__HOURLY_FORECAST_TTL_SEC
+# Cache TTL for hourly forecasts (1 hour).
+hourly_forecast_ttl_sec = 3600
+
```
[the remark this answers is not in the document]
jrconlin: Is this true? 
AIUI: this value is stored as `AccuweatherBackend.cached_hourly_forecast_ttl_sec`, which is later used in `AccuweatherBackend.get_hourly_forecast()` in the `self.request_upstream()` call to specify the cache period. Is that not correct?

merino/providers/suggest/weather/backends/accuweather/backend.py
```diff
@@ -707,13 +751,26 @@ async def as_awaitable(val: Any) -> Any:
...
+                return WeatherReport(
+                    city_name=city_name if city_name else location.localized_name,
+                    region_code=admin_area,
+                    current_conditions=current_conditions,
+                    forecast=forecast,
+                    hourly_forecasts=None,
+                    ttl=ttl,
+                )
+
+            if has_requested_hourly_forecasts and hourly_forecasts:
```
jrconlin: I think we can drop the first condition here. We're already at the point where we know `has_requested_hourly_forecasts` must be True, and it's confusing to read this with that additional condition present. (I'd also argue renaming the variable as `cached_hourly_forecasts`.)

merino/web/api_v1.py
jrconlin: We will never want more or less hours than 5? I can see limiting to something like 24 or 48 (From what I understand anything beyond 60 gets iffy according to most weather models.) but saying "The Number Shall Be 5 and 5 Can Only Be The Number" seems less like a setting and more like a boolean.

merino/web/api_v1.py
```diff
@@ -138,6 +139,8 @@ async def suggest(
         "location" or "weather" string. For "location" it will get location completion
         suggestion. For "weather" it will return weather suggestions. If omitted, it defaults
         to weather suggestions.
+    - `forecast_hours`: [Optional] Number of hourly forecast hours to return when requesting weather.
+        Only excepted value is 5 for now.
```
jrconlin:
```suggestion
        Only accepted value is 5 for now.
```

tests/unit/providers/suggest/weather/backends/test_accuweather.py
```diff
@@ -2026,6 +2191,66 @@ async def test_get_forecast_error(accuweather: AccuweatherBackend, language: str
...
+    )
+
+    result: list[HourlyForecast] | None = await accuweather.get_hourly_forecast(
+        location_key, language, forecast_hours
+    )
+
+    # Assertions
+    assert result is not None
+    assert isinstance(result, list)
+    assert len(result) == 5  # First 5 forecasts are processed (DEFAULT_FORECAST_HOURS)
```
jrconlin: Shouldn't we match against `forecast_hours` we specified earlier?

tests/unit/providers/suggest/weather/backends/test_utils.py
```diff
@@ -0,0 +1,191 @@
...
+            {
+                "date_time": forecast["DateTime"],
+                "epoch_date_time": forecast["EpochDateTime"],
+                "temperature_unit": forecast["Temperature"]["Unit"].lower(),
+                "temperature_value": forecast["Temperature"]["Value"],
+                "icon_id": forecast["WeatherIcon"],
+                "url": add_partner_code(forecast["Link"], PARTNER_PARAM_ID, PARTNER_CODE_NEWTAB),
+            }
+            for forecast in accuweather_hourly_forecasts_response[
+                :5
```
jrconlin: I get that right now, 5 is the magic number, but it feels like we should consolidate these to the defined default from `merino.providers.suggest.backends.accuweather.utils.DEFAULT_FORECAST_HOURS` so that we're not chasing through code if it ever changes.

PR #1259, opened 2026-01-30
[DISCO-3809] Establish a common Elasticsearch adapter for providers and jobs

misaniwere:
## References

JIRA: [DISCO-3809](https://mozilla-hub.atlassian.net/browse/DISCO-3809)

## Description
This PR introduces a common Elasticsearch adapter. As more components rely on Elasticsearch, we want to standardize how clients are initialized and how common operations (index lifecycle, bulk, search, etc.) are handled, while still leaving room for future flexibility.

### Direction
We are keeping the Elasticsearch adapter configurable, allowing it to be initialized with an arbitrary DSN and API key. If Merino providers and jobs are using the same Elasticsearch cluster, the intended direction is that they should also use a shared credential set (DSN + API key), rather than defining provider-specific credentials. We may need to distinguish read vs write access (e.g., separate API keys). We can confirm if we do, but the goal is to keep the number of credentials small and well-defined.

As a follow-up, we plan to:
- Consolidate Elasticsearch credentials into one or two shared config variables (e.g. read and write).
- Update providers and jobs to consume these shared credentials by default.
- Update Airflow jobs to use the unified credentials.

This approach allows us to:
- Standardize behavior and reduce duplication.
- Avoid locking ourselves into a single global client prematurely.
- Preserve flexibility for future multi-cluster or least-privilege needs.

Followup 
- update sports provider to use the new adapter
- consolidate credentials for all providers (make them point to one config)

## PR Review Checklist

[DISCO-3809]: https://mozilla-hub.atlassian.net/browse/DISCO-3809?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/MC-2052)

### DIFF
--- a/merino/search/async_elastic.py
+++ b/merino/search/async_elastic.py
@@ -1,0 +1,158 @@
+"""Async Elasticsearch service utilities."""
+
+from typing import Any, Optional, cast
+
+from elasticsearch import AsyncElasticsearch
+
+
+class AsyncElasticSearchAdapter:
+    """Wrapper around AsyncElasticsearch.
+
+    - Lazily creates and caches an AsyncElasticsearch client instance.
+    - Exposes async helpers for operations (search, close).
+    """
+
+    def __init__(
+        self,
+        *,
+        url: str,
+        api_key: str,
+    ) -> None:
+        self._url = url
+        self._api_key = api_key
+        self._client: Optional[AsyncElasticsearch] = None
+
+    def create_client(self) -> AsyncElasticsearch:
+        """Create an AsyncElasticsearch client."""
+        self._client = AsyncElasticsearch(self._url, api_key=self._api_key)
+        return self._client
+
+    def get_client(self) -> AsyncElasticsearch:
+        """Return the cached AsyncElasticsearch client, creating it if needed."""
+        if self._client is None:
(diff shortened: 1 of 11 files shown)


merino/search/async_elastic.py
```diff
@@ -0,0 +1,158 @@
...
+        url: str,
+        api_key: str,
+    ) -> None:
+        self._url = url
+        self._api_key = api_key
+        self._client: Optional[AsyncElasticsearch] = None
+
+    def create_client(self) -> AsyncElasticsearch:
+        """Create an AsyncElasticsearch client."""
+        self._client = AsyncElasticsearch(self._url, api_key=self._api_key)
```
jrconlin: *Question*: I wonder if we should consider adding [connection pooling](https://elasticsearch-py.readthedocs.io/en/7.x/async.html#asynctransport) using `elasticsearch.AsyncTransport()`?
(Probably not required for this version, but might help for things like connection retries)

PR #1265, opened 2026-02-04
[DISCO-3914] Integrate PII into Suggest

tiftran:
## References

JIRA: [DISCO-3914](https://mozilla-hub.atlassian.net/browse/DISCO-3914)

## PR Review Checklist

[DISCO-3914]: https://mozilla-hub.atlassian.net/browse/DISCO-3914?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/MC-2058)

### DIFF
--- a/merino/web/api_v1.py
+++ b/merino/web/api_v1.py
@@ -215,10 +215,14 @@ async def suggest(
 
     match pii_type:  # noqa
         case PIIType.EMAIL:
-            metrics_client.increment(f"suggestions.query.pii_detected.{pii_type.name.lower()}")
+            metrics_client.increment(
+                "suggestions.query.pii_detected", tags={"type": pii_type.name.lower()}
+            )
             return build_suggestion_response(client_variants, search_from, [])
         case PIIType.NUMERIC:
-            metrics_client.increment(f"suggestions.query.pii_detected.{pii_type.name.lower()}")
+            metrics_client.increment(
+                "suggestions.query.pii_detected", tags={"type": pii_type.name.lower()}
+            )
             is_soft_pii = True
         case _:
             pass
@@ -267,7 +271,8 @@ async def suggest(
 
     if is_soft_pii and len(suggestions):
         metrics_client.increment(
-            f"suggestions.query.pii_detected.{pii_type.name.lower()}.false_positive"
+            "suggestions.query.pii_detected.false_positive",
+            tags={"type": pii_type.name.lower()},
         )
 
     emit_suggestions_per_metrics(metrics_client, suggestions, search_from)
--- a/metrics.yaml
+++ b/metrics.yaml
@@ -2,18 +2,14 @@ version: 1.0
 
 service: Merino
 suggest:
-  query_pii_detected_email:
+  query_pii_detected:
     description: |
-      Measurement for queries that contain emails
+      Measurement for queries that contain pii
     type: counter
-    scope: |
-      The "api/v1/suggest" API endpoint.
-    alert_policy: []
-  query_pii_detected_numeric:
-    description: |
-      Measurement for queries that contain numbers
-    type:
-      counter
+    labels:
+      type:
+        - email
+        - numeric
     scope: |
       The "api/v1/suggest" API endpoint.
     alert_policy: []
--- a/tests/integration/api/v1/suggest/test_suggest.py
+++ b/tests/integration/api/v1/suggest/test_suggest.py
@@ -369,7 +369,7 @@ def test_suggest_with_invalid_geolocation_ip(
         (
             "/api/v1/suggest?q=test@example.com",
             [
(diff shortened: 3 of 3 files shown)


merino/web/api_v1.py
```diff
@@ -196,8 +196,28 @@ async def suggest(
...
+                    client_variants.split(",", maxsplit=CLIENT_VARIANT_MAX)[:CLIENT_VARIANT_MAX]
+                    if client_variants
+                    else []
+                ),
+            )
+            return ORJSONResponse(content=jsonable_encoder(response))
+        case PIIType.NUMERIC:
+            is_soft_pii = True
+        case _:
+            pass
```
Herraj: nit: I'm not sure but if there are gonna be other cases where we match a `PIITYPE` and return an empty suggestion response like on 206, we can maybe abstract that to a helper method! For now this is fine

merino/web/api_v1.py
```diff
@@ -196,8 +196,28 @@ async def suggest(
...
-    if pii_inspect(q):
-        pass
+
+    pii_type = pii_inspect(q)
+    is_soft_pii = False
+
+    match pii_type:  # noqa
+        case PIIType.EMAIL:
+            metrics_client.increment(f"suggestions.pii.{pii_type.name.lower()}")
+            response = SuggestResponse(
```
ncloudioj: Let's make a helper function for this to avoid duplication.

merino/web/api_v1.py
```diff
@@ -196,8 +196,28 @@ async def suggest(
...
     user_agent: UserAgent = request.scope[ScopeKey.USER_AGENT]
-    if pii_inspect(q):
-        pass
+
+    pii_type = pii_inspect(q)
+    is_soft_pii = False
+
+    match pii_type:  # noqa
+        case PIIType.EMAIL:
+            metrics_client.increment(f"suggestions.pii.{pii_type.name.lower()}")
```
ncloudioj: How about "`suggestion.query.pii_detected` with `pii_type.name` as a metric label? It will be easier to document and consume it.

merino/web/api_v1.py
```diff
@@ -196,8 +196,28 @@ async def suggest(
...
+                # [:CLIENT_VARIANT_MAX] filter at end to drop any trailing string beyond max_split.
+                client_variants=(
+                    client_variants.split(",", maxsplit=CLIENT_VARIANT_MAX)[:CLIENT_VARIANT_MAX]
+                    if client_variants
+                    else []
+                ),
+            )
+            return ORJSONResponse(content=jsonable_encoder(response))
+        case PIIType.NUMERIC:
+            is_soft_pii = True
```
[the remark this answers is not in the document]
tiftran: this would mean numerator over # of flight requests + weather location key request + maybe some amp suggestions. the amp suggestions that contain numbers should be small.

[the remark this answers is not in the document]
ncloudioj: Right, there is location key requests. Do you think we should exempt those requests from pii inspection for now? I.e. if the provider == accuweather and request_type == ??, then skip the inspection. Longer term, weather should be fine as we're planning to split it out off the suggest endpoint.

merino/providers/suggest/weather/provider.py
```diff
@@ -166,8 +166,12 @@ async def query(self, srequest: SuggestionRequest) -> list[BaseSuggestion]:
         weather_report: WeatherReport | None = None
         location_completions: list[LocationCompletion] | None = None
         weather_context = WeatherContext(geolocation, languages, request_source=source)
+        is_soft_pii: bool = srequest.is_soft_pii
         try:
             with self.metrics_client.timeit(f"providers.{self.name}.query.backend.get"):
+                if is_location_completion_request and is_soft_pii:
```
tiftran: this would mean locations like "100 mile house" would not be searchable, but given that there is no internal look up for location requests first, I think this is the option we should take

ncloudioj (commented):
LGTM!

merino/web/api_v1.py
```diff
@@ -253,8 +265,18 @@ async def suggest(
     if len(suggestions) == 1 and suggestions[0] is NO_LOCATION_KEY_SUGGESTION:
         return Response(status_code=204)
 
+    if is_soft_pii and len(suggestions):
+        metrics_client.increment(
+            f"suggestions.query.pii_detected.{pii_type.name.lower()}.false_positive"
```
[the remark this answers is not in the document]
tiftran: oops missed that sorry, will do

PR #1287, opened 2026-02-26
[HNT-1826] Follow ups for hourly forecasts endpoint

Herraj:
## References

JIRA: [HNT-1826](https://mozilla-hub.atlassian.net/browse/HNT-1826)

## Description
Follow ups to HNT-1161. Changes:

- Refactored backend method to have all logic one `try` block and the respective `except`s at the bottom. The functionality is unchanged.
- Added some metrics and update the `metrics.yaml` file.
- Added circuit breaker. (_please see the comments below_)
- Added and updated tests.

## PR Review Checklist

[HNT-1826]: https://mozilla-hub.atlassian.net/browse/HNT-1826?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/MC-2080)

### DIFF
--- a/merino/providers/suggest/weather/backends/accuweather/backend.py
+++ b/merino/providers/suggest/weather/backends/accuweather/backend.py
@@ -981,10 +981,16 @@ class AccuweatherBackend:
             return None
 
         self.metrics_client.increment("accuweather.cache.hit.hourly_forecasts")
+
+        # convert binary json from cache to regular json.
         hourly_forecasts_as_json = orjson.loads(hourly_forecasts_cached)["hourly_forecasts"]
+
+        # build HourlyForecast objects from the converted json above.
         hourly_forecasts: list[HourlyForecast] = create_hourly_forecasts_from_json(
             hourly_forecasts_as_json[:DEFAULT_FORECAST_HOURS]
         )
+
+        # casting the ttl to an int even though it is an int because linter does not trust binary int.
         return HourlyForecastsWithTTL(hourly_forecasts=hourly_forecasts, ttl=cast(int, ttl_cached))
 
     async def _fetch_hourly_forecasts_from_upstream(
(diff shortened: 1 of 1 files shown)


merino/web/api_v1.py
```diff
@@ -533,10 +534,15 @@ async def get_hourly_forecasts(
         hourly_forecasts: list[HourlyForecast] = []
         ttl = 0
```
Herraj: Now I'm not sure of the behaviour here. This will still trip the circuit breaker in the provider but then the exceptions raised by provider will be caught here and the client will be sent a `200 OK` with `[]` as a response. Is this the correct behaviour?

Or should we not catch the exception here and let FastAPI deal with it and have it return a `500`?

merino/providers/suggest/weather/provider.py
```diff
@@ -223,6 +223,9 @@ def build_suggestion(
                 locations=data,
             )
 
+    @WeatherCircuitBreaker(
+        name="weather", fallback_function=_hourly_forecasts_fallback
+    )  # Expect `AccuweatherError`
```
mmiermans: Do you want to share the `"weather"` circuit breaker name with the legacy query?

Some questions to help decide this:
- Is the most important shared resource that you aim to protect Accuweather credit, or memory/CPU? If it's Accuweather credits, then separate circuit breaker names would be useful.
- Are the `query` and `get_hourly_forecasts` functions more likely to fail independently or together? If they fail independently, then separate circuit breaker names would be better.

jpetto (changes requested):

PR #1380, opened 2026-04-08
[AIMOD-291] Response level section experiments

mlplyler:
## References

JIRA: [AIMOD-291](https://mozilla-hub.atlassian.net/browse/AIMOD-291?atlOrigin=eyJpIjoiZjI4OTBkYTBmNjgzNDQyY2EzZmEyNTBiMjFiNmYwYmUiLCJwIjoiaiJ9)

## Description
Parses sections and looks for A/B pairs where A is the canonical id and B has a '__exp<name>' suffix. One type of section experiment is implemented: 5050. in the 5050 experiment, A or B is served in any given response at random. The canonical id is attached to version B (also already assigned to version A) so that block and follow is respected.

For non-US locales, there is a locale suffix attached to the id. The locale suffix must be a suffix and is last. We are changing section object construction such that instead of all __ suffix being dropped, only the __l locale suffix is dropped. The __exp suffix is dropped later. There is an additional hard drop of __ so that we keep the behavior of no __ surviving. 

Enum of assumptions:
* experimental sections must follow pattern <base>__exp<name>__l<locale>
* malformed experiments id are dropped. <base>__exp<name>__<anything>__l<locale > is dropped
* there can be only A/B versions for a base name. A/B/C is not supported and outcomes will be non-deterministic
* an experiment id without a matching base name is dropped. no solo experiment sections
* we always pass the base name to the user

## PR Review Checklist

- [x ] [Functional and performance test](https://github.com/mozilla-services/merino-py/blob/main/docs/dev/testing.md) coverage has been expanded and maintained (if applicable)

[AIMOD-291]: https://mozilla-hub.atlassian.net/browse/AIMOD-291?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/merino/curated_recommendations/corpus_backends/protocol.py
+++ b/merino/curated_recommendations/corpus_backends/protocol.py
@@ -1,5 +1,7 @@
 """Protocol for the Corpus provider backend."""
 
+from __future__ import annotations
+
 from enum import Enum, unique
 from typing import Protocol
 
@@ -95,6 +97,16 @@ class CorpusSection(BaseModel):
     iab: IABMetadata | None = None
     externalId: str
     createSource: CreateSource
+    experimentVariant: int = Field(
+        default=0,
+        exclude=True,
+        description="Internal experiment variant for the section. 0 means the canonical section.",
+    )
+    alternateSection: CorpusSection | None = Field(
+        default=None,
+        exclude=True,
+        description="Internal link to an alternate slate for the canonical section.",
+    )
     followable: bool = Field(default=True, description="Whether users can follow this section.")
     allowAds: bool = Field(
         default=True, description="Whether ads can be displayed in this section."
--- a/merino/curated_recommendations/corpus_backends/sections_backend.py
+++ b/merino/curated_recommendations/corpus_backends/sections_backend.py
@@ -35,6 +35,24 @@ from merino.providers.manifest import Provider as ManifestProvider
 logger = logging.getLogger(__name__)
 
 
+def parse_section_external_id(raw_external_id: str) -> tuple[str, int]:
+    """Normalize a raw section ID into its canonical ID and experiment variant."""
+    # Strip any locale suffix (e.g., "__lEN_GB", "__lEN_CA") from externalId if present.
+    external_id = raw_external_id.split("__l", 1)[0]
+
+    marker = "__exp"
+    idx = external_id.rfind(marker)
+    if idx <= 0:
+        return external_id.split("__", 1)[0], 0
+
+    base_id = external_id[:idx]
+    variant_id = external_id[idx + len(marker) :]
+    if variant_id.isdigit():
+        return base_id, int(variant_id)
+
+    return external_id.split("__", 1)[0], 0
+
+
 class SectionsBackend(SectionsProtocol):
     """Backend for fetching corpus sections using the getSections query."""
 
@@ -126,15 +144,13 @@ class SectionsBackend(SectionsProtocol):
             raise CorpusGraphQLError(f"Sections API returned GraphQL errors {data['errors']}")
 
         utm_source = get_utm_source(surface_id)
-        sections_list = []
+        parsed_sections = []
         for section in data["data"]["getSections"]:
             if not section.get("active") or section.get("externalId", "").endswith("_crawl"):
                 logger.info(f"Skipping inactive section {section['externalId']} for {surface_id}")
                 continue
 
-            # Strip any locale suffix (e.g., "__lEN_GB", "__lEN_CA") from externalId if present.
-            # Keep non-locale suffixes such as "__exp5050" intact for downstream processing.
-            external_id = section["externalId"].split("__l", 1)[0]
+            external_id, experiment_variant = parse_section_external_id(section["externalId"])
 
             section_obj = CorpusSection(
                 externalId=external_id,
@@ -144,6 +160,7 @@ class SectionsBackend(SectionsProtocol):
                 heroSubtitle=section.get("heroDescription"),
                 iab=section["iab"],
                 createSource=section["createSource"],
+                experimentVariant=experiment_variant,
                 followable=section["followable"],
                 allowAds=section["allowAds"],
                 sectionItems=[
@@ -153,6 +170,28 @@ class SectionsBackend(SectionsProtocol):
                     for section_item in section["sectionItems"]
                 ],
             )
-            sections_list.append(section_obj)
+            parsed_sections.append(section_obj)
+
+        sections_list = []
+        base_sections_by_id: dict[str, CorpusSection] = {}
+        pending_alternates: dict[str, CorpusSection] = {}
+
+        for section in parsed_sections:
+            if section.experimentVariant == 0:
+                sections_list.append(section)
+                base_sections_by_id[section.externalId] = section
+
+                alternate_section = pending_alternates.pop(section.externalId, None)
+                if alternate_section is not None and section.alternateSection is None:
+                    section.alternateSection = alternate_section
+                continue
+
+            base_section = base_sections_by_id.get(section.externalId)
+            if base_section is not None:
+                if base_section.alternateSection is None:
+                    base_section.alternateSection = section
+                continue
+
+            pending_alternates.setdefault(section.externalId, section)
 
         return sections_list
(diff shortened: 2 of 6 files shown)


merino/curated_recommendations/corpus_backends/sections_backend.py
```diff
@@ -132,9 +132,9 @@ async def fetch(self, surface_id: SurfaceId) -> list[CorpusSection]:
                 logger.info(f"Skipping inactive section {section['externalId']} for {surface_id}")
                 continue
 
-            # Strip any suffix (e.g., "__lEN_GB", "__lEN_CA") from externalId if present
-            # This handles locale suffixes and any future suffix patterns
-            external_id = section["externalId"].split("__")[0]
+            # Strip any locale suffix (e.g., "__lEN_GB", "__lEN_CA") from externalId if present.
```
rolf-moz: We are making the assumption here that experiment ID is always after the locale ID. 
It is probably good to use a utility function with a clear spec on supported and unsupported extension order. Maybe best to support both?

merino/curated_recommendations/corpus_backends/sections_backend.py
```diff
@@ -132,9 +132,9 @@ async def fetch(self, surface_id: SurfaceId) -> list[CorpusSection]:
...
+            # Keep non-locale suffixes such as "__exp5050" intact for downstream processing.
+            external_id = section["externalId"].split("__l", 1)[0]
 
             section_obj = CorpusSection(
```
rolf-moz, line 159: Looks good but I wanted to suggest one possible improvement.

CorpusSection is an internal representation of a section. A different data type Section is returned to the Fx clients. Our previous google doc for the section ID in the new tab, we discussed having non-zero variant_id for a section indicating that it is a section variant.

Maybe we can start with implementing some of this now in Merino? Adding additional data structures in section_backend will keep repeat string comparisons happening every request.

Specifically, we could add to CorpusSection  an int field experiment_variant that defaults to 0 or None and store the original canonical section id. The alt section could be part of the list, or be linked as an attribute of CorpusSection (i.e. corpus section linking to an alternative slate). 

dedupe_experiment_sections could made much simpler with these changes. For example, it could go through the sections, and for a section that has a variant, decide which one to swap into the list.

In this method the section id would be normalized and you'd have two of the same section id in the list. Therefore you would never need to do a deep copy to create a new section during the request period.

merino/curated_recommendations/sections.py
```diff
@@ -201,6 +199,76 @@ def _process_corpus_sections(
...
+        parsed_eid = clean_exp_id(eid)
+        if parsed_eid is None:
+            continue
+        can_eid, exp_type = parsed_eid
+        # Check if there is a matching non-experimental section.
+        if can_eid in id_to_section:
+            # Pick the winning section ID.
+            kept_id = resolve_section_experiment(can_eid, eid, exp_type)
+            # Replace the value while keeping the canonical ID.
+            id_to_result[can_eid] = deepcopy(id_to_section[kept_id])
```
rolf-moz: I think deep-copy will copy all the items individually and thus be computationally expensive. That would be bad... it's best to avoid deepcopy if possible... see my suggestion above. If you need to do a deep copy, maybe implement a custom function so you can be sure it does not recursively copy the list of items.

rolf-moz (changes requested):

merino/curated_recommendations/corpus_backends/sections_backend.py
mlplyler: We assume exp is BEFORE locale. 

Downstream from this line, we filter any id with exp that doesn't meet our specs. Are you suggestion we make this filter standalone and move to this location?

merino/curated_recommendations/sections.py
mlplyler: Ah didn't realize the sections are cached. Let me try your suggestion.

merino/curated_recommendations/corpus_backends/sections_backend.py
```diff
@@ -153,6 +170,28 @@ async def fetch(self, surface_id: SurfaceId) -> list[CorpusSection]:
                     for section_item in section["sectionItems"]
                 ],
             )
-            sections_list.append(section_obj)
+            parsed_sections.append(section_obj)
+
+        sections_list = []
```
rolf-moz: I see how we are a little more efficient this way, this code is called infrequently so we can maybe focus on readability

```
base_sections = [s for s in parsed_sections if s.experimentVariant == 0]
experimental_sections = [s for s in parsed_sections if s.experimentVariant != 0]

base_sections_by_id: dict[str, CorpusSection] = {
    s.externalId: s for s in base_sections
}

for section in experimental_sections:
    base = base_sections_by_id.get(section.externalId)
    if base and base.alternateSection is None:
        base.alternateSection = section

sections_list = base_sections
```

or if you can add a utility function for corpus sections.

This entire code block could be wrapped in a utility function also
```
base_sections = [s for s in parsed_sections if not s.is_experiment_section()]
experimental_sections = [s for s in parsed_sections if s.is_experiment_section()]

base_sections_by_id: dict[str, CorpusSection] = {
    s.externalId: s for s in base_sections
}

for section in experimental_sections:
    base = base_sections_by_id.get(section.externalId)
    if base and base.alternateSection is None:
        base.alternateSection = section

sections_list = base_sections
```

merino/curated_recommendations/corpus_backends/sections_backend.py
rolf-moz, line 97: Add comment here indicating experimental sections are removed from result and referenced separately

rolf-moz (approved):

rolf-moz:
One side effect of the design that I suggested is that the legacy users (non sections) and mobile will get the base experiment without the variant. I don't know if there is enough traffic there to cause an issue.  The previous method would have likely merged both, increasing the number of items for the experimental sections.

It might be worth adding a to the integration tests for the legacy provider a case with an experimental section like this.

merino/curated_recommendations/corpus_backends/sections_backend.py
mlplyler: ok took this thanks!

merino/curated_recommendations/corpus_backends/sections_backend.py
```diff
@@ -35,6 +35,24 @@
...
+    idx = external_id.rfind(marker)
+    if idx <= 0:
+        return external_id.split("__", 1)[0], 0
+
+    base_id = external_id[:idx]
+    variant_id = external_id[idx + len(marker) :]
+    if variant_id.isdigit():
+        return base_id, int(variant_id)
+
+    return external_id.split("__", 1)[0], 0
```
mmiermans, line 53: Could you write unit tests against this function for any relevant edge case?

merino/curated_recommendations/corpus_backends/protocol.py
```diff
@@ -95,6 +97,16 @@ class CorpusSection(BaseModel):
     iab: IABMetadata | None = None
     externalId: str
     createSource: CreateSource
+    experimentVariant: int = Field(
+        default=0,
+        exclude=True,
+        description="Internal experiment variant for the section. 0 means the canonical section.",
+    )
```
mmiermans, line 104: `exclude=True` will prevent this field from being sent to the client, right? How will clients report which branch they're in? Via the section name?

rolf-moz: At this point we don't have any reporting from the client. That is a client change that we'll do in the future. We know which items were in which variant, so that is how we'll measure the results.

rolf-moz: As far as i can tell the CorpusSection class is never directly serialized/unserialized in any way, so exclude=true simply indicates that the value is more of a transient data.

merino/curated_recommendations/corpus_backends/sections_backend.py
mlplyler: Added TestParseSectionExternalId

merino/curated_recommendations/corpus_backends/protocol.py
mmiermans: Gotcha. That means the A/B test is assumed to have distinct items?

mmiermans (approved):

PR #1393, opened 2026-04-13
feat(gcs): add SyncedGcsBlobV2

kschelonka:
## References
Broken off from https://mozilla-hub.atlassian.net/browse/DISCO-3988

## Description
According to ADR the goal is to unify cached blob access with one class. SyncedGcsBlob requires some updates to comply with suggest provider expectations and SRE requirements around metrics:
* use the async `Storage` client
* reduce metrics cardinality by using labels
* add unit and integration tests

The metrics updates are breaking changes so a new version was created instead, which existing code paths can migrate to. The existing metrics from SyncedGcsBlob were preserved but use labels to differentiate namespace instead of the metric name. The rest of the methods were mostly preserved with updates to the following:

Some other small improvements include:
* Requiring a callback function in the class constructor to eliminate potential for developer error (previously it required manually setting the callback after the instance was constructed)
* Implement error handling with additional metrics
* Store the cached data directly on the class, rather than requiring implementation in the callback_fn
* New helper: Adding a factory method for JSON blobs - automatically parse and validate against a model

## PR Review Checklist

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/MC-2202)

### DIFF
--- a/merino/utils/synced_gcs_blob_v2.py
+++ b/merino/utils/synced_gcs_blob_v2.py
@@ -1,0 +1,216 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+"""Module providing a class to manage synchronized blobs from Google Cloud Storage."""
+
+import asyncio
+import logging
+import time
+from datetime import datetime, timezone
+from json import JSONDecodeError
+from typing import Callable, Generic, Optional, TypeVar
+
+import aiohttp
+import orjson
+from aiodogstatsd import Client as StatsdClient
+from aiodogstatsd.typedefs import MTags
+from gcloud.aio.storage import Blob, Bucket, Storage
+from pydantic import BaseModel, ValidationError
+
+from merino.utils import cron
+
+
+logger = logging.getLogger(__name__)
+
+
+LAST_UPDATED_INITIAL_VALUE = datetime.min.replace(tzinfo=timezone.utc)
+
+T = TypeVar("T", bound=BaseModel)
+
+
+class SyncedGcsBlobV2(Generic[T]):
+    """Class to manage a synchronized Google Cloud Storage blob.
+
+    This class periodically fetches data from a GCS blob in the background.
+    """
+
+    cron_task: asyncio.Task
+    last_updated: datetime
+    metrics_namespace = "gcs.sync"
+    _data: Optional[T]
+
+    def __init__(
+        self,
+        storage_client: Storage,
+        metrics_client: StatsdClient,
+        bucket_name: str,
+        blob_name: str,
+        max_size: int,
+        cron_interval_seconds: float,
+        cron_job_name: str,
+        fetch_callback: Callable[[bytes], T],
+    ) -> None:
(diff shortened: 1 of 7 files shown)


merino/utils/synced_gcs_blob_v2.py
```diff
@@ -0,0 +1,217 @@
...
+        # garbage collected because asyncio's runtime only holds a weak reference to it.
+        self.cron_task = asyncio.create_task(cron_job())
+
+    @property
+    def update_count(self) -> int:
+        """Return the number of times the data has been updated."""
+        return self._update_count
+
+    @property
+    def data(self) -> Optional[T]:
```
kschelonka, line 100: new behavior - previous implementation required intercepting the result of fetch_callback and storing it elsewhere

merino/utils/synced_gcs_blob_v2.py
```diff
@@ -0,0 +1,217 @@
...
+    def __init__(
+        self,
+        storage_client: Storage,
+        metrics_client: StatsdClient,
+        bucket_name: str,
+        blob_name: str,
+        max_size: int,
+        cron_interval_seconds: float,
+        cron_job_name: str,
+        fetch_callback: Callable[[bytes], T],
```
kschelonka, line 52: change: this is required for class construction, and the signature updated to return the data rather than `None`; previous implementation set it after instantiating the class (possible error if forget to set) and required that the callback method stores the response somewhere accessible outside of this class

merino/utils/synced_gcs_blob_v2.py
```diff
@@ -0,0 +1,219 @@
...
+        self._data = None
+
+    def initialize(self) -> None:
+        """Start the background cron job to get new data."""
+        if hasattr(self, "cron_task"):
+            return  # already initialized
+        cron_job = cron.Job(
+            name=self.cron_job_name,
+            interval=self.cron_interval_seconds,
+            condition=lambda: True,
```
ncloudioj, line 87: [Non-blocking] Re: sync failure handling: when a temporary sync failure occurs, the condition check can be used to retry sooner before the next sync tick, this can be particularly useful for large sync intervals.

`cron.Job` was designed with a primary cron interval and an additional condition check to facilitate retrying. For example, for a hourly cron job whose job handler might fail, we can make the cron job tick more frequently (by using a much smaller interval) and set another condition check so it will skip the job if the condition is not met (e.g. the last update timestamp is within an hour).

As a followup, this new class might be able to leverage that for better failure handling.

ncloudioj (approved):
Looks good!

merino/utils/synced_gcs_blob_v2.py
kschelonka: yeah I noticed that, currently the way that the original class was set up (which I preserved) is that it didn't pull the file unless the updated_at timestamp was newer than the last known value. so theoretically we could have the job tick much more frequently than the actual desired update interval. Or we could have the forked behavior for error handling specifically as before (or both). thoughts on this?

kschelonka: I'm going to go ahead and merge for now but we can revisit this!

ncloudioj: Yup, this can be revisited later. Sorry for the delay!

PR #1403, opened 2026-04-15
[AIMOD-1043] Implement Query Normalization Experiment

vazish:
## References

JIRA: [AIMOD-1043](https://mozilla-hub.atlassian.net/browse/AIMOD-1043)

## Description
This PR allows us to test the new query normalization pipeline using the `&client_variants=query_norm_treatment` parameter via Nimbus. At present, we're only targeting "Finance" (Polygon) and "Sports" providers.

Artifacts are loaded from GCS to build the autocomplete index and canonical set for each provider.

See experiment proposal here: [Doc](https://docs.google.com/document/d/158GidXdX9dkirtYPHL1yqittSoQW1tC35cVPLiRDRqw/edit?tab=t.0)
See repo for implementation: [Repo](https://github.com/mozilla/query_normalization)
See impact analysis here: [Doc](https://mozilla-hub.atlassian.net/wiki/spaces/FAAMT/pages/2477523204/Query+Normalization+Impact+Analysis)
Draft Experimenter Link: [Link](https://experimenter.services.mozilla.com/nimbus/suggest-query-normalization-experiment/summary/)

Good places to start review:
`init_pipeline()`: Downloads artifacts, builds autocomplete indices and canonical set that we check against.
`class NormalizePipeline`: Class that uses built indices and canonical set to do the normalization on the user query

To test:
```
# Setup
# Terminal 1
make install
make docker-compose-up # needed for Sports (Elasticsearch)

# Terminal 2
# for local test 
make dev 

# or for gcs test with artifacts loaded
# might require "gcloud auth application-default login" first
MERINO_QUERY_NORMALIZATION__DATA_SOURCE=remote \
MERINO_QUERY_NORMALIZATION__GCS_BUCKET=merino-ml-data-stage \
make dev

# Terminal 3
# control, no match, finance
curl -s "http://localhost:8000/api/v1/suggest?q=dow+jone" | python3 -m json.tool

# control, no match, sports
curl -s "http://localhost:8000/api/v1/suggest?q=bulls+scor" | python3 -m json.tool

# treatment, match, finance
curl -s "http://localhost:8000/api/v1/suggest?q=dow+jone&client_variants=query_norm_treatment" | python3 -m json.tool

# treatment, match, sports
curl -s "http://localhost:8000/api/v1/suggest?q=bulls+scor&client_variants=query_norm_treatment" | python3 -m json.tool

# Run unit tests
MERINO_ENV=testing uv run pytest tests/unit/query_normalization/ -v
```

Also did some locust testing. These were the results:
Screenshot 2026-04-14 at 16 57 33

## PR Review Checklist

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/MC-2210)

[AIMOD-1043]: https://mozilla-hub.atlassian.net/browse/AIMOD-1043?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


rolf-moz:
Should I get a result for:
`curl -s "http://localhost:8000/api/v1/suggest?q=bulls+scores" | python3 -m json.tool`
(base case without experiment flag). I don't get a result for that.

merino/configs/default.toml
```diff
@@ -642,6 +676,10 @@ dummy_attempted_count = 1000
 # Whether to check `client_variants` for Thompson sampling
 check_client_variants = true
 
+# MERINO_PROVIDERS__ADM__THOMPSON__ENGAGEMENT_GUIDED_SUGGESTIONS
+# The client_variant value that enables Thompson sampling
+engagement_guided_suggestions = "engagement_guided_suggestions"
```
rolf-moz: Does this need to be a pref? Shouldn't it be a constant instead of in the settings? I don't see why you'd want different values for this between staging/production/dev etc.

merino/providers/suggest/adm/provider.py
```diff
@@ -52,6 +52,7 @@ class IABCategory(str, Enum):
 FALLBACK_COUNTRY_CODE: str = "US"
 CLIENT_VARIANTS_ALLOW_LIST = frozenset(settings.web.api.v1.client_variant_allow_list)
 TS_DRY_RUN: bool = settings.providers.adm.thompson.dry_run
+ENGAGEMENT_GUIDED_SUGGESTIONS: str = settings.providers.adm.thompson.engagement_guided_suggestions
```
rolf-moz: again, it feels a little odd to have as a pref, but maybe you can leave it.

tests/unit/providers/suggest/adm/test_provider_thompson.py
```diff
@@ -22,7 +22,7 @@
 USER_AGENT = UserAgent(form_factor="desktop", browser="firefox", os_family="macos")
 
 
-CLIENT_VARIANTS: list[str] = ["engagement_guided_suggestions"]
+CLIENT_VARIANTS: list[str] = ["engagement_guided_suggestions", "query_norm_treatment"]
```
rolf-moz: use constant here for each, or get from settings if you are keeping in settings

tests/unit/query_normalization/test_pipeline.py
```diff
@@ -0,0 +1,477 @@
...
+    assert _try_join_normalize(tokens, canonical) is None
+
+
+def test_join_no_match(canonical: set[str]) -> None:
+    """Non-canonical merges should return None."""
+    tokens = ["purple", "elephant"]
+    assert _try_join_normalize(tokens, canonical) is None
+
+
+def test_join_single_token(canonical: set[str]) -> None:
```
rolf-moz, line 67: This is because we're only normalizing multi-word right now? Maybe add a note

merino/web/api_v1.py
```diff
@@ -242,9 +241,19 @@ async def suggest(
     geolocation = refine_geolocation_for_suggestion(request, city, region, country)
     client_variants_list = client_variants.split(",") if client_variants else []
 
+    # Query normalization experiment (Phase 1: sports + finance only)
+    pipeline = get_pipeline()
+    client_variant = settings.query_normalization.client_variant
```
rolf-moz: ^as before this feels better as a constant than a setting. When we have multiple experiments it will get more complicated as a setting.

merino/web/api_v1.py
```diff
@@ -242,9 +241,19 @@ async def suggest(
...
+    use_normalization = pipeline is not None and client_variant in client_variants_list
```
rolf-moz: do we need a use_normalization flag at all? Here is a suggestion.

```
normalization_pipeline = get_normalization_pipeline()
normalization_providers = {}
if  normalization_pipeline is not None:
     normalization_providers = frozenset(settings.query_normalization.providers)

```
    
```
q_for_provider = q_normalized if p.name in normalization_providers else q
```

If you need to check a flag you can use  len(settings.query_normalization.providers) > 0

merino/web/api_v1.py
```diff
@@ -287,6 +296,21 @@ async def suggest(
 
     emit_suggestions_per_metrics(metrics_client, suggestions, search_from)
 
+    # Normalization experiment metrics (Phase 1 providers only)
```
rolf-moz: could this be a separate function or part of emit_suggestions_per_metrics?

merino/utils/query_processing/normalization/pipeline.py
```diff
@@ -0,0 +1,406 @@
...
+
+    Build once at startup; normalize() is cheap at query time.
+    """
+
+    def __init__(
+        self,
+        canonical: set[str],
+        finance_bm25: BM25Index | None = None,
+        canonical_prefix_index: (dict[str, tuple[str, int, int]] | None) = None,
+    ) -> None:
```
rolf-moz, line 342: Is there any benefit for normalizing differently for each provider?

rolf-moz:
Can you link in the code to the repo/pipeline that generates all of the text data files we're using?

merino/web/api_v1.py
```diff
@@ -242,9 +241,19 @@ async def suggest(
     geolocation = refine_geolocation_for_suggestion(request, city, region, country)
     client_variants_list = client_variants.split(",") if client_variants else []
 
+    # Query normalization experiment (Phase 1: sports + finance only)
```
rolf-moz, line 258: Do we want to skip normalization completely if none of the providers we're searching in search_from will use the normalization?

tests/unit/providers/suggest/adm/test_provider_thompson.py
tiftran: I think `"query_norm_treatment"` doesn't need to be added here. Existing tests should pass without the add?

merino/web/api_v1.py
tiftran: I like this idea! Might be a me thing, took me a while to figure out the logic here. We're collecting metrics for queries that were normalized that produced a suggestion for a given provider? Could we add a comment 🥺

tests/unit/query_normalization/test_pipeline.py
```diff
@@ -0,0 +1,477 @@
...
+    _try_join_normalize,
+    _try_split_normalize,
+    _try_split_token,
+    _try_wordsegment,
+    build_prefix_index,
+    tier_a,
+)
+
+
+# tier_a
```
tiftran, line 18: we do like to use `@pytest.parametrize` for repetitive tests, but not really a big deal

merino/query_normalization/__init__.py
ncloudioj, line 1: For consistency, can you move the normalization module to `merino/utils/query_processing` and make it a sub-module of it?

merino/providers/suggest/__init__.py
```diff
@@ -40,6 +41,9 @@ async def init_providers() -> None:
             extra={"providers": [*providers.keys()], "elapsed": timer() - start},
         )
 
+    # init query normalization pipeline
+    await init_pipeline()
```
ncloudioj, line 44: For remote mode, any network or data parsing errors will get raised here and prevent Merino from initializing the app. Since query normalization is currently optional in the suggest request path, let's add some exception handling to `QueryNormRemoteFileManager` or `init_pipeline()` so it won't throw here.

ncloudioj: Also, let's time the init duration for this similar to other initializers.

merino/query_normalization/pipeline.py
```diff
@@ -0,0 +1,406 @@
...
+        # return if candidate is in canonical set
+        if candidate in canonical or merged in canonical:
+            hits.append(candidate)
+    # only return if unambiguous
+    return hits[0] if len(hits) == 1 else None
+
+
+# Step 4: segment query into tokens with wordsegment
+# maxsize can be changed to tune perf.
+@functools.lru_cache(maxsize=10_000)
```
ncloudioj: Let's make maxsize a configuration and add it to `default.toml`.

merino/utils/query_processing/normalization/pipeline.py
```diff
@@ -0,0 +1,406 @@
+"""Precision-first query normalization cascade.
+
+Pipeline (in order):
```
ncloudioj, line 7: Could you briefly outline the pipeline protocol (i.e. how the pipeline components get chained) in the doc, so readers don't need to find out that from reading the source below? For example, if a pipeline component finishes (or couldn't) normalization, its return value will be sent down to the next step or if there is any short circuiting to terminate the entire normalization.

merino/query_normalization/pipeline.py
```diff
@@ -0,0 +1,406 @@
...
+        "\u00a0": " ",
+        "\u200b": "",
+        "\u200c": "",
+        "\u200d": "",
+        "\ufeff": "",
+    }
+)
+
+# detect urls so we can skip tier_a()
+_URL_RE = re.compile(
```
ncloudioj: FWIW, Firefox already performs the URL check and will not send any queries to Merino. Maybe we can skip this.

merino/query_normalization/pipeline.py
```diff
@@ -0,0 +1,406 @@
...
+    return " ".join(_wordsegment.segment(tok))
+
+
+def _try_wordsegment(tokens: list[str], canonical: set[str]) -> str | None:
+    """Try wordsegment on each fused token; accept if result is canonical.
+
+    Accepts the rewrite if either the full rebuilt query OR the segmented
+    portion alone is in canonical.
+    """
+    result_tokens = list(tokens)
```
ncloudioj: Nit: `tokens` is of type `list[str]`, no need to listify it.

merino/configs/production.toml
```diff
@@ -112,6 +112,11 @@ cache = "redis"
 # If `redis`, the global Redis settings must be set. See redis.server.
 cache = "redis"
 
+[production.query_normalization]
+enabled = true
```
ncloudioj: Let's disable this in prod for now. We can make a followup change to turn it on once verified in stage.

merino/utils/query_processing/normalization/pipeline.py
```diff
@@ -0,0 +1,406 @@
...
+        self._canonical_words: set[str] = {w for phrase in canonical for w in phrase.split()}
+
+        # pre-warm cache on canonical words
+        _wordsegment.load()
+        for phrase in canonical:
+            for tok in phrase.split():
+                if len(tok) >= 5 and tok not in _wordsegment.UNIGRAMS:
+                    _ws_segment(tok)
+
+    def normalize(self, query: str) -> str:
```
ncloudioj, line 361: Merino currently will reject any queries that are more than [500 characters](https://github.com/mozilla-services/merino-py/blob/main/merino/configs/default.toml#L122) in size. Do you think we should apply a similar max-query-length-for-normalization here? Longer query strings would definitely take longer for Merino to normalize and it'd leave a potential attacking surface for bad actors to send those queries to overload the service.

merino/query_normalization/pipeline.py
```diff
@@ -0,0 +1,406 @@
...
+            # skip short tokens
+            continue
+        merged = tokens[i] + tokens[i + 1]
+        candidate_tokens = tokens[:i] + [merged] + tokens[i + 2 :]
+        candidate = " ".join(candidate_tokens)
+        # return if candidate is in canonical set
+        if candidate in canonical or merged in canonical:
+            hits.append(candidate)
+    # only return if unambiguous
+    return hits[0] if len(hits) == 1 else None
```
ncloudioj: `itertools.pairwise()` can simplify the index twiddling and the case of len(tokens) < 2. 
```suggestion
    hits: list[str] = []
    
    for i, (left, right) in enumerate(itertools.pairwise(tokens)):
        if len(left) < 2 or len(right) < 2:
            continue
        merged = f"{left}{right}"
        # form a new candidate with the joined adjacent tokens and the rest of other tokens  
        candidate = " ".join([*tokens[:i], merged, *tokens[i+2:]])
        if candidate in canonical or merged in canonical:
             hits.append(candidate)
     
     # only return if unambiguous
     return hits[0] if len(hits) == 1 else None
```

merino/query_normalization/pipeline.py
```diff
@@ -0,0 +1,406 @@
...
+    result_tokens = list(tokens)
+    for i, tok in enumerate(tokens):
+        if len(tok) < 5 or tok in canonical:
+            continue
+        segmented = _ws_segment(tok)
+        if segmented == tok:
+            continue
+        # e.g. "redsox game" -> "red sox game"
+        candidate_tokens = result_tokens[:i] + segmented.split() + result_tokens[i + 1 :]
+        candidate = " ".join(candidate_tokens)
```
ncloudioj: Nit:
```suggestion
        candidate = " ".join([*result_tokens[:i], segmented, *result_tokens[i+1:]])
```

merino/query_normalization/pipeline.py
```diff
@@ -0,0 +1,406 @@
...
+    result_tokens = list(tokens)
+
+    for i, tok in enumerate(tokens):
+        if len(tok) < 5 or tok in canonical:
+            continue
+
+        split = _try_split_token(tok, canonical)
+        if split is not None:
+            candidate_tokens = result_tokens[:i] + split.split() + result_tokens[i + 1 :]
+            candidate = " ".join(candidate_tokens)
```
ncloudioj: Ditto, if you're fine with `" ".join([*a, split, *b])`.

ncloudioj (changes requested):
Looks good, thanks!

Finished the first half of the review and wanted to share some early comments.

vazish:
> Should I get a result for: `curl -s "http://localhost:8000/api/v1/suggest?q=bulls+scores" | python3 -m json.tool` (base case without experiment flag). I don't get a result for that.


> Should I get a result for: `curl -s "http://localhost:8000/api/v1/suggest?q=bulls+scores" | python3 -m json.tool` (base case without experiment flag). I don't get a result for that.

Try `curl -s "http://localhost:8000/api/v1/suggest?q=bulls+score" | python3 -m json.tool`

merino/configs/default.toml
vazish: This is for the current thompson sampling experiment so we don't mix branches. I'll move to const.

tests/unit/providers/suggest/adm/test_provider_thompson.py
vazish: Yea they should. I added it to test preventing mixing branches. In the previous implementation, we were checking in the list, so clients in the query norm branch would get ts outputs. I'll remove.

merino/providers/suggest/adm/provider.py
vazish: Will move to const.

tests/unit/query_normalization/test_pipeline.py
vazish: There's nothing to join with a single word :)

merino/web/api_v1.py
vazish: In prod, `settings.query_normalization.providers` will be `["polygon", "sports"]`. So if we do above, we'll always return normalized version for 100% of population.

merino/web/api_v1.py
vazish: Sorry Tif! I'll make this more clear and comment. These metrics are just so we have some visibility on queries that were actually changed by the normalization pipeline. I don't know where the telemetry lives yet...

merino/utils/query_processing/normalization/pipeline.py
vazish: These just build the indices (indexes?) but each provider does have slightly different handling in the current main, probably some idiosyncrasies of particular matching provider.

vazish:
> Can you link in the code to the repo/pipeline that generates all of the text data files we're using?

They are built from the sports provider API (had to create a free trial for this one) and the existing merino finance tickers. I'll add the code to our test query norm repo.

merino/web/api_v1.py
vazish: Hmm the new normalization sits on top of the provider specific normalization atm, so it might not be easy to do so.

merino/providers/suggest/__init__.py
vazish: Will do!

merino/query_normalization/pipeline.py
vazish: Ok cool yes, then we don't need this.

merino/utils/query_processing/normalization/pipeline.py
vazish: Sounds good, I'll add a max limit.

merino/query_normalization/pipeline.py
vazish: Nice! Didn't know about this, will update.

merino/utils/query_processing/normalization/pipeline.py
```diff
@@ -0,0 +1,406 @@
...
+# Step 5: Prefix Complete (multi-token queries only)
+_AUTOCOMPLETE_MIN_PREFIX_LEN = 4  # minimum prefix length before autocomplete
+_AUTOCOMPLETE_MIN_ABS_FREQ = 3_000  # minimum absolute frequency from vocab to allow autocomplete
+_AUTOCOMPLETE_MIN_FREQ_RATIO = 2.0  # minimum gating ratio between best and second best
+_AUTOCOMPLETE_COMMON_WORD_FREQ = (
+    8_000_000  # wordsegment frequency guard to prevent autocomplete of common words
+)
+
+
+def build_prefix_index(
```
ncloudioj, line 187: Do we have an idea of the vocabulary size in production? Two notes for medium or large vocabs.
- This prefix index could be pre-built to save some index building time
- The result prefix index could have large redundancy, for example, for the term "mozilla firefox", all the prefixes from "mozi" to "mozilla firefox" would share the same value `("mozilla firefox", best_freq, second_best_freq)`, storing a lot of those entries in memory can be wasteful.

To build prefix indices for large vocabs, we have more efficient ways implemented in [merino-ext](https://github.com/mozilla-services/moz-merino-ext).

merino/query_normalization/pipeline.py
```diff
@@ -0,0 +1,406 @@
...
+
+
+def _apply_prefix_complete(
+    query: str,
+    prefix_index: dict[str, tuple[str, int, int]],
+    allowlist: set[str],
+    min_abs_freq: int = _AUTOCOMPLETE_MIN_ABS_FREQ,
+) -> str:
+    """Complete the last token only when one completion clearly dominates."""
+    tokens = query.split()
```
ncloudioj: Just noticed that there are various tokenization performed in multiple steps, is that true? If so, maybe we can do this once at the beginning of normalization and pass the tokens down the pipeline to avoid repetitive tokenization.

vazish:
> Can you link in the code to the repo/pipeline that generates all of the text data files we're using?

Here! https://github.com/mozilla/query_normalization/pull/2

merino/utils/query_processing/normalization/pipeline.py
vazish: It's pretty small (`~41 KB` - `2,869 words`), so might be ok? This is because I'm filtering for the canonical terms only from the finance tickers and the sports teams, making autocomplete changes only when we're sure and against canonical.

Here's the current one in the bucket: 
```
gsutil cp gs://merino-ml-data-prod/query_normalization/word_freq.csv ~/Desktop/
```

I think once we start adding larger canonical sets from Yelp, etc, then it might be better / faster to build offline.

vazish: It's also per word, so it only happens on the last incomplete token. It's not very smart but does help a bit.

merino/configs/default.toml
vazish: Done!

merino/providers/suggest/adm/provider.py
vazish: Done!

tests/unit/providers/suggest/adm/test_provider_thompson.py
vazish: Removed!

merino/web/api_v1.py
vazish: Done!

merino/web/api_v1.py
vazish: Done!

merino/query_normalization/__init__.py
vazish: Done!

merino/providers/suggest/__init__.py
vazish: Done!

merino/query_normalization/pipeline.py
vazish: Done!

merino/utils/query_processing/normalization/pipeline.py
vazish: Done!

merino/query_normalization/pipeline.py
vazish: Removed!

merino/configs/production.toml
vazish: Done!

merino/utils/query_processing/normalization/pipeline.py
vazish: Added 200 since max query size is probably lower and this will prevent wordsegment from taking long.

merino/query_normalization/pipeline.py
vazish: Done!

merino/query_normalization/pipeline.py
vazish: Done!

merino/query_normalization/pipeline.py
vazish: Done!

merino/query_normalization/pipeline.py
vazish: Yea, it's redone here. I updated the params to pass both the query and the tokens.

merino/utils/query_processing/normalization/pipeline.py
ncloudioj: Love it, thanks!

merino/utils/query_processing/normalization/pipeline.py
```diff
@@ -0,0 +1,406 @@
...
+    }
+)
+
+
+# Step 1: do canonicalization on the query string
+def tier_a(query: str) -> str:
+    """NFKC + punctuation normalization + casefold + whitespace collapse."""
+    query = unicodedata.normalize("NFKC", query)
+    query = query.translate(_PUNCT_MAP)
+    return re.sub(r"\s+", " ", query.casefold().strip())
```
ncloudioj: Nit: let's define a constant for this regex pattern to avoid repetitively defining it for every single call as this will be called a lot.

merino/utils/query_processing/normalization/pipeline.py
```diff
@@ -0,0 +1,406 @@
...
+    """
+    hits: list[str] = []
+    for i, (left, right) in enumerate(pairwise(tokens)):
+        if len(left) < 2 or len(right) < 2:
+            continue
+        merged = f"{left}{right}"
+        # form a new candidate with the joined adjacent tokens and the rest
+        candidate = " ".join([*tokens[:i], merged, *tokens[i + 2 :]])
+        if candidate in canonical or merged in canonical:
+            hits.append(candidate)
```
ncloudioj, line 92:
```suggestion
            hits.append(candidate)
            if len(hits) > 1:
                # abort normalization if ambiguous
                return None
```

merino/utils/query_processing/normalization/pipeline.py
```diff
@@ -0,0 +1,406 @@
...
+            prev = 0
+            for cut in cuts:
+                parts.append(token[prev:cut])
+                prev = cut
+            parts.append(token[prev:])
+
+            if all(len(p) >= min_seg for p in parts):
+                candidate = " ".join(parts)
+                if candidate in canonical:
+                    return candidate
```
ncloudioj: We can break early whenever getting a part that `len(part) < min_seg` 
```suggestion
            parts = []
            prev = 0
            cuts += (len(token),)  # a sentinel so that the tail can be appended inside of the loop
            for cut in cuts:
                if cut - prev < min_seg:
                    break
                parts.append(token[prev:cut])
                prev = cut
            else:
                # this block is only entered when the above loop exhausts all items without breaking
                candidate = " ".join(parts)
                if candidate in canonical:
                    return candidate
```

ncloudioj (commented):
Thanks for resolving all those review comments, we're almost there :)

Just added a couple of followup comments. Going to review the last bit of the pipeline (BM25) tomorrow. Thanks for your patience!

merino/utils/query_processing/normalization/pipeline.py
```diff
@@ -0,0 +1,406 @@
...
+    """Okapi BM25 over a fixed string corpus built once at startup."""
+
+    def __init__(self, corpus: list[str], k1: float = 1.5, b: float = 0.75) -> None:
+        """Build the BM25 inverted index."""
+        self.corpus = corpus  # keyword list (e.g. ["dow jones stock"])
+        self.k1 = k1  # frequency saturation, how much repeating words matter
+        self.b = b  # document length saturation, how much longer documents are penalized
+        self._tokenized = [doc.split() for doc in corpus]
+        self._corpus_set: set[str] = set(corpus)
+        n = len(self._tokenized)
```
ncloudioj: Nit: i think this reads clearer.

```suggestion
        n = len(self.corpus)
```

merino/utils/query_processing/normalization/pipeline.py
```diff
@@ -0,0 +1,406 @@
...
+        self._avgdl = sum(len(d) for d in self._tokenized) / max(
+            n, 1
+        )  # average document length across corpus
+        self._doc_lens = [len(d) for d in self._tokenized]
+
+        df: dict[str, int] = {}
```
ncloudioj: Nite: since you also use `Counter` below, it can be used here to be consistent.

```
df = Counter()
for doc in self._tokenized:
    df.update(set(doc))
```

merino/utils/query_processing/normalization/pipeline.py
```diff
@@ -0,0 +1,406 @@
...
+        # counts how many times each term appears in the corpus
+        for doc in self._tokenized:
+            for term in set(doc):
+                df[term] = df.get(term, 0) + 1
+
+        # scores rarer terms more highly
+        self._idf: dict[str, float] = {
+            term: math.log((n - cnt + 0.5) / (cnt + 0.5) + 1) for term, cnt in df.items()
+        }
+        # builds inverted index
```
ncloudioj: Can you also document the inverted index, especially the tuples in the value?

merino/utils/query_processing/normalization/pipeline.py
```diff
@@ -0,0 +1,406 @@
...
+            return None
+
+        # get top score and matching document index
+        top_idx = max(scores, key=lambda i: scores[i])
+        if scores[top_idx] == 0.0:
+            return None
+
+        # guard to make sure we have the same number of tokens
+        # for document and query
+        if q_sorted != sorted(self._tokenized[top_idx]):
```
ncloudioj: query token sorting can be moved down to here.
```suggestion
        if sorted(q_tokens) != sorted(self._tokenized[top_idx]):
```

merino/utils/query_processing/normalization/pipeline.py
```diff
@@ -0,0 +1,406 @@
...
+                    / (tf + self.k1 * (1 - self.b + self.b * dl / self._avgdl))
+                )
+                scores[doc_idx] = scores.get(doc_idx, 0.0) + score
+
+        if not scores:
+            return None
+
+        # get top score and matching document index
+        top_idx = max(scores, key=lambda i: scores[i])
+        if scores[top_idx] == 0.0:
```
ncloudioj: While the score is a real value (could even be negative), in which case could it be 0?

merino/utils/query_processing/normalization/pipeline.py
ncloudioj: Thanks! Shall we be even more aggressive? FWIW, the average query length is 14 and most of the canonical keywords are no more than 50 characters.

merino/web/api_v1.py
```diff
@@ -242,9 +247,20 @@ async def suggest(
...
+    pipeline = get_pipeline()
+    use_normalization = (
+        pipeline is not None and NORMALIZATION_CLIENT_VARIANT in client_variants_list
+    )
+    q_normalized = pipeline.normalize(q) if use_normalization and pipeline else q
+    normalization_providers = frozenset(settings.query_normalization.providers)
```
ncloudioj: Dynaconf can be slow when accessing deeply nested configs in a hot path, we can define a constant for this at the module level to mitigate that.

```
QUERY_NORM_PROVIDERS: fronzenset[str] = fronzenset(settings.query_normalization.providers)
```

ncloudioj (commented):
kk, finished the review with a few minor comments.

merino/utils/query_processing/normalization/pipeline.py
vazish: Nice, yep

merino/utils/query_processing/normalization/pipeline.py
vazish: Hmm it can't be 0 or negative here because of the +1 in `idf`... will remove.

vazish:
> kk, finished the review with a few minor comments.

Addressed!

merino/utils/api/metrics.py
```diff
@@ -23,3 +23,31 @@ def emit_suggestions_per_metrics(
...
+    the normalized query produced a suggestion (matched) and whether
+    the pipeline actually changed the query (query_changed). This lets
+    us monitor how often normalization fires and whether it leads to
+    new provider matches.
+    """
+    suggestion_providers = {s.provider for s in suggestions}
+    for provider in searched_providers:
+        if provider.name in normalization_providers:
+            metrics_client.increment(
+                "normalization.experiment.provider_match",
```
ncloudioj, line 47: We document all the metrics in metrics.yaml. Could you do that for those new metrics?

ncloudioj (approved):
Looks good, thanks!

Just one more ask to document the new metrics, after that, feel free to merge. 🚢

merino/utils/api/metrics.py
vazish: Yes let me add, thanks!

PR #1408, opened 2026-04-16
feat(logos): add logos resolution for sports and flights

kschelonka:
Add back the change -- reverting did not show any results for sports and airlines on stage. So I can't test this in stage but it doesn't seem to be the cause of no data

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/MC-2215)


tiftran (approved):
hah i get to review your pretty logos 😁

PR #1410, opened 2026-04-16
fix(logos): local gcs url resolution

kschelonka:
Updates the GCS emulator paths to treat "logos" like the bucket, so the returned URLs will resolve to the real file locally

Fixes tests that weren't being run before

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/MC-2217)


ncloudioj (approved):
Thanks!

PR #1413, opened 2026-04-17
feat: Use term translation dictionary to normalize discrepencies

jrconlin:
## References

JIRA: [DISCO-4049](https://mozilla-hub.atlassian.net/browse/DISCO-4049)

## Description

The provider is not consistent in term names or values. Ideally, each
`Sport` subclass encapsulates the logic, or peculiarities of the class,
allowing for the process to be as generic as possible.

This intoduces the `translate_terms` dictionary to help resolve
discrepencies between individual sports.

Closes DISCO-4049

## PR Review Checklist

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/MC-2220)

[DISCO-4049]: https://mozilla-hub.atlassian.net/browse/DISCO-4049?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


merino/providers/suggest/sports/backends/sportsdata/common/data.py
```diff
@@ -79,7 +85,7 @@ def from_data(cls, team_data: dict[str, Any], term_filter: list[str], team_ttl:
         name = team_data["Name"]
         fullname = team_data.get("FullName") or f"{locale} {team_data["Name"]}"
         logger.debug(f"{LOGGING_TAG} - Team: {fullname}")
-        team_id = team_data.get("GlobalTeamID", team_data.get("GlobalTeamId"))
```
kschelonka, line 92: 🫢

merino/providers/suggest/sports/backends/sportsdata/common/data.py
```diff
@@ -199,6 +205,11 @@ class Sport:
     # and ownership reasons.
     term_filter: list[str] = []
     cache_dir: str | None
+    # Each sport may use a different term for these values.
+    # You should prefer to use the `Global*` version when possible, but not all sports
+    # provide that value, nor do all returned data sets.
+    # This array is used by both Schedule and Score lookups.
+    translate_terms: dict = {}
```
kschelonka: comment(non-blocking): `translate` makes me think of language translations, what about `normalized_terms`, `canonical_keys`, or `term_lookup` etc.

tests/unit/providers/suggest/sports/backends/common/test_sports.py
```diff
@@ -329,6 +1258,9 @@ def _mk_team(key: str, name: str, locale: str, id: int) -> Team:
     )
 
 
+# = Start Tests: ====
+
+
```
kschelonka, line 1263: comment(non-blocking): maybe instead move the fixtures to another file and import them?

tests/unit/providers/suggest/sports/backends/common/test_sports.py
```diff
@@ -430,75 +1358,69 @@ async def test_nhl_update_teams_with_none_season(
...
     """Test ucl team updates."""
     ucl = UCL(settings=settings.providers.sports)
+    teams_payload = soccer_teams_payload()
     get_data = mocker.patch(
         "merino.providers.suggest.sports.backends.sportsdata.common.sports.get_data",
-        side_effect=[
-            nhl_nba_teams_payload,
-            nhl_nba_teams_payload,
-        ],  # called twice per code
+        side_effect=[teams_payload],  # called twice per code
```
kschelonka:
```suggestion
        side_effect=[teams_payload],
```

tests/unit/providers/suggest/sports/backends/common/test_sports.py
```diff
@@ -650,69 +1651,124 @@ async def test_nba_update_events(
...
     mocker: MockerFixture,
 ) -> None:
     """Test MLB event updates."""
     sport = MLB(settings=settings.providers.sports)
-    sport.load_teams_from_source(generic_teams_payload)
+    sport.load_teams_from_source(mlb_teams_payload())
     sport.season = "2026PRE"
     sport.event_ttl = timedelta(weeks=2)
 
+    # schedules_payload = mlb_schedule_payload()
```
kschelonka, line 1663: no schedules?

merino/providers/suggest/sports/backends/sportsdata/common/data.py
```diff
@@ -231,6 +242,16 @@ def __init__(
...
+        self.translate_terms = {
+            "GameID": "GameID",  # This value _MUST_ match between the Schedule and Sport if both are used.
+            "AwayTeamID": "AwayTeamID",
+            "AwayTeamKey": "AwayTeam",
+            "AwayTeamScore": "AwayTeamScore",
+            "HomeTeamID": "HomeTeamID",
+            "HomeTeamKey": "HomeTeam",
+            "HomeTeamScore": "HomeTeamScore",
+            "TeamID": "TeamID",
+        }.copy()
```
kschelonka: I don't think this copy is necessary

merino/providers/suggest/sports/backends/sportsdata/common/data.py
```diff
@@ -231,6 +242,16 @@ def __init__(
         self.team_ttl = team_ttl or timedelta(weeks=settings.get("team_ttl_weeks", TEAM_TTL_WEEKS))
         self.term_filter = term_filter
         self.cache_dir = cache_dir
+        self.translate_terms = {
```
kschelonka: Ah, ok so this is just a default term dictionary. I understand now.

I think since this class is meant to be subclassed, I'd suggest accepting term overrides in the constructor, and using it to build this dictionary with the defaults base. Rather than updating the superclass parameter when constructing the subclass.

tests/unit/providers/suggest/sports/backends/common/test_sports.py
```diff
@@ -739,3 +1795,12 @@ async def test_sportsdata_errors() -> None:
 
     error = SportsDataError("Foo")
     assert str(error) == "SportsDataError: Foo"
+
+
+def test_sport_term_pollution() -> None:
```
kschelonka, line 1800: a bit confused about this one, can you explain?

merino/providers/suggest/sports/backends/sportsdata/common/data.py
```diff
@@ -453,17 +459,18 @@ def load_schedules_from_source(
...
-                "GlobalHomeTeamId"
-            )
-            away_id = event_description.get("GlobalAwayTeamID") or event_description.get(
-                "GlobalAwayTeamId"
-            )
+            game_id = event_description[self.translate_terms["GameID"]]
+            home_id = event_description.get(self.translate_terms["HomeTeamID"])
+            away_id = event_description.get(self.translate_terms["AwayTeamID"])
+            # home_name = event_description.get(self.translate_terms["HomeTeamKey"])
+            # away_name = event_description.get(self.translate_terms["AwayTeamKey"])
```
kschelonka:
```suggestion
```

comment should be removed?

merino/providers/suggest/sports/backends/sportsdata/common/data.py
jrconlin: Yeah, it's weird. I'll go with `normalized_terms` since there appears to be no idea of canonical from the source and `term_lookup` doesn't quite feel right either.

tests/unit/providers/suggest/sports/backends/common/test_sports.py
jrconlin: Well, I'm not really a fan of making singleton fixtures, so I didn't do that. I'm also not a huge fan of hiding data, so having the data where the code is made sense to me. I would _LOVE_ to shuffle tests around so that each sport had it's tests lumped together, but that would be a nightmare to diff, so I didn't do that.

tests/unit/providers/suggest/sports/backends/common/test_sports.py
jrconlin: Thanks, legacy.

tests/unit/providers/suggest/sports/backends/common/test_sports.py
jrconlin: Turns out, for MLB, no. We only pull scores. It's a bit of a special case.

This may change at some point in the future so I am leaving this here for now.

merino/providers/suggest/sports/backends/sportsdata/common/data.py
jrconlin: Turns out, it is. The data gets handed off as being mutable so every instance gets it's own copy. Copilot [flagged](https://github.com/mozilla-services/merino-py/pull/1413#discussion_r3114256301) it initially, and I tested it. Sure enough it was right. The copy basically resolves that without doing other shenanigans.

merino/providers/suggest/sports/backends/sportsdata/common/data.py
jrconlin: I thought about that, but no. Each subclass would contain the overrides (since each subclass contains the business logic of how to deal with itself). Having the term dictionary in the `__init__` is cleaner because then we would just instantiate the class, without passing specialty items, which would make instantiation more complex. (Otherwise, you could argue that all special behavior belonged in parameters and that would make instantiation REALLY ugly fast.)

tests/unit/providers/suggest/sports/backends/common/test_sports.py
jrconlin: This is the test that shows you want to make sure that the `translate_terms` is copied from the superclass, rather than just updated. (How I wish that python had COW dictionaries)

merino/providers/suggest/sports/backends/sportsdata/common/data.py
```diff
@@ -195,6 +201,18 @@ def serialize(self) -> dict[str, Any]:
         return result
 
 
+SportNormalizedTerms = {
+    "GameID": "GameID",  # This value _MUST_ match between the Schedule and Sport if both are used.
```
ncloudioj: Suggestion: we can define all those key literals as a str enum to avoid typos as they are used throughout the module.

```python
from enum import StrEnum

class SportTerms(StrEnum):
    GAME_ID = "GameID"
    AWAY_TEAM_ID = "AwayTeamID"
    AWAY_TEAM_KEY = "AwayTeam"

SportNormalizedTerms = {
    GAME_ID = "GameID",
    ...
}
```

jrconlin: I would love to, but they're going to be specific to each sport, since each sport can have slight variations (e.g. "AwayTeamID" vs. "AwayTeamId"), or do you mean the "normalized" labels?

merino/providers/suggest/sports/backends/sportsdata/common/data.py
```diff
@@ -403,25 +415,22 @@ def load_scores_from_source(
...
-                        event_description, ["AwayTeamScore", "AwayScore", "AwayTeamRuns"]
-                    ),
+                    home_score=home_score,
+                    away_score=away_score,
                     status=GameStatus.parse(event_description["Status"]),
                     expiry=utc_time_from_now(self.event_ttl),
                     updated=updated,
                 )
-            self.events[event.id] = event
+                self.events[event.id] = event
```
ncloudioj: Was this indentation intended?
```suggestion
            self.events[event.id] = event
```

merino/providers/suggest/sports/backends/sportsdata/common/data.py
ncloudioj: No, i meant the those keys. As they've being used extensively elsewhere. Folks need to come back to make sure they use the right string literals, which can be error-prone. If each sport has its specific keys, we can define its own key enum for them.

```
# before
                home_id = event_description.get(self.normalized_terms["HomeTeamID"])
                away_id = event_description.get(self.normalized_terms["AwayTeamID"])
                home_name = event_description.get(self.normalized_terms["HomeTeamKey"])
                away_name = event_description.get(self.normalized_terms["AwayTeamKey"])
                home_score = event_description.get(self.normalized_terms["HomeTeamScore"])
                away_score = event_description.get(self.normalized_terms["AwayTeamScore"])

# after
                home_id = event_description.get(self.normalized_terms[SportsTermKeys.HOME_TEAM_ID])
                ...
```

ncloudioj (approved):
r+wc.

merino/providers/suggest/sports/backends/sportsdata/common/data.py
jrconlin: Ah! Good catch! Thanks.

merino/providers/suggest/sports/backends/sportsdata/common/data.py
kschelonka: python mutability has so many footguns

merino/providers/suggest/sports/backends/sportsdata/common/data.py
kschelonka: specifically it's the super attribute mutation which smells to me. I hear your point on constructor complexity but it makes a lot more sense to me to live as an option on the base class constructor. accepting difference in opinion

PR #1416, opened 2026-04-20
[DISCO-4086] feat(es): remove query retries and introduce es error metrics

kschelonka:
## References

JIRA: [DISCO-4086]

## Description
Troubleshooting performance for wikipedia, I discovered through tracing that 429 responses were being retried by the elasticsearch client, further overloading the cluster. This PR removes retries for reads altogether (due to low latency requirements and frequent querying on keystroke), and introduces additional metrics to understand and alert on elasticsearch performance.

In an ideal world both providers would use the same base and we could implement the metrics only once, but to avoid refactoring for now just add them in both clients on the relevant call paths.

Additional performance optimizations may be forthcoming, but this is a first start.

## PR Review Checklist

- [x]  `[load test: (abort|skip|warn)]` keywords are applied to the last commit message (if applicable)

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/MC-2223)

[DISCO-4086]: https://mozilla-hub.atlassian.net/browse/DISCO-4086?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/merino/providers/suggest/wikipedia/backends/elastic.py
+++ b/merino/providers/suggest/wikipedia/backends/elastic.py
@@ -86,7 +86,9 @@ class ElasticBackend:
             }
         }
 
-        self._metrics_client.increment(f"{ES_SEARCH_METRIC_NAME}.count", tags={"index": index_id})
+        # Add this if wikipedia gets more than one index; for now it's covered by the provider
+        # search count
+        # self._metrics_client.increment(f"{ES_SEARCH_METRIC_NAME}.count", tags={"index": index_id})
         try:
             res = await self.elasticsearch.search(
                 index=index_id,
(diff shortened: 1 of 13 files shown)


merino/providers/suggest/wikipedia/backends/elastic.py
```diff
@@ -76,14 +86,25 @@ async def search(self, q: str, language_code: str) -> list[dict[str, Any]]:
             }
         }
 
+        self._metrics_client.increment(f"{ES_SEARCH_METRIC_NAME}.count", tags={"index": index_id})
```
ncloudioj: This would overlap with the counter on `provider.search()` and there is only one index for Wikipedia, maybe we can save a counter here.

ncloudioj (approved):
LGTM.

PR #1417, opened 2026-04-21
[Disco-3946] chore: Upgrade Python 3.14 and dependencies 

ncloudioj:
## References

JIRA: [DISCO-3946](https://mozilla-hub.atlassian.net/browse/DISCO-3946)

## Description

Change highlights:
- Upgrade Merino to Python 3.14 (Happy Py Version!)
- Bump up dependency versions to adapt 3.14
- Bump Dockerfile accordingly
- Bump dockerflow
- Code lint/format changes from the latest ruff and mypy

Deprecations:
- FastAPI has deprecated "ORJSONResponse" to favor `response_model` and Pytantic model objects. As a result, all of its uses have been dropped in Merino.
- FastAPI now encourages consumers to set custom HTTP headers via the "temporal response object". This used to be done by specifying the `headers` argument of `ORJSONResponse` in Merino.

Other noticeable changes:
- Docker build takes much longer locally (40secs -> 140secs), especially for `uv sync ...`
- The docker image size also increases significantly (with torch takes 600MB and a bunch of other deps). We've seen this before when torch was first added to the project, couldn't figure out a way to de-bloat it yet.

To make the review easier, the RR consists of three logical parts, which can be followed for code review.

Migration for your local dev environment should be as simple as run `uv sync`.      

## PR Review Checklist

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/MC-2224)

[DISCO-3946]: https://mozilla-hub.atlassian.net/browse/DISCO-3946?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


Dockerfile
```diff
@@ -21,12 +21,13 @@ COPY . $APP_HOME
 
 # Install maxmind db and app dependencies. Clean up build tools after
 RUN apt-get update && \
-    apt-get install --yes build-essential libmaxminddb0 libmaxminddb-dev && \
+    apt-get install --yes build-essential libxml2-dev libxslt-dev python3-dev libffi-dev libmaxminddb0 libmaxminddb-dev && \
```
ncloudioj, line 24: Somehow, for 3.14, all those dependencies need to be installed for uv to build the dependencies.

pyproject.toml
```diff
@@ -179,10 +180,10 @@ filterwarnings = [
   "error",
   # Ignored due to https://github.com/encode/starlette/discussions/2603
   "ignore::pytest.PytestUnraisableExceptionWarning",
-  # Ignored as `robobro` still uses an old version of `werkzeug`
-  "ignore:'OrderedMultiDict' is deprecated",
```
ncloudioj, line 186: We don't use robobro any more.

tests/unit/providers/suggest/sports/backends/common/test_sports.py
```diff
@@ -373,7 +373,7 @@ async def test_nfl_update_teams(
 
     assert nfl.season == "2025REG"
     assert nfl.week == "3"
-    assert set(nfl.teams.keys()) == {28, 20}
+    assert set(nfl.teams.keys()) == {28, 20}  # type: ignore[unreachable]
```
ncloudioj: mypy claimed that this code is unreachable, no idea.

merino/configs/app_configs/config_logging.py
```diff
@@ -77,7 +77,7 @@ def configure_logging() -> None:
     )
 
 
-class GCPCompatibleJSONFormatter(dockerflow_logging.JsonLogFormatter):
+class GCPCompatibleJSONFormatter(dockerflow_logging.MozlogFormatter):
```
ncloudioj, line 80: JsonLogFormatter has been renamed to MozlogFormatter.

merino/curated_recommendations/protocol.py
```diff
@@ -263,7 +263,7 @@ class CuratedRecommendation(CorpusItem):
 
     __typename: TypeName = TypeName.RECOMMENDATION
 
-    experiment_flags: Annotated[set[str] | None, Field(default_factory=set, exclude=True)] = None
+    experiment_flags: Annotated[set[str] | None, Field(default_factory=set, exclude=True)]
```
ncloudioj, line 266: Pydantic now doesn't support having both `default` and `default_factory` for a field.

merino/web/api_v1.py
```diff
@@ -553,10 +550,8 @@ async def get_hourly_forecasts(
             # this failure before re-raising, so the breaker will still trip as expected.
             pass
 
-        return ORJSONResponse(
-            content=jsonable_encoder(hourly_forecasts),
-            headers={"Cache-Control": (f"private, max-age={ttl}")},
-        )
+        response.headers["Cache-Control"] = f"private, max-age={ttl}"
```
ncloudioj: @Herraj Notice a potential bug while refactoring the code here, when `.get_hourly_forecasts(...)` throws, this will return an empty response with a cache header set to 0. Would that confuse the client or affect how it retries?

merino/providers/suggest/sports/backends/sportsdata/protocol.py
```diff
@@ -28,7 +28,7 @@ class SportTeamDetail(BaseModel):
 def build_query(event: dict[str, Any]) -> str:
     """Build the search query from the event information"""
     date = datetime.fromisoformat(event["date"]).strftime("%d %b %Y")
-    return f"""{event.get("sport")} {event.get("away_team",{}).get("name","")} at {event.get("home_team", {}).get("name", "")} {date}"""
+    return f"""{event.get("sport")} {event.get("away_team", {}).get("name", "")} at {event.get("home_team", {}).get("name", "")} {date}"""
```
kschelonka, line 31: comment(non-blocking): noticing a lot of formatting, shouldn't this have been caught by our linter?

merino/web/api_v1.py
```diff
@@ -104,7 +107,7 @@ async def suggest(
     client_variants: str | None = Query(default=None, max_length=CLIENT_VARIANT_CHARACTER_MAX),
     sources: tuple[dict[str, BaseProvider], list[BaseProvider]] = Depends(get_suggest_providers),
     request_type: Annotated[str | None, Query(pattern="^(location|weather)$")] = None,
-) -> Response:
+) -> Any:
```
kschelonka: question(blocking): this seems like a type regression, can you explain why it's necessary?

merino/web/api_v1.py
```diff
@@ -302,15 +311,10 @@ def build_suggestion_response(client_variants, search_from, suggestions) -> Resp
...
-    # response headers
-    response_headers = {}
     # could be specific or default
     ttl = get_ttl_for_cache_control_header_for_suggestions(search_from, suggestions)
-    response_headers["Cache-control"] = f"private, max-age={ttl}"
-    return ORJSONResponse(
-        content=jsonable_encoder(response, exclude_none=True),
-        headers=response_headers,
-    )
+    response.headers["Cache-Control"] = f"private, max-age={ttl}"
```
kschelonka: comment(non-blocking): the mutation and var scope of `response` is confusing, and isn't understandable imo without additional context. Could you please update the documentation to be more like the router suggest endpoint method?

kschelonka: Personally I'd prefer to avoid mutating variables passed as args in functions, especially when they aren't returned and rely on python's shared memory references. I don't know if it's idiomatic in python but it seems like an easy way to introduce bugs. Is there an alternative to set the headers in the caller instead?

merino/web/api_v1.py
```diff
@@ -453,8 +457,10 @@ async def curated_content_legacy_fx_114(
     response_model=ManifestData,
 )
 async def get_manifest(
-    request: Request, provider: ManifestProvider = Depends(get_manifest_provider)
-) -> ORJSONResponse:
+    request: Request,
+    response: Response,
```
kschelonka: see https://github.com/mozilla-services/merino-py/pull/1417/files#r3125420442

(edit: there's a few more instances but I'm going to stop commenting on them specifically)

pyproject.toml
```diff
@@ -4,32 +4,32 @@ name = "merino-py"
 version = "0.1.0"
 description = "Service for Firefox Suggest at Mozilla"
 authors = [{ name = "Mozilla" }]
-requires-python = "~=3.13.0"
+requires-python = "~=3.14.0"
```
kschelonka, line 7: there's some pretty big bumps in here, would feel good about deploying into staging for QA first

merino/providers/suggest/sports/backends/sportsdata/protocol.py
ncloudioj: Yup, i noted in the PR description. Newer version of ruff seems more capable of formatting. All the formatting changes look reasonable to me.

merino/web/api_v1.py
ncloudioj: FastAPI has deprecated `ORJSONReponse` and recommends to return data of "Response Model" (`SuggestResponse` in this case). There is also a special case that the handler could return early with a `Response(status_code=204)`. Per [FastAPI's doc](https://fastapi.tiangolo.com/tutorial/response-model/#response-model-parameter),

> If you have strict type checks in your editor, mypy, etc, you can declare the function return type as Any.
>
> That way you tell the editor that you are intentionally returning anything. But FastAPI will still do the data documentation, validation, filtering, etc. with the response_model.

`Any` can be used to appease the type checker. FastAPI will do the right thing as long as `response_model` is declared in the router path operator.

The alternative is to replace `ORJSONResponse` with `JSONResponse` in `build_suggestion_response()` and manually decode the pydantic object to JSON, the following note of FastAPI made me choose the current approach:

> You will normally have much better performance using a [Response Model](https://fastapi.tiangolo.com/tutorial/response-model/) than returning a JSONResponse directly, as that way it serializes the data using Pydantic, in Rust.

I don't have a strong preference though, if you believe the latter is better, we can do that. Let me know.

merino/web/api_v1.py
ncloudioj: I agree it's odd to set the headers this way. It's actually a FastAPI pattern (for the new version of the framework) and the [recommended way](https://fastapi.tiangolo.com/advanced/response-headers/) to set custom headers and cookies for the response. The other option is to directly create responses and set the headers as arguments.

I've noted it the reason i opted to the "Response Model" in my last [comment](https://github.com/mozilla-services/merino-py/pull/1417/changes#r3125951298). If you prefer the other option, more than happy to switch over.

pyproject.toml
ncloudioj: Yup, we will. Already tested it locally w/ and w/o Docker, so far so good.

merino/web/api_v1.py
kschelonka: No that seems reasonable in context. I just really dislike `Any`

merino/web/api_v1.py
kschelonka: I dislike it, but if it's idiomatic and recommended then it makes sense. This relates more broadly to python linguistic feature and safety

kschelonka (approved):

kschelonka:
Let's make sure to put a hold on prod sync while we QA this

merino/web/api_v1.py
ncloudioj: You've convinced me! Not a fan of this idiomatic, either. Will switch back to the alternative option. It will address that `Any` type regression as well.

merino/web/api_v1.py
```diff
@@ -646,7 +654,7 @@ async def get_picture_of_the_day(
 )
 async def get_game_particle(
     request: Request, provider: ParticleProvider = Depends(get_particle_provider)
-) -> ORJSONResponse:
+) -> Response:  # pragma: no cover
```
ncloudioj, line 659: Will fix via https://mozilla-hub.atlassian.net/browse/HNT-2508

merino/web/api_v1.py
```diff
@@ -617,7 +624,7 @@ async def get_hourly_forecasts(
 async def get_picture_of_the_day(
     request: Request,
     provider: WikimediaPictureOfTheDayProvider = Depends(get_wikimedia_potd_provider),
-) -> ORJSONResponse:
+) -> Response:  # pragma: no cover
```
ncloudioj, line 629: Will fix via https://mozilla-hub.atlassian.net/browse/HNT-2509

Dockerfile
ncloudioj: The reason is that some dependencies (e.g. `mechanicsoup`) do not have pre-built wheels for 3.14, therefore uv needs to build them from source which in turn requires those dev dependencies installed first.

merino/web/api_v1.py
ncloudioj: @kschelonka okie doke, i've switched the code back to the old fashion way of sending response headers. Mind taking another look?

merino/web/api_v1.py
```diff
@@ -98,12 +98,15 @@
 WEATHER_PROVIDER = settings.providers.accuweather.backend
 SOURCE_TYPE = Literal["urlbar", "newtab", "unknown"]
 
+MANIFEST_TTL_SEC = settings.runtime.default_manifest_response_ttl_sec
```
jrconlin, line 101: Did you forget to use this in `get_manifest`?

merino/web/api_v1.py
```diff
@@ -338,8 +345,8 @@ def build_suggestion_response(client_variants, search_from, suggestions) -> Resp
     response_headers = {}
     # could be specific or default
     ttl = get_ttl_for_cache_control_header_for_suggestions(search_from, suggestions)
-    response_headers["Cache-control"] = f"private, max-age={ttl}"
-    return ORJSONResponse(
+    response_headers["Cache-Control"] = f"private, max-age={ttl}"
```
jrconlin, line 352: Heh, since you poked me about having constants for frequently used strings, I'm wondering if we should have a separate PR to replace these with constants as well. (Honestly a bit surprised that none of our dependencies declare this.)

tests/integration/jobs/test_cli.py
```diff
@@ -8,12 +8,6 @@
 runner = CliRunner()
 
 
-def test_cli_no_args():
```
jrconlin, line 11: I take it this is no longer a concern?

jrconlin (approved):
r+ w/?

merino/web/api_v1.py
ncloudioj: Thanks, the change got undone in a rebase.

merino/web/api_v1.py
ncloudioj: Yup, we could do that. They don't provide this maybe because http headers are case insensitive?

tests/integration/jobs/test_cli.py
ncloudioj: Correct.

tests/unit/providers/suggest/sports/backends/common/test_sports.py
```diff
@@ -1303,7 +1303,7 @@ async def test_nfl_update_teams(mock_client: AsyncClient, mocker: MockerFixture)
 
     assert nfl.season == "2025REG"
     assert nfl.week == "3"
-    assert set(nfl.teams.keys()) == {1, 2}
+    assert set(nfl.teams.keys()) == {1, 2}  # type: ignore[unreachable]
```
ncloudioj, line 1306: This appears to be a false alarm. Going to ignore it for now.

PR #1423, opened 2026-04-22
DISCO-4032 Adding ETag to /manifest endpoint

gruberb:
## References

JIRA: [DISCO-4032](https://mozilla-hub.atlassian.net/browse/DISCO-4032)

## Description
We want to make the `/manifest` endpoint being able to be called as cheap as possible. The first step is to pass an `ETag` to the endpoint, so we can check if the Manifest file changed or not. This allows clients to store the `ETag` in the response, and send it back in the following requests. If the `ETag` is the same, we know the file hasn't changed. 

## PR Review Checklist

[DISCO-4032]: https://mozilla-hub.atlassian.net/browse/DISCO-4032?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/merino/web/api_v1.py
+++ b/merino/web/api_v1.py
@@ -7,7 +7,7 @@ from itertools import chain
 from typing import Annotated, Literal
 
 from asgi_correlation_id.context import correlation_id
-from fastapi import APIRouter, Depends, Query, Header
+from fastapi import APIRouter, Depends, Header, HTTPException, Query, status
 from fastapi.encoders import jsonable_encoder
 from fastapi.responses import ORJSONResponse
 from starlette.requests import Request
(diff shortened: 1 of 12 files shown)


merino/web/api_v1.py
```diff
@@ -470,22 +479,34 @@ async def get_manifest(
...
-                },
+                headers=headers,
             )
 
         metrics_client.increment("manifest.request.error")
         logger.error("Manifest file not found")
         return ORJSONResponse(
             content=jsonable_encoder(manifest_data),
-            status_code=404,
+            status_code=status.HTTP_404_NOT_FOUND,
```
ncloudioj, line 536: Q: when 404 is returned, `manifest_data` should be None, correct? So we are returning a JSON string `null` to the client.

ncloudioj (approved):

merino/web/api_v1.py
gruberb: Hm, yes. But now thinking through it again, we could also raise a `HTTPException`:
```python
raise HTTPException(
    status_code=status.HTTP_404_NOT_FOUND,
    detail="Manifest not found",
)
```

ncloudioj: Would 404 confuse the consumer especially if the manifest is just temporarily unavailable? I wonder if 204 is more suitable in this case.

gruberb: I think this case would probably not happen, but the most accurate would likely be `503: Service Unavailable`. Which truly reflects that: There should be something, but we don't have it right now.

`404`, as you pointed out, means the consumer is wrong to call this endpoint. And `204` means no content, which is more like for a `PUT` or `DELETE.

ncloudioj: One downside of using 5xx errors is that many other infra pieces could return them during service degradation and outages and we have alerts specifically for them. If we can return 503 here and the request volume is significant, the alerts will be triggered. Though this is not the service disruption that we should take immediate actions, correct?

Re: `204`, actually, it's been widely used between Firefox and our services including Merino to indicate the service has no content for a particular request. For example, Merino would return 204 if it couldn't fetch weather forecast for a user (could be geolocation lookup failed or the vendor doesn't have data for the location).

As i thought about this, maybe 404 with a message would be the better way to convey this like you suggested above.

PR #1425, opened 2026-04-22
DISCO-4108 - Add key to missing logo metrics

gruberb:
## References

JIRA: [DISCO-4108](https://mozilla-hub.atlassian.net/browse/DISCO-4108)

## Description
In Grafana, we see that we are  missing some logos for different sports. But it would be good to see which teams we are missing to fetch.

Screenshot 2026-04-22 at 15 03 08

## PR Review Checklist

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/MC-2232)

[DISCO-4108]: https://mozilla-hub.atlassian.net/browse/DISCO-4108?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/merino/utils/logos.py
+++ b/merino/utils/logos.py
@@ -74,8 +74,12 @@ def get_logo_url(category: LogoCategory, key: str) -> Optional[HttpUrl]:
     logo = load_manifest().get(category, key)
     if logo is None:
         logger.warning(f"Logo does not exist for category={category} and key={key}")
+        # Normalize to the manifest's uppercase storage form so "aa" and "AA"
+        # collapse to a single series. Cardinality is bounded by the known key
+        # spaces (IATA codes, sports team keys), so this is safe for Prometheus.
         metrics_client.increment(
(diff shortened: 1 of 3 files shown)


ncloudioj (approved):

kschelonka:
comment(non-blocking): we log these already -- it seems like it would be easier to just pull them from the logs? Or I'd suggest using Sentry because then we could be notified and connect the event to jira easily.

gruberb:
Looking at this again, I agree with @kschelonka. We should probabaly log an `error` instead of a `warning` and get notified via Sentry. This metric I think is not helpful, since we should never have a miss, and if we do, we need to fix it right away.

gruberb:
@kschelonka I added a sentry error for missing logos.

PR #1427, opened 2026-04-22
[DISCO-4071]Add ingestion of kw/advertiser engagement data

tiftran:
## References

JIRA: [DISCO-4071](https://mozilla-hub.atlassian.net/browse/DISCO-4071)

## Description
Ingestion of the keyword/advertiser engagement data. There a lot of duplication with the engagement advertiser level stuff, but it would probs be removed in [4072](https://mozilla-hub.atlassian.net/browse/DISCO-4072)

## PR Review Checklist

[DISCO-4071]: https://mozilla-hub.atlassian.net/browse/DISCO-4071?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/merino/providers/suggest/adm/backends/protocol.py
+++ b/merino/providers/suggest/adm/backends/protocol.py
@@ -24,6 +24,28 @@ class EngagementData(BaseModel):
     amp_aggregated: dict[str, int]
 
 
+class KeywordMetrics(BaseModel):
+    """Impressions and clicks for a single time window."""
+
+    impressions: int
+    clicks: int
+
+
+class KeywordEntry(BaseModel):
+    """Live and historical engagement metrics for a single advertiser/keyword pair."""
+
+    live: KeywordMetrics
+    historical: KeywordMetrics
+
+
+class KeywordEngagementData(BaseModel):
+    """Model for keyword-level engagement data file content."""
+
+    amp: dict[str, KeywordEntry] = {}
+    amp_aggregated: dict[str, int] = {}
+    wiki_aggregated: dict[str, int] = {}
+
+
 class SuggestionContent(BaseModel):
     """Class that holds the result from a fetch operation."""
 
--- a/merino/providers/suggest/adm/backends/protocol.py
+++ b/merino/providers/suggest/adm/backends/protocol.py
@@ -1,7 +1,7 @@
 """Protocol for the AdM provider backends."""
 
 from enum import Enum
-from typing import Protocol
+from typing import Optional, Protocol
 
 from pydantic import BaseModel, ConfigDict
 from moz_merino_ext.amp import AmpIndexManager
@@ -34,8 +34,8 @@ class KeywordMetrics(BaseModel):
 class KeywordEntry(BaseModel):
     """Live and historical engagement metrics for a single advertiser/keyword pair."""
 
-    live: KeywordMetrics
-    historical: KeywordMetrics
+    live: Optional[KeywordMetrics]
+    historical: Optional[KeywordMetrics]
 
 
 class KeywordEngagementData(BaseModel):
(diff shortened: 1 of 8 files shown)


merino/providers/suggest/adm/backends/protocol.py
```diff
@@ -24,6 +24,28 @@ class EngagementData(BaseModel):
...
+
+    impressions: int
+    clicks: int
+
+
+class KeywordEntry(BaseModel):
+    """Live and historical engagement metrics for a single advertiser/keyword pair."""
+
+    live: KeywordMetrics
+    historical: KeywordMetrics
```
misaniwere: these should both be optional

merino/providers/suggest/adm/backends/protocol.py
```diff
@@ -24,6 +24,28 @@ class EngagementData(BaseModel):
...
+
+    impressions: int
+    clicks: int
+
+
+class KeywordEntry(BaseModel):
+    """Live and historical engagement metrics for a single advertiser/keyword pair."""
+
+    live: Optional[KeywordMetrics]
+    historical: Optional[KeywordMetrics]
```
misaniwere: What do you think about defaulting each field to `None` and adding a validator to ensure at lease one is present?

```suggestion
    live: Optional[KeywordMetrics] = None
    historical: Optional[KeywordMetrics] = None
    
    @model_validator(mode="after")
    def at_least_one_present(self) -> "KeywordEntry":
        if self.live is None and self.historical is None:
            raise ValueError("at least one of live or historical must be set")
        return self
```

merino/providers/suggest/adm/backends/protocol.py
```diff
@@ -24,6 +24,28 @@ class EngagementData(BaseModel):
...
+
+
+class KeywordEngagementData(BaseModel):
+    """Model for keyword-level engagement data file content."""
+
+    amp: dict[str, KeywordEntry] = {}
+    amp_aggregated: dict[str, int] = {}
+    wiki_aggregated: dict[str, int] = {}
```
misaniwere: should we omit this field since this is the `adm` file?

merino/utils/gcs/engagement/filemanager.py
```diff
@@ -24,6 +24,28 @@ class EngagementData(BaseModel):
...
+    historical: KeywordMetrics
+
+
+class KeywordEngagementData(BaseModel):
+    """Model for the keyword-level engagement data file stored in GCS."""
+
+    amp: dict[str, KeywordEntry] = {}
+    amp_aggregated: dict[str, int] = {}
+    wiki_aggregated: dict[str, int] = {}
+
```
misaniwere, line 34: these are already defined in `protocol.py`

misaniwere (changes requested):
looks good!  just a few changes below

merino/utils/gcs/engagement/filemanager.py
tiftran: lol i noticed that `EngagementData` is also defined again above. Is there a special case for that? i was just trying to follow convention 😅

misaniwere: oh true. so the filemanager `EngagementData` is defining the model for the full file that we are getting from GCS so we can validate the model in `get_file` and then the specific fields are defined in `adm/protocol` (`amp` and `amp_aggregated`) and `wikipedia/protocol` (`wiki_aggregated`) which is used by each of their providers.

tiftran: sound like i'm dropping `wiki_aggregated` in `protocol.py` and keeping this then

misaniwere: yeah and we can pull in `KeywordEntry` and `KeywordMetrics` from `amp/protocol.py` since those are amp related and already defined there, so we don't have them defined twice

misaniwere (approved):
Looks good!

PR #1432, opened 2026-04-23
feat: [DISCO-4076] Add FIFA support for Sports Suggestions

jrconlin:
## References

JIRA: [DISCO-4076](https://mozilla-hub.atlassian.net/browse/DISCO-4076)

## Description

Add FIFA support to Merino Suggest

Closes DISCO-4076

## PR Review Checklist

[DISCO-4076]: https://mozilla-hub.atlassian.net/browse/DISCO-4076?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


merino/providers/suggest/sports/backends/sportsdata/common/sports.py
```diff
@@ -670,6 +671,103 @@ async def update_events(self, client: AsyncClient):
...
+                SportTerms.COLOR2: "ClubColor2",
+                SportTerms.COLOR3: "ClubColor3",
+                SportTerms.COLOR4: "ClubColor4",
+            }
+        )
+
+    async def get_season(self, client: AsyncClient):
+        """Get the current season (which is just the current year)"""
+        self.season = str(datetime.now(tz=timezone.utc).year)
+        return self
```
kschelonka:
```suggestion
        return self.season
```

Should this getter return the season?

merino/providers/suggest/sports/backends/sportsdata/common/sports.py
```diff
@@ -41,7 +41,8 @@
     "NBA": SportCategory.Basketball,
     "UCL": SportCategory.Soccer,
     "MLB": SportCategory.Baseball,
-    "EPL": SportCategory.Soccer,
+    # "EPL": SportCategory.Soccer,
+    "FIFA": SportCategory.Soccer,
```
ncloudioj: Per the legal guidance, let's avoid all FIFA trademarks and IPs in our code. Favour generic terms such as "World Tournament (WT)".

merino/providers/suggest/sports/backends/sportsdata/common/sports.py
```diff
@@ -670,6 +671,103 @@ async def update_events(self, client: AsyncClient):
         return self
 
 
+class FIFA(Sport):
```
ncloudioj: Ditto

merino/providers/suggest/sports/backends/sportsdata/common/sports.py
```diff
@@ -670,6 +671,103 @@ async def update_events(self, client: AsyncClient):
...
+                SportTerms.HOME_TEAM_KEY: "HomeTeamKey",
+                SportTerms.TEAM_ID: "GlobalTeamId",
+                SportTerms.COLOR1: "ClubColor1",
+                SportTerms.COLOR2: "ClubColor2",
+                SportTerms.COLOR3: "ClubColor3",
+                SportTerms.COLOR4: "ClubColor4",
+            }
+        )
+
+    async def get_season(self, client: AsyncClient):
```
ncloudioj, line 674: Let's annotate the return type and drop `async` from the method signature.

merino/providers/suggest/sports/backends/sportsdata/common/sports.py
```diff
@@ -670,6 +671,103 @@ async def update_events(self, client: AsyncClient):
...
+        return self
+
+    async def get_team(self, id: int) -> Team | None:
+        """Fetch a team from the thread locked source"""
+        async with self._lock:
+            return self.teams.get(id)
+
+    async def update_teams(self, client: AsyncClient):
+        """Fetch active team information"""
+        logger = logging.getLogger(__name__)
```
ncloudioj: Nit: by convention for this codebase, `logger` can be initialized once at the module level and then used throughout the module.

merino/providers/suggest/sports/backends/sportsdata/common/sports.py
```diff
@@ -670,6 +671,103 @@ async def update_events(self, client: AsyncClient):
...
+        await self.get_season(client=client)
+        logger.debug(f"{LOGGING_TAG} Getting {self.name} teams ")
+        url = f"{self.base_url}/Teams/{self.name.lower()}?key={self.api_key}"
```
ncloudioj: If the request fails, the API key could be exposed in logs or other error reports. All the methods of the HTTP client take an argument `params` for query params. You can send a query param dict to `get_data()` to mitigate that.

merino/providers/suggest/sports/backends/sportsdata/common/sports.py
```diff
@@ -670,6 +671,103 @@ async def update_events(self, client: AsyncClient):
...
+
+        response = await get_data(
+            client=client,
+            url=url,
+            ttl=timedelta(hours=4),
+            cache_dir=self.cache_dir,
+        )
+        async with self._lock:
+            self.load_teams_from_source(response)
+        return self
```
ncloudioj: Is `self` returned for chaining? If so, plz annotate it accordingly.

merino/providers/suggest/sports/backends/sportsdata/common/sports.py
```diff
@@ -670,6 +671,103 @@ async def update_events(self, client: AsyncClient):
...
+
+    async def update_events(self, client: AsyncClient, allow_no_teams: bool = False):
+        """Update schedules and game scores for this sport"""
+        await self.get_season(client=client)
+        logger = logging.getLogger(__name__)
+        logger.debug(f"{LOGGING_TAG} Getting {self.name} schedules")
+        url = f"{self.base_url}/SchedulesBasic/{self.name}/{self.season}?key={self.api_key}"
+        local_timezone = ZoneInfo("UTC")
+        response = await get_data(
```
ncloudioj, line 705: Just noticed that all `get_data()` calls in this patch didn't capture exceptions on HTTP request failures, would that be handled by the callers of `update_events()` and `update_teams()`? If so, could you note that in the docstring for future reference?

merino/providers/suggest/sports/backends/sportsdata/common/sports.py
```diff
@@ -670,6 +671,103 @@ async def update_events(self, client: AsyncClient):
...
+            day = event.date.strftime("%Y-%b-%d").upper()
+            if not event.status.is_scheduled() and day not in date_list:
+                url = f"{self.base_url}/GamesByDate/{self.name}/{day}?key={self.api_key}"
+                response = await get_data(
+                    client=client,
+                    url=url,
+                    ttl=timedelta(minutes=5),
+                    cache_dir=self.cache_dir,
+                )
+                self.load_scores_from_source(
```
ncloudioj, line 727: Blocking: this could raise an exception of mutating an object inside of its iterator since `self.load_scores_from_source()` could mutate `self.events` [here](https://github.com/mozilla-services/merino-py/blob/main/merino/providers/suggest/sports/backends/sportsdata/common/data.py#L584).

ncloudioj (changes requested):

merino/providers/suggest/sports/backends/sportsdata/common/sports.py
jrconlin: We never actually use the return result, since it modifies the value contained in "self". Having this return a reference to `self` would allow us to chain these together.

merino/providers/suggest/sports/backends/sportsdata/common/sports.py
jrconlin: I use `async` here mostly for consistency, since other implementations may require a REST call. This is why the `@abstractmethod` definition in `class Sport` is `async`.  None of the calls use the value that is returned, so technically this function should not return any value, but I feel returning `self` here is safer, and allows us to potentially chain functions.

merino/providers/suggest/sports/backends/sportsdata/common/sports.py
jrconlin: This replicates the pattern from other classes. I can fix this to be a global, but I would need to update all other classes in order to maintain that consistency.

merino/providers/suggest/sports/backends/sportsdata/common/sports.py
jrconlin: No, it should be handled in the `get_data` since that's the bottleneck for these things. I assumed that those exceptions would percolate out to the middleware, is that not correct?

ncloudioj: `response.raise_for_status()` in `get_data()` could raise.

Yes, if this runs in the Merino's web app, FastAPI and Sentry will capture the exception(s) and turn them into HTTP errors (this is the main reason to not hardcode the API key in the URL as it could be included in the Sentry report). If this runs in the Airflow job, no middleware available there, unhandled exceptions will simply crash the job.

merino/providers/suggest/sports/backends/sportsdata/common/sports.py
jrconlin: Changing this to WCS for now.

merino/providers/suggest/sports/backends/sportsdata/common/sports.py
jrconlin: Not sure that's a bad thing. If there's an error fetching content such that it results in a failed call, processing should stop, since no other calls would work, and we would run the risk of corrupting data.

merino/providers/suggest/sports/backends/sportsdata/common/sports.py
jrconlin: Wrapping the `.items()` in a `list()` should solve for this.

merino/providers/suggest/sports/backends/sportsdata/common/sports.py
kschelonka: This method is only called by airflow though right? It's not served by merino http publicly.

ncloudioj: Yup, i believe so.

Okay, if failing soon is preferred, feel free to leave it this way.

ncloudioj (approved):
LGTM, thanks!

jrconlin:
Huh, just noticed that "merge when ready" doesn't autosquash. That explains a few things.

PR #1437, opened 2026-04-27
fix: Do not process unnecessary sporting events

jrconlin:
Closes: DISCO-4123

## References

JIRA: [DISCO-4123](https://mozilla-hub.atlassian.net/browse/DISCO-4123)

## Description
Skip Canceled , NotNecessary events (These can cause the recorded events to be changed.)

## PR Review Checklist

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/MC-2244)

[DISCO-4123]: https://mozilla-hub.atlassian.net/browse/DISCO-4123?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/merino/providers/suggest/sports/backends/sportsdata/common/data.py
+++ b/merino/providers/suggest/sports/backends/sportsdata/common/data.py
@@ -382,6 +382,9 @@ class Sport:
         end_window = datetime.now(tz=timezone.utc) + self.event_ttl
         for event_description in data:
             game_id = event_description[self.normalized_terms[SportTerms.GAME_ID]]
+            status = GameStatus.parse(event_description["Status"])
+            if status in [GameStatus.NotNecessary, GameStatus.Canceled]:
+                continue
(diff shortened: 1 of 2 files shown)


Herraj (approved):

kschelonka (commented):
I won't block the fix on this, but I'd like to see a regression test. From reading the error it seems like mutability is the primary culprit -- I'd also consider refactoring in the future to avoid mutating the initial state

jrconlin:
Mutability is the source of the problem, but determining the source of the mutability was also important.
In this case, it was the fact that events that are not necessary may appear in "Scores", but not "Schedules".
As a safeguard, extracting a list of the events rather than iterating over them should prevent similar things from happening in the future.

PR #1442, opened 2026-04-27
fix: Have MLB check schedules

jrconlin:
## References

JIRA: [DISCO-4124](https://mozilla-hub.atlassian.net/browse/DISCO-4124)

## Description
Previously MLB would error on schedule checks, but the normalize_terms should help.

## PR Review Checklist

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/MC-2249)

[DISCO-4124]: https://mozilla-hub.atlassian.net/browse/DISCO-4124?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/merino/providers/suggest/sports/backends/sportsdata/common/data.py
+++ b/merino/providers/suggest/sports/backends/sportsdata/common/data.py
@@ -383,7 +383,7 @@ class Sport:
         for event_description in data:
             game_id = event_description[self.normalized_terms[SportTerms.GAME_ID]]
             status = GameStatus.parse(event_description["Status"])
-            if status in [GameStatus.NotNecessary, GameStatus.Canceled]:
+            if status in [GameStatus.Canceled, GameStatus.NotNecessary]:
                 continue
             # only update the scores.
             if game_id in self.events:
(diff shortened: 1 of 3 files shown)


merino/providers/suggest/sports/backends/sportsdata/common/sports.py
```diff
@@ -615,58 +615,33 @@ async def get_team(self, id: int) -> Team | None:
...
             client=client,
             url=url,
             ttl=timedelta(minutes=5),
             cache_dir=self.cache_dir,
         )
-        self.load_scores_from_source(response, event_timezone=local_timezone)
+        self.load_schedules_from_source(response, event_timezone=local_timezone)
+        date_list = []
+        # update scores based on events:
+        # Events may cross multiple days, so we should update those scores.
```
kschelonka, line 631: ⁉️

kschelonka (approved):

merino/providers/suggest/sports/backends/sportsdata/common/data.py
```diff
@@ -400,7 +400,9 @@ def load_scores_from_source(
                 away_id = event_description.get(self.normalized_terms[SportTerms.AWAY_TEAM_ID])
                 home_name = event_description.get(self.normalized_terms[SportTerms.HOME_TEAM_KEY])
                 away_name = event_description.get(self.normalized_terms[SportTerms.AWAY_TEAM_KEY])
-                logger.warning(f"Adding game...{away_name} at {home_name} :: {status}")
+                logger.warning(
```
kschelonka, line 403: this feels more like a debug or info

jrconlin: Possibly, but it's also potentially anomalous, since it's an event that didn't appear in `/schedules`, and also impact the list of events for that sport, which could potentially cause suggestions around that event to behave oddly. 

I'd like to follow this PR up with a set of tests for this situation so that we can better understand the implications, but I don't have the bandwidth to do that right now, thus the warning.

merino/providers/suggest/sports/backends/sportsdata/common/sports.py
jrconlin: Yeah. Basically, baseball games are logged by inning. This means that they are not time bound. (In addition, you can get things like weather delays introduced into an event, which can cause play to begin hours later.)
This can lead to very, very long games.

PR #1451, opened 2026-04-30
[DISCO-4080] WCS Publish: List of teams

Herraj:
## References

JIRA: [DISCO-4080](https://mozilla-hub.atlassian.net/browse/DISCO-4080)

## Description
Adding the `.../wcs/teams` endpoint that serves static teams data to unblock UI development.

## PR Review Checklist

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/MC-2258)

[DISCO-4080]: https://mozilla-hub.atlassian.net/browse/DISCO-4080?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


merino/data/wcs_teams.json
```diff
@@ -0,0 +1,1490 @@
+[
```
Herraj, line 1: This list was copied from [here](https://drive.google.com/drive/folders/1OqBKCN4eAwtu8PfHbzlrvlxl5oSgrj1m). It originally had 64 entries but I removed the teams that weren't in the tournament and also added a `Group` key for each team.

merino/providers/wcs/fake_data.py
```diff
@@ -58,6 +60,41 @@ def _team(
 }
 
 
+def _team_from_json(entry: dict) -> TeamInfo:
```
kschelonka, line 63: Recommend adding some comments to clearly indicate that this is static data for UI development

kschelonka (approved):

merino/data/wcs_teams.json
jrconlin: I thought we weren't adding `group`? As noted, that data is not easily available in the SportsData feed. We can add it if required, but it will require special handling.

merino/providers/wcs/fake_data.py
```diff
@@ -58,6 +60,41 @@ def _team(
...
+    return TeamInfo(
+        key=entry["Key"],
+        global_team_id=entry["GlobalTeamId"],
+        name=entry["Name"],
+        region=entry["Key"],
+        colors=colors,
+        icon_url=_icon(entry["Key"]),
+        group=entry["Group"],
+        eliminated=False,
+        standing={"wins": 0, "losses": 0, "draws": 0, "points": 0},
```
jrconlin: I thought that we were not returning Standing. 
Again, this is not easily available data. We can get this data, but it requires additional calls.

jrconlin (changes requested):

merino/providers/wcs/fake_data.py
Herraj: I can remove `standing` as I see it is marked as optional in the api contract.

merino/data/wcs_teams.json
Herraj: For `group`, I do see the it in the api contract. Should I just return an empty string for it, for now? I can double check the UI designs to see if it is needed.

Herraj: looks like the UI does show the group for the group stage matches 🤔

Herraj: I removed the `group` key from the response.

jrconlin: Also a bit curious about one other thing.
That list was taken from [the SportsData.io endpoint](https://sportsdata.io/developers/api-documentation/soccer#team-profiles--by-competition) That's a reasonable source of truth for this data, and I'm ok with not having to fetch it constantly.
I already have some code that parses the result, and stores the data as a dictionary keyed by the GlobalTeamId into the Redis store, but it could easily live as a python dictionary.  
If we're going to cache that data, doesn't it make sense to do the work to reduce and index the cache data rather than replicate the index build every time we start?

(for that matter, there are a number of fields here that we will never use, which would just take up space and processing time. we should probably reduce things so that the load times are better.)

jrconlin: FWIW: Here's the WIP for the WCS/Redis caching stuff: https://github.com/mozilla-services/merino-py/pull/1453/changes

jrconlin: Nevermind, getting ahead of myself.
This is disposable code for the UI.

jrconlin (approved):

PR #1453, opened 2026-04-30
Feat/[DISCO-413] WCS in Redis support

jrconlin:
## References

JIRA: [DISCO-4131](https://mozilla-hub.atlassian.net/browse/DISCO-4131)

## Description

This PR contains the WIP for the WCS widget support. 

Currently, this focuses on expanding the RedisAdapter, as well as adding some methods and caching for retrieving and storing data. It is very much still a work in progress. 

## PR Review Checklist

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/MC-2260)

[DISCO-4131]: https://mozilla-hub.atlassian.net/browse/DISCO-4131?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


merino/cache/redis.py
```diff
@@ -127,6 +129,199 @@ async def scard(self, key: str) -> int:
...
+            ) from exc
+
+    async def zrem(
+        self,
+        key: str,
+        *field: str,
+    ) -> int:
+        """Remove a field from a zrange key"""
+        try:
+            return await self.replica.zrem(
```
gruberb: Isn't the replica read-only? This would fail in production, right? We should do this operation on `self.primary`?

merino/cache/redis.py
```diff
@@ -127,6 +129,199 @@ async def scard(self, key: str) -> int:
...
+
+    async def zremrangebyscore(
+        self,
+        key: str,
+        min: int,
+        max: int,
+    ) -> int:
+        """Remove any values that fall between the min and max inclusively"""
+        try:
+            return await self.replica.zremrangebyscore(
```
gruberb: Isn't the replica read-only? This would fail in production, right? We should do this operation on `self.primary`?

merino/providers/suggest/sports/backends/sportsdata/common/sports.py
```diff
@@ -671,15 +689,156 @@ def __init__(self, settings: LazySettings, *args, **kwargs):
...
+        fact that they used the same key code for two different countries).
+
+        """
+        meta_data = await self.cache.hgetall(f"{self.cache_prefix}:meta")
+
+        # Has it been over a year since we last updated the meta info?
+        lock_period = 365
+        if not force and meta_data:
+            updated = int(meta_data.get(b"meta_updated") or 0)
+            if updated and updated < int(
```
gruberb: Should be `>` instead of `<`, here, right? We want to skip if it's less than 365?

merino/providers/suggest/sports/backends/sportsdata/common/sports.py
```diff
@@ -671,15 +689,156 @@ def __init__(self, settings: LazySettings, *args, **kwargs):
...
+                (datetime.now() - timedelta(days=lock_period)).timestamp()
```
gruberb: `datetime.now()` returns a naive datetime in the "local" time. Then later we do a `datetime.now(tz)` with a timestamp. I think the datetimes are all over the place. We should maybe make it clear which timezone we expect and use it everywhere?

`int(datetime.now(tz=timezone.utc).timestamp())`

merino/providers/suggest/sports/backends/sportsdata/common/sports.py
```diff
@@ -671,15 +689,156 @@ def __init__(self, settings: LazySettings, *args, **kwargs):
...
+            ):
+                return
+        now = datetime.now(tz=timezone.utc)
+        mylock = int((now + timedelta(seconds=30)).timestamp())
+        meta_key = f"{self.cache_prefix}:meta"
+        prev = await self.cache.hget(meta_key, "lock")
+        if force or (prev is not None and int(prev) < int(now.timestamp())):
```
gruberb: I think this could be a race condition here, where multiple workers acquire a stale lock. 

We could use the native Redis `SET NX EX` (https://redis.io/docs/latest/develop/clients/patterns/distributed-locks/) in the `set` function in `merin/cache/redis.py`:

```python
 async def set(
      self,
      key: str,
      value: bytes | str,
      ttl: timedelta | None = None,
      nx: bool = False,
  ) -> bool:
      """Set a key. Returns True on write, False when `nx` and the key existed."""
      try:
          result = await self.primary.set(
              key,
              value,
              ex=int(ttl.total_seconds()) if ttl else None,
              nx=nx,
          )
          return bool(result)
      except RedisError as exc:
          raise CacheAdapterError(f"Failed to SET `{key!r}` with error: `{exc}`") from exc


  async def delete(self, key: str) -> int:
      """Delete a key. Returns 1 if it existed."""
      try:
          return await self.primary.delete(key)
      except RedisError as exc:
          raise CacheAdapterError(f"Failed to DEL `{key!r}` with error: `{exc}`") from exc
```


In `merin/cache/none.py`:
```python
  async def set(
      self,
      key: str,
      value: bytes | str,
      ttl: timedelta | None = None,
      nx: bool = False,
  ) -> bool:
      return True


  async def delete(self, key: str) -> int:
      return 0
```

And then:

```python
lock_key = f"{self.cache_prefix}:meta:lock"
lock_ttl = timedelta(minutes=10)
token = uuid.uuid4().hex

acquired = await self.cache.set(lock_key, token, ttl=lock_ttl, nx=True)
if not acquired:
    return

try:
    await self.load_areas(client)
    await self.update_teams(client)
    await self.cache_teams()
    await self.cache.hset(
        meta_key,
        {"meta_updated": int(datetime.now(tz=timezone.utc).timestamp())},
    )
finally:
    await self.cache.delete(lock_key)
```


Haven't ran this locally, but roughly that should work?

merino/cache/redis.py
```diff
@@ -127,6 +129,199 @@ async def scard(self, key: str) -> int:
...
+                min=min,
+                max=max,
+            )
+        except RedisError as exc:
+            raise CacheAdapterError(
+                f"Failed to {cast(types.FrameType, inspect.currentframe()).f_code.co_name.upper()} {key} with error: {exc}"
+            ) from exc
+
+    # ==
+    async def scan(self, pattern: str, limit: int = 100) -> list[Any]:
```
gruberb, line 323: This would just return the first batch?

```python
 async def scan(self, pattern: str, batch: int = 100) -> list[bytes]:
      items: list[bytes] = []
      try:
          async for key in self.replica.scan_iter(match=pattern, count=batch):
              items.append(key)
      except RedisError as exc:
          raise CacheAdapterError(f"Failed to SCAN `{pattern!r}` with error: `{exc}`") from exc
      return items
```

merino/providers/suggest/sports/backends/sportsdata/common/sports.py
```diff
@@ -639,12 +645,23 @@ async def update_events(self, client: AsyncClient):
...
+    def __init__(
+        self,
+        settings: LazySettings,
+        *args,
+        cache: RedisAdapter | NoCacheAdapter = NoCacheAdapter(),
+        **kwargs,
+    ):
         # name = self.__class__.__name__
-        name = "FIFA"
+        name = "fifa"
```
gruberb, line 667: Should we rename this to something like `competition_key: str` for the sportsdata url? And then do 

```python
name = self.__class__.__name__   # "WCS"
```

for the caching etc, so we align with our python names, and just use `fifa` for the API?

merino/jobs/sportsdata_jobs/__init__.py
```diff
@@ -217,6 +228,18 @@ async def update(self) -> None:
...
+        """Fetch widget based info and store into the cache"""
+        # we only deal with WCS for now.
+        sport = self.sports.get("WCS")
+        if sport is None:
+            return
+        await self.store.startup()
+        await sport.init_cache(client=self.client)  # type: ignore
+        if not sport.teams:
+            await sport.update_teams(client=self.client)
+        await sport.cache_teams()  # type: ignore
```
gruberb, line 239: Do we also want to `await sport.update_events(client=self.client)`. here?

merino/providers/suggest/manager.py
```diff
@@ -369,6 +369,21 @@ def _create_provider(provider_id: str, setting: Settings) -> BaseProvider:
...
+                RedisAdapter(
+                    *create_redis_clients(
+                        settings.redis.server,
+                        settings.redis.replica,
+                        settings.redis.max_connections,
+                        settings.redis.socket_connect_timeout_sec,
+                        settings.redis.socket_timeout_sec,
+                    )
+                )
+                if settings.get("cache") == "redis"
```
gruberb: Should be `setting` instead of `settings`.

gruberb (commented):
First round of review. Since it's draft, I am waiting for the final state :)

merino/cache/redis.py
```diff
@@ -127,6 +129,199 @@ async def scard(self, key: str) -> int:
...
             raise CacheAdapterError(f"Failed to SCARD {key} with error: {exc}") from exc
 
+    # == Hash Functions
+    async def hexists(self, key: str, field: str) -> int:
+        """Check if a hash field exists"""
+        try:
+            return await self.replica.hexists(key, field)
+        except RedisError as exc:
+            raise CacheAdapterError(
+                f"Failed to {cast(types.FrameType, inspect.currentframe()).f_code.co_name.upper()} {key} with error: {exc}"
```
ncloudioj: I'd suggest to not inspect the stackframe here. As the consumer might prefer to fail fast. Especially, logging is usually done by them at the calling site, which should give them enough context already.

merino/providers/suggest/sports/backends/sportsdata/common/sports.py
jrconlin: Yeah, I was using hset partly because I was thinking that i'd keep all the meta key values in a simple hash, but I'm fine using `set` and a unique key.

merino/cache/redis.py
jrconlin: yes, I could add a cursor, but for the most part the stuff I'm scanning is less than 100 items, so "good enough for now". I'll add the `scan_iter` anyway.

merino/providers/suggest/sports/backends/sportsdata/common/sports.py
jrconlin: Well, it's already the odd one out since it's class does not match the competition. A lot of other things presume that the `.name` field is the parameter to use in URLs and what-not. 
I'd rather make one change here than change all the other classes.

merino/jobs/sportsdata_jobs/__init__.py
jrconlin: No, not really. This is more about initializing things. We will be grabbing the schedules and scores more often than the teams and areas, so it makes sense to "pre-load" them here. I was planning on changing the `get_team()` function for WCS to try pulling from the cache first, so having that available quickly would help.

merino/cache/redis.py
jrconlin: Fair, just trying to provide a bit of help at the point of failure rather than making someone dig through an airflow string, and I wanted something I could cut and paste quickly.

merino/cache/redis.py
```diff
@@ -85,6 +85,17 @@ async def get(self, key: str) -> bytes | None:
...
+    async def delete(self, key: str) -> int:
+        """Delete the value associated with the key from Redis.
+
+        Raises:
+            - `CacheAdapterError` if Redis returns an error.
+        """
+        try:
+            return await self.primary.delete(key)
+        except RedisError as exc:
+            raise CacheAdapterError(f"Failed to get `{repr(key)}` with error: `{exc}`") from exc
```
ncloudioj:
```suggestion
            raise CacheAdapterError(f"Failed to delete `{repr(key)}` with error: `{exc}`") from exc
```

merino/cache/none.py
```diff
@@ -10,6 +10,9 @@ class NoCacheAdapter:  # pragma: no cover
     async def get(self, key: str) -> bytes | None:  # noqa: D102
         return None
 
+    async def delete(self, key: str) -> bytes | None:  # noqa: D102
```
ncloudioj:
```suggestion
    async def delete(self, key: str) -> int | None:  # noqa: D102
```

merino/cache/redis.py
ncloudioj: Note that the scan might take longer than we expect because per Redis' [doc](https://redis.io/docs/latest/commands/scan/#the-match-option):

> It is important to note that the MATCH filter is applied after elements are retrieved from the collection, just before returning data to the client. This means that if the pattern matches very little elements inside the collection, SCAN will likely return no elements in most iterations.

A `scan` will iterate over all the keys in the database, (all the Merino Redis producers share the same database, i.e. `db 0` and it holds approx. 500-600K keys). It will go through a lot of unrelated keys (e.g. those weather entries) before fetching the target keys matching the given pattern.

merino/providers/suggest/sports/backends/sportsdata/common/sports.py
```diff
@@ -671,15 +692,160 @@ def __init__(self, settings: LazySettings, *args, **kwargs):
...
+            },
+            {
+                "AreaId": 2,
+                "CountryCode": "ASI",
+                "Name": "Asia",
+                "Competitions": [...]
+            }
+        ]
+        """
+        self.countries = {}
```
ncloudioj: Nit: let's annotate `self.countries` in the class body and also initialize it in `__init__()`.

merino/providers/suggest/sports/backends/sportsdata/common/sports.py
```diff
@@ -639,12 +648,23 @@ async def update_events(self, client: AsyncClient):
...
+    def __init__(
+        self,
+        settings: LazySettings,
+        *args,
+        cache: RedisAdapter | NoCacheAdapter = NoCacheAdapter(),
+        **kwargs,
+    ):
         # name = self.__class__.__name__
-        name = "FIFA"
+        name = "fifa"
```
ncloudioj, line 667: My apologies for being extremely pesky on this – let's replace "fifa" with "wcs" throughout the code.

merino/providers/suggest/sports/backends/sportsdata/common/sports.py
```diff
@@ -671,15 +692,160 @@ def __init__(self, settings: LazySettings, *args, **kwargs):
...
+            # We got the lock, we can initialize things.
+            # to initial stuff.
+            await self.load_areas(client)
+            await self.update_teams(client)
+            await self.cache_teams()
+            complete = int(datetime.now(tz=timezone.utc).timestamp())
+            logging.info(f"{LOGGING_TAG} Marking db initialized as of {complete}")
+            # We're done, carry on...
+            await self.cache.set(f"{meta_key}:updated", complete.to_bytes(8))
+            await self.cache.delete(f"{meta_key}:lock")
```
ncloudioj: If any of the above blows up, the lock will be kept around for 30 seconds which might be undesired. The lock can be dropped in a try..finally block.

merino/providers/suggest/sports/backends/sportsdata/common/sports.py
```diff
@@ -671,15 +692,160 @@ def __init__(self, settings: LazySettings, *args, **kwargs):
...
+            await self.cache.get(f"{meta_key}:updated") or int(0).to_bytes()
+        )
+
+        # Has it been over a year since we last updated the meta info?
+        lock_period = 365
+        if not force and last_update:  # pragma: no cache "widget"
+            if last_update > int(
+                (datetime.now(tz=timezone.utc) - timedelta(days=lock_period)).timestamp()
+            ):
+                return
```
ncloudioj, line 726: Still not super clear after reading the docstring. Can you explain what we wanted to initialize here? Particularly, the 365-day TTL bit. If we don't expect those data points to be updated, why would we cache them in the first place? Can we just keep them elsewhere to simplify things?

merino/providers/suggest/sports/backends/sportsdata/common/sports.py
```diff
@@ -671,15 +692,160 @@ def __init__(self, settings: LazySettings, *args, **kwargs):
...
+            await self.cache.set(f"{meta_key}:updated", complete.to_bytes(8))
+            await self.cache.delete(f"{meta_key}:lock")
+
+    async def load_areas(self, client) -> None:
+        """Fetch and load the countries to the cache"""
+        # Fetch the area info, specifically the country code.
+        # Each event and team has an `AreaId` which we will use to map
+        # to the individual country. This should not change, so we only need
+        # to update once, and maybe once per year.
+        url = f"{self.base_url}/Areas?key={self.api_key}"
```
ncloudioj: Let's not hardcode API credentials in the URL.

merino/providers/suggest/sports/backends/sportsdata/common/sports.py
```diff
@@ -671,15 +692,160 @@ def __init__(self, settings: LazySettings, *args, **kwargs):
...
+        logging.info(f"{LOGGING_TAG} Pre Loading Countries")
+        for area in response:
+            # build the reverse index to get the country code and id.
+            self.countries[area["AreaId"]] = {
+                "name": area["Name"],
+                "code": area["CountryCode"],
+            }
+            # cache this for later, we're gonna need them for teams.
+            await self.cache.hset(
+                f"{self.cache_prefix}:area:{area['AreaId']}", self.countries[area["AreaId"]]
```
ncloudioj: [Suggestion]: Adding a version str e.g. `v1` to the cache keys would make it easier to handle breaking changes in the future if needed.

merino/providers/suggest/sports/backends/sportsdata/common/sports.py
```diff
@@ -730,6 +925,39 @@ async def update_events(self, client: AsyncClient, allow_no_teams: bool = False)
...
+            # import pdb; pdb.set_trace()
+            event_dict = self.event_as_str(event)
+            await self.cache.set(f"{self.cache_prefix}:event:{event_id}", event_dict.encode())
+            # Add the event to the zorder for date lookups
+            await self.cache.zadd(
+                f"{self.cache_prefix}:calendar",
+                {f"{self.cache_prefix}:event:{event_id}": int(event.date.timestamp())},
+                gt=True,
+            )
+        return self
```
ncloudioj: Nit: the method signature suggests this returns None, `self` is returned instead.

merino/cache/redis.py
```diff
@@ -102,6 +113,29 @@ async def set(
         except RedisError as exc:
             raise CacheAdapterError(f"Failed to set `{repr(key)}` with error: `{exc}`") from exc
 
+    async def setnx(
+        self,
+        key: str,
+        value: bytes,
+        ttl: timedelta | None = None,
+        nx: bool = False,
```
ncloudioj, line 133: `nx` with `False` as default can be confusing as it will behave exactly as `set()`. I'd recomend remove `nx` and always pass True to `self.primary.set()` below.

merino/cache/redis.py
```diff
@@ -127,6 +161,168 @@ async def scard(self, key: str) -> int:
...
+
+    # Technically, dict[str, Any] works fine, but mypy complains.
+    async def hset(self, key: str, values: dict[str, Any]) -> int:
+        """Return all fields for a hash key"""
+        try:
+            return await self.primary.hset(key, mapping=values)  # type: ignore
+        except RedisError as exc:
+            raise CacheAdapterError(f"Failed to cache {key} with error: {exc}") from exc
+
+    async def hsetnx(self, key: str, field: str, value: Any, expiry: int | None = None) -> int:
```
ncloudioj: `expiry` is unused and can be removed. FWIW, our Redis instance is pinned to 7.2, which doesn't support setting TTL for individual hash fields.

ncloudioj (commented):
Did a first scan and added some comments for now.

merino/providers/suggest/sports/backends/sportsdata/common/sports.py
kschelonka: it's the key for the path on sportsdata.io

merino/cache/redis.py
jrconlin: Yep. A _LOT_ of these will probably get dropped since we're not going to need them. They're there just so that I don't have to implement them later. (It was easy to cut * paste these originally using a macro). For now, I'll drop `setnx` and just add the `nx` param to `set`.

merino/cache/redis.py
jrconlin: Yeah, hopefully, I won't need to ever do a scan. It's there partly for dev reasons now, but also because I wasn't sure if we'll need it for some of the other operations.

merino/providers/suggest/sports/backends/sportsdata/common/sports.py
jrconlin: yep, we can't avoid that. I mean, we could do screwy things to obscure the string value (e.g. `%67%69%67%61`), but that seems pretty cheezy.

merino/configs/production.toml
```diff
@@ -144,4 +144,4 @@ bucket_name = "merino-ml-data-prod"
 gcp_project = "moz-fx-merino-prod-5de4"
 
 # GCS bucket that contains cohort model data
-bucket_name = "merino-ml-data-prod"
+qbucket_name = "merino-ml-data-prod"
```
kschelonka:
```suggestion
bucket_name = "merino-ml-data-prod"
```

merino/jobs/sportsdata_jobs/__init__.py
```diff
@@ -112,14 +118,19 @@ def __init__(
                     sport = UCL(settings)
                 case "MLB":
                     sport = MLB(settings)
+                case "WCS":
+                    sport = WCS(settings, cache=cache)
                 # case "EPL":
                 #    sport = EPL(settings)
+                # case "FIFA":
+                #    sport = FIFA(settings)
```
kschelonka:
```suggestion
```

merino/providers/suggest/sports/backends/sportsdata/common/sports.py
jrconlin: We need to cache this data because the resulting data will be referenced. (e.g. country codes based on RegionIDs, Team info, and possibly more.)
We only need to grab and update the information once. 
Because I have had a history of doing long term support for disposable code, I add a way for the data to be reasonably refreshed.

I'm fine dropping or expanding the total lock period, but I'm also betting that this is a pattern we'll see crop up again.

merino/providers/suggest/sports/backends/sportsdata/common/sports.py
jrconlin: I'll add the try..finally, but I'd still prefer if there was a time lock. 30 seconds isn't that long, particularly when there are other instances that will immediately retry once that lock expires. (In addition, this might reduce the blast radius since we're gating retry attempts, so that if there's some other issue, we don't wind up with a bunch of boxes doing high CPU work to also immediately fail.)

merino/providers/suggest/sports/backends/sportsdata/common/sports.py
jrconlin: Or drop the local `countries` in favor of reading them from the cache.

merino/providers/suggest/sports/backends/sportsdata/common/sports.py
ncloudioj: Oh, I didn't realize `name` was used to form the request URL. NVM.

merino/cache/redis.py
jrconlin: Sadly, that turned out to be a bad idea, since `set` is used a fair bit. I'm going to keep `setnx` for now to keep the size of the PR smaller.

merino/configs/production.toml
jrconlin: ah! good catch. sorry.

merino/providers/suggest/sports/backends/sportsdata/common/sports.py
ncloudioj: > We only need to grab and update the information once.

Sounds like static data from the vendor? Does it make sense to vendor static data points directly into Merino to save all the ceremony here? I guess it depends on if the data is already available on their end now.

merino/providers/suggest/sports/backends/sportsdata/common/sports.py
ncloudioj: Oh, would all Merino instance run this code to init the cache? I thought that the centralized job would do this in a more controlled way.

jrconlin: No, this should only really be done by jobs, and the lock essentially says "first past the gate, or first to succeed wins".

merino/cache/redis.py
ncloudioj: Oh, I was suggesting to just drop the argument `nx` from `setnx()` and always set it to True in `self.primary.set(..., nx=True)`. It's fine to have both `setnx()` and `set()` for the adapter.

>  A LOT of these will probably get dropped since we're not going to need them.

Those `CacheAdapter` additions can be retained for future uses IMO. :)

merino/providers/suggest/sports/backends/sportsdata/README.md
```diff
@@ -0,0 +1,61 @@
+# World Cup Widget modifications
```
ncloudioj, line 1: Thanks for adding this doc! Very helpful.

merino/providers/suggest/sports/backends/sportsdata/README.md
```diff
@@ -0,0 +1,61 @@
...
+|name| type | description |
+| --- | --- | --- |
+| lock | _timestamp_ | initialization lock (Set on NX) |
+| meta_updated | _timestamp_ | timestamp for when the last initialization completed |
+
+### Team information
+
+This is a reference table for team information. (May include standings?)
+
+key: **sport:wcs:team:{ _teamId_ }**
```
ncloudioj, line 26: Nit: let's add the version str to the key prefix to reflect the implementation.

merino/providers/suggest/sports/backends/sportsdata/common/sports.py
```diff
@@ -695,7 +866,56 @@ async def update_teams(self, client: AsyncClient):
...
+
+    def team_as_str(self, team: Team) -> str:  # pragma: no cover "widget"
+        """Serialize a team as a dictionary for the widget"""
+        # TODO: Populate the eliminated field (from old team info?)
+        serialized = dict(
+            team.__dict__
+        )  # Because BaseModel is clever and helpful instead of dumb and useful.
+        serialized["expiry"] = team.expiry.timestamp()
+        serialized["updated"] = team.updated.timestamp()
+        return json.dumps(serialized)
```
ncloudioj, line 879: In Merino, if bytes JSON serialization is preferred, we usually do `orjson.dumps()` instead of `json.dumps().encode()`.

merino/providers/suggest/sports/backends/sportsdata/common/sports.py
```diff
@@ -695,7 +866,56 @@ async def update_teams(self, client: AsyncClient):
...
         async with self._lock:
-            self.load_teams_from_source(response)
+            await self.async_load_teams_from_source(response)
+
+    def team_as_str(self, team: Team) -> str:  # pragma: no cover "widget"
+        """Serialize a team as a dictionary for the widget"""
+        # TODO: Populate the eliminated field (from old team info?)
+        serialized = dict(
+            team.__dict__
+        )  # Because BaseModel is clever and helpful instead of dumb and useful.
```
ncloudioj, line 876: `dict(team.__dict__)` will not dump nested models.
```suggestion
        serialized = team.model_dump()
```

merino/providers/suggest/sports/backends/sportsdata/common/sports.py
```diff
@@ -695,7 +866,56 @@ async def update_teams(self, client: AsyncClient):
...
+        serialized["expiry"] = team.expiry.timestamp()
+        serialized["updated"] = team.updated.timestamp()
+        return json.dumps(serialized)
+
+    def event_as_str(self, event: Event) -> str:  # pragma: no cover "widget"
+        """Serialize an event as a dictionary for the widget"""
+        # import pdb;pdb.set_trace()
+        # TODO: add more team info?
+        serialized = dict(event.__dict__)
```
ncloudioj, line 885:
```suggestion
        serialized = event.model_dump()
```

merino/providers/suggest/sports/backends/sportsdata/common/sports.py
```diff
@@ -695,7 +866,56 @@ async def update_teams(self, client: AsyncClient):
...
+        ids = []
+        for teamId, team in self.teams.items():
+            id = f"{self.cache_prefix}:team:{teamId}"
+            await self.cache.set(id, self.team_as_str(team).encode())
+            ids.append(id)
+        # for now, dump the list of team ids into a meta list so that we can save doing a scan later.
+        # TODO: replace with vadd when available.
+        await self.cache.set(f"{self.cache_prefix}:meta:team_ids", json.dumps(ids).encode())
+
+    async def get_all_teams(self) -> dict[int, Any]:
```
ncloudioj, line 904: Question: looking at the `Team` model, the team info seems quite static. If so, does it make sense to cache the Redis response here?

`functools.cache` doesn't support async methods, but we can easily cache it as a class member.

ncloudioj (approved):
LGTM, r=nanj.

The remaining comments can be considered as followups.

merino/providers/suggest/sports/backends/sportsdata/README.md
jrconlin: It should be https://github.com/mozilla-services/merino-py/blob/5f97e91673a2496d14f75dd1c89e65e5bef918f4/merino/providers/suggest/sports/backends/sportsdata/common/sports.py#L656-L657

(Ah, this is the ReadMe, nvm).

merino/providers/suggest/sports/backends/sportsdata/common/sports.py
jrconlin: Ideally, yes, although there may be some Team data points that could change (e.g. if a team is eliminated or not). Team elimination could be updated daily, though, so we could have the local cache expire at midnight UTC.

PR #1485, opened 2026-05-08
[DISCO-4180] chore: Allow dual redis connections

ncloudioj:
## References

JIRA: [DISCO-4180](https://mozilla-hub.atlassian.net/browse/DISCO-4180)

## Description

Webservice-infra changes: https://github.com/mozilla/webservices-infra/pull/10896

## PR Review Checklist

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/MC-2291)

[DISCO-4180]: https://mozilla-hub.atlassian.net/browse/DISCO-4180?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/merino/jobs/sportsdata_jobs/__init__.py
+++ b/merino/jobs/sportsdata_jobs/__init__.py
@@ -304,8 +304,8 @@ if elastic_credentials.validate():
         cache = (
             RedisAdapter(
                 *create_redis_clients(
-                    settings.redis.server,
-                    settings.redis.replica,
+                    settings.redis.wcs_server,
+                    settings.redis.wcs_replica,
                     settings.redis.max_connections,
                     settings.redis.socket_connect_timeout_sec,
                     settings.redis.socket_timeout_sec,
--- a/merino/providers/suggest/manager.py
+++ b/merino/providers/suggest/manager.py
@@ -373,8 +373,8 @@ def _create_provider(provider_id: str, setting: Settings) -> BaseProvider:
             cache = (
                 RedisAdapter(
                     *create_redis_clients(
-                        settings.redis.server,
-                        settings.redis.replica,
+                        settings.redis.wcs_server,
+                        settings.redis.wcs_replica,
                         settings.redis.max_connections,
                         settings.redis.socket_connect_timeout_sec,
                         settings.redis.socket_timeout_sec,
--- a/merino/configs/default.toml
+++ b/merino/configs/default.toml
@@ -251,6 +251,12 @@ socket_connect_timeout_sec = 3
 # Timeout to interact with Redis in seconds
 socket_timeout_sec = 3
 
+# MERINO_REDIS__WCS_SERVER - URI to the WCS Redis primary endpoint.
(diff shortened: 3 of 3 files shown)


gruberb (approved):

merino/jobs/sportsdata_jobs/__init__.py
```diff
@@ -304,8 +304,8 @@ async def update_and_cache_wcs(self) -> None:
         cache = (
             RedisAdapter(
                 *create_redis_clients(
-                    settings.redis.server,
-                    settings.redis.replica,
+                    settings.redis.wcs_server,
```
gruberb, line 307: @ncloudioj  Question: Did we use Redis before WCS, or was it always just intended for world cup? Or we tried to re-use the existing Redis, but we cannot from the airflow job, and now we use it inside the cron job with our own Redis instance just for world cup?

ncloudioj: For Sports, Redis is new. We only added it in for the WCS widget endpoints. The WCS ETL job also needs to access it to populate the data from the vendor. Airflow doesn't have access to any of Merino's Redis instances, so the Airflow job just failed.

jrconlin: Technically, you also need it for the ingestion job since that's what is writing the values into Redis

merino/providers/suggest/manager.py
```diff
@@ -373,8 +373,8 @@ def _create_provider(provider_id: str, setting: Settings) -> BaseProvider:
             cache = (
                 RedisAdapter(
                     *create_redis_clients(
-                        settings.redis.server,
-                        settings.redis.replica,
+                        settings.redis.wcs_server,
```
jrconlin, line 376: I'm confused.

We're running the WCS code separately from the normal Merino code, correct? Are we having both connect to the same servers at any point or is this something that just needs different configuration file values?

merino/jobs/sportsdata_jobs/__init__.py
ncloudioj: Yes, the WCS ETL job does the ingestion, which uses the new connection [here](https://github.com/mozilla/webservices-infra/pull/10896/changes).

merino/providers/suggest/manager.py
ncloudioj: > We're running the WCS code separately from the normal Merino code, correct?

Correct, once we have the WCS deployment up and running, the WCS code will be running separately from the normal Merino code. Before we get there, the normal Merion will be running both the wcs/* and the other endpoints (e.g. suggest), the former will access the new Redis instance and the latter will access the existing one, which is why we'd need separate configurations for those.

BTW, is Redis required for the Sports Suggestion Provider? I am not sure about that but saw it got initialized here so went ahead update the settings regardless.

PR #1478, opened 2026-05-07
[DISCO-4177] Add and serve hex codes for team colours

Herraj:
## References

JIRA: [DISCO-4177](https://mozilla-hub.atlassian.net/browse/DISCO-4177)

## Description
Add and team colours in hex code format. These colours were sourced from this [codepen](https://codepen.io/tmirson/pen/wBzboEw). `USA`, `Canada` & `Mexico` will be updated soon.

These colours won't change and will be wired into the logic for serving non-fake data in follow up work.

## PR Review Checklist
- [x] This PR conforms to the [Contribution Guidelines](https://github.com/mozilla-services/merino-py/blob/main/CONTRIBUTING.md)
- [x] The PR title starts with the JIRA issue reference, format example `[DISCO-####]`, and has the same title (if applicable)
- [ ] `[load test: (abort|skip|warn)]` keywords are applied to the last commit message (if applicable)
- [ ] [Documentation](https://github.com/mozilla-services/merino-py/tree/main/docs) has been updated (if applicable)
- [ ] [Metric documentation](https://github.com/mozilla-services/merino-py/tree/main/metrics.yaml) has been updated (if applicable)
- [ ] [Functional and performance test](https://github.com/mozilla-services/merino-py/blob/main/docs/dev/testing.md) coverage has been expanded and maintained (if applicable)

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/MC-2284)

[DISCO-4177]: https://mozilla-hub.atlassian.net/browse/DISCO-4177?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #1487, opened 2026-05-08
DISCO-4130 Return real wcs events from redis

gruberb:
## References

JIRA: [DISCO-4130](https://mozilla-hub.atlassian.net/browse/DISCO-4130)

## Description
This PR wires the WCS widget API to the Redis cache populated by the WCS ETL job, replacing the static event path with cached matches.

## PR Review Checklist

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/MC-2293)

[DISCO-4130]: https://mozilla-hub.atlassian.net/browse/DISCO-4130?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


merino/providers/suggest/sports/backends/sportsdata/common/data.py
```diff
@@ -410,6 +437,10 @@ def load_scores_from_source(
                 event.away_score = event_description.get(
                     self.normalized_terms[SportTerms.AWAY_TEAM_SCORE]
                 )
+                event.status = status
+                event.updated = self.updated_at(event_description, event_timezone)
+                for field, value in self.event_details(event_description).items():
```
jrconlin, line 437: We should ensure that the value is not `None` here before we update the value.

merino/providers/suggest/sports/backends/sportsdata/common/sports.py
```diff
@@ -53,6 +54,12 @@
 # Global logger
 logger = logging.getLogger(__name__)
 
+# Open-ended sorted-set score bounds. WCS event scores are Unix timestamps,
```
jrconlin: Let's rephrase this, because "scores" has multiple meanings and contexts.
In this case, we are referring to the "scores" associated with the sorted-set. We should refer to them as either "timestamps" or "set-scores", and keep the redis sorted-set specific reference to "scores" as close to the redis function as possible.

merino/providers/suggest/sports/backends/sportsdata/common/sports.py
```diff
@@ -53,6 +54,12 @@
...
+# so these signed 64-bit sentinels comfortably bracket all possible events
+# while keeping RedisAdapter.zrange's integer contract unchanged.
```
gruberb: Per https://redis.io/docs/latest/commands/zadd/

merino/providers/suggest/sports/backends/sportsdata/common/sports.py
```diff
@@ -847,6 +866,11 @@ async def async_load_teams_from_source(self, data: list[dict[str, Any]]) -> dict
                     team_ttl=self.team_ttl,
                     normalized_terms=self.normalized_terms,
                 )
+                if team_data.get("Type") == "National":
```
jrconlin, line 873: This logic seems very Soccer specific and I wonder if there's a way to migrate it to the UCL/WCS classes. (This is similar to the idea of having the term translation dictionary).

merino/providers/suggest/sports/backends/sportsdata/common/data.py
```diff
@@ -595,6 +618,25 @@ def load_schedules_from_source(
...
+            )
+        if event_description.get("Updated"):
+            return datetime.fromisoformat(event_description["Updated"]).replace(
+                tzinfo=event_timezone
+            )
+        return None
+
+    def event_details(self, event_description: dict[str, Any]) -> dict[str, Any]:
+        """Return optional sport-specific event details."""
+        return {}
```
Herraj, line 644: Do we expect this to return something else in follow up work?

merino/providers/suggest/sports/backends/sportsdata/common/data.py
```diff
@@ -595,6 +612,25 @@ def load_schedules_from_source(
...
```
Herraj, line 644: Do we expect this to return something else in follow up work?

merino/providers/suggest/sports/backends/sportsdata/common/sports.py
```diff
@@ -697,6 +704,18 @@ def __init__(
...
+        clock = event_description.get("ClockDisplay") or event_description.get("Clock")
+        return {
+            "period": event_description.get("Period"),
+            "home_extra": event_description.get("HomeTeamScoreExtraTime"),
+            "away_extra": event_description.get("AwayTeamScoreExtraTime"),
+            "home_penalty": event_description.get("HomeTeamScorePenalty"),
+            "away_penalty": event_description.get("AwayTeamScorePenalty"),
+            "clock": str(clock) if clock is not None else None,
+        }
+
```
Herraj, line 722: This could be a gap in my knowledge but this looks like it is only building extra time event details and not the regular period details?

merino/providers/suggest/sports/backends/sportsdata/common/data.py
```diff
@@ -71,6 +71,8 @@ class Team(BaseModel):
     expiry: datetime
     # Country / Area
     country: str | None
+    # Whether this team has been eliminated from tournament play.
```
jrconlin: FWIW, I'm not sure we get this field. Each team has an `Active` field, but I don't know if that is seasonal, or overall active.

merino/providers/suggest/sports/backends/sportsdata/common/sports.py
```diff
@@ -873,24 +897,11 @@ async def update_teams(self, client: AsyncClient):
...
     def team_as_str(self, team: Team) -> str:  # pragma: no cover "widget"
         """Serialize a team as a dictionary for the widget"""
-        # TODO: Populate the eliminated field (from old team info?)
-        serialized = dict(
-            team.__dict__
-        )  # Because BaseModel is clever and helpful instead of dumb and useful.
-        serialized["expiry"] = team.expiry.timestamp()
-        serialized["updated"] = team.updated.timestamp()
-        return json.dumps(serialized)
+        return json.dumps(team.model_dump(mode="json"))
```
Herraj: non-blocking: wondering if `orjson` works here and if it is worth it for the performance? Not sure if it'll make a big difference.

merino/providers/suggest/sports/backends/sportsdata/common/sports.py
```diff
@@ -904,19 +915,35 @@ async def cache_teams(self) -> None:  # pragma: no cover "widget"
...
+                    try:
+                        team = Team(**json.loads(bteam))
+                    except (json.JSONDecodeError, ValidationError, TypeError) as ex:
+                        logger.warning(
+                            "%s Could not load cached WCS team %s: %s",
+                            LOGGING_TAG,
+                            team_key,
+                            ex,
+                        )
+                        continue
```
Herraj, line 988: non-blocking: same thing, wondering if `orjson.loads()` could be useful here but it does load as bytes and not as json I think.

tests/unit/providers/suggest/sports/backends/common/test_sports.py
```diff
@@ -2199,6 +2216,128 @@ async def test_wcs_update_events(
...
+
+@freezegun.freeze_time("2026-06-10T00:00:00", tz_offset=0)
+@pytest.mark.asyncio
+async def test_wcs_sportsdata_schedule_and_games_by_date_payloads_match_expected_shape() -> None:
+    """Validate representative FIFA payloads from SchedulesBasic and GamesByDate."""
+    sport = WCS(settings=settings.providers.sports)
+    sport.event_ttl = timedelta(weeks=8)
+    await sport.async_load_teams_from_source(wcs_static_teams_payload({"SWE", "TUN"}))
+
+    schedule = {
```
jrconlin: I would prefer if we used "real" captured data from SportsData like we do for the other tests. There can be very subtle differences for returned key names and some values (e.g. "GlobalGameI[Dd]", etc.) This means not doing short cuts like we're doing below where we extend data with what we assume are the new fields.

merino/providers/wcs/protocol.py
```diff
@@ -12,7 +29,24 @@ class TeamInfo(BaseModel):
...
+    def from_event_team(cls, team: dict[str, Any]) -> "TeamInfo":
+        """Build widget team info from the compact team dict stored on events."""
+        key = str(team["key"])
+        raw_icon_url = team.get("icon_url")
+        return cls(
+            key=key,
+            global_team_id=int(team["id"]),
+            name=str(team["name"]),
+            region=str(team.get("region") or team.get("country") or key),
+            colors=[str(color) for color in team["colors"] if color],
```
jrconlin, line 46: So we're not going to convert the string color names to hex values?

merino/providers/suggest/sports/backends/sportsdata/common/sports.py
```diff
@@ -954,19 +986,48 @@ async def update_events(self, client: AsyncClient, allow_no_teams: bool = False)
...
+        stale_event_keys = [
+            event_key
+            for raw_event_key in indexed_event_keys
+            if (event_key := raw_event_key.decode("utf-8", errors="replace"))
+            not in active_event_keys
+        ]
+        if not stale_event_keys:
+            return
+        await self.cache.delete(*stale_event_keys)
+        await self.cache.zrem(calendar_key, *stale_event_keys)
```
Herraj, line 1082: non-blocking: this method was a little tricky to read, could use some 1-2 liner comments! Maybe if you can also refactor lines 1021 through 1026 to be a little easier to read :)

jrconlin (approved):
A few comments, but otherwise wfm.

merino/providers/suggest/sports/backends/sportsdata/common/sports.py
```diff
@@ -979,12 +1040,41 @@ async def get_events_by_team(  # pragma: no cover "widget"
...
+                # prune_stale_events clears the orphaned calendar entry on that refresh.
+                continue
+            event_key_str = event_key.decode("utf-8", errors="replace")
+            try:
+                result.append(Event(**json.loads(raw_event)))
+            except (json.JSONDecodeError, ValidationError, TypeError) as ex:
+                logger.warning(
+                    f"{LOGGING_TAG} Could not load cached WCS event {event_key_str}: {ex}"
+                )
+        return result
```
Herraj, line 1133: Are lines 1045 through 1048 setting the `start` and `end` variables as the current (server) time? Meaning the time of this request being processed if `start` and `end` are not provided 🤔

merino/providers/wcs/protocol.py
```diff
@@ -12,7 +29,24 @@ class TeamInfo(BaseModel):
...
+            icon_url=HttpUrl(raw_icon_url) if raw_icon_url else _icon(key),
+            eliminated=bool(team.get("eliminated", False)),
+        )
```
Herraj, line 49: Can the return type here not be `TeamInfo` or `self`? What does having it in the `""` mean?

merino/providers/wcs/protocol.py
```diff
@@ -36,6 +70,31 @@ class EventInfo(BaseModel):
...
+            home_extra=event.home_extra,
+            away_extra=event.away_extra,
+            home_penalty=event.home_penalty,
+            away_penalty=event.away_penalty,
+            clock=event.clock or "",
+            updated=int(updated.timestamp()),
+            status=event.status.as_str(),
+            status_type=event.status.as_ui_status(),
+            query=f"{home_team.name} vs {away_team.name}",
+        )
```
Herraj, line 96: Nice you added the `query` property. One less thing for me 😄

Herraj (commented):
I haven't reviewed the tests yet. Source code looks good to me so far. I would recommend adding some more 1-2 line comments for some blocks of logic because the business logic is not clear sometimes or the syntax is a little complex.

Will review the tests after a break.

merino/providers/suggest/sports/backends/sportsdata/common/sports.py
gruberb: This already lives in the `WCS` class (which reminds me, we should probably split this huge file into submodules).

merino/providers/wcs/protocol.py
Herraj: We are! I have this [PR](https://github.com/mozilla-services/merino-py/pull/1478) approved. It just missing the host country colours. Once UX gets back to me with those, I can rebase and merge.

merino/providers/wcs/protocol.py
jrconlin: Some iterations of python (older ones) would only allow for methods that returned the current class to be returned as a quoted string, since the class has yet to be fully instantiated.
I believe the current version of Python allows for unquoted class names, but the behavior should still persist. 

As for the return type, this is a `@classmethod` so it needs to return the class' method. Technically, since it _is_ a `@classmethod` it should behave like `__init__` and the returned class should be implied, but again, it's better to be safe in case of older versions of Python.

merino/providers/suggest/sports/backends/sportsdata/common/sports.py
jrconlin: Not going to argue about splitting things. Just hadn't gotten around to doing that and it can be handy to cut & paste from existing code when getting a new sport started.

merino/providers/suggest/sports/backends/sportsdata/common/sports.py
jrconlin: "clock" is mapped to the SportsData.io `ClockDisplay` string. This contains the completed minutes of regular play plus any extra time (appended with a "+`#`")
In short, it contains both regular and extra time.

merino/providers/suggest/sports/backends/sportsdata/common/sports.py
jrconlin: Not quite. The data provider is not consistent about things like dates. While we do set the appropriate timezone on ingestion, we need to ensure that we are using the expected, normalized time when doing the lookups. I presume that those are guardrails to ensure that we are using normalized timezones for any searches.

merino/data/wcs_games_by_date.json
```diff
@@ -0,0 +1,222 @@
...
+    "HomeTeamPointSpreadPayout": -115,
+    "AwayTeamPointSpreadPayout": -113,
+    "OverUnder": 2.50,
+    "OverPayout": 123,
+    "UnderPayout": -175,
+    "Attendance": null,
+    "Updated": "2026-04-29T10:26:26",
+    "UpdatedUtc": "2026-04-29T14:26:26",
+    "GlobalGameId": 90086918,
+    "GlobalAwayTeamId": 90001067,
```
jrconlin: Non blocking, but for simplicity between this and the schedule, I will change the home team to be *1 and away team to be *2 (e.g. `90000001`  and `90000002`) These can be close enough to the "real" data that they don't cause problems. This is something we can adjust in a future PR if needed.

jrconlin (approved):

merino/providers/suggest/sports/backends/sportsdata/common/sports.py
```diff
@@ -905,19 +917,35 @@ async def cache_teams(self) -> None:  # pragma: no cover "widget"
...
+            str_teams = await self.cache.mget(team_keys)
             if str_teams:
-                for bteam in str_teams:
-                    team = Team(**json.loads(bteam))
+                for team_key, bteam in zip(team_keys, str_teams, strict=False):
+                    if not bteam:
+                        logger.warning("%s Missing cached WCS team %s", LOGGING_TAG, team_key)
+                        continue
+                    try:
+                        team = Team(**json.loads(bteam))
```
ncloudioj: `Team.model_validate_json(bteam)` ([doc](https://pydantic.dev/docs/validation/latest/concepts/json#json-parsing)) can parse and validate json bytes directly.

merino/providers/suggest/sports/backends/sportsdata/common/sports.py
```diff
@@ -980,12 +1042,41 @@ async def get_events_by_team(  # pragma: no cover "widget"
...
+
+        result: list[Event] = []
+        for event_key, raw_event in zip(event_keys, cached_events, strict=False):
+            if not raw_event:
+                # The event payload can expire before the next authoritative refresh;
+                # prune_stale_events clears the orphaned calendar entry on that refresh.
+                continue
+            event_key_str = event_key.decode("utf-8", errors="replace")
+            try:
+                result.append(Event(**json.loads(raw_event)))
```
ncloudioj: Ditto: `Event.model_validate_json(raw_event)`

merino/providers/suggest/sports/backends/sportsdata/common/sports.py
```diff
@@ -54,6 +55,13 @@
 # Global logger
 logger = logging.getLogger(__name__)
 
+# Open-ended timestamp bounds for the WCS calendar index. The calendar is a
+# Redis sorted set keyed by Unix timestamps, so these signed 64-bit sentinels
+# comfortably bracket all possible events while keeping RedisAdapter.zrange's
+# integer contract unchanged.
+ZSET_MIN_TIMESTAMP = -(2**63)
```
ncloudioj, line 62: Nit: Per Redis' [doc](https://redis.io/docs/latest/develop/data-types/sorted-sets/), "[0, -1]" is the idiomatic way to fetch all the entries from a zset  `zrange foo 0 -1`.

merino/providers/suggest/sports/backends/sportsdata/common/sports.py
```diff
@@ -955,19 +988,48 @@ async def update_events(self, client: AsyncClient, allow_no_teams: bool = False)
...
+
+    async def prune_stale_events(self, active_event_keys: set[str]) -> None:
+        """Remove cached event entries that no longer exist in the latest source data."""
+        calendar_key = f"{self.cache_prefix}:calendar"
+        indexed_event_keys = cast(
+            list[bytes],
+            await self.cache.zrange(
+                calendar_key,
+                min=ZSET_MIN_TIMESTAMP,
+                max=ZSET_MAX_TIMESTAMP,
```
ncloudioj: This could make Redis return all 104 games, is there any concerns about that? If so, there is `zscan` available.

ncloudioj (approved):
Looks good, r+wc.

PR #1506, opened 2026-05-12
[DISCO-4201] fix(wcs): query string updates for world cup

kschelonka:
## References

JIRA: [DISCO-4201]

## Description
fix(wcs): query string updates for world cup

Use full month for dates instead of short name (this affects all sport responses)

Use 'World Cup 2026' as the prefix instead of just 'World Cup'

Use 'vs' instead of 'at' as the team separator

This is propagated to the /wcs/matches query response as well, so they are both unified. This is not yet in the /live endpoint because of the way the fake data is constructed, but when we use real data it should go through the EventInfo.from_event method and be propagated through

## PR Review Checklist

[DISCO-4021]: https://mozilla-hub.atlassian.net/browse/DISCO-4021?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[DISCO-4201]: https://mozilla-hub.atlassian.net/browse/DISCO-4201?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/merino/providers/suggest/sports/backends/sportsdata/protocol.py
+++ b/merino/providers/suggest/sports/backends/sportsdata/protocol.py
@@ -46,7 +46,7 @@ def build_query(event: dict[str, Any]) -> str:
         sport_query_name = "World Cup 2026"
     else:
         sport_query_name = event.get("sport", "")
-    return f"""{sport_query_name} {event.get("away_team", {}).get("name", "")} vs {event.get("home_team", {}).get("name", "")} {date}"""
+    return f"""{sport_query_name} {event.get("home_team", {}).get("name", "")} vs {event.get("away_team", {}).get("name", "")} {date}"""
 
 
 class SportEventDetail(BaseModel):
--- a/merino/providers/suggest/sports/backends/sportsdata/protocol.py
+++ b/merino/providers/suggest/sports/backends/sportsdata/protocol.py
@@ -39,11 +39,14 @@ class SportTeamDetail(BaseModel):
 
 def build_query(event: dict[str, Any]) -> str:
     """Build the search query from the event information"""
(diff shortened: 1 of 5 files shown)


merino/providers/suggest/sports/backends/sportsdata/protocol.py
```diff
@@ -39,11 +39,14 @@ class SportTeamDetail(BaseModel):
...
-    date = datetime.fromisoformat(event["date"]).strftime("%d %b %Y")
+    date = datetime.fromisoformat(event["date"]).strftime("%d %B %Y")
     # catch pre-stored values
     if event.get("sport") == "fifa":
         event["sport"] = "World Cup"
-    return f"""{event.get("sport")} {event.get("away_team", {}).get("name", "")} at {event.get("home_team", {}).get("name", "")} {date}"""
+        sport_query_name = "World Cup 2026"
+    else:
+        sport_query_name = event.get("sport", "")
+    return f"""{sport_query_name} {event.get("away_team", {}).get("name", "")} vs {event.get("home_team", {}).get("name", "")} {date}"""
```
Herraj: non-blocking: Not that it matters but since we don't have the ` ... at ...` template, for the ` ... vs ...` it is usually Home Team vs Away Team.

Herraj (approved):
:shipit:

merino/providers/suggest/sports/backends/sportsdata/protocol.py
kschelonka: done

PR #1516, opened 2026-05-13
[DICSO-4203] feat(wcs): add tournament stage to widget response

kschelonka:
## References

JIRA:  [DISCO-4203]

## Description
Add stage name to the /wcs/matches response. This is not available on the SchedulesBasic or Scores payload, but they do have the RoundId. The Competitions payload has a list of rounds, so I decided to use that in a similar way to how Areas is used. We expect this data to be essentially static since all the Rounds are already planned for the competition (even if we don't know what teams are playing).

## PR Review Checklist

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/MC-2318)

[DISCO-4203]: https://mozilla-hub.atlassian.net/browse/DISCO-4203?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/merino/providers/suggest/sports/backends/sportsdata/common/sports.py
+++ b/merino/providers/suggest/sports/backends/sportsdata/common/sports.py
@@ -822,7 +822,7 @@ class WCS(Sport):
         """
         # Sample Response
         """
-        {
+        [{
             "CompetitionId": 21,
             ...
             "Key": "FIFA",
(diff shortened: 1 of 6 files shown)


Herraj (approved):

merino/providers/suggest/sports/backends/sportsdata/common/sports.py
```diff
@@ -811,6 +815,63 @@ async def load_areas(self, client) -> None:
...
+        competition = next(
+            (c for c in response if c.get("Key") == self.name.upper()),
+            None,
+        )
+        seasons: list[dict] = (competition or {}).get("Seasons") or []
+        season_data = next(
+            (s for s in seasons if str(s.get("Season")) == self.season),
+            None,
+        )
+        rounds: list[dict] = (season_data or {}).get("Rounds") or []
```
ncloudioj: If you prefer, you can pattern match against the response to extract the rounds:

```py
match response:
    case {
        "Key": key,
        "Seasons": [
            {
                "Season": season,
                "Rounds": rounds,
            }
        ], 
    } if key == self.name.upper() and season == self.season:
        self.rounds = {
            int(r["RoundId"]): r["Name"]
            for r in rounds
            if r.get("RoundId") is not None and r.get("Name")
        }
    case _:
        # handle the negative case

```

ncloudioj (approved):

merino/providers/suggest/sports/backends/sportsdata/common/sports.py
kschelonka: Oh I didn't know python did that, thanks. Looks closer to scala my beloved :)

PR #1520, opened 2026-05-13
DISCO-4163 enable live teams for wcs teams endpoint

gruberb:
## References

JIRA: [DISCO-4163](https://mozilla-hub.atlassian.net/browse/DISCO-4163)

## Description
This updates the WCS teams endpoint to build from cached SportsData teams and schedule data while preserving our static grouping, region, color, and icon metadata. Elimination is derived from knockout fixtures and explicit winner data instead of SportsData’s `active` flag. 

## PR Review Checklist

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/MC-2322)

[DISCO-4163]: https://mozilla-hub.atlassian.net/browse/DISCO-4163?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/merino/providers/suggest/sports/backends/sportsdata/common/wcs_elimination.py
+++ b/merino/providers/suggest/sports/backends/sportsdata/common/wcs_elimination.py
@@ -1,0 +1,138 @@
+"""WCS tournament elimination helpers."""
+
+from collections.abc import Iterable
+import json
+from importlib.resources import files
+
+import orjson
+
+from merino.providers.suggest.sports.backends.sportsdata.common.data import Event
+
+_FIRST_KNOCKOUT_ADVANCER_COUNT = 32
+_ELIMINATED_TEAM_KEYS_META_KEY = "meta:eliminated_team_keys"
+
+
+def eliminated_team_keys_cache_key(cache_prefix: str) -> str:
+    """Return the Redis key used for cached eliminated WCS team keys."""
+    return f"{cache_prefix}:{_ELIMINATED_TEAM_KEYS_META_KEY}"
+
+
+def serialize_eliminated_team_keys(team_keys: Iterable[str]) -> bytes:
+    """Serialize eliminated team keys for Redis."""
+    return orjson.dumps(sorted(set(team_keys)))
+
+
+def parse_eliminated_team_keys(raw: bytes | str | None) -> set[str]:
+    """Parse cached eliminated team keys."""
+    if not raw:
+        return set()
+    payload = orjson.loads(raw)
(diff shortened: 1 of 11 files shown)


merino/providers/wcs/provider.py
```diff
@@ -96,9 +102,42 @@ async def get_live_matches(self, team_keys: frozenset[str] | None) -> LiveMatche
...
+                    eliminated=cached_team.key in eliminated_keys,
+                    region=cached_team.country or roster_team.region,
+                )
+            )
+        return TeamsResponse(teams=response)
+
+    async def _get_eliminated_team_keys(self) -> set[str]:
+        """Return team keys that no longer have a tournament path."""
+        try:
+            events = await self.sport.get_events_by_date()
```
ncloudioj: Non-blocking: `get_events_by_date()` will do a zrange followed by another mget under the hood, meaning that each request to the `teams` endpoint will cause a large read amplification to Redis. Perhaps we can get the cron job to manage and persist eliminated teams so the provider can simplify use them.

ncloudioj (approved):

merino/providers/wcs/provider.py
gruberb: Added it to the wcs cron job and added Redis cache entries for the elimanted teams.

merino/providers/suggest/sports/backends/sportsdata/common/wcs_elimination.py
```diff
@@ -0,0 +1,134 @@
...
+
+def _team_key(team: dict[str, object]) -> str | None:
+    """Return a team key from a compact event team dictionary."""
+    key = team.get("key")
+    return key if isinstance(key, str) and key else None
+
+
+def _roster_team_keys() -> set[str]:
+    """Return tournament roster team keys from static WCS metadata."""
+    data = json.loads((files("merino.data") / "wcs_teams.json").read_text())
```
ncloudioj, line 137: I assume this would only be called by the cronjob, correct? If so, could you add a note to clarify that. Otherwise, if the provider also uses it to serve the endpoint requests, we need to be mindful of not doing repetitive blocking IO in the web app.

ncloudioj (approved):
Thanks!

I forgot to suggest doing it as a followup, but you are FAST!

ncloudioj (commented):
Looks good, thanks!

Forgot to mention to do that as a follow up, but you're just too FAST to wait :)

PR #1527, opened 2026-05-15
DISCO-4079 return real live matches for /live endpoint

gruberb:
## References

JIRA: [DISCO-4079](https://mozilla-hub.atlassian.net/browse/DISCO-4079)

## Description
This PR wires `/api/v1/wcs/live` to the Redis-backed WCS event cache while leaving the mock response in place as the production default. 

On the cron side, WCS.update_events now refreshes `GamesByDate/fifa/<day>` for any event whose day falls inside the active UTC window (today plus or minus one day) instead of only when `SchedulesBasic` already reports a non-scheduled status, with the TTL on that call lowered to two minutes so live status, score, and clock land within one cron tick. 

The endpoint reads a window of cached events, filters by `event.status.is_in_progress()`, and returns a new `{"current": [EventInfo, ...]}` response, with the Redis path gated behind `providers.wcs.live_data_enabled` so we can flip the cutover via config without a code change.

## PR Review Checklist

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/MC-2329)

[DISCO-4079]: https://mozilla-hub.atlassian.net/browse/DISCO-4079?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/merino/configs/default.toml
+++ b/merino/configs/default.toml
@@ -400,6 +400,11 @@ providers = ["sports", "polygon"]
 # GCS bucket for remote data files (used when data_source = "remote").
 gcs_bucket = ""
 
+# Sections under `[*.providers.<name>]` are loaded by the suggest provider
+# manager only when they expose a `type` field. Sibling subsystem configs can
+# share this namespace without `type` and initialize themselves elsewhere, as
+# `[*.providers.wcs]` does for the World Cup widget API.
+
 [default.providers.accuweather]
 # MERINO_PROVIDERS__ACCUWEATHER__TYPE
 # The type of this provider, should be `accuweather`.
--- a/merino/configs/default.toml
+++ b/merino/configs/default.toml
@@ -516,6 +516,12 @@ cache_dir="/tmp"
 # base_url.nfl="https://api.sportsdata.io/v3/nfl/scores/json"
 # base_url.ucl="https://api.sportsdata.io/v4/soccer/scores/json"
 
+[default.providers.wcs]
+# MERINO_PROVIDERS__WCS__LIVE_DATA_ENABLED
+# Keep the /wcs/live mock response by default. DISCO-4231 can flip this to
+# serve Redis-backed in-progress matches after launch approval.
+live_data_enabled = false
+
 
 [default.manifest]
 # MERINO_PROVIDERS__MANIFEST__CRON_INTERVAL_SEC
--- a/merino/providers/suggest/manager.py
+++ b/merino/providers/suggest/manager.py
@@ -424,7 +424,12 @@ def load_providers(disabled_providers_list: list[str]) -> dict[str, BaseProvider
     """
     providers: dict[str, BaseProvider] = {}
     for provider_id, setting in settings.providers.items():
-        # Do not initialize provider if disabled in config.
+        # Skip sibling subsystems that share the settings.providers namespace
+        # but are not suggest providers (e.g. wcs). The contract is that a
+        # suggest provider entry exposes a `type` field; anything else here
+        # belongs to another subsystem (WCS, etc.) and initializes itself.
+        if "type" not in setting:
+            continue
         if provider_id.lower() not in disabled_providers_list:
             providers[provider_id] = _create_provider(provider_id, setting)
     return providers
--- a/merino/providers/suggest/manager.py
+++ b/merino/providers/suggest/manager.py
@@ -426,8 +426,9 @@ def load_providers(disabled_providers_list: list[str]) -> dict[str, BaseProvider
     for provider_id, setting in settings.providers.items():
         # Skip sibling subsystems that share the settings.providers namespace
         # but are not suggest providers (e.g. wcs). The contract is that a
-        # suggest provider entry exposes a `type` field; anything else here
-        # belongs to another subsystem (WCS, etc.) and initializes itself.
+        # suggest provider entry exposes a `type` field. This is documented in
+        # default.toml near the providers section; anything else here belongs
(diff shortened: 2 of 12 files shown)


merino/providers/suggest/sports/backends/sportsdata/common/sports.py
```diff
@@ -1090,28 +1094,50 @@ async def update_events(self, client: AsyncClient, allow_no_teams: bool = False)
...
-                    ttl=timedelta(minutes=5),
-                    cache_dir=self.cache_dir,
-                    headers=self.api_headers(),
-                )
-                self.load_scores_from_source(
-                    response,
-                    event_timezone=local_timezone,
-                )
-                date_list.add(day)
+        for day in self._games_by_date_slugs_to_refresh(
```
gruberb, line 1214: Crucial change, not sure if needed: Instead of relying on the overall schedule we fetch every tick, I thought: "Why not update todays games in an interval, so we are sure we get the scores". Maybe the overall schedule is changing slower etc.

merino/configs/default.toml
```diff
@@ -516,6 +516,12 @@ cache_dir="/tmp"
 # base_url.nfl="https://api.sportsdata.io/v3/nfl/scores/json"
 # base_url.ucl="https://api.sportsdata.io/v4/soccer/scores/json"
 
+[default.providers.wcs]
+# MERINO_PROVIDERS__WCS__LIVE_DATA_ENABLED
+# Keep the /wcs/live mock response by default. DISCO-4231 can flip this to
+# serve Redis-backed in-progress matches after launch approval.
+live_data_enabled = false
```
gruberb, line 528: If we want to switch to live data, we can just override an `ENV`. We can choose to also not have this flag and simply put up a PR to switch it over (need one anyway to remove the fake data)

kschelonka: agree, we'll still need to redeploy affected pods if we use env vars, but this prevents us from redeploying all of merino.

merino/providers/suggest/manager.py
```diff
@@ -424,7 +424,12 @@ def load_providers(disabled_providers_list: list[str]) -> dict[str, BaseProvider
     """
     providers: dict[str, BaseProvider] = {}
     for provider_id, setting in settings.providers.items():
-        # Do not initialize provider if disabled in config.
+        # Skip sibling subsystems that share the settings.providers namespace
+        # but are not suggest providers (e.g. wcs). The contract is that a
+        # suggest provider entry exposes a `type` field; anything else here
+        # belongs to another subsystem (WCS, etc.) and initializes itself.
+        if "type" not in setting:
+            continue
```
kschelonka, line 433: Is this contract documented anywhere? The settings are large and difficult to reason about, especially in code since there aren't any types/linting that works on it

merino/providers/suggest/sports/backends/sportsdata/common/sports.py
kschelonka: agree, this data is not 100% trustworthy so I'd like to err on the side of over-refreshing

merino/providers/wcs/protocol.py
```diff
@@ -62,7 +62,7 @@ def from_team(
         )
 
     @classmethod
-    def from_event_team(cls, team: dict[str, Any]) -> "TeamInfo":
+    def from_event_team(cls, team: dict[str, Any], *, group: str | None = None) -> "TeamInfo":
```
kschelonka: Think this needs rebase for https://github.com/mozilla-services/merino-py/pull/1529/

merino/providers/wcs/provider.py
```diff
@@ -94,16 +102,37 @@ async def get_matches(
...
+            ]
+            return LiveMatchesResponse(matches=matches)
+
+        now = datetime.now(UTC)
+        window_start = now - _LIVE_MATCH_LOOKBACK
+        window_end = now + _LIVE_MATCH_LOOKAHEAD
+        try:
+            events = await self.sport.get_events_by_date(start=window_start, end=window_end)
+        except CacheAdapterError as ex:
+            self._record_cache_error("live", ex)
```
kschelonka, line 124: this seems an actionable error so it should be reported to Sentry, so we can fix it quickly

merino/providers/wcs/provider.py
```diff
@@ -94,16 +102,37 @@ async def get_matches(
...
+            return LiveMatchesResponse(matches=[])
+
+        live_events: list[EventInfo] = []
+        for event in sorted(events, key=lambda e: e.date):
+            if not event.status.is_in_progress():
+                continue
+            event_info = EventInfo.from_event(event)
+            if team_keys is not None and not _has_team(event_info, team_keys):
```
kschelonka: I'm confused by this boolean -- shouldn't it be that we want `_has_team` to return true?

merino/providers/wcs/provider.py
```diff
@@ -165,3 +194,8 @@ def _event_date(event: Event) -> date:
 def _has_team(event: EventInfo, team_keys: frozenset[str]) -> bool:
     """Return True if either side of `event` plays for one of `team_keys`."""
     return event.home_team.key in team_keys or event.away_team.key in team_keys
+
+
+def _matches_team_filter(event: EventInfo, team_keys: frozenset[str] | None) -> bool:
```
kschelonka, line 201: ah I understand now

kschelonka (approved):

PR #1528, opened 2026-05-15
DISCO-4235 TBD teams for WCS

gruberb:
## References

JIRA: [DISCO-4235](https://mozilla-hub.atlassian.net/browse/DISCO-4235)

## Description
World Cup knockout games appear on the SportsData feed before both teams are drawn, and Merino's base `event_from_row` drops any row missing a team `id`, so the widget never sees the bracket until matchups are
  finalized. 

This change overrides `event_from_row` on the `WCS` subclass to detect knockout placeholders (`SeasonType` ==
  `3`, `Group is None`, at least one null team id) and emit an `Event` whose unknown side carries `{"key": "TBD", "name": "TBD", "colors": [], "id": 0}` instead of being dropped, with `_event_team_keys` in `wcs_elimination` filtering TBD out so it never pollutes the eliminated-team set or the active-path calculation. 

The `/matches?date=...` endpoint is unchanged and `TeamInfo.from_event_team` handles the placeholder naturally, returning `icon_url: null` since TBD is absent from the logo manifest and emitting a click query of the form World Cup 2026 TBD vs TBD 05 July 2026.

## PR Review Checklist

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/MC-2330)

[DISCO-4235]: https://mozilla-hub.atlassian.net/browse/DISCO-4235?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/merino/providers/suggest/sports/backends/sportsdata/common/sports.py
+++ b/merino/providers/suggest/sports/backends/sportsdata/common/sports.py
@@ -33,6 +33,7 @@ from merino.providers.suggest.sports.backends.sportsdata.common.data import (
     sportsdata_day_slug,
 )
 from merino.providers.suggest.sports.backends.sportsdata.common.wcs_elimination import (
+    TBD_TEAM_KEY,
     eliminated_team_keys,
     eliminated_team_keys_cache_key,
     parse_eliminated_team_keys,
@@ -40,7 +41,6 @@ from merino.providers.suggest.sports.backends.sportsdata.common.wcs_elimination 
 )
 
 FORCE_IMPORT = ""
-_TBD_TEAM_KEY = "TBD"
 _SEASON_TYPE_POSTSEASON = 3
 
 # When creating a new sport class, add its entry to SPORT_CATEGORY_MAP below.
@@ -987,7 +987,7 @@ class WCS(Sport):
     ) -> tuple[dict[str, Any], str] | None:
         """Return compact team data and search terms for one placeholder row side."""
         if team_id is None:
-            return self._tbd_team_minimal(), _TBD_TEAM_KEY.lower()
+            return self._tbd_team_minimal(), TBD_TEAM_KEY.lower()
 
         team = self.teams.get(team_id)
         if team is None:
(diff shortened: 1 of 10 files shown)


merino/providers/suggest/sports/backends/sportsdata/common/sports.py
```diff
@@ -38,6 +39,8 @@
 )
 
 FORCE_IMPORT = ""
+_TBD_TEAM_KEY = "TBD"
```
kschelonka: comment(blocking): Should discuss contract with mobile team -- Sarah mentioned having pushed an implementation which handles null values explicitly

gruberb: For Suggest, we need to serve `TBD`, for the widget, we will serve `null`.

merino/providers/suggest/sports/backends/sportsdata/common/wcs_elimination.py
```diff
@@ -10,6 +10,7 @@
 
 _FIRST_KNOCKOUT_ADVANCER_COUNT = 32
 _ELIMINATED_TEAM_KEYS_META_KEY = "meta:eliminated_team_keys"
+_PLACEHOLDER_TEAM_KEYS = {"TBD"}
```
kschelonka: comment(non-blocking): soft-blocking, please avoid duplicating magic strings and use _TBD_TEAM_KEY (maybe rename to not be private)

merino/providers/wcs/protocol.py
```diff
@@ -14,6 +14,7 @@
 # Pin WCS flag URLs to the production image bucket so stage and prod render the
 # same assets.
 _LOGO_HOST = "https://storage.googleapis.com/merino-images-prod"
+_TBD_TEAM_KEY = "TBD"
```
kschelonka: comment: please avoid duplicating magic strings

merino/providers/suggest/sports/backends/sportsdata/common/sports.py
kschelonka: This raises a question for me about how these will be served by suggest... testing locally the only way I'm able to get it is using "tbd game" -- "fifa game" and "world cup game" and "july 7 game" all don't work.

I won't block on this but it seems unlikely to show up in suggest

kschelonka (approved):
Approved pending magic string deduplication

PR #1546, opened 2026-05-21
[DISCO-4322] Add circuit breaker for wcs

tiftran:
## References

JIRA: [DISCO-4322](https://mozilla-hub.atlassian.net/browse/DISCO-4322)

## Description
Add circuit breaker to WCS

## PR Review Checklist

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/MC-2348)

### DIFF
--- a/merino/configs/testing.toml
+++ b/merino/configs/testing.toml
@@ -127,7 +127,12 @@ resync_interval_sec = 10800
 domain_data_source = "local"
 
 [testing.providers.wikipedia]
+circuit_breaker_failure_threshold = 1
+circuit_breaker_recover_timeout_sec = 3
 # The backend of the provider. Either "elasticsearch" or "test".
+backend = "test"
+
+[testing.providers.wcs]
 circuit_breaker_failure_threshold = 1
 circuit_breaker_recover_timeout_sec = 3
 backend = "test"
--- a/merino/configs/testing.toml
+++ b/merino/configs/testing.toml
@@ -132,13 +132,14 @@ circuit_breaker_recover_timeout_sec = 3
 # The backend of the provider. Either "elasticsearch" or "test".
 backend = "test"
 
+cron_interval_sec = 60
+resync_interval_sec = 86400
+
 [testing.providers.wcs]
 circuit_breaker_failure_threshold = 1
 circuit_breaker_recover_timeout_sec = 3
 backend = "test"
 
-cron_interval_sec = 60
-resync_interval_sec = 86400
 
 [testing.curated_recommendations.gcs]
 bucket_name = "test-bucket"
--- a/merino/governance/circuitbreakers.py
+++ b/merino/governance/circuitbreakers.py
@@ -29,7 +29,7 @@ Each breaker can be defined as follows:
 from circuitbreaker import CircuitBreaker
 
 from merino.configs import settings
-from merino.exceptions import BackendError
+from merino.exceptions import BackendError, CacheAdapterError
 from merino.providers.suggest.flightaware.backends.errors import FlightawareError
 from merino.providers.suggest.weather.backends.accuweather.errors import (
     AccuweatherError,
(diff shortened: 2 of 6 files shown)


merino/providers/wcs/provider.py
```diff
@@ -27,6 +30,22 @@
 logger = logging.getLogger(__name__)
 
 
+# Because the WCS return multiple different response, these fallback will be used to override the default fallback function
+async def _wcs_matches_fallback_fn(*args, **kwargs) -> MatchesResponse:
```
tiftran: needed to have these here, cause circular dependency problems

merino/web/api_v1_wcs.py
```diff
@@ -39,7 +39,8 @@ async def get_wcs_matches(
     """
     target_date = date or datetime.now(UTC).date()
     team_keys = _parse_team_keys(teams)
-    return await provider.get_matches(target_date, limit, team_keys)
+    matches: MatchesResponse = await provider.get_matches(target_date, limit, team_keys)
```
tiftran, line 42: Had lint issues, wouldn't go away until I explicitly had the type here

merino/governance/circuitbreakers.py
```diff
@@ -93,3 +93,14 @@ class SportsCircuitBreaker(CircuitBreaker):
...
     # When the breaker is open, use this to simply return an empty suggestion list to the caller.
     FALLBACK_FUNCTION = _suggest_provider_fallback_fn
+
+
+class WCSCircuitBreaker(CircuitBreaker):
+    """Circuit breaker for the WCS provider."""
+
+    FAILURE_THRESHOLD = settings.providers.wcs.circuit_breaker_failure_threshold
+    RECOVERY_TIMEOUT = settings.providers.wcs.circuit_breaker_recover_timeout_sec
+    # BackendError is raised by elasticsearch query errors
```
kschelonka: outdated copy-paste: CacheAdapterError and redis

merino/configs/testing.toml
```diff
@@ -127,7 +127,12 @@ resync_interval_sec = 10800
 domain_data_source = "local"
 
 [testing.providers.wikipedia]
+circuit_breaker_failure_threshold = 1
```
kschelonka, line 130: comment(blocking): this change split up the wikipedia config, and it now lost `cron_interval_sec` and `resync_interval_sec` below (absorbed into testing.providers.wcs)

merino/governance/circuitbreakers.py
```diff
@@ -93,3 +93,14 @@ class SportsCircuitBreaker(CircuitBreaker):
...
+    EXPECTED_EXCEPTION = CacheAdapterError
+    # When the breaker is open, use this to simply return an empty suggestion list to the caller.
+    FALLBACK_FUNCTION = _suggest_provider_fallback_fn
```
kschelonka: comment(blocking): I don't think this is appropriate as a default. Recommend removing this line and relying on passing the appropriate fallback method in the decorator

Unfortunately I don't think this class/python linter has a way of requiring the fallback function in the decorator as-is. From reading the source it will raise if circuitbreaker is open and fallback isn't defined. I think since we'll catch in sentry that's preferable than returning a wrong-shaped response accidentally if we forget to override

kschelonka: actually, now that I'm thinking about it more... wouldn't we want to expose errors to the clients? I imagine that it would be a better experience for them to handle errors and simply not refresh data vs. refreshing with empty fallback data

Suggest is a different beast with the multiple providers, not wanting one provider erroring to cause the whole thing to fail

merino/governance/circuitbreakers.py
tiftran: 😓

merino/configs/testing.toml
tiftran: 🤦‍♀️

merino/governance/circuitbreakers.py
tiftran: it's overridden in the decorator currently:https://github.com/mozilla-services/merino-py/pull/1546/changes#diff-3cb8c2f73d41be008afbaa1bcc4f12d8ec4d734950fd4bc1d4de011d930805ffR81 :) 

The latter comment, I think might be a broader team and/or team + mobile discussion

tiftran:
I think the changes would now result in a 500 internal server error. Maybe we want to return a 503 instead? If so, I can do that as a follow up PR?

kschelonka (approved):

kschelonka:
> I think the changes would now result in a 500 internal server error. Maybe we want to return a 503 instead? If so, I can do that as a follow up PR?

Yeah sounds good. 503 would probably be appropriate most of the time, although technically it would depend on what caused the error. we can also populate retry-after hint

PR #1555, opened 2026-05-26
[HNT-2623] Add a publisher constraint to popular today

mlplyler:
## References

JIRA: [HNT-2623](https://mozilla-hub.atlassian.net/browse/HNT-2623)

## Description
Adds a max number per publisher constraint to Popular Today. This is effectively off by default by setting to 100. For de_DE, this is set as 1 meaning we can have only 1 article per publisher in Popular Today.

## PR Review Checklist

- [ X] This PR conforms to the [Contribution Guidelines](https://github.com/mozilla-services/merino-py/blob/main/CONTRIBUTING.md)
- [ X] The PR title starts with the JIRA issue reference, format example `[DISCO-####]`, and has the same title (if applicable)

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/MC-2357)

[HNT-2623]: https://mozilla-hub.atlassian.net/browse/HNT-2623?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/merino/curated_recommendations/article_balancer.py
+++ b/merino/curated_recommendations/article_balancer.py
@@ -1,7 +1,8 @@
 """Balancers for curated recommendation articles."""
 
-from collections import defaultdict
+from collections import Counter, defaultdict
 import math
+import random
 
 from merino.curated_recommendations.corpus_backends.protocol import Topic
 from merino.curated_recommendations.article_balancer_configs import (
@@ -12,13 +13,20 @@ from merino.curated_recommendations.protocol import CuratedRecommendation
 
 
 class ArticleBalancer:
-    """Balance articles by multiple criteria."""
+    """Balance articles by multiple criteria.
+
+    Topic and aggregate counters use the controlled feature_counts namespace.
+    Publisher counters are tracked separately because publisher names are arbitrary
+    external strings and can collide with topic values or aggregate keys like "evergreen".
+    """
 
     def __init__(self, expected_num_articles: int, config: ArticleBalancerConfig) -> None:
         """Initialize limits for target number of articles."""
         self.config = config
         self.article_list: list[CuratedRecommendation] = []
         self.feature_counts: defaultdict[str, int] = defaultdict(int)
+        self.publisher_counts: Counter[str] = Counter()
+        self.enforce_publisher = True
         self.num_expected = 0
         self.evergreen_topics = set(config.evergreen_topics)
         self.subtopic_checker = config.subtopic_checker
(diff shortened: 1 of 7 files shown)


tests/unit/curated_recommendations/fixtures.py
```diff
@@ -54,7 +54,7 @@ def generate_recommendations(
             title="little larry",
             excerpt="is failing english",
             topic=topics[i] if topics is not None else random.choice(list(Topic)),
-            publisher="cohens",
+            publisher=f"publisher:{item_id}",
```
rolf-moz: we do have publisher capping for the 'grid' view. Are you sure this change doesn't affect those tests?

merino/curated_recommendations/article_balancer.py
```diff
@@ -89,6 +103,9 @@ def _does_meet_spec(self, info_dict) -> bool:
             return False
         if info_dict.get("blocked_topics", 0) > self.max_blocked_topics:
             return False
+        for key, count in info_dict.items():
```
rolf-moz: I think might be a little cleaner if _does_meet_spec had separate object for publishers. That way you wouldn't have to create a composite key

publisher_counter = Counter()

if len(publisher_counter) > 0:
   item, count = counts.most_common(1)[0]
   if count and count > self.max_per_publisher:
       return False

rolf-moz: If you do keep as is, maybe create a utility function for the composite key

rolf-moz (changes requested):
Suggested change: Add to the config publisher_enforcement_likelyhood (defaults to 0.85)
In the TopStoriesArticleBalancer constructor it reads the configs and sets a enforce_publisher flag at random based on the likelyhood

tests/unit/curated_recommendations/fixtures.py
mlplyler: Are there tests we should check outside the CI?

merino/curated_recommendations/article_balancer.py
mlplyler: publisher is on its own counter now

tests/unit/curated_recommendations/fixtures.py
rolf-moz: the tests are in the CI. They must not be affected by this change. They are mostly in test_curated_recommendations.py

rolf-moz (approved):

PR #1624, opened 2026-06-18
AIMOD-1340 Update bayes smoothing constant for inferred personalization

rolf-moz:
## References

JIRA: https://mozilla-hub.atlassian.net/browse/AIMOD-1340

## Description
User interests are shared via an interest vector in Merino.

Previously we shared per interest CTR data that was rescaled to 1 and then thresholded to 4 values.

Several weeks ago we transitioned to Bayesian smoothing, where we find the average ctr for that user and use that as a basis for looking at variations off of that. Analysis shows that with that change information content in the interest vector has dropped, and level 1,2,3 of an interest doesn't accurately show the magnitude of it.

The issue is that our smoothing value may be to low for the low clickthrough that rates we have. We are changing it from k=100 to 500, which should lead the 1,2,3 values to show true intensity of interest rather than being driven by a single click.

## PR Review Checklist

- [x ] This PR conforms to the [Contribution Guidelines](https://github.com/mozilla-services/merino-py/blob/main/CONTRIBUTING.md)

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/MC-2426)


mlplyler (approved):

mlplyler:
> Analysis shows that with that change information content in the interest vector has dropped, and level 1,2,3 of an interest doesn't accurately show the magnitude of it.

When I read this, I thought you were going less bayesian, but the next paragraph and the prior change implies a stronger prior...so almost "more" bayesian?

PR #1625, opened 2026-06-19
[HNT-2476] Logic for potd image manifest file

Herraj:
## References

JIRA: [HNT-2476](https://mozilla-hub.atlassian.net/browse/HNT-2476)

## Description
Implements the POTD image manifest pipeline: fetches the Wikimedia RSS feed, downloads and uploads images to GCS, and writes a daily JSON manifest that the provider reads on startup to serve the picture of the day.

- Hardcoded stub is only returned in prod env. Real upload pipeline: fetch RSS feed → parse → download images → upload to GCS → write JSON manifest, will be tested in stage environments in follow up work.
- Added date validation in `parse_potd()`. It rejects entries where published is not equal to today.
- Provider now loads `self.potd` from GCS on startup.
- `get_picture_of_the_day()` simplified to sync.
- Added and expanded unit + integration test coverage for all new methods.

## PR Review Checklist

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/MC-2427)

[HNT-2476]: https://mozilla-hub.atlassian.net/browse/HNT-2476?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


merino/providers/rss/wikimedia_potd/backends/wikimedia_potd.py
```diff
@@ -41,33 +43,72 @@ def __init__(
         self.http_client = http_client
         self.gcs_uploader = gcs_uploader
 
-    async def get_picture_of_the_day(self) -> PictureOfTheDay | None:
-        """Fetch the current Wikimedia Picture of the Day.
+    async def orchestrate_picture_of_the_day_upload(self) -> bool:
```
mmiermans: _Orchestration_ sounds a bit heavy for what I think this function does. What do you think about this name?
```suggestion
    async def upload_picture_of_the_day(self) -> bool:
```

merino/providers/rss/wikimedia_potd/provider.py
```diff
@@ -35,15 +36,30 @@ def __init__(
...
+                thumbnail_image_url=HttpUrl(
+                    "https://prod-images.merino.prod.webservices.mozgcp.net/rss/wikimedia_potd/POTD_2026_04_13.jpg"
+                ),
+                high_res_image_url=HttpUrl(
+                    "https://prod-images.merino.prod.webservices.mozgcp.net/rss/wikimedia_potd/POTD_hi_res_2026_4_13.jpg"
+                ),
+            )
+        # TODO uncomment for stage env testing.
+        # elif self.potd is None:
+        #     self.potd = await asyncio.to_thread(self.backend.fetch_potd_from_gcs_bucket)
```
mmiermans: Is there a Jira task you can reference for this TODO?

merino/providers/rss/wikimedia_potd/backends/wikimedia_potd.py
```diff
@@ -121,3 +166,53 @@ def upload_image(self, image: Image, is_thumbnail: bool) -> str | None:
...
+                content=manifest_json,
+                destination_name=f"rss/wikimedia_potd/POTD_{today}.json",
+                content_type="application/json",
+                forced_upload=True,
+            )
+
+            return True
+        except Exception as ex:
+            sentry_sdk.capture_exception(ex)
+            return False
```
mmiermans: [upload_content](https://github.com/mozilla-services/merino-py/blob/main/merino/utils/gcs/gcs_uploader.py#L47) catches all exceptions and returns the `destination_blob`, so no exceptions will bubble up here. The function will always return `True`, even on upload failures. I believe this affects both `upload_potd_image` and `upload_potd_manifest`.

merino/providers/rss/wikimedia_potd/backends/wikimedia_potd.py
```diff
@@ -41,33 +43,72 @@ def __init__(
...
+            thumbail_url, hi_res_url = image_urls
+
+            potd_to_upload = potd.model_copy(
+                update={"thumbnail_image_url": thumbail_url, "high_res_image_url": hi_res_url}
+            )
+
+            return self.upload_potd_manifest(potd_to_upload)
+        except Exception as ex:
+            sentry_sdk.capture_exception(ex)
+            return False
```
mmiermans, line 77: I think swallowing errors with `except Exception as ex` is generally not a good pattern, because it puts the responsibility on every caller in the call chain to propagate errors. This requires boilerplate code from `try ... except` blocks and error guards such as `if potd is None:`. It also increases the likelihood of bugs, because it is not clear from function names whether errors are returned as `None` or `False`, or thrown as errors. 

What's the downside of letting exceptions bubble up, using middleware to send them to Sentry, and letting Fastly return a 500 error? Is there a requirement to have a graceful fallback response, if we can't deliver the picture of the day?

tests/integration/providers/rss/potd/test_wikimedia_potd.py
```diff
@@ -36,18 +41,157 @@ def fixture_backend(
...
+
+        sentry_capture = mocker.patch(
+            "merino.providers.rss.wikimedia_potd.backends.wikimedia_potd.sentry_sdk.capture_exception"
+        )
+
+        result = await backend.orchestrate_picture_of_the_day_upload()
+        assert result is False
+        sentry_capture.assert_called_once_with(unexpected_error)
+
+
```
mmiermans, line 236: Could you add an integration test for the case where the GCP upload fails? I believe that's currently not working as expected. Here's Claude's stab at one:

```python
@freezegun.freeze_time("2026-06-24")
@pytest.mark.asyncio
async def test_orchestrate_picture_of_the_day_upload_returns_false_when_gcs_upload_fails(
    self,
    backend: WikimediaPictureOfTheDayBackend,
    gcs_storage_client,
    gcs_storage_bucket,
    mocker: MockerFixture,
) -> None:
    """A failed GCS upload must be reported as failure and must not publish a manifest."""
    client_mock: AsyncMock = cast(AsyncMock, backend.http_client)
    client_mock.get.return_value = Response(
        status_code=200,
        content=TEST_RSS_FEED,
        request=Request(method="GET", url=FEED_URL),
    )
    # Images "download" fine; only the GCS upload should fail.
    mocker.patch.object(backend, "download_potd_image").return_value = Image(
        content=b"255", content_type="Image/png"
    )

    # Simulate a real GCS write failure (network / 5xx-after-retries / quota)
    # at the storage layer, so the real GcsUploader.upload_content path runs.
    mocker.patch(
        "google.cloud.storage.Blob.upload_from_string",
        side_effect=Exception("GCS upload failed"),
    )

    result = await backend.orchestrate_picture_of_the_day_upload()

    # Intended contract: the upload failed, so orchestration reports failure.
    assert result is False

    # Intended contract: no manifest is published pointing at images that never uploaded.
    assert backend.fetch_potd_from_gcs_bucket() is None
    blob_names = [
        blob.name for blob in gcs_storage_client.get_bucket(gcs_storage_bucket.name).list_blobs()
    ]
    assert "rss/wikimedia_potd/POTD_2026-06-24.json" not in blob_names
```

tests/data/rss/wikimedia_potd/potd_feed.py
```diff
@@ -0,0 +1,183 @@
+"""Module containing a sample rss response from Wikimedia Picture of the Day feed, used for testing."""
```
mmiermans: Instead of encoding XML in a Python file, could you store this as an XML file? What avoids having to encode the string, which could introduce a discrepancy with the actual RSS received in production.

mmiermans (commented):
I think this code works  as intended, at least in the happy path. I have some questions/suggestions to simplify code.

mmiermans (approved):
Thanks for implementing feedback!

PR #1669, opened 2026-07-13
[DISCO-4331] feat(common): add AsyncBatchQueue

kschelonka:
## References

JIRA: [DISCO-4331]

## Description
Method for processing batches serially and asynchronously, e.g. to make batch requests to an API.

## PR Review Checklist

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/MC-2470)

[DISCO-4331]: https://mozilla-hub.atlassian.net/browse/DISCO-4331?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/merino-common/merino_common/utils/async_batch_queue.py
+++ b/merino-common/merino_common/utils/async_batch_queue.py
@@ -187,7 +187,7 @@ class AsyncBatchQueue(Generic[T]):
         batch: list[T] = []
         loop = asyncio.get_running_loop()
         deadline = loop.time() + self.collection_delay_s
-        shutdown = asyncio.ensure_future(self._shutdown.wait())
+        shutdown = asyncio.create_task(self._shutdown.wait())
         try:
             while len(batch) < self.max_batch_size:
                 remaining = deadline - loop.time()
@@ -289,7 +289,7 @@ class AsyncBatchQueue(Generic[T]):
         within a running event loop.
         """
         if self._current_task is None:
-            self._current_task = asyncio.get_running_loop().create_task(self.run())
+            self._current_task = asyncio.create_task(self.run())
             # Yield once so run() reaches its first await and is_running() is
             # True by the time start() returns.
             await asyncio.sleep(0)
@@ -316,7 +316,7 @@ class AsyncBatchQueue(Generic[T]):
                     await in_flight
                     in_flight = None
                 if batch:
-                    in_flight = asyncio.get_running_loop().create_task(self._process(batch))
+                    in_flight = asyncio.create_task(self._process(batch))
             # On shutdown, finish the in-flight request and then drain the rest.
             if in_flight is not None:
                 await in_flight
--- a/merino-common/merino_common/utils/async_batch_queue.py
+++ b/merino-common/merino_common/utils/async_batch_queue.py
@@ -2,16 +2,17 @@
 
 import asyncio
 import logging
-from collections.abc import Callable, Iterable
+from collections.abc import Callable
 from typing import Generic, TypeVar, Awaitable, Any
 from asyncio import QueueFull
 from opentelemetry.metrics import (
-    CallbackOptions,
     Counter,
     MeterProvider,
-    Observation,
+    _Gauge,
     get_meter_provider,
 )
+import time
+from pydantic import PositiveFloat, PositiveInt
 
 
 class AsyncBatchException(Exception):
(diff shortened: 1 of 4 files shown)


merino-common/merino_common/utils/async_batch_queue.py
```diff
@@ -0,0 +1,380 @@
...
+
+    Must be called from a running event loop; not thread-safe.
+
+    Args:
+        on_batch (Callable[[list[T]], Awaitable[Any]]): async callback to invoke whenever a batch
+            is ready. There is no deadline on this callback, so the ``on_batch`` method itself
+            should handle hung processes.
+        on_error (Callable[[list[T], Exception], Awaitable[Any]], optional): async callback to invoke
+            whenever ``on_batch`` raises. If not provided, errors are logged and otherwise ignored.
+        max_batch_size (int, optional): The max number of items to process in a batch. Must be
```
ncloudioj, line 73: Let's note that if `max_batch_size` is set to be greater than `max_queue_size`, each batch would only be processed when `collection_delay_s` elapses.

merino-common/merino_common/utils/async_batch_queue.py
```diff
@@ -0,0 +1,380 @@
...
+        """Collect a batch, blocking for items until the delay elapses.
+
+        Returns as soon as the batch is full, the schedule delay expires, or
+        shutdown is requested. On shutdown it returns immediately rather than
+        waiting out any remaining delay.
+        """
+        batch: list[T] = []
+        loop = asyncio.get_running_loop()
+        deadline = loop.time() + self.collection_delay_s
+        shutdown = asyncio.ensure_future(self._shutdown.wait())
```
ncloudioj: Nit: per the [doc](https://docs.python.org/3/library/asyncio-future.html#asyncio.ensure_future), `asyncio.create_task()` is preferred than `asyncio.ensure_future()` for creating new tasks.


```suggestion
        shutdown = asyncio.create_task(self._shutdown.wait())
```

merino-common/merino_common/utils/async_batch_queue.py
```diff
@@ -0,0 +1,380 @@
...
+        try:
+            while len(batch) < self.max_batch_size:
+                remaining = deadline - loop.time()
+                if remaining <= 0:
+                    break
+                getter = asyncio.ensure_future(self._queue.get())
```
ncloudioj: Ditto.

merino-common/merino_common/utils/async_batch_queue.py
```diff
@@ -0,0 +1,380 @@
...
+        in_flight: asyncio.Task | None = None
+        try:
+            while not self._shutdown.is_set():
+                batch = await self._wait_for_batch()
+                # Ensure processing batches is done serially
+                if in_flight is not None:
+                    await in_flight
+                    in_flight = None
+                if batch:
+                    in_flight = asyncio.get_running_loop().create_task(self._process(batch))
```
ncloudioj: `asyncio.create_task()` by default spawns a new task off the current running loop.
```suggestion
                    in_flight = asyncio.create_task(self._process(batch))
```

merino-common/merino_common/utils/async_batch_queue.py
```diff
@@ -0,0 +1,380 @@
...
+            self._record_processed(len(batch))
+
+    async def start(self) -> None:
+        """Start the background batch loop.
+
+        Idempotent: repeat calls while running are a no-op. Must be called from
+        within a running event loop.
+        """
+        if self._current_task is None:
+            self._current_task = asyncio.get_running_loop().create_task(self.run())
```
ncloudioj:
```suggestion
            self._current_task = asyncio.create_task(self.run())
```

merino-common/merino_common/utils/async_batch_queue.py
```diff
@@ -0,0 +1,380 @@
...
+
+    async def _process(self, batch: list[T]) -> None:
+        """Run the batch callback, routing failures to the error callback."""
+        try:
+            await self._batch_callback(batch)
+        except Exception as e:
+            self.logger.error("Error processing batch", exc_info=True)
+            self._record_processed(len(batch), error=e)
+            if self._error_callback is not None:
+                await self._error_callback(batch, e)
```
ncloudioj: [blocking] Let's wrap this call inside of try/catch otherwise a raising `_error_callback` would silently kill the run task (as the exception propagates out of the `_process` task) and clear `self._is_running`.

merino-common/merino_common/utils/async_batch_queue.py
```diff
@@ -0,0 +1,380 @@
...
+        """Drain and process everything left in the queue, without delays."""
+        while not self._queue.empty():
+            batch = self._collect_batch_sync()
+            if batch:
+                await self._process(batch)
+
+    def _init_metrics(self, meter_provider: MeterProvider) -> None:
+        """Register OTel instruments for queue size, capacity, and processed count."""
+        meter = meter_provider.get_meter(__name__)
+        meter.create_observable_up_down_counter(
```
ncloudioj: [blocking] We might need an identifier for each queue instance and set that as a metric label, otherwise metric registration collides across instances.

merino-common/merino_common/utils/async_batch_queue.py
```diff
@@ -0,0 +1,380 @@
...
+            self._record_processed(len(batch), error=e)
+            if self._error_callback is not None:
+                await self._error_callback(batch, e)
+        else:
+            self._record_processed(len(batch))
+
+    async def start(self) -> None:
+        """Start the background batch loop.
+
+        Idempotent: repeat calls while running are a no-op. Must be called from
```
ncloudioj, line 351: After `stop()` gets called, `_shutdown` stays set and `_current_task` stays non-None, so `start()` is a permanent no-op. Let's note here that a queue can't be restarted once stopped.

ncloudioj (changes requested):
Overall it looks good, nice abstraction! Once fix a few minor comments & nits, this is good to land.

merino-common/merino_common/utils/async_batch_queue.py
kschelonka: I thought about this, and ultimately decided on using the meter provider to inject the desired resource values. but I can require a name as part of the constructor.

merino-common/merino_common/utils/async_batch_queue.py
kschelonka: I don't know why you'd set it to be greater than the max queue, but yeah that's expected if you do since you will never fill the batch. Might be better instead to clamp to max queue size or throw validation error, what do you think?

merino-common/merino_common/utils/async_batch_queue.py
ncloudioj: I see, that works, too. 

> I can require a name as part of the constructor

Yup, it can still be useful to name each queue so we can attach it to logs and stuff.

merino-common/merino_common/utils/async_batch_queue.py
ncloudioj: Yup, throwing would be great. It's a utility class with free constructor args for consumers to specify, we can help them avoid footguns.

ncloudioj (approved):

PR #1684, opened 2026-07-17
[HNT-2158] Add localization logic for potd

Herraj:
## References

JIRA: [DISCO-2158](https://mozilla-hub.atlassian.net/browse/DISCO-2158)

## Description
- Adding an additional api call to the Wikimedia Commons Action API to discover which languages have an authored picture of the day description for a given day.
- Adding a loop to make N number of requests for the picture of the day Featured API depending on the number of localized languages available.
- Downloads and upload logic remains the same for thumbnail and hi-res images.
- Parses `accept-languages` request header and determines the `description` field on the `potd` based on that and what's available.

## PR Review Checklist

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/MC-2485)

[DISCO-2158]: https://mozilla-hub.atlassian.net/browse/DISCO-2158?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/merino/providers/rss/wikimedia_potd/backends/protocol.py
+++ b/merino/providers/rss/wikimedia_potd/backends/protocol.py
@@ -68,11 +68,12 @@ class WikimediaPictureOfTheDayBackend(Protocol):
         """
         ...
 
-    async def discover_languages(self, date_str: str) -> list[str]:  # pragma: no cover
+    async def discover_languages(self, date_str: str) -> set[str]:  # pragma: no cover
         """Discover the languages that have an authored POTD description for `date_str`.
 
         Returns:
-            A list of language codes, always including the default language.
+            A set of language codes with an authored description, excluding the default
+            language (en).
         """
         ...
 
--- a/merino/providers/rss/wikimedia_potd/backends/protocol.py
+++ b/merino/providers/rss/wikimedia_potd/backends/protocol.py
@@ -21,10 +21,16 @@ class PictureOfTheDay(BaseModel):
         description="High resolution URL of the picture of the day image."
     )
     published_date: str = Field(description="Date when the image was published.")
-    description: str = Field(description="Description of the image in the default language.")
+    description: str = Field(
+        description="Description of the image in the default (en) or the requested language."
+    )
     localized_descriptions: dict[str, str] = Field(
         default_factory=dict,
-        description="Localized descriptions keyed by language code (e.g. 'de', 'es').",
+        description=(
+            "Localized descriptions keyed by language code (e.g. 'de', 'es'). Server-side "
+            "NOTE: only used to localize `description` for the client's Accept-Language and is "
+            "omitted from the API response."
+        ),
     )
     author: str = Field(default="", description="Name of the image's artist.")
     file_page: HttpUrl | None = Field(default=None, description="Commons file page URL.")
--- a/merino/providers/rss/wikimedia_potd/backends/protocol.py
+++ b/merino/providers/rss/wikimedia_potd/backends/protocol.py
@@ -21,7 +21,11 @@ class PictureOfTheDay(BaseModel):
         description="High resolution URL of the picture of the day image."
     )
     published_date: str = Field(description="Date when the image was published.")
-    description: str = Field(description="Description of the image.")
+    description: str = Field(description="Description of the image in the default language.")
+    localized_descriptions: dict[str, str] = Field(
+        default_factory=dict,
+        description="Localized descriptions keyed by language code (e.g. 'de', 'es').",
+    )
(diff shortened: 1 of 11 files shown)


merino/providers/rss/wikimedia_potd/backends/protocol.py
```diff
@@ -21,7 +21,17 @@ class PictureOfTheDay(BaseModel):
         description="High resolution URL of the picture of the day image."
     )
     published_date: str = Field(description="Date when the image was published.")
-    description: str = Field(description="Description of the image.")
+    description: str = Field(
+        description="Description of the image in the default (en) or the requested language."
+    )
+    localized_descriptions: dict[str, str] = Field(
```
jpetto, line 27: this is nice

merino/providers/rss/wikimedia_potd/backends/wikimedia_potd.py
```diff
@@ -75,6 +92,87 @@ async def upload_picture_of_the_day(self) -> bool:
...
+
+        try:
+            response: Response = await self.http_client.get(
+                self.commons_api_url, params=params, headers=WIKIMEDIA_REQUEST_HEADERS
+            )
+            response.raise_for_status()
+            discovered_languages = parse_discovered_languages(response.json())
+
+            # make sure we don't include a duplicate "en"
+            if "en" in discovered_languages:
```
jpetto: guessing "en" is not guaranteed to be in `discovered_languages`? if so, might be good to add a comment

jpetto: also i feel like i'm getting nit-picky, but - 

if you moved the "is `en` in the set?" check to post (after you pull all the languages from discovered), you could just do `discovered_languages.extend('en')` instead of having two different language lists.

merino/providers/rss/wikimedia_potd/backends/protocol.py
```diff
@@ -21,7 +21,17 @@ class PictureOfTheDay(BaseModel):
         description="High resolution URL of the picture of the day image."
     )
     published_date: str = Field(description="Date when the image was published.")
-    description: str = Field(description="Description of the image.")
+    description: str = Field(
+        description="Description of the image in the default (en) or the requested language."
```
jpetto, line 25: question - would it be simpler to keep all descriptions in `localized_descriptions`? is special-casing english necessary?

merino/providers/rss/wikimedia_potd/backends/wikimedia_potd.py
```diff
@@ -75,6 +92,87 @@ async def upload_picture_of_the_day(self) -> bool:
...
+                )
+                continue
+
+            # wikimedia featured api call returns English description as a fallback,
+            # making sure not to add duplicate English descriptions as we have it on the default potd already.
+            # so a call for potd to the Featured API, in Italian ("it"),
+            # could return the fall back "en" version even if the language discovery call returned with "it"
+            # as one of the supported languages for this potd.
+            if description and lang_code != "en":
+                localized_descriptions[lang_code] = description
```
jpetto, line 173: related to last comment - e.g. could we drop logic like this if all languages were kept in a single dict / didn't special case the english description?

jpetto (changes requested):
didn't review the tests yet - wanted to get your thoughts on comments so far.

merino/providers/rss/wikimedia_potd/backends/wikimedia_potd.py
Herraj: So I ended up switching to a `set()` instead of ` list[str]`. This gives us key uniqueness by default.

I've refactored the logic such that `en` is never in the set since the `en` description is always on the base / default `potd` object.

merino/providers/rss/wikimedia_potd/backends/protocol.py
Herraj: The API contract only defines a description field, so returning every localized description in the response would inflate the payload. `localized_descriptions` is a backend only field used to swap the right value into description. It's stripped from the response JSON to keep the contract intact.

jpetto: ah, yes, sorry for being vague. i was referring to storing all (including english) in `localized_descriptions`, then setting the proper `description` on API response based on the language request header. but i don't think there's a big issue with doing it the way you have it.

jpetto (approved):
nice!

PR #1692, opened 2026-07-22
[DISCO-4333] feat(fleece): add an endpoint for search terms submission

ncloudioj:
## References

JIRA: [DISCO-4333](https://mozilla-hub.atlassian.net/browse/DISCO-4333)

## Description

This adds a new endpoint `api/v1/search_terms` to Fleece to receive search terms from Merino for sanitization and logging.

- The new endpoint only validates the request and replies a response code. Sanitization and logging will be added as a followup.
- It records a metric to count the number of search terms received using OTEL metics. Various metric test utility functions are moved to `merino-common.testing` for reusing.
- Also move various Suggest logging models to `merino-common` for the same reason.
- Add `metrics.yaml` to `merino-fleece` to document Fleece's metrics.

Changes are split into two parts to facilitate review.

## PR Review Checklist

[DISCO-4333]: https://mozilla-hub.atlassian.net/browse/DISCO-4333?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/merino-common/merino_common/models/suggest_logging.py
+++ b/merino-common/merino_common/models/suggest_logging.py
@@ -1,0 +1,39 @@
+"""Shared log data models for Suggest logging."""
+
+from datetime import datetime
+
+from pydantic import BaseModel, field_serializer
+
+
+class LogDataModel(BaseModel):
+    """Shared generic log data model."""
+
+    errno: int
+    time: datetime
+    path: str
+    method: str
+
+    @field_serializer("time")
+    def serialize_time(self, v: datetime, **kwargs):
+        """Return a datetime value as an iso formatted str."""
+        return v.isoformat()
+
+
+class SuggestLogDataModel(LogDataModel):
+    """Log metadata specific to Suggest logs."""
+
+    sensitive: bool
+    query: str | None = None
+    code: int
+    rid: str  # Provided by the asgi-correlation-id middleware.
+    session_id: str | None = None
+    sequence_no: int | None = None
+    client_variants: str
+    requested_providers: str
+    country: str | None = None
+    region: str | None = None
+    city: str | None = None
+    dma: int | None = None
+    browser: str
+    os_family: str
+    form_factor: str
--- a/merino-common/merino_common/models/suggest_logging.py
+++ b/merino-common/merino_common/models/suggest_logging.py
@@ -1,12 +1,13 @@
 """Shared log data models for Suggest logging."""
 
 from datetime import datetime
+from typing import Any
 
-from pydantic import BaseModel, field_serializer
+from pydantic import BaseModel, field_serializer, model_serializer
 
 
-class LogDataModel(BaseModel):
-    """Shared generic log data model."""
+class MozlogDataModel(BaseModel):
+    """Generic fields for Mozlog."""
 
     errno: int
     time: datetime
@@ -19,10 +20,9 @@ class LogDataModel(BaseModel):
         return v.isoformat()
 
 
-class SuggestLogDataModel(LogDataModel):
-    """Log metadata specific to Suggest logs."""
+class SuggestRequestParams(BaseModel):
+    """Suggest request parameters specific to Suggest logs."""
 
-    sensitive: bool
     query: str | None = None
     code: int
     rid: str  # Provided by the asgi-correlation-id middleware.
@@ -39,7 +39,27 @@ class SuggestLogDataModel(LogDataModel):
     form_factor: str
 
 
+class SuggestLogDataModel(BaseModel):
+    """Log metadata specific to Suggest logs."""
+
+    # The Suggest search term data log is always flagged as sensitive for
+    # Merino's search terms data log routing. Note that this field should
+    # _not_ be used to flag the search term sanitization result.
+    sensitive: bool
+    mozlog: MozlogDataModel
+    request_params: SuggestRequestParams
+
+    @model_serializer
+    def serialize_flat(self) -> dict[str, Any]:
+        """Dump to a flat dict for backward-compatible logging output."""
+        return {
+            "sensitive": self.sensitive,
+            **self.mozlog.model_dump(),
+            **self.request_params.model_dump(),
+        }
+
+
 class SearchTermsSubmission(BaseModel):
     """Request body for submitting search terms for sanitization."""
 
-    search_terms: list[SuggestLogDataModel]
+    search_terms: list[SuggestRequestParams]
--- a/merino-common/merino_common/models/suggest_logging.py
+++ b/merino-common/merino_common/models/suggest_logging.py
@@ -37,3 +37,9 @@ class SuggestLogDataModel(LogDataModel):
     browser: str
     os_family: str
     form_factor: str
+
+
+class SearchTermsSubmission(BaseModel):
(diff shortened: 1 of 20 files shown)


merino-common/merino_common/models/suggest_logging.py
```diff
@@ -0,0 +1,45 @@
+"""Shared log data models for Suggest logging."""
+
+from datetime import datetime
+
+from pydantic import BaseModel, field_serializer
+
+
+class LogDataModel(BaseModel):
```
kschelonka: comment(non-blocking): I love having shared data models here. I wonder if there's an easy way for us to generate this code automatically from the endpoint so that we don't have to worry about keeping things in sync. Could be a future extension.

merino-common/merino_common/testing/metrics.py
```diff
@@ -0,0 +1,62 @@
+"""Helpers for asserting on OpenTelemetry metrics via an ``InMemoryMetricReader``.
```
kschelonka, line 1: 👍

merino-fleece/metrics.yaml
```diff
@@ -0,0 +1,26 @@
...
+      A counter recording the number of search terms received in each submission
+      to the search-terms endpoint. Incremented once per request by the size of
+      the submitted batch.
+    type: counter
+    labels: []
+    scope: |
+      The "api/v1/search-terms" API endpoint of "merino-fleece".
+    alert_policy: []
+
+pii:
```
kschelonka, line 17: Since we're already using open telemetry for the counter, can we simplify and also use it for the timer rather than aiodogstatsd?

merino-common/merino_common/models/suggest_logging.py
```diff
@@ -0,0 +1,45 @@
...
+    @field_serializer("time")
+    def serialize_time(self, v: datetime, **kwargs):
+        """Return a datetime value as an iso formatted str."""
+        return v.isoformat()
+
+
+class SuggestLogDataModel(LogDataModel):
+    """Log metadata specific to Suggest logs."""
+
+    sensitive: bool
```
kschelonka: comment(blocking): we should have a separate input type for the fleece api request. It doesn't really make sense for sensitive to be a required field that is sent to Fleece since the detection is going to be happening there.

merino-common/merino_common/models/suggest_logging.py
ncloudioj: Yup, the idea is that these log models can be shared between Merino (the producer) and Fleece and other consumers.

> I wonder if there's an easy way for us to generate this code automatically from the endpoint so that we don't have to worry about keeping things in sync. Could be a future extension.

Which endpoint are you referring to? The Suggest log schema is built loosely on top of MozLog for internal uses only, I don't recall we've exposed that via an endpoint anywhere.

merino-fleece/metrics.yaml
ncloudioj: Yes, that's the plan! I didn't convert the other one here to keep the PR scoped. I filed [DISCO-4376](https://mozilla-hub.atlassian.net/browse/DISCO-4376) for that as well as removing the aiodogstatsd stuff and the statsd-exporter sidecar from Fleece as a whole.

merino-common/merino_common/models/suggest_logging.py
ncloudioj: The `sensitive` field of the Suggest log is always set to true for Merino's log routing (see [here](https://github.com/mozilla/global-platform-admin/blob/main/webservices-high/tf/prod/logs.tf#L23)). All search terms data collected from Merino is deemed "sensitive" regardless of whether they have PII or not, so this field is not used to flag PII detection for each request.

I can add a comment to note that here, does that make sense?

merino-common/merino_common/models/suggest_logging.py
kschelonka: [SearchTermsSubmission](https://github.com/mozilla-services/merino-py/blob/disco-4333/merino-common/merino_common/models/suggest_logging.py#L45) is made of list of SuggestLogDataModel, which is what we send to [fleece endpoint ](https://github.com/mozilla-services/merino-py/blob/disco-4333/merino-fleece/merino_fleece/api/v1/search_terms.py#L31)

merino-common/merino_common/models/suggest_logging.py
kschelonka: Maybe it would help to separate the fleece /search-terms input type from this log structure then? Just to make it clear what Merino needs to provide vs. what can be computed by fleece

merino-common/merino_common/models/suggest_logging.py
ncloudioj: Oh, you were referring to the `search_terms` endpoint. Thanks for the clarification!

> I wonder if there's an easy way for us to generate this code automatically from the endpoint so that we don't have to worry about keeping things in sync.

Can you expand on the code generation part? Do you mean to generate those models from the endpoint? If so, I am a little confused as we also use them (e.g. `SearchTermsSubmission` which in turn depends on `SuggestLogDataModel`) to implement the `search_terms` endpoint, I wonder how we can generate model code from that.

merino-common/merino_common/models/suggest_logging.py
ncloudioj: Agreed, the Fleece endpoint input should be decoupled from the log structure. Technically, Fleece only needs those fields of  `SuggestLogDataModel` but not `LogDataModel`, I keep it this way purely for the sake of backward compatibility as the downstream log consumer might use fields from `LogDataModel`, if so, Merino needs to pass them to Fleece still.

Let me check that, if not, I can do the separation in this PR.

ncloudioj: Refactored the models via 27a9003e9dc4986b653c42769e524746f5158486 to decouple Fleece/search-terms input model from the suggest log one. Let me know what you think.

merino-common/merino_common/models/suggest_logging.py
kschelonka: Maybe this isn't a python/fastAPI thing, but in typescript you can generate code models from the openapi yaml contract (not internal, the api interfaces) for example

again this is not a blocking thing, just wondering if there's a way to make it easier to keep models and api contract up to date and usable for both internal and external clients

kschelonka (approved):
thanks!

merino-common/merino_common/models/suggest_logging.py
kschelonka: Looks good, thanks!

merino-common/merino_common/models/suggest_logging.py
ncloudioj: Ah, I see. Thanks again for clarifying!

FastAPI does a similar thing but other way around, it generates openapi docs based on your API's data models (i.e. Pydantic models). For Merino, all its API models are documented in the "schema" section [here](https://merino.services.mozilla.com/docs#). We normally treat those models as the single source of truth and let FastAPI generate openapi docs for them as part of the API docs.

PR #1702, opened 2026-07-31
[Disco-4339] feat(merino-fleece): sanitize search terms in the background 

ncloudioj:
## References

JIRA: [DISCO-4339](https://mozilla-hub.atlassian.net/browse/DISCO-4339)

## Description

This implements full search term sanitization (email, digit, and NER) for Fleece.

Change summaries:
- Sanitization is performed in the background off the path of the `search-terms` handler.
- Modeled after the `search_terms` message handler for message queuing and batch processing.
- Add a new method to the pii module for batch inference via `spaCy.nlp()`. All NER detection calls are executed on a thread pool to avoid blocking the main thread / event loop.

For reviewers:
- Commits are divided into logically related units, feel free to check them out in that order.
- Logging sanitized search terms is left out in this PR, will do that in a followup.

## PR Review Checklist

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/MC-2503)

[DISCO-4339]: https://mozilla-hub.atlassian.net/browse/DISCO-4339?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/merino-common/tests/unit/utils/test_async_batch_queue.py
+++ b/merino-common/tests/unit/utils/test_async_batch_queue.py
@@ -697,3 +697,69 @@ async def test_error_callback_metric_records_failed_error_callback() -> None:
     assert attributes is not None
     # error.type reflects the error callback's own exception, not on_batch's.
     assert attributes["error.type"] == "RuntimeError"
+
+
+@pytest.mark.asyncio
+async def test_remaining_capacity_tracks_queue_headroom() -> None:
+    """`remaining_capacity` reports headroom, shrinking as items are queued and
+    never dropping below zero once the queue is full.
+
+    The batcher is deliberately not started, so nothing drains the queue and the
+    reported headroom is attributable solely to the puts.
+    """
+    batcher: AsyncBatchQueue[int] = AsyncBatchQueue(
+        on_batch=on_batch_noop,
+        max_batch_size=2,
+        max_queue_size=3,
+        collection_delay_s=LONG_COLLECTION_DELAY_S,
+        meter_provider=None,
+    )
+
+    assert batcher.remaining_capacity() == 3, "a fresh queue has its full capacity"
+
+    batcher.put(1)
+    assert batcher.remaining_capacity() == 2
+
+    batcher.put(2)
+    batcher.put(3)
+    assert batcher.remaining_capacity() == 0, "a full queue reports no headroom"
+
+    # The queue is full, so the next put is rejected rather than making the
+    # remaining capacity go negative.
+    with pytest.raises(QueueFullException):
+        batcher.put(4)
+    assert batcher.remaining_capacity() == 0
+
+
+@pytest.mark.asyncio
+async def test_remaining_capacity_recovers_as_batches_are_collected() -> None:
+    """Headroom is returned once queued items are collected into a batch, so a
+    producer that backs off on a full queue can make progress again.
+    """
+    processed: list[int] = []
+
+    async def on_batch(batch: list[int]) -> None:
+        processed.extend(batch)
+
+    batcher: AsyncBatchQueue[int] = AsyncBatchQueue(
+        on_batch=on_batch,
+        max_batch_size=2,
+        max_queue_size=2,
+        collection_delay_s=0.02,
+        meter_provider=None,
+    )
+    await batcher.start()
+
+    batcher.put(1)
+    batcher.put(2)
+    assert batcher.remaining_capacity() == 0
+
+    async def batch_collected() -> None:
+        while batcher.remaining_capacity() < 2:
+            await asyncio.sleep(0.005)
+
+    await asyncio.wait_for(batch_collected(), timeout=5.0)
+    await batcher.stop()
+
+    assert processed == [1, 2]
+    assert batcher.remaining_capacity() == 2
--- a/merino-fleece/merino_fleece/api/v1/search_terms.py
+++ b/merino-fleece/merino_fleece/api/v1/search_terms.py
@@ -1,15 +1,25 @@
 """Search terms submission endpoint."""
 
-from fastapi import APIRouter
+import logging
+
+from fastapi import APIRouter, Depends, HTTPException
 from opentelemetry import metrics
 from pydantic import BaseModel
 
 from merino_common.models.suggest_logging import SearchTermsSubmission
+from merino_common.utils.async_batch_queue import AsyncBatchException
+from merino_fleece.message_handlers.search_terms import MessageHandler, get_message_handler
+
+logger = logging.getLogger(__name__)
 
 _meter = metrics.get_meter("fleece")
 _search_terms_received_counter = _meter.create_counter(
     name="api.search_terms.receive.count",
     description="Number of search terms received by the submission endpoint.",
+)
+_search_terms_rejected_counter = _meter.create_counter(
+    name="api.search_terms.reject.count",
+    description="Number of search terms rejected because they could not be queued.",
 )
 
 router = APIRouter(tags=["search-terms"])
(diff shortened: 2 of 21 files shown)


merino-common/tests/unit/utils/test_async_batch_queue.py
```diff
@@ -697,3 +697,69 @@ async def error_callback_recorded() -> None:
     assert attributes is not None
     # error.type reflects the error callback's own exception, not on_batch's.
     assert attributes["error.type"] == "RuntimeError"
+
+
+@pytest.mark.asyncio
+async def test_remaining_capacity_tracks_queue_headroom() -> None:
```
kschelonka, line 703: points for thoroughness, but rather heavyweight for a math function 😂

merino-fleece/merino_fleece/api/v1/search_terms.py
```diff
@@ -28,13 +38,47 @@ class SearchTermsResponse(BaseModel):
...
+    Raises:
+        HTTPException: 503 when the terms cannot be queued, so the submitter can
+            back off rather than have its data silently dropped.
     """
     count = len(body.search_terms)
     _search_terms_received_counter.add(count)
 
+    # Admit the whole submission or none of it. Enqueuing term by term and failing
+    # partway would leave some terms queued while the submitter sees an error and
+    # drops the batch, so neither side's accounting would be right.
```
kschelonka, line 61: agree on wanting to not have partial batches.

alternatively you could spill over into the pubsub queue. it depends on how fastapi works exactly (need to look it up) but if we're operating without much headroom you might still encounter this error from another request thread racing for the same resources

merino-fleece/merino_fleece/api/v1/search_terms.py
```diff
@@ -28,13 +38,47 @@ class SearchTermsResponse(BaseModel):
...
+            "Rejected search term submission: insufficient queue capacity",
+            extra={"submitted_count": count, "remaining_capacity": capacity},
+        )
+        raise HTTPException(status_code=503, detail="Search term queue is at capacity")
+
+    for index, term in enumerate(body.search_terms):
+        try:
+            handler.put(term)
+        except (AsyncBatchException, RuntimeError) as ex:
+            # The capacity check above is exact for this request -- there is no await
```
kschelonka, line 75: aha, pre-empted https://github.com/mozilla-services/merino-py/pull/1702/files#r3693209302

seems like a good place to spill over to pubsub

merino-fleece/merino_fleece/message_handlers/search_terms/handler.py
```diff
@@ -0,0 +1,145 @@
...
+
+_meter = metrics.get_meter("fleece")
+_sanitize_counter = _meter.create_counter(
+    name="search_terms.sanitize",
+    unit="{item}",
+    description="Number of search terms sanitized, labeled by the detected PII type.",
+)
+
+
+class MessageHandler:
```
kschelonka, line 32: This abstraction feels unnecessary to me. `sanitize_batch` doesn't use `self` so could be a module function, and the rest of the class just wraps the underlying queue functions. `AsyncBatchQueue` was already meant to fulfill the role of `MessageHandler` (constructed with its callbacks at app start and owned by the app lifecycle) , which is why it's generic and why `on_batch`/`on_error`are constructor params. Injecting the queue itself should work here, and would remove a lot of duplicated code and tests. It would also simplify error handling and the associated test code (no more RuntimeError on queue not started).

if there's a constraint I'm missing that informs this design I'd like to hear more, otherwise I'd push us towards deduplication to reduce maintenance complexity

merino-fleece/merino_fleece/api/v1/search_terms.py
ncloudioj: Good call out!

FastAPI runs on asyncio, which is single-threaded cooperative multitasking (rather than preemptive). An async coroutine runs until it hits an await statement or finishes, meaning that `handler.remaining_capacity()` and `handler.put()` down below will be executed sequentially (no await statements between the two calls) without being preempted by other request coroutines.

Now `handler.put()` could raise the following exceptions:
- `RuntimeError` if `handler._queue` is None.
- `QueueShutDownException` if `handler._queue` has been shut down.
- `QueueFull` if `handler._queue` is full, which is already guarded by the prior check with `remaining_capacity()`.
 
So it shouldn't cause any partial submissions that would require a spill over, correct?

merino-fleece/merino_fleece/message_handlers/search_terms/handler.py
ncloudioj: Exactly, we will be adding several members to `self` to facilitate `sanitize_batch` in the near future. For instance, all the live AMP keywords are deemed non-PII and shouldn't be sanitized, we need to download them from the source beforehand and `handler.start()` is a good place for that. Having a message handler to manage all its states (including the queue) and the life cycle callbacks seems clearer to me.

Does that make sense?

merino-common/tests/unit/utils/test_async_batch_queue.py
ncloudioj: lol, i disclosed the intention when spec'd out `remain_capacity()`, claude took that in full action! Apparently, it got heavily inspired by other tests here as well. :)

merino-fleece/merino_fleece/message_handlers/search_terms/handler.py
kschelonka: Yeah that makes sense, but I'd argue that's sanitizer state, not queue state. A `Sanitizer` class could own the AMP keywords and whatever refresh logic they need, then get passed in as `AsyncBatchQueue(on_batch=sanitizer.sanitize_batch)`. Decoupling sanitization from the queue makes it easier to unit test and lets us drop the tests that only re-cover `AsyncBatchQueue`.

Somewhat more speculative but I think keeping sanitization separate and modular gets more useful as the logic grows. AMP allowlisting, entity detection, and regex matching each carry their own state, and I'd expect them to end up as separate functions composed into the batch callback (each stage handing the still-unclassified terms to the next) rather than more members on one handler which would make bootstrapping and testing more difficult.

merino-fleece/merino_fleece/api/v1/search_terms.py
kschelonka: gotcha, so then it's less about partial submissions and more about "who handles the overload case" -- delegating back to caller if we think that we'll recover the space or handling it internally. I don't have strong opinions here, but I guess since we're already going to be having the backup queue that Merino calls then it would make sense to just let Merino handle the retries (and if it keeps getting queue full 503 then Merino will put onto pubsub anyway)

merino-fleece/merino_fleece/message_handlers/search_terms/handler.py
kschelonka: I won't block on this because I can see where we might want a wrapper class in the future to decorate behavior with metrics or handle dependent shutdown/startup state. It's just the current passthrough state that I question

kschelonka (approved):

merino-fleece/merino_fleece/api/v1/search_terms.py
ncloudioj: Exactly, will note that in the docstring for clarity.

merino-fleece/merino_fleece/message_handlers/search_terms/handler.py
ncloudioj: Agreed with your points. I am still tweaking the boundaries among the sanitizers, the queue, other supporting components, and the driver (currently modeled as the message handler). Should be able to land on a better implementation in future iterations.

PR #1717, opened 2026-08-10
[Disco 4414] feat(fleece): add exemption for search term sanitization

ncloudioj:
## References

JIRA: [DISCO-4414](https://mozilla-hub.atlassian.net/browse/DISCO-4414)

## Description

For feature parity, this allows Fleece to exempt AMP keywords from search term sanitization. As a next step, we can also add other sanitization exempts (e.g. Sports/Stocks/Flights keywords) in the future.

Note that it doesn't use the compact keyword indices from the Merino Rust extension as they carry a lot of unused data, we can add a special index type for this in the future if necessary.

**For reviewers**, please review by commits as I had to move some common utilities (cron and http client) from merino to merino-common.  

## PR Review Checklist

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/MC-2518)

[DISCO-4414]: https://mozilla-hub.atlassian.net/browse/DISCO-4414?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


merino-fleece/merino_fleece/sanitize/exempts/amp.py
```diff
@@ -0,0 +1,232 @@
...
+            self.cron_task = None
+        await self.http_client.aclose()
+
+    def is_exempt(self, search_term: str) -> bool:
+        """Return whether the search term matches an AMP keyword.
+
+        MARS keywords are lowercase, so the term is normalized the same way merino's adm
+        provider normalizes queries before looking them up.
+        """
+        return search_term.strip().lower() in self.keywords
```
gruberb: I think we need to do `lstrip()` here, or we loose trailing spaces, which we need?

merino-fleece/merino_fleece/sanitize/exempts/amp.py
```diff
@@ -0,0 +1,232 @@
...
+            if not suggestions:
+                logger.warning(f"MARS returned empty suggestions for {segment}")
+                _fetch_counter.add(1, {**tags, "status": "empty_response"})
+                return None, False
+
+            keywords = {
+                normalized
+                for suggestion in suggestions
+                for keyword in suggestion.get("keywords") or ()
+                if (normalized := keyword.strip().lower())
```
gruberb: We need to remove `strip()` or we loose trailing spaces

merino-fleece/merino_fleece/sanitize/exempts/amp.py
```diff
@@ -0,0 +1,232 @@
...
+                headers=headers,
+            )
+
+            if response.status_code == 304:
+                _fetch_counter.add(1, {**tags, "status": "not_modified"})
+                return None, False
+
+            response.raise_for_status()
+
+            suggestions = response.json()["suggestions"]
```
gruberb: We have to check if `keywords` is a list of strings. If we, by accident, get a string (e.g. "iphone 15"), we would iterate over characters, and create exemptions for i and 5 etc.

So something like:

```python
payload = response.json()
suggestions = payload["suggestions"]

if not isinstance(suggestions, list):
    raise ValueError("MARS 'suggestions' must be a list")

keywords: set[str] = set()

for suggestion in suggestions:
    if not isinstance(suggestion, dict):
        raise ValueError("Each MARS suggestion must be an object")

    raw_keywords = suggestion.get("keywords")
    if not isinstance(raw_keywords, list) or not all(
        isinstance(keyword, str) for keyword in raw_keywords
    ):
        raise ValueError("MARS suggestion 'keywords' must be a list of strings")

    for keyword in raw_keywords:
        if keyword.strip(): 
            keywords.add(keyword.lower()) 
```

gruberb (dismissed):
I think we need to remove `strip` twice, and safe guard against a possible String in the response.

merino-fleece/merino_fleece/sanitize/exempts/amp.py
ncloudioj: Yup, good catch, thanks!

merino-fleece/merino_fleece/sanitize/exempts/amp.py
ncloudioj: Makes sense. I will do some basic checks here for now, the long-term plan is to get merino-ext to handle the ingestion and indexing.

ncloudioj: Resolved via 1974307c1c7ca6f7efe25cc03a041d3a18315d71.

merino-fleece/merino_fleece/sanitize/exempts/amp.py
```diff
@@ -0,0 +1,256 @@
...
+"""
+
+import asyncio
+import contextlib
+import logging
+import time
+from urllib.parse import urljoin
+
+import httpx
+from merino_common.utils import cron
```
gruberb, line 14: Moving cron under `merino_common` takes its logger outside the namespace configured by `configure_logging`. `dictConfig` defaults to disabling existing unlisted loggers, and Fleece imports this module before configuring logging, so `merino_common.utils.cron` becomes disabled.

merino-fleece/merino_fleece/message_handlers/search_terms/handler.py
```diff
@@ -125,6 +128,12 @@ async def sanitize_batch(self, batch: list[SuggestRequestParams]) -> None:
 
         for term in batch:
             query = (term.query or "")[:QUERY_CHARACTER_MAX]
+
+            # Exemption wins over detection by definition, and is the cheaper check.
+            if query and exempts.is_exempt(query):
```
gruberb, line 133: We check the exemption against the truncated query, but later log the original full query. Could we either classify the full query or only log the part we actually classified, so an unchecked suffix cannot end up in the logs?

gruberb (commented):
I think we have to fix those two as well. But double check, it's a large PR, I might have missed some information.

merino-fleece/merino_fleece/message_handlers/search_terms/handler.py
ncloudioj: Good catch, will drop that truncation here as a followup to favor prefiltering those queries upstream in Merino ([DISCO-4413](https://mozilla-hub.atlassian.net/browse/DISCO-4413)).

merino-fleece/merino_fleece/sanitize/exempts/amp.py
ncloudioj: Ah, thanks! This one is nasty, fixed via 6b5f0c4fa3ecf8afcb9a2c14f7819070cd9a030c.

gruberb (approved):

PR #1720, opened 2026-08-11
[HNT-2920] feat: remove the large tile from Popular Today behind an experiment

mmiermans:
## References

- JIRA: [HNT-2920](https://mozilla-hub.atlassian.net/browse/HNT-2920)
- Nimbus experiment: [new-tab-2-column-no-large-card](https://experimenter.services.mozilla.com/nimbus/new-tab-2-column-no-large-card), branch `treatment`

## Description

Nova changed the layout breakpoints, so more users land on the 2-column layout, where Popular Today leads with a **large tile spanning both columns** — no ad above the fold. This serves a Popular Today layout without that tile, so the first row becomes an organic tile plus an ad. The experiment `new-tab-2-column-no-large-card` exists only to let us QA this internally: only its **treatment** branch gets the new layout, and all other clients are unaffected. Once the layout is approved, the gate comes out and the layout applies to everyone.

### Layout comparison

layout_comparison

### Notes for review

- **Why 8 tiles, not 7.** A large tile is 2 columns wide. With 7 mediums the 4-column grid has a partial row, and the client hides trailing partial rows, so 3 tiles would not render. Eight fills 4, 2 and 1 columns exactly.
- **3 columns is unchanged** — there is no large tile there. The 8th tile only keeps the position set consistent across breakpoints, which the client requires, and lands in a hidden trailing row. Without the train hop it renders as a row with 1 of 3 columns filled, until the Sept 1 release.
- **The tile order is deliberate.** `position` is one index shared by stories and ads, and ads are placed from the 1-column layout only, so ads sit at positions 1 and 5 everywhere. Neither ad moves at 4 or 3 columns, so the experiment isolates the large tile; at 2 columns the row-1 ad moves, which is the change being measured.

## QA

Scott [tested](https://mozilla.slack.com/archives/C09PBCCN0M6/p1786551850135679?thread_ts=1786456151.759449&cid=C09PBCCN0M6) the new layout in Firefox:

<table>
<tr>
<td width="38%">widest window</td>
<td width="34%">narrower window</td>
<td width="28%">narrowest window</td>
</tr>
</table>

## PR Review Checklist

[HNT-2920]: https://mozilla-hub.atlassian.net/browse/HNT-2920


merino/curated_recommendations/layouts.py
```diff
@@ -200,3 +200,63 @@
...
+layout_8_tiles_2_ads = Layout(
+    name="8-double-row-2-ad",
+    responsiveLayouts=[
+        ResponsiveLayout(
+            columnCount=4,
+            tiles=[
+                Tile(size=TileSize.MEDIUM, position=0, hasAd=False, hasExcerpt=False),
+                Tile(size=TileSize.MEDIUM, position=2, hasAd=False, hasExcerpt=False),
+                Tile(size=TileSize.MEDIUM, position=3, hasAd=False, hasExcerpt=False),
+                Tile(size=TileSize.MEDIUM, position=1, hasAd=True, hasExcerpt=False),
```
jpetto, line 216: non-blocking, but wondering why these aren't listed in `position` order 😅

tests/integration/api/v1/curated_recommendations/test_curated_recommendations.py
```diff
@@ -1896,6 +1896,14 @@ def test_sections_inferred_contextual_ranking_result_for_cohort(
...
+                "experimentName": ExperimentName.NO_LARGE_CARD_EXPERIMENT.value,
+                "experimentBranch": "treatment",
+            },
+            {
+                "experimentName": ExperimentName.NO_LARGE_CARD_EXPERIMENT.value,
+                "experimentBranch": "control",
+            },
         ],
     )
     def test_sections_layouts(self, sections_payload, experiment_payload, client: TestClient):
```
jpetto, line 1895: this test seems to have a number of parameterized inputs that aren't checked at all in the test body - should/could those be removed?

tests/integration/api/v1/curated_recommendations/test_curated_recommendations.py
```diff
@@ -1920,10 +1928,18 @@ def test_sections_layouts(self, sections_payload, experiment_payload, client: Te
...
-            experiment_payload.get("experimentName") == ExperimentName.ML_SECTIONS_EXPERIMENT.value
-            or experiment_payload.get("experimentName") is None
+            experiment_name == ExperimentName.NO_LARGE_CARD_EXPERIMENT.value
+            and experiment_branch == "treatment"
+        ):
+            # HNT-2920: Popular Today drops its large tile in the treatment branch
+            assert first_section["layout"]["name"] == "8-double-row-2-ad"
+        elif experiment_name in (
+            None,
+            ExperimentName.ML_SECTIONS_EXPERIMENT.value,
```
jpetto: i don't see this experiment referenced anywhere else in the code - should/could we remove it entirely?

jpetto (changes requested):
one very small tidiness nit, and a couple questions about clean up that we should do while we're here if possible.

merino/curated_recommendations/layouts.py
mmiermans: Good question.. `position` maps a tile to a recommendation, with the purpose of putting the most engaging recommendations in the most visible spots. The order of tiles is what the client turns into CSS order, to horizontally stack them. Unfortunately the client doesn't support ads changing position across column counts; it uses the 1-column layout's `hasAd` positions only. 🙃 I know..

I'll add a comment to this effect here. This is a good reason to bring the layout responsibility to the client, because there are several client-specific constraints that we didn't anticipate, and didn't capture in the layout schema. If we weren't moving the layout to the client, then I'd advocate for doing more (in the form of tests or a playground) but I think we should leave it at a comment.

tests/integration/api/v1/curated_recommendations/test_curated_recommendations.py
```diff
@@ -2424,45 +2425,6 @@ def test_sections_interest_picker(
...
-            assert (
-                interest_picker_response["title"] == "Follow topics to fine-tune your experience"
-            )
-        else:
-            assert interest_picker_response is None
-        # Ensure top_stories_section always has receivedFeedRank == 0
-        top_stories_section = data["feeds"].get("top_stories_section")
-        assert top_stories_section is not None
-        assert top_stories_section["receivedFeedRank"] == 0
-
```
mmiermans, line 2465: After removing the `new-tab-ml-sections` experiment, this test becomes a duplicate of `test_sections_interest_picker`, which is preserved. (If Pytest parameterize had been used, we could've avoided the duplicate code.)

jpetto (approved):
ty for the cleanup!

PR #1729, opened 2026-08-17
[DISCO-4408] Update the data source for Wikipedia Indexer job

misaniwere:
## References

JIRA: [DISCO-4408](https://mozilla-hub.atlassian.net/browse/DISCO-4408)

## Description
- Upstream notice: https://dumps.wikimedia.org/other/cirrussearch/DEPRECATED.txt
- New location: https://dumps.wikimedia.org/other/cirrus_search_index/

Wikimedia relocated and reshaped the CirrusSearch index dumps, so the
`wikipedia-indexer copy-export` job can no longer find them.

This PR updates the data source to use the new location with the following approach:
- `copy-export` traverses the dated snapshots newest-first and takes the first one
that is both newer than what's on GCS and complete (`_SUCCESS` present)
- The shards are concatenated into one GCS object byte-for-byte. bzip2 allows
compressed streams to be joined and still read as a single file, so no
decompress/recompress happens during the copy

## PR Review Checklist

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/MC-2530)

[DISCO-4408]: https://mozilla-hub.atlassian.net/browse/DISCO-4408?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/merino/jobs/wikipedia_indexer/filemanager.py
+++ b/merino/jobs/wikipedia_indexer/filemanager.py
@@ -1,12 +1,12 @@
 """Manage files between Wikimedia exports and gcs bucket"""
 
+import bz2
 import logging
 import re
 from datetime import datetime as dt
-from gzip import GzipFile
 from html.parser import HTMLParser
 from typing import Generator, Optional, Pattern
-from urllib.parse import urljoin
+from urllib.parse import unquote, urljoin
 from merino.configs import settings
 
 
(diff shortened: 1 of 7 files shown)


merino/jobs/wikipedia_indexer/filemanager.py
```diff
@@ -80,43 +93,71 @@ def __init__(
...
+        snapshots.sort(key=lambda pair: pair[1], reverse=True)
+        return snapshots
 
-        return None
+    def _list_shards(self, snapshot_url: str) -> list[str]:
+        """List the ordered shard URLs for this language within a snapshot.
+
+        Returns an empty list if the export is missing or not yet fully published.
+        """
+        index_url = urljoin(snapshot_url, f"index_name={self.language}wiki_content/")
```
kschelonka, line 139: you could get a small performance benefit here by checking for the success marker first (e.g. with a HEAD request) but the juice is probably not worth the squeeze for payloads of this size (couple kilobytes or less)

merino/jobs/wikipedia_indexer/filemanager.py
```diff
@@ -154,36 +195,77 @@ def stream_latest_dump_to_gcs(self, latest_gcs: Optional[Blob] = None) -> Blob:
...
+                ProgressReporter(logger, "Copy", self.base_url, name, content_len)
+                if content_len > 0
+                else None
+            )
+            completed = 0
+            writer: BlobWriter
+            with blob.open("wb") as writer:
+                for shard_url in shard_urls:
+                    with requests.get(  # nosec
+                        shard_url, stream=True, headers=WIKIMEDIA_REQUEST_HEADERS
```
kschelonka, line 262: this would benefit significantly from async rather than a blocking loop (like Promise.allSettled in tyepscript), since it's multiple large files.

misaniwere: hmm i'm not sure async would work here since the shards have to be in the right order

misaniwere: actually, let me double check that because if they do not have to be in order, then your suggestion might work

kschelonka: you could still download them concurrently even if they have to be written in order, but yes good to check since I wouldn't expect ordering to matter with file indexing (but I haven't investigated the files so that's just a hunch)

misaniwere: you're right, they do not need to be downloaded in order. although, the current approach streams each shard through in 40MB chunks, so nothing is ever held. downloading whole shards concurrently would mean temporarily holding them somewhere before they're written, and at ~600MB per shard, we'll have to confirm if we have enough memory or disk space on airflow to do that.

misaniwere: we could also upload each shard temporarily to gcs and then merge them into a single blob. we'd need to come up with a way to clean up incomplete uploads or failures

kschelonka: we could also just stream the shards directly vs. merging them into a whole file. but we don't need to prematurely optimize, it's totally reasonable for a job of this frequency to see what the performance impact is and decide

kschelonka (approved):

PR #1732, opened 2026-08-18
HNT-2939: add circuit breaker to corpus recommendations fetches

jpetto:
## References

JIRA: [HNT-2939](https://mozilla-hub.atlassian.net/browse/HNT-2939)

## Description

the overall intent of this PR is to reduce corpus load/down time during a merino deploy, which doubles the k8s pod count, which can overwhelm the corpus with requests. importantly, the intent is _not_ to reduce the number of errors reported - the circuitbreaker decorator will continue to raise, just without a corpus API call.

this PR makes two notable changes:

1. adds a circuitbreaker decorator to the sections backend `fetch` method.

the circuitbreaker decorator sits between the retry decorator and the stale while revalidate decorator. the intention is to stop attempting to hit the corpus while the circuitbreaker is open, which will happen after one round of retries (currently scoped at 3) fails.

the circuitbreaker must raise, as any other return value will be cached by the stale while revalidate decorator.

2. stops retrying `fetch` failures on the sections backend `fetch` method if a `ValueError` raises

a `ValueError` while processing the results of the `fetch` call (e.g. parsing JSON) should not be retried, as there is very little chance a retry will help anything. a `ValueError` would be better raised immediately rather than making 2 more requests against the corpus.

## Testing

...or the lack thereof.

due to the stale while revalidate decorator, testing the circuitbreaker end to end is challenging. the ideal imo is to test the interplay of the decorators, but the combined complexity makes this onerous - retries must fail and the circuitbreaker must open and the SWR cache must handle both cold start and warmed scenarios. testing is also error prone due to the SWR decorator implementation, which kicks off async API requests that are specifically _not_ awaited (making timing/call counts flaky).

the circuitbreaker itself is basic in its configuration and functionality - i'm not sure any test directly against the circuitbreaker would do more than verify the python circuitbreaker package works.

while i don't feel great having _no_ tests here, i am hard pressed to come up with a testing scenario that is both meaningful and predictable given the implementation of the SWR decorator. i'm open to ideas.

### Update

had the claude robot take a swing at adding tests, and it did a good job of finding ways to verify the circuit breaker directly (using the circuitbreaker package API) in addition to testing HTTP call counts. cleaner than modeling after other approaches/patterns. (though i did clean up some of the implementation.)

## PR Review Checklist

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/MC-2533)

[HNT-2939]: https://mozilla-hub.atlassian.net/browse/HNT-2939?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/merino/curated_recommendations/corpus_backends/caching.py
+++ b/merino/curated_recommendations/corpus_backends/caching.py
@@ -165,11 +165,6 @@ async def _get_or_update_cache(
         cache_obj[key] = entry
 
         async with entry.lock:
-            if datetime.now() < entry.expiration and entry.value is not None:
-                # The entry is specifically given a value of None above - so
-                # will this return ever happen? Can we drop this if block?
-                return entry.value
-
             return await _fetch_and_store(func, entry, args, kwargs, wait_expiration)
 
 
--- a/merino/curated_recommendations/corpus_backends/caching.py
+++ b/merino/curated_recommendations/corpus_backends/caching.py
@@ -125,6 +125,12 @@ async def _get_or_update_cache(
     if key in cache_obj:
         entry = cache_obj[key]
         if datetime.now() > entry.expiration and not entry.lock.locked():
+            # If the cache has expired, kick off an async task to update it and
+            # try to return the stale value while the update is in progress.
+            # This is the "stale while revalidate" part.
+            #
+            # If the current cache entry has a value of None (e.g. on a cold
+            # start error), BackendError is raised.
             task = asyncio.create_task(
                 _update_cache(func, key, args, kwargs, wait_expiration, cache_obj)
             )
@@ -142,15 +148,28 @@ async def _get_or_update_cache(
                     # This line is reached if the background task successfully updated the cache.
                     return entry.value  # type: ignore[unreachable]
                 else:
-                    # This line is reached if the update failed, and there was no stale value.
+                    # This line is reached if the update failed OR if the update
+                    # is still in progress, and there was no stale value.
+                    #
+                    # In the case of a cold start with an initial failure (as
+                    # seen during merino deploys, which double pod count), the
+                    # cache key will exist with a value of None (see the `else`
+                    # block below). Subsequent calls to _update_cache (line 132
+                    # above) spawn a background task which is specifically
+                    # *not* awaited. While this background task executes, an
+                    # error will continue to be raised here.
                     raise BackendError(f"Failed to obtain value for {key}")
     else:
         # This is the first request for this cache key, so an entry must be created.
         entry = CacheEntry(value=None, expiration=datetime.min, lock=asyncio.Lock())
         cache_obj[key] = entry
+
         async with entry.lock:
             if datetime.now() < entry.expiration and entry.value is not None:
+                # The entry is specifically given a value of None above - so
+                # will this return ever happen? Can we drop this if block?
                 return entry.value
+
             return await _fetch_and_store(func, entry, args, kwargs, wait_expiration)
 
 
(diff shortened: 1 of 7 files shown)


merino/curated_recommendations/corpus_backends/circuitbreaker.py
```diff
@@ -0,0 +1,54 @@
...
+
+class CuratedRecommendationsCircuitBreaker(CircuitBreaker):
+    """Circuit breaker for the Curated Recommendations backends.
+
+    this circuit breaker is intended to avoid a thundering herd problem
+    when merino deploys - which doubles the pod count, all of which exist
+    in a cold-start/empty cache state.
+
+    this decorator sits between two other decorators on the `fetch` functions
+    of any corpus backend (sections and scheduled surfaces, currently). the
```
mmiermans, line 29: It's not yet added to scheduled_surface_backend.py.

merino/curated_recommendations/corpus_backends/caching.py
```diff
@@ -142,15 +148,28 @@ async def _get_or_update_cache(
...
         # This is the first request for this cache key, so an entry must be created.
         entry = CacheEntry(value=None, expiration=datetime.min, lock=asyncio.Lock())
         cache_obj[key] = entry
+
         async with entry.lock:
             if datetime.now() < entry.expiration and entry.value is not None:
+                # The entry is specifically given a value of None above - so
+                # will this return ever happen? Can we drop this if block?
                 return entry.value
+
```
mmiermans:
```suggestion
```
Good catch. This can be removed because there's no suspension between `_get_or_update_cache` being called and this condition; in particular `async with entry.lock` does not suspend.

mmiermans (commented):
Nice work! Overall looks good to me.

merino/curated_recommendations/corpus_backends/circuitbreaker.py
jpetto: ah yes - wanted to get the pattern sorted on sections (which is the heavier call) to keep PR small. once this merges i'll file a follow up to add it to the scheduled surface backend.

tests/integration/api/v1/curated_recommendations/corpus_backends/test_sections_backend.py
```diff
@@ -218,3 +284,157 @@ async def test_fetch_schedules_spindle_refresh(
...
+    ):
+        """The fallback must raise so @stale_while_revalidate can serve stale data."""
+        caplog.set_level(logging.ERROR)
+        backend = make_sections_backend(sections_http_client)
+        fresh = await backend.fetch(SurfaceId.NEW_TAB_EN_US)
+        assert sections_http_client.post.call_count == 1
+
+        # Expire the cached value, then open the breaker.
+        for entry in backend._cache.values():
+            entry.expiration = datetime.min
```
mmiermans: It's not ideal to tie a test to the internal implementation of the backend, but I'm assuming that this was done because advancing time proved too difficult? I think it's acceptable if the alternative is too complex or messy.

tests/integration/api/v1/curated_recommendations/corpus_backends/test_sections_backend.py
```diff
@@ -218,3 +284,157 @@ async def test_fetch_schedules_spindle_refresh(
...
+        sections_http_client: AsyncMock,
+        fixture_request_data,
+        sections_circuit_breaker,
+    ):
+        """After the recovery timeout the breaker half-opens, then closes on a good fetch."""
+        await trip_breaker(make_sections_backend, fixture_request_data)
+        assert sections_circuit_breaker.opened
+
+        # circuitbreaker measures the recovery window with time.monotonic().
+        mocker.patch("circuitbreaker.monotonic", return_value=monotonic() + RECOVERY_TIMEOUT + 1)
```
mmiermans: This ties the test to an internal implementation of a third party, which makes the test slightly more fragile and harder to read (although the comment helps).

Could you try to convert this to a freezegun approach? Claude says it got it to work using `tick()`:
```python
with freezegun.freeze_time("2026-08-18", tick=True) as freezer:
    await trip_breaker(make_sections_backend, fixture_request_data)
    assert sections_circuit_breaker.opened

    freezer.tick(RECOVERY_TIMEOUT + 1)
    assert sections_circuit_breaker.state == STATE_HALF_OPEN
    ...
```

jpetto: hmm, claude told me freezegun wouldn't work. can always ask the robot again!

tests/integration/api/v1/curated_recommendations/corpus_backends/test_sections_backend.py
jpetto: yeah, i can see if we can advance time to expire the cache (instead of overriding the expiration time as is done currently)

tests/integration/api/v1/curated_recommendations/corpus_backends/test_sections_backend.py
jpetto: just to add some detail here - the above wouldn't pass for me, but did for claude. i increased the timeout (`+ 5` instead of `+ 1`), which got it passing for me.

mmiermans (approved):

PR #1733, opened 2026-08-18
fix(wikipedia): add successful search count metric

kschelonka:
## References

JIRA: [DISCO-4430]

## Description
Add successful search count so the % requests resulting in error can be derived. This was already implemented for sports.

## PR Review Checklist

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/MC-2534)

[DISCO-4430]: https://mozilla-hub.atlassian.net/browse/DISCO-4430?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/merino/providers/suggest/wikipedia/backends/elastic.py
+++ b/merino/providers/suggest/wikipedia/backends/elastic.py
@@ -110,7 +110,9 @@ class ElasticBackend:
             raise BackendError(
                 f"Failed to search from Elasticsearch for {language_code}: {e}"
             ) from e
-
+        self._metrics_client.increment(
+            f"{ES_SEARCH_METRIC_NAME}.count", tags={"index": index_id}
+        )
         if "suggest" in res:
             return [
                 self.build_article(q, doc, language_code)
--- a/merino/providers/suggest/wikipedia/backends/elastic.py
+++ b/merino/providers/suggest/wikipedia/backends/elastic.py
@@ -86,9 +86,7 @@ class ElasticBackend:
             }
         }
 
-        # Add this if wikipedia gets more than one index; for now it's covered by the provider
-        # search count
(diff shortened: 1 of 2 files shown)


merino/providers/suggest/wikipedia/backends/elastic.py
```diff
@@ -110,7 +110,9 @@ async def search(self, q: str, language_code: str) -> list[dict[str, Any]]:
             raise BackendError(
                 f"Failed to search from Elasticsearch for {language_code}: {e}"
             ) from e
-
+        self._metrics_client.increment(
```
ncloudioj: This would only record the happy path. I see another one commented out on line 91, maybe re-enable that.

ncloudioj (approved):
r+wc.

merino/providers/suggest/wikipedia/backends/elastic.py
kschelonka: the metric description says that it only records the happy path and to use the sum of error + count for total. I agree it's not the most obvious convention.

I can update it but that would also need to change sports and the dashboards (historical data would be broken but that's probably ok). wydt?

ncloudioj: Ah, i see. Let's keep it as is then. Feel free to remove that comment if you don't mind.

kschelonka: oh such a good catch I did not notice that commented line

kschelonka: ok re-reading this code I think I misunderstood, it should be the "obvious"/expected convention. thanks for pushing on this, updating

kschelonka:
thanks @ncloudioj I believe I misread the metrics, it's updated for the expected convention

PR #1736, opened 2026-08-19
[DISCO-4428] refactor(fleece): extract sanitization code for reuse

ncloudioj:
## References

JIRA: [DISCO-4428](https://mozilla-hub.atlassian.net/browse/DISCO-4428)

## Description

This refactor extracts the sanitization & logging code so it can be used by both the fleece app and the fleece worker.
- The sanitization core has been relocated from "message_handlers/search_terms/handler.py" to "sanitize/sanitizer.py". As a result, `MessageHandler` has been replaced with a singleton `AsyncBatchQueue`.
- The fleece worker is refactored to run as an async app to facilitate other async sanitization dependencies (e.g. the NER threadpool, the exempt registry, etc.). Since streaming pull runs on its own thread, it adds a little complexity in passing the event loop around as well as async callback dispatching. Comments are provided in the source for clarification. This can be further simplified as the async support is getting added to the pub/sub client.   

[Update]: Apologies for disorganization of this PR. I ended up squashing the refactor and the wiring commits while experimenting with the stacked PR feature 😓. 

## PR Review Checklist

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/MC-2537)

[DISCO-4428]: https://mozilla-hub.atlassian.net/browse/DISCO-4428?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


merino-fleece/merino_fleece/sanitize/worker/main.py
```diff
@@ -1,38 +1,87 @@
...
+    consult both -- before ``start()`` returns.
+    """
+    configure_logging(
+        log_format=settings.logging.format,
+        level=settings.logging.level,
+        can_propagate=settings.logging.can_propagate,
+        current_env=settings.current_env,
+        logger_name="merino_fleece",
+    )
+    configure_sentry(
```
gruberb, line 51: I think this can leak search terms to sentry?  Via `include_local_variables=False` and can we exclude `web.suggest.sanitized`?

merino-fleece/merino_fleece/configs/default.toml
```diff
@@ -158,3 +158,16 @@ heartbeat_path = ""
...
+# nacking the message for redelivery. Keep comfortably under the subscription's max ack
+# deadline, since a blocked callback holds the message's lease.
+sanitize_timeout_sec = 30.0
+
+# MERINO_FLEECE_PUBSUB__MAX_MESSAGES
+# Pub/Sub flow-control cap on messages leased concurrently. The streaming pull dispatches
+# callbacks on its own thread pool, but they all funnel into the single NER thread pool
+# sized by `pii.executor_max_workers`, so leasing far more than that pool can chew through
+# just queues work until callbacks hit `sanitize_timeout_sec`. Keep it near the pool size.
+max_messages = 4
```
gruberb: `settings.pii.executor_max_workers` defaults to `. Should we derive the `max_messages` from `settings.pii.executor_max_workers`  instead of having this separate config here?

merino-fleece/merino_fleece/sanitize/worker/worker.py
```diff
@@ -166,10 +171,55 @@ def _connect(self) -> None:
...
+            # redeliver forever.
+            logger.exception("Dropping invalid message")
+            message.ack()
+            return
+
+        future = asyncio.run_coroutine_threadsafe(
+            sanitize_batch(submission.search_terms), self.loop
+        )
+        try:
+            future.result(timeout=self.sanitize_timeout_s)
```
gruberb: As far as I understand, the Pub/Sub callback thead calls `_callback()`, which than runs `run_coroutine_threadsafe` and schedules `sanitize_batch` on the asyncio event loop. The `sanitize_batch()` sends the call to the `ThreadPoolExecutor`. The timeout only controls how long the `callback` waits for the `run_coroutine_threadsafe` call. But by the time the time is running out, work has already scheduled on antoher OS thread and keeps running. 

Is the timeout needed here? Or do we think it's not an issue if work is running even though we think we cancelled it? 

```python
future = asyncio.run_coroutine_threadsafe(
    sanitize_batch(submission.search_terms),
    self.loop,
)

try:
    future.result() 
except Exception:
    logger.exception("Failed to sanitize search terms; message will be redelivered")
    message.nack()
    return

message.ack()
```

merino-fleece/merino_fleece/sanitize/worker/main.py
ncloudioj: Yes, it can. https://github.com/mozilla-services/merino-py/pull/1743 will take care of both redacting all the locals in this module and excluding `web.suggest.stanitized` and `web.suggest.log` from Sentry's log integration.

merino-fleece/merino_fleece/sanitize/worker/worker.py
ncloudioj: > As far as I understand, the Pub/Sub callback thead calls _callback(), which than runs run_coroutine_threadsafe and schedules sanitize_batch on the asyncio event loop. The sanitize_batch() sends the call to the ThreadPoolExecutor. The timeout only controls how long the callback waits for the run_coroutine_threadsafe call.

Correct!

>  Is the timeout needed here?

I think so. Here is my reasoning: `_callback()` runs on one of pub/sub callback threads and blocks there until the coroutine finishes. Without a deadline, a coroutine that never completes (for whichever reason) blocks that thread forever, which also prevents the message from being redelivered. To err on the side of caution, I think we should guard this callback with a timeout (30s which should be more than enough given the average sanitization duration is 50ms per batch).  

> But by the time the time is running out, work has already scheduled on antoher OS thread and keeps running. Or do we think it's not an issue if work is running even though we think we cancelled it?

Good call out! Note the cancel call for timeout error handling, it would only cancel the only awaiting point (`await detect_person_batch()`) within `sanitize_batch()`. Both side effects, recording metrics and logging, come after it. So a cancelled batch can only be interrupted during NER, before anything is emitted. Granted, if the timeout happens after the await (which would be rather rare), it will cause that message to be re-delivered.

Does that make sense?

merino-fleece/merino_fleece/configs/default.toml
ncloudioj: Resolved via 2b4647e29c0cdad1c6c5177a7e65e7c65b633827.

now, `max_message = max_workers * messages_per_ner_worker`

merino-fleece/merino_fleece/sanitize/worker/main.py
```diff
@@ -1,38 +1,89 @@
...
+        heartbeat_interval=settings.pubsub.heartbeat_interval,
     )
+    for sig in (signal.SIGTERM, signal.SIGINT):
+        # Handled on the loop rather than by interrupting whichever thread is active.
+        loop.add_signal_handler(sig, _handle_exit, worker, sig)
+
+    try:
+        # The streaming pull is blocking and synchronous, so it gets a thread of its own;
+        # this keeps the loop free to run the sanitization callbacks.
+        await asyncio.to_thread(worker.start)
```
kschelonka: non-blocking: the streaming pull reference reads a little misleading to me -- the `start` method is synchronous and blocking because of the endless restart loop. the streaming pull itself is executed in a background thread and is fully async. I think now that sanitization is async we should just remove the blocking restart loop from `worker.start` altogether and just let k8s restart the pod once the heartbeat goes stale. simplifies the code too.

merino-fleece/merino_fleece/configs/default.toml
```diff
@@ -158,3 +158,20 @@ heartbeat_path = ""
 # How often to refresh the heartbeat file. Keep the probe's staleness threshold
 # comfortably above `heartbeat_interval + restart_backoff`.
 heartbeat_interval = 10  # seconds
+
+# MERINO_FLEECE_PUBSUB__SANITIZE_TIMEOUT_SEC
+# Max seconds a message callback waits for its search terms to be sanitized before
+# nacking the message for redelivery. Keep comfortably under the subscription's max ack
+# deadline, since a blocked callback holds the message's lease.
+sanitize_timeout_sec = 30.0
```
kschelonka: blocking: the ack deadline on the queue is 30s -- let's set this lower

merino-fleece/merino_fleece/sanitize/worker/main.py
ncloudioj: > the streaming pull itself is executed in a background thread and is fully async.

It's executed in a background thread managed by the google pub/sub client (i.e. it's thread-based rather than asyncio-based), thread gymnastics is still needed b/c sanitization execution needs to be dispatched by the main loop running on the main thread. 

In 34e9ffe68f235c5e40beecf94ade5e07095278b9, I switched over to the community client `gcloud-aio-pubsub` (we already use `gcloud-aio-storage` for merino), which significantly simplifies the code thanks to its native asycnio support. However, the downsides are:
- This client pings the pub/sub REST pull endpoints instead of using the grpc-based streaming of the google client. As a result, it needs us to set a couple of knobs to control how the subscriber pulls the messages. I added `SubscriberTuning` to facilitate that based on our pub/sub configurations.  
- It doesn't support certain advanced features such as "exactly-once" delivery.

I'd lean towards simplicity to use the community client as we don't need those fancy features for now.

@kschelonka @gruberb Let me know your thoughts.

kschelonka: apologies I think I wasn't clear. this comment was targeted towards updating the code architecture I wrote initially to eliminate the need for additional threading by making the subscription non-blocking (via getting rid of blocking restart loop). will review your change with the new library

kschelonka (approved):

PR #1759, opened 2026-08-31
[HNT-3124] Resize potd hi-res image to bounded WebP before upload

mmiermans:
## References

JIRA: [HNT-3124](https://mozilla-hub.atlassian.net/browse/HNT-3124)

## Description

The potd job re-hosted the untouched Wikimedia original on the wallpaper path. On 2026-08-31 that was a **110MB, 18018×8275 (~149 megapixel) JPEG**, which New Tab clients downloaded in full and re-decoded on every new-tab activation (several seconds per paint, per dholbert's Slack report). It also feeds directly into the CDN cost concerns tracked in [HNT-3056](https://mozilla-hub.atlassian.net/browse/HNT-3056).

The `update-wikimedia-potd` job now processes the hi-res image before uploading to GCS:

- **Downscale to fit within 3840px on the longest edge** (4K UHD width), preserving aspect ratio and never upscaling. The POTD widget blueprint requires ≥1920px width and prefers ≥2560px, so 3840 keeps ample headroom; today's image becomes 3840×1764 — the same rendition Wikimedia itself offers as a download.
- **Re-encode as WebP quality 75**, matching the quality the New Tab image cache (img-getpocket.cdn.mozilla.net) uses for content thumbnails. The manifest's `high_res_image_url` now points at `hi_res.webp`.
- **Strip metadata** (EXIF etc.), after applying the EXIF orientation to the pixels; the ICC profile is kept so colors survive.
- **Fail safe**: a source that cannot be decoded or exceeds a 500 megapixel bound fails the day's update, so the provider keeps serving the previous day's picture rather than shipping an unprocessed original. JPEG sources are decoded at a reduced DCT scale (`draft()`), so even very large pictures never materialize at full resolution in memory.

The thumbnail (already a small Wikimedia rendition) is uploaded unmodified, as before.

Also folded in per SREIN-1732: **image Cache-Control goes from `max-age=3600` to `public, max-age=31536000, immutable`**. Each day's images live at a fresh dated path and are never modified, so hourly re-fetching only burned CDN transfer. Note this closes the "swap a bad image in place" remediation used on 2026-08-31 — remediation now means publishing the fixed image at a new path via the manifest, which stays short-cached as the mutable pointer.

Verified against the actual 2026-08-31 image:

| | before | after |
|---|---|---|
| bytes | 109,951,518 | 1,461,482 (**75× smaller**) |
| dimensions | 18018×8275 | 3840×1764 |
| format | JPEG | WebP q75 |

Processing took 2.2s with ~580MB peak RSS in the job process (dominated by two in-memory copies of the 110MB source; the decode buffer itself stays ~28MB thanks to DCT draft). Worth confirming the job pod's memory limit accommodates this.

### Why these choices (data from a full-year scan of the Featured API, 2025)

- **PNG support is kept deliberately**: 98.3% of days are JPEG, but ~4-5 days/year are PNG and the Featured API does *not* normalize the source format. SVG (~2 days/yr) and GIF (~1 every few years) sources are already rejected by the pre-existing URL extension check and fail safe.
- **The pixel-bound override is exercised, not theoretical**: ~2.2% of 2025 days exceeded Pillow's default ~89MP decompression-bomb guard (which would otherwise warn/error), including a 1.02-gigapixel day (23181×44120) that exceeds even the 500MP bound and fails safe.

### Notes for reviewers

- The proposed ["How the POTD widget should crop pictures" ADR](https://mozilla-hub.atlassian.net/wiki/spaces/FPS/pages/2981003324) documents keeping the untouched original on the wallpaper path; this PR deliberately revises that decision in response to the incident.
- Alternative considered: wrapping the source in the image cache (thumbor). Rejected for this fix because its `MAX_SOURCE_SIZE` is ~19.5MB and it runs in a 30s/2GB Lambda, so it would refuse exactly the oversized sources that motivate this change, and would add a cross-repo deploy dependency.
- Possible follow-up: request an appropriately sized rendition from Wikimedia's thumb service instead of downloading the original, which would avoid the 110MB fetch entirely and also cover the rare SVG/gigapixel days.
- New config: `image_max_dimension` (3840) and `image_webp_quality` (75) under `rss_providers.wikimedia_potd`.

## PR Review Checklist

[HNT-3124]: https://mozilla-hub.atlassian.net/browse/HNT-3124?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[HNT-3056]: https://mozilla-hub.atlassian.net/browse/HNT-3056?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/merino/merino/configs/default.toml
+++ b/apps/merino/merino/configs/default.toml
@@ -1250,6 +1250,18 @@ commons_api_url = "https://commons.wikimedia.org/w/api.php"
 # Cache control headers for stored images in the bucket (1hr)
 cache_control = "public, max-age=3600"
 
+# MERINO_RSS_PROVIDERS__WIKIMEDIA_POTD__IMAGE_MAX_DIMENSION
+# Upper bound (in pixels) on the longest edge of the hi-res image uploaded to the bucket.
+# Larger sources are downscaled to fit, preserving aspect ratio; smaller sources are kept
+# at their original size. 3840 matches the width of a 4K UHD display, the largest common
+# resolution the New Tab wallpaper renders at.
+image_max_dimension = 3840
+
+# MERINO_RSS_PROVIDERS__WIKIMEDIA_POTD__IMAGE_WEBP_QUALITY
+# WebP quality (0-100) used when re-encoding the hi-res image, matching the quality the
+# New Tab image cache (img-getpocket.cdn.mozilla.net) uses for content thumbnails.
+image_webp_quality = 75
+
 [default.games_providers.particle]
 # MERINO_GAMES_PROVIDERS__PARTICLE__TYPE
 # The type of this provider, should be `particle`.
--- a/apps/merino/merino/configs/default.toml
+++ b/apps/merino/merino/configs/default.toml
@@ -1247,8 +1247,10 @@ featured_api_base = "https://api.wikimedia.org/feed/v1/wikipedia"
 commons_api_url = "https://commons.wikimedia.org/w/api.php"
 
 # MERINO_RSS_PROVIDERS__WIKIMEDIA_POTD__CACHE_CONTROL
-# Cache control headers for stored images in the bucket (1hr)
-cache_control = "public, max-age=3600"
+# Each day's images are uploaded to a fresh dated path and never modified, so clients and
+# the CDN may cache them forever; hourly re-fetching was a major CDN cost driver (SREIN-1732).
+# Remediating a bad image requires publishing it at a new path via the manifest.
+cache_control = "public, max-age=31536000, immutable"
 
 # MERINO_RSS_PROVIDERS__WIKIMEDIA_POTD__IMAGE_MAX_DIMENSION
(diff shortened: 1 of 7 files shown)


jpetto:
while we're here, can we also update the cache for this image to be immutable and (for all intents and purposes) forever? there's never a need for the cache to expire on one of these images, right?

currently [the cache is set at 1 hour](https://github.com/mozilla-services/merino-py/blob/b00186b8952cfcc575d2a7bc4abae8d5c71f02e2/apps/merino/merino/configs/default.toml#L1251).

Herraj (approved):
Paired and reviewed this live with Matt.

PR #1763, opened 2026-09-01
[HNT-3134] Fix flaky stale-while-revalidate cache tests

mmiermans:
## References

JIRA: [HNT-3134](https://mozilla-hub.atlassian.net/browse/HNT-3134)

## Description

`TestScheduledSurfaceCircuitBreaker::test_open_breaker_serves_stale_cache` fails intermittently in CI ([example run](https://github.com/mozilla-services/merino-py/actions/runs/33528342756/job/99924813210)). The test ticks the frozen clock forward by exactly `cache_time_to_live_max` to expire the stale-while-revalidate cache. However, freezegun's `tick()` advances from the moment the freeze started rather than from the current time, so real time spent on the first fetch is not counted. When the random TTL draws near its maximum on a slow runner, the cache is then still fresh after the tick, no revalidation task is spawned, and the expected "Returning stale data" log never appears.

The fix ticks one minute past the maximum TTL. The sections backend test, `test_cache_returned_on_subsequent_calls`, and the caching unit tests used the same pattern and get the same margin, and the breaker recovery tests get the same margin on their recovery ticks. The two stale-cache tests now also request the existing breaker reset fixture, since they previously left the shared circuit breaker open. The margins are frozen time, so test durations are unchanged.

I reproduced the failure locally by randomly pausing the test process with SIGSTOP to mimic CI scheduler stalls. Before the fix, the unmodified test failed 1 time in 45 runs. After the fix, 413 runs of this test and 60 runs of the sections clone under the same stress all passed. If the failure rate were unchanged, the chance of that outcome would be below 0.01 percent.

## PR Review Checklist

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/MC-2564)

[HNT-3134]: https://mozilla-hub.atlassian.net/browse/HNT-3134?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/merino/tests/integration/api/v1/curated_recommendations/corpus_backends/test_scheduled_surface_backend.py
+++ b/apps/merino/tests/integration/api/v1/curated_recommendations/corpus_backends/test_scheduled_surface_backend.py
@@ -6,7 +6,7 @@ import logging
 import pytest
 
 from circuitbreaker import STATE_HALF_OPEN, CircuitBreakerMonitor
-from datetime import datetime
+from datetime import datetime, timedelta
 from httpx import AsyncClient, HTTPStatusError, Response
 from pydantic import HttpUrl
 from tests.types import FilterCaplogFixture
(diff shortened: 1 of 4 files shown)


jpetto (approved):
playing with time is always a tricky business.

PR #1794, opened 2026-09-21
[HNT-3238] ctrpred engb, time-of-day propensity

mlplyler:
## References

JIRA: [HNT-3238](https://mozilla-hub.atlassian.net/browse/HNT-3238)

## Description
This experiment tests time-of-day propensity in en_GB. The adjusted_field in the engagement table is now time-of-day adjusted for the treatment region. This PR routes a user response to the experiment engagement table. 

We are also cleaning out the concluded german publisher constraint experiment.

We are also removing global engagement blending for the experiment branches. This is to avoid info leakage between treatment and control. Note it is possible our treatment adjusted_impressions could leak to other regions where the same article is served. This is technically mixing units incorrectly, but the units are on the same scale and this should be a rare event. This is not worth correcting for.

## PR Review Checklist

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/MC-2595)

[HNT-3238]: https://mozilla-hub.atlassian.net/browse/HNT-3238?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/merino/merino/curated_recommendations/article_balancer_configs.py
mmiermans: Remove this unused parameter?

apps/merino/merino/curated_recommendations/rankers/ranker.py
```diff
@@ -79,6 +80,10 @@ def get_regional_prior(
         Branch-specific priors are request-level artifacts. If a branch prior exists,
         use it for the full ranking pass; otherwise fall back to the base country prior.
         """
+        if is_ctr_prediction_engagement_region(prior_region):
+            # Do not substitute a prior fitted from another experiment cohort.
+            return self.prior_backend.get(prior_region) or ConstantPrior().get()
```
mmiermans: Claude suggests this, because experiment priors don't exist yet in the backend:
```suggestion
            return self.prior_backend.get(prior_region) or  self.prior_backend.get(region) or ConstantPrior().get()
```

But I'm guessing you are going to create those before turning the experiment on?

apps/merino/merino/curated_recommendations/utils.py
```diff
@@ -138,29 +133,32 @@ def derive_region(locale: Locale, region: str | None = None) -> str | None:
...
     branch = request.experimentBranch
 
     if (
-        PUBLISHER_CONSTRAINT_IN_GERMANY_BRANCH_ENGAGEMENT_ENABLED
-        and region == PUBLISHER_CONSTRAINT_IN_GERMANY_REGION
+        get_recommendation_surface_id(request.locale, request.region, request)
+        == SurfaceId.NEW_TAB_EN_GB
         and branch is not None
-        and branch in PUBLISHER_CONSTRAINT_IN_GERMANY_BRANCHES
+        and branch in {"control", "treatment"}
```
mmiermans: Is this line necessary? Can you be in the `CTR_PREDICTION_ENGB_EXPERIMENT` experiment and not be in control or treatment?

mmiermans (approved):
I only have some minor suggestions and a question.

apps/merino/merino/curated_recommendations/rankers/ranker.py
mlplyler: This is a good fallback suggestion. Going with the suggestion.

apps/merino/merino/curated_recommendations/utils.py
mlplyler: You're right its unnecessary.

I think we are probably protected enough without it because the experiment branches in experimenter determine which engagement-regions exist, so its better to not define here.

